PR #12, opened 2019-04-16
Convert to TypeScript

JeremyRH:
First off, I apologize for the major changes to the code. It was difficult to add types the way it was written before. I bumped the version of a few dependencies and also made a slight change to the `api` object returned from `create`:

```diff
[useStore, {
  getState,
+ setState,
  subscribe,
  destroy
}] = create((set, get) => ({ value: null }))
```

Additionally, the `useStore` hook was modified slightly. I removed the need for memoizing selectors. The way I did it was by using a ref object. Example: 
```js
function useSomeHook(value) {
  const ref = useRef(value)
  if (ref.current !== value) ref.current = value // ref is always updated when needed
  const fn = useCallback(() => ref.current, []) // function always has latest value
```
This means store subscriptions will only unsubscribe when the component unmounts and still allows selector functions to change.

I also bumped the minor version number to `0.1.0` to follow semantic versioning rules.

Let me know if you have any questions or concerns and feel free to say what you like and what you don't.


JeremyRH:
I realized a major change to the API wasn't worth the "perfect" type definitions and reverted some of my changes. The `set` and `get` functions passed into the store creator use a generic `Record<string, any>` type so there's not much help when using them. The `api.getState` and `api.setState` methods do have proper type inference and should just work.

drcmda:
This looks great! You think it’s ready to merge? I’m on a small travel right now, no laptop with me, but from looking over the changes everything seems reasonable.

JeremyRH:
I believe it is ready to merge. I've manually tested it pretty thoroughly but there is always a chance I missed something. The few unit tests cover most of the functionality but there are gaps. Safest bet would be to manually review and test this when you are back.

I'm available to fix any issues that arise from my changes though so you wouldn't be left alone to deal with it.

drcmda:
im back from a small trip, will give it a go today - thanks for holding out : )

JeremyRH:
Did you have a chance to look at this and test it yet?

drcmda:
yes, worked flawless, thank you so much! published it, too

drcmda:
@JeremyRH could you give me another quick rundown how you managed to remove selector dependencies?

For instance

    const person = useStore(state => state.persons[id])

Previously that would remove and register the selector as a new listener on every render, unless we append `, [id]`. How do you avoid it?

JeremyRH:
@drcmda Consider the following code:
```js
function useListener(fn) {
  const functionRef = useRef();
  // update functionRef.current if it has changed
  if (functionRef.current !== fn) functionRef.current = fn;

  useEffect(() => {
    function listener() {
      // functionRef.current references the current fn
      functionRef.current();
    }
    document.addEventListener("something", listener);
    return () => {
      document.removeEventListener("something", listener);
    };
  }, [] /* no deps */);
}
```
`fn` can be updated without removing and reattaching the event listener. This is done by referencing the property of a ref object which will always hold the value of the last `fn` passed in.

drcmda:
oh wow, this is such a neat idea. they've even just added deps to redux hooks. maybe we should pass this on.

src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  function useStore<U>(selector?: StateSelector<State, U>) {
+    // Gets entire state if no selector was passed in
+    const selectState = typeof selector === 'function' ? selector : getState
+    // Using functional initial b/c selected itself could be a function
+    const [stateSlice, setStateSlice] = React.useState(() => selectState(state))
+    // Prevent useEffect from needing to run when values change by storing them in a ref object
+    const refs = React.useRef({ stateSlice, selectState }).current
+
+    if (refs.stateSlice !== stateSlice) refs.stateSlice = stateSlice
+    if (refs.selectState !== selectState) refs.selectState = selectState
```
aleclarson: Mutating refs during render is discouraged, because any render can be interrupted by "time slicing" and "suspense" features that aren't released yet.

https://twitter.com/dan_abramov/status/1102182582362521600

drcmda: useEffect(() => (functionRef.current = fn), [fn])

Would work : )

JeremyRH: @aleclarson Thanks for sharing this. Here's my initial thoughts:

The issue is that we shouldn’t depend on the render phase to update a ref. Why is this bad? In the case of `useInterval`, you don't want a delayed or interrupted render to cause your callback to not be updated. So is @gaearon saying that `useEffect` is not delayed or interrupted by Suspense and Time Slicing? That seems great but I would like to verify this is true.

In the case of zustand, `useStore` subscribes to the store by adding a state setter to a list. The values of the store are able to be updated outside of React by exposing a function that iterates over the list and calls the state setters. This means if an update to the store happens during a delay, state setters can be called when they "shouldn’t". I put "shouldn’t" in quotes because if the component needed to change its subscription, it was already rendering or set to render anyway. So calling its state setter should just cause the render to be rescheduled. I feel like whatever reconciliation happens during/after Suspense and Time Slicing could handle this.

Anyway, I could be completely wrong about these assumptions. We should just do what @drcmda suggests and put the ref update in `useEffect`.

src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    React.useEffect(() => {
+      return subscribe(() => {
```
timkindberg: Is it supposed to receive the updated `state` in this listener?

JeremyRH: `state` is a `let` variable so it can be reassigned. It is declared once on `create()`. The callback passed into `subscribe` will always have the latest `state`.

timkindberg: So, not arguing just trying to learn...

The useEffect will only be run once because the dependencies is an empty array. That means subscribe will only be called once, which means the listener function will only be created once. Won't it lock the state variable to what it was when the listener was created using JavaScript scoping? 10+ doing JavaScript and I'm still learning about scoping...

JeremyRH: No problem, I'm happy to share :)
This listener is created once but `let` and `var` variables declared in a parent scope are not frozen to their values. MDN has a good explanation of this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures#Emulating_private_methods_with_closures

src/index.ts
aleclarson: > In the case of `useInterval`, you don't want a delayed or interrupted render to cause your callback to not be updated. So is @gaearon saying that `useEffect` is not delayed or interrupted by Suspense and Time Slicing?

In the case of `useInterval`, you don't want your refs updated until the render has been applied to the DOM, which is what `useEffect` guarantees.

JeremyRH: Ah I see. Thanks.

PR #16, opened 2019-04-29
Update refs in useEffect, add more tests, fix pre-commit lint

JeremyRH:
Just minor updates. Adding a test for `destroy` made me question something. The store can be used after `destroy` is called. Is this expected behavior?

Also, I did not update the package.json version. I noticed `yarn publish` gives you the option to bump the version when deploying the package so I figured I would leave that to @drcmda

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/esm/index.js": {
-    "bundled": 2298,
-    "minified": 910,
-    "gzipped": 483,
+    "bundled": 2417,
+    "minified": 898,
+    "gzipped": 470,
     "treeshaked": {
       "rollup": {
         "code": 14,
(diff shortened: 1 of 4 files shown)


drcmda:
true, haven't given that much thought. destroy is intended for manual usecases when apps cave in, so that some state can be freed by the GC earlier. 

@JeremyRH interested in having access to it with shared authorship in package.json, git-access and npm? we can move this onto a neutral org as well. loved the last PR, and it's always good when more people have oversight. we're going forward with this now in a bigger project, so i'll be here helping to maintain, but could use your good input.

drcmda:
and yes, usually i do

    npm version patch
    git push --follow-tags
    npm publish <---it builds by itself

JeremyRH:
@drcmda Wow, thanks! Yeah, I'm interested in helping out any way I can. I am also using zustand in a larger project and everything is running smoothly. A neutral org is pretty common among popular packages so it's probably a good idea.

My npm username is the same as my github: [jeremyrh](https://www.npmjs.com/~jeremyrh).

drcmda:
moved it to the react-spring org, github and npm access should work now 🙂

PR #19, opened 2019-05-10
Update state slice when selector changes

### LINKED ISSUE #18: I think we need the dependencies array back...
I think we need dependencies back. When I upgraded to 0.1.1 my code broke. I like to create selectors in my state and those selectors use external variable dependencies. It's probably easiest to just look at a working example of the issue.

https://codesandbox.io/s/8x2ylkzpj8

I posted this in a comment, but then realized I should probably open an issue.

JeremyRH:
The latest selector was only used in the subscribe callback. It should also be used when `useStore` is called. Should fix #18.

This change should be reviewed and tested by others (@drcmda, @timkindberg) to make sure it actually fixes the issue and does not cause regressions. I only added one test.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/esm/index.js": {
-    "bundled": 2379,
-    "minified": 885,
-    "gzipped": 470,
+    "bundled": 2443,
+    "minified": 929,
+    "gzipped": 491,
     "treeshaked": {
       "rollup": {
         "code": 14,
@@ -14,8 +14,8 @@
     }
   },
   "dist/cjs/index.js": {
-    "bundled": 2751,
-    "minified": 1089,
-    "gzipped": 527
+    "bundled": 2843,
+    "minified": 1148,
+    "gzipped": 550
   }
 }
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/esm/index.js": {
-    "bundled": 2443,
-    "minified": 929,
-    "gzipped": 491,
+    "bundled": 2756,
+    "minified": 989,
+    "gzipped": 509,
     "treeshaked": {
       "rollup": {
         "code": 14,
@@ -14,8 +14,8 @@
     }
   },
   "dist/cjs/index.js": {
-    "bundled": 2843,
-    "minified": 1148,
-    "gzipped": 550
+    "bundled": 3079,
+    "minified": 1126,
+    "gzipped": 527
   }
 }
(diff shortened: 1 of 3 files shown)


drcmda:
calling the selector on every render is a good thing imo. perhaps we could still do usememo with optional deps? like

    const slice = useMemo(() => selectState(state), deps || [{}])

this way selector calls *could* be optimized if they're maybe super heavy or something.

JeremyRH:
Would it need to be called the same way in the subscribe callback? We cant use hooks there. Although we could only update `selectState` if the optional dependencies change:
```js
// deps is readonly any[] or undefined
const selectState = useCallback(typeof selector === 'function' ? selector : getState, deps)
```
`useCallback` does nothing when called with only one argument which is what we want.

JeremyRH:
Yep, `useMemo` will run every render when deps are omitted. So `useCallback` to keep the selector in sync for when it is called in the subscribe callback and `useMemo` to optionally restrict the selector from being called in the `useStore` hook. Should I make another PR for adding optional dependencies?

drcmda:
amending this one would be best i think.

JeremyRH:
Alright, it's ready to be tested.

timkindberg:
The code looks right to me... I'll test when I get a second, maybe tonight.

drcmda:
@JeremyRH tried it out in CSB and TS complained about "readonly", when i removed it it worked as expected. I don't know TS good enough, is this CSB's fault? Here's the sandbox: https://codesandbox.io/s/o7nky5wwj5

JeremyRH:
@drcmda It might be the TypeScript parser in CSB is not up to date. You can change it from `readonly any[]` to `ReadonlyArray<any>`.

I found an issue yesterday. If you update the state without updating the dependencies, the selector is not called and you get the old state from `useStore`. [Here is an example.](https://codesandbox.io/s/k5jpowk4kv)

JeremyRH:
I think I fixed it now. 

You can specify the selector doesn't need to be called in `useStore` with optional dependencies:
```js
function Component() {
  // empty deps. selector only called in subscribe callback and on initialization
  const foo = useStore(state => state.foo, [])
  return <div>{foo}</div>
}
```

There is also another way to prevent the selector from being called in `useStore`:
```js
function selector(state) {
  return state.foo
}
function Component() {
  // static reference to selector. only called in subscribe callback and on initialization
  const foo = useStore(selector)
  return <div>{foo}</div>
}
```

drcmda:
Nice! Thanks Jeremy!

JeremyRH:
Ah, there's a bug because I forgot to update the dependencies ref. @drcmda if you haven't, don't publish to npm yet. I'll fix it in a few hours when I get home.

drcmda:
merged the other PR and it's out now as 0.1.3.

PR #22, opened 2019-05-24
Improve Typescript types and generics

### LINKED ISSUE #21: Improve Typescript support and only make the first generic type required.
As soon as you start messing with the first generic, you are quickly hit by the second and third required generics. This could probably be improved by giving those default types.

Also exporting some of the types and an interface for the store api would be a nice addition. This way we can create functions that require the store api as a parameter for example.

I've already tried implementing some of these, but sadly broke the auto-generation of the `State` generic, so no auto-completion for `set` and `get` and so on, which probably isn't so nice for none typescript users.

https://github.com/Luxizzle/zustand/commit/b095e8cd63ea56854ae8cd61dc28faa604049ac8

Luxizzle:
Fixes #21 

Exports a few types for usage outside of the module and adds default types for generics.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,10 @@ export type State = Record<string, any>
 export type SetState<T> = (partialState: PartialState<T>) => void
 export type GetState<T> = () => T
 
-export type UseStore<T> = <U = T>(selector?: StateSelector<T, U>) => U
+export type UseStore<T> = {
+  (): T
+  <U>(selector: StateSelector<T, U>, dependencies?: ReadonlyArray<any>): U
+}
 
 export interface StoreApi<T> {
   getState: GetState<T>
--- a/src/index.ts
+++ b/src/index.ts
@@ -59,7 +59,7 @@ export default function create<TState extends State>(
   function useStore<U>(
     selector?: StateSelector<TState, U>,
     dependencies?: ReadonlyArray<any>
-  ) {
+  ): TState | U {
     // State selector gets entire state if no selector was passed in
     const stateSelector = typeof selector === 'function' ? selector : getState
     const selectState = useCallback(
--- a/src/index.ts
+++ b/src/index.ts
@@ -101,5 +101,5 @@ export default function create<TState extends State>(
   )
   const api = { destroy, getState, setState, subscribe }
 
-  return [useStore, api] as [typeof useStore, typeof api]
+  return [useStore, api]
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,20 +1,35 @@
 import { useCallback, useLayoutEffect, useReducer, useRef } from 'react'
 import shallowEqual from './shallowEqual'
 
-type StateListener<T> = (state: T) => void
-type StateSelector<T, U> = (state: T) => U
-type PartialState<T> = Partial<T> | ((state: T) => Partial<T>)
+export type StateListener<T> = (state: T) => void
+export type StateSelector<T, U> = (state: T) => U
+export type PartialState<T> = Partial<T> | ((state: T) => Partial<T>)
+
+export type State = Record<string, any>
+export type SetState<T> = (partialState: PartialState<T>) => void
+export type GetState<T> = () => T
+
+export type UseStore<T> = <U = T>(selector?: StateSelector<T, U>) => U
+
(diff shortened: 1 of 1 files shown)


drcmda:
Looks good to me, we need these types as well at work, duplicated them for now. @JeremyRH how do you feel?

JeremyRH:
The reason for not wanting to expose all the types is mostly for type names. If we change a name, it's a breaking change. I do see the value for not duplicating types though so I'm thinking we should expose all the types.

 I'll test out these changes today and try to get this merged. Thanks @Luxizzle!

dm385:
When using "zustand" in one of our projects, I had defined the types UseStore<T> and StoreApi<T> in globals.d.ts myself. Now I came across this pull request. If I now compile the project with Luxizzle's solution directly with tsc, I get some compiler errors. These can also be reproduced in this repo when compiling the file test.tsx. For this reason, I have adapted the definitions in my own branch and adapted the configuration so that it can be compiled directly with tsc (yarn build:tsc).

You can have a look at it if you like:
[improve-typescript](https://github.com/dm385/zustand/tree/improve-typescript)

Please correct me if I am wrong.

JeremyRH:
@dm385 You're right, the tests were failing with Luxizzle's changes. I asked them to create a PR so we could work on fixing the issues together. I have a version of his changes working without changing the tests now.

And thank you for the fixes on your fork. I'll see if there's anything there that I can add to this.

src/index.ts
```diff
@@ -1,20 +1,35 @@
...
 const reducer = <T>(state: any, newState: T) => newState
 
 export default function create<
-  State extends Record<string, any>,
-  SetState extends (partialState: PartialState<Record<string, any>>) => void,
-  GetState extends () => Record<string, any>
->(createState: (set: SetState, get: GetState) => State) {
-  const listeners: Set<StateListener<State>> = new Set()
+  TState extends State,
+  TSetState extends SetState<TState> = SetState<TState>,
```
JeremyRH: These `TSetState` and `TGetState` generics should just be removed. I was trying to use them to generate correct function signatures of the `set` and `get` functions passed into `createState`. I couldn’t get it to work though so we can simplify this bit by removing them:
```ts
export type CreateState<T> = (set: SetState<State>, get: GetState<State>) => T
...
export default function create<TState>(
  createState: CreateState<TState>
): [UseStore<TState>, StoreApi<TState>] {
```

src/index.ts
```diff
@@ -81,7 +96,7 @@ export default function create<
     return stateSlice
   }
 
-  let state = createState(setState as SetState, getState as GetState)
+  let state = createState(setState as TSetState, getState as TGetState)
   const api = { destroy, getState, setState, subscribe }
 
   return [useStore, api] as [typeof useStore, typeof api]
```
JeremyRH: `as [typeof useStore, typeof api]` is not needed because you correctly defined the return type in the function signature.
```diff
- const api = { destroy, getState, setState, subscribe }
-
- return [useStore, api] as [typeof useStore, typeof api]
+ return [useStore, { destroy, getState, setState, subscribe }]
```

src/index.ts
```diff
@@ -81,7 +96,7 @@ export default function create<
     return stateSlice
   }
 
-  let state = createState(setState as SetState, getState as GetState)
+  let state = createState(setState as TSetState, getState as TGetState)
```
JeremyRH: My proposed changes to the `createState` type will require this change:
```ts
let state = createState(
  setState as SetState<State>,
  getState as GetState<State>
)
```

src/index.ts
```diff
@@ -34,16 +49,16 @@ export default function create<
...
   function useStore<U>(
-    selector: StateSelector<State, U>,
+    selector: StateSelector<TState, U>,
     dependencies?: ReadonlyArray<any>
   ): U
   function useStore<U>(
-    selector?: StateSelector<State, U>,
+    selector?: StateSelector<TState, U>,
     dependencies?: ReadonlyArray<any>
   ) {
```
JeremyRH: We should add a return type to useStore: `TState | U`.

Luxizzle:
I updated everything with the proposed changes you gave. Thanks for all the explanations by the way, they really helped!

Issue #23, opened 2019-05-29
Feature request: api.subscribe(callback, /*selector*/)


drcmda:
We need a way to bind components to the store without causing updates (for rapid micro updates that would cause lots of render requests). See: https://twitter.com/0xca0a/status/1133329007800397824

In theory this can be done using subscribe, but it can't select state. My suggestion would be an optional second arg:

```jsx
const [, api] = create(set => ({
  a: 1,
  b: 2,
  set
}))
const unsub = api.subscribe(state => console.log(state), state => state.b)
api.getState().set(state => ({ ...state, b: 3 }))

// console would say: 3
```

Now components could receive transient updates and also disconnect on unmount via useEffect:

```jsx
function Person({ id }) {
  useEffect(() =>
    api.subscribe(state => /*do something*/, state => state.persons[id]),
    [id]
  )
```

@JeremyRH what's your opinion? And if you're fine with it, could you help getting this in? I'm still learning TS and i'm getting some typing issues, i had to set the internal listeners array to "StateListener-any" but it looked good so far:

```jsx
  const listeners: Set<StateListener<any>> = new Set()

  const setState = (partialState: PartialState<State>) => {
    state = Object.assign(
      {},
      state,
      typeof partialState === 'function' ? partialState(state) : partialState
    )
    listeners.forEach(listener => listener(state))
  }

  const getState = () => state

  function subscribe<U>(
    callback: StateListener<U>,
    selector?: StateSelector<State, U>
  ) {
    let listener = callback
    if (selector) {
      let stateSlice = selector(state)
      listener = () => {
        const selectedSlice = selector(state)
        if (!shallowEqual(stateSlice, (stateSlice = selectedSlice)))
          callback(stateSlice)
      }
    }
    listeners.add(listener)
    return () => void listeners.delete(listener)
  }
```

JeremyRH:
This is great! It would be good to test the performance when React is not in development mode. I was seeing comparable speeds with the production version of React.

But this is definitely worth exploring. I was just prototyping something that involved updating the store without calling React's `dispatch` (not performance related). It seems like we should enable full features for reading and writing state outside of React. We might need to make some of the types less strict. I'll look into this today.

JeremyRH:
[...]

PR #25, opened 2019-06-01
Add selector to api.subscribe

### LINKED ISSUE #24: api.subscribe calls every subscriber even if state doesn't change
Adding an external subscriber using `api.subscribe` calls the subscriber on every `setState` call even if the state does not change. @drcmda Is this expected behavior? Fixing it could cause regressions if someone relies on it.

JeremyRH:
Fixes #24 and adds feature #23.

This was a bit more painful than I thought it would be. Expected changes were made to `setState` and `subscribe`. Unexpected changes were made to `useStore`. For performance reasons, I decided to not use `getState` as a default selector if none is passed in. If I didn't do this, a `shallowEqual` check would be performed on the entire store for components that use `useStore` without a selector. This led to a cascade of required changes that should be tested more thoroughly.

Additionally, I removed the `state = {}` in `destroy` due to the fact that it invisibly breaks the store when using TypeScript (Type is still `TState` but value is `{}`). External references to the store or objects in the store still won't be garbage collected if they can be reached so I'm not sure if setting the store to an empty object helps much but please correct me if I'm wrong.

Documentation also needs to be added. Just a reminder, the selector argument goes first:
```js
api.subscribe(state => state.foo, foo => console.log(foo))
```

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/esm/index.js": {
-    "bundled": 2627,
-    "minified": 963,
-    "gzipped": 506,
+    "bundled": 3293,
+    "minified": 1084,
+    "gzipped": 566,
     "treeshaked": {
       "rollup": {
         "code": 14,
@@ -14,8 +14,8 @@
     }
   },
   "dist/cjs/index.js": {
-    "bundled": 2875,
-    "minified": 1082,
-    "gzipped": 528
+    "bundled": 3698,
+    "minified": 1230,
+    "gzipped": 590
   }
 }
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/esm/index.js": {
-    "bundled": 3293,
-    "minified": 1084,
-    "gzipped": 566,
+    "bundled": 3092,
+    "minified": 1031,
+    "gzipped": 538,
     "treeshaked": {
       "rollup": {
         "code": 14,
@@ -14,8 +14,8 @@
     }
   },
   "dist/cjs/index.js": {
-    "bundled": 3698,
-    "minified": 1230,
-    "gzipped": 590
+    "bundled": 3425,
+    "minified": 1149,
+    "gzipped": 556
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -196,14 +196,16 @@ You can use it with or without React out of the box.
 const [, api] = create({ n: 0 })
 
 // Getting fresh state
-const n = api.getState().n
+const num = api.getState().n
 // Listening to changes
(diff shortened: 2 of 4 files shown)


drcmda:
Beautiful! This will make edge case so much faster. Should we merge?

JeremyRH:
The concern that a user will do this: `set(state => ({ ...state, something: 123 }))` and hurt performance is valid. Also using `immer` would break this optimization because an entirely new copy of the object is set.

This problem exists with the current version and it could actually be much worse. The current version does a shallow equality check inside every listener that is attached from `useStore`. So two components using `userStore()` without a selector will cause two shallow equality checks across the whole store every time `setState` is called.

JeremyRH:
I just realized this update breaks your demo: https://codesandbox.io/s/amazing-hofstadter-e38rw

The issue is in the `turn` method. You make a copy of state but none of the enumerable properties change. At least one of the properties needs to not be equal for the state to change. This is how shallow equality checks work and it wasn't being used for `api.subscribe` or `setState` before.

We had inconsistencies between `useStore` and `api.subscribe` and unifying some of the behaviors is a breaking change.

drcmda:
Oh, correct, it mutates the state, didn't see that. Fixed the demo.

I found this bit in the redux docs, i guess they're avoiding a shallow-root-check for the same reason:

> Just saw this in the Redux docs:

subscribe(listener)

Adds a change listener. It will be called any time an action is dispatched, and some part of the state tree may potentially have changed. You may then call getState() to read the current state tree inside the callback.

> This problem exists with the current version and it could actually be much worse. The current version does a shallow equality check inside every listener that is attached from useStore. So two components using useStore() without a selector will cause two shallow equality checks across the whole store every time setState is called.

If no selector is given, can we just avoid the check alltogether and fire?

JeremyRH:
Yeah, this sounds good to me. No selector, no shallow equal check. I think we should still do an identity check though.

JeremyRH:
I think this is a minor version change: `0.1.3` -> `0.2.0`. These are the potential breaking changes:
- No shallow equal check on `useStore()` with no selector.
- Generic type `T` must now extend `State`.
- Selector called twice instead of once on initialization. Once in `useReducer` (3rd argument) and once in subscribe to get initial state slice.
- Identity check in `setState` before calling all listeners.
- `destroy` doesn't set state to `{}`.

All of these are pretty minor in my opinion.

One last thing, we should update the readme for `api.subscribe` and possibly suggest the selector function be pure.

drcmda:
Sounds good! Sure, we could mention selector purity. 👍

JeremyRH:
@drcmda I've published this as v0.2.0 on npm. I hope that was okay.

drcmda:
sure, thanks a lot!

PR #27, opened 2019-06-03
Use the state generic in SetState and GetState types

Magellol:
After upgrading to `0.2.0`, I've noticed a few type issues on our end.
It seems like we needed to use the generic in `SetState` and `GetState`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,7 +28,7 @@ export interface StoreApi<T> {
 const reducer = <T>(state: any, newState: T) => newState
 
 export default function create<TState extends State>(
-  createState: (set: SetState<State>, get: GetState<State>) => TState
+  createState: (set: SetState<TState>, get: GetState<TState>) => TState
 ): [UseStore<TState>, StoreApi<TState>] {
   const listeners: Set<StateListener<TState>> = new Set()
 


drcmda:
Let's ping @JeremyRH

JeremyRH:
This will break inferring the `TState` type. It will default to `Record<string, any>` and non TypeScript users will lose autocompletion on `useStore`, `api.getState`, and `api.subscribe`. How did this break your types? Can you provide an example?

Magellol:
Sounds good.

I've tried to replicate it in a code sandbox with minimal code. https://codesandbox.io/s/zustandimmertypeissue020-02e5r

You'll notice that TS isn't happy in the `create()` call. Passing the `TState` to `SetState` and `GetState` fixed it but I can see how it could break autocompletion for non TS users

JeremyRH:
Yeah, setting the generic manually on `create<MyState>` is no longer supported. You have to define the types of `get` and `set` manually.

I've tried to get it to work with your example and can't.

Magellol:
Okay, fair enough!
To anyone running into this, this is how I solved it

```ts
import create, {SetState} from 'zustand';

const [useStore] = create<AppState>((set: SetState<AppState>) => ({...})
```

I'm guessing we can close this then :). Thank you.

JeremyRH:
@Magellol Does changing `create<AppState>(` to `create(` cause any issues? I'm hoping TypeScript can infer that generic. The only things I can't get it to infer are `set` and `get`.

Magellol:
@JeremyRH Not setting it explicitely does not provide type safety on the shape of the store

```ts
interface AppState {
  count: number;
  actions: {
    dispatch(action: Action): void;
    add(n: number): void;
    setCount(n: number): void;
  };
}

const [useStore] = create((set: SetState<AppState>) => ({
  count: 0,
  actions: {
    dispatch(action) {
      set(state => reducer(state, action));
    },
    add(n) {
      set(state => ({ count: state.count + n }));
    }
    // setCount(n) {
    //   set({ count: n });
    // }
  }
}));
```

This does not generate a type issue but it technically should.
We're also losing type safety on the members of the state 

image

JeremyRH:
What version of TypeScript are you using? I'm not seeing the same result.

Screen Shot 2019-06-04 at 8 43 22 AM

Magellol:
It's the one from the sandbox, I've updated the example https://codesandbox.io/s/zustandimmertypeissue020-02e5r
I'm guessing it's the latest one? The type issue in on the `add` method for instance.

I personally think this should not be allowed by TS but somehow is if we don't explicitely pass the State generic to `create`.
```ts
interface AppState {
  count: number;
}

const [useStore] = create((set: SetState<AppState>) => ({}));
```

JeremyRH:
Yeah I see. We should support setting the generic in this case because the inferred type is not enough. I will keep looking into this. Thanks for the info.

JeremyRH:
I think I've run up against limitations in TypeScript and can't get inference to work properly on `get` and `set`. I might try again someday but I'm going to close this PR for now.

PR #34, opened 2019-06-22
Replace useLayoutEffect to useEffect in server side to fix ssr issue

codemilli:
Replace useLayoutEffect to useEffect in server side not to print below error message.

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.

```

https://github.com/react-spring/zustand/issues/11

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/esm/index.js": {
-    "bundled": 3159,
-    "minified": 1031,
-    "gzipped": 538,
+    "bundled": 3181,
+    "minified": 1037,
+    "gzipped": 542,
     "treeshaked": {
       "rollup": {
         "code": 14,
@@ -14,8 +14,8 @@
     }
   },
   "dist/cjs/index.js": {
-    "bundled": 3492,
-    "minified": 1149,
-    "gzipped": 556
+    "bundled": 3514,
+    "minified": 1155,
+    "gzipped": 561
   }
 }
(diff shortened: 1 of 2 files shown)


a-eid:
not sure why `useLayoutEffect` was used here instead of `useEffect`. 
but I'm wondering how does this affect `react-native` apps ? @codemilli

codemilli:
@a-eid `useLayoutEffect` is called right after dome updated (after commit phase), and then `useEffect` is called after very short timeout. So I think the usually we want to run the logic in effect after DOM updated that's why `useLayoutEffect` is used. Could you verify this @drcmda ?

codemilli:
@a-eid and actually I don't have enough knowledge to answer how react-native handle `useLayoutEffect`, but I think there is no big difference.

a-eid:
I'm just wondering why you had to replace it here, `useEffect` is recommended in almost all cases.

codemilli:
Actaully this commit does not replace `useEffect` with `useLayoutEffect`, it replaces `useLayoutEffect` with `useEffect` because in server side (Node.js), because `useLayoutEffect` do nothing. 
you can find more here https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

a-eid:
@codemilli I'm sorry, I must have read it wrong.

codemilli:
@a-eid No worries 😉

PR #35, opened 2019-06-22
Make set & get types use generic state

### LINKED ISSUE #32: SetState & GetState Typescript declaration in create function arguments
The `create` function typescript definition has a generic parameter that helps to define the final state that will be used : `TState extends State`. This state has to extend `State` (`Record<string, any>`) which is relevant for  a state.

Unfortunately both `set` and `get` parameters types don't use this generic parameter and instead are bound to `State` as we can see :
```  createState: (set: SetState<State>, get: GetState<State>, api: any) => TState```

It is not very anoying for the `set` parameter, but it is really a pity that when one uses the `get` method, the definition does not simply type the return to `TState` instead of the very neutral `State`.

Is it just a mistake ? Or is there a reason not to have used `TState` in `set: SetState<State>` and `get: GetState<State>` ? (I must admit that I would expect `set: SetState<TState>` and `get: GetState<TState>` instead).

I've created a (trivial) pull request : https://github.com/react-spring/zustand/pull/31

### PR DESCRIPTION
**Pull request by JeremyRH**
Hopefully this fixes #32 and still allows type inference in JS files.
@jfbrazeau @drcmda  @Magellol @dm385

### CODE REVIEW

**Comment by drcmda** [comment]
I'm not knowledgable about TS to review this, you're the pro 🙂 but if it solves the issue, let's merge.

File: src/index.ts [hunk 1]
```diff
@@ -29,9 +26,13 @@ export interface StoreApi<T> {
 }
 
 const reducer = <T>(state: any, newState: T) => newState
+const useIsomorphicLayoutEffect =
+  typeof window !== 'undefined' ? useLayoutEffect : useEffect
 
 export default function create<TState extends State>(
-  createState: (set: SetState<State>, get: GetState<State>, api: any) => TState
+  createState: keyof TState extends never
```
Magellol: Just to understand, is it to detect if we're in JS vs TS land?

I just tested it and it does fix the type inference issue in TS. We still have to pass the state generic explicitely to `create` to not get implicit anys but I think that's a good DX behaviour.

JeremyRH: > Just to understand, is it to detect if we're in JS vs TS land?

Essentially yeah. It checks to see if the state generic was passed in or not.

jfbrazeau: Hi,
The fix works fine with a very basic sample like this one :
Screen Shot 2019-07-03 at 17 55 54

But I'm afraid that th fix is not enough because it is very easy to get a non working case. Imagine you have several stores, and imagine you want to do something generic on them. You will create a function that performs the "generic" things and then call that function for each store you have.

This will lead you to create a generic function that encapsulates the zustand `create` function invokation. But as far as I can see, it doesn't work.

Just take a look at this really trivial sample in which we try to do something generic during store intialization : 
Screen Shot 2019-07-03 at 17 56 45

I hope you will agree with me that this sample code is trivial (10 lines of effective code), and it highlights the problem...

In other words, as soon as you want to add even a thin layer over zustand for your own purpose in Typescript world, you meet that problem.

So on my side, I still have the same issue as the one I raised in issue #32.

jfbrazeau: I have put all this (trivial) code on codesandbox : https://codesandbox.io/s/j2nys
(`basic.ts` and `initStore.ts` files)

Magellol: You're right. I've tested without passing the generic `TState` from `initStore` and it works.

```ts
interface State {
  foo: string;
}

export function initStore(initialState: State) {
  return create<State>((set: SetState<State>, get: GetState<State>) => {
    return {
      foo: "bar"
    };
  });
}
```

But Typescript complain as soon as we pass a generic down the line:

```ts
export function initStore<S>(initialState: S) {
  return create<S>((set: SetState<S>, get: GetState<S>) => {
    return {};
  });
}
```

It looks like the conditional type isn't kicking in correctly when a generic is  used.

PR #37, opened 2019-06-24
try/catch selectors in order to prevent zombie children

drcmda:
See: https://react-redux.js.org/next/api/hooks#stale-props-and-zombie-children
And reduxes reference impl: https://github.com/reduxjs/react-redux/blob/master/src/hooks/useSelector.js#L84-L103

Without this, taking away state on which certain selector could rely on forces them to always check for the existence of state, independently of the component structure. 

This is b/c state reconciling with hooks falls out of sync with component reconciling (where each component is executed in order).

I don't know about the performance impact of this. Though seeing that redux has merged this, it seems to me they must've profiled and accepted it for what it is.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -108,7 +108,13 @@ export default function create<TState extends State>(
         ? subscribe(
             // Truthy check because it might be possible to set selectorRef to
             // undefined and call this subscriber before it resubscribes
-            () => (selectorRef.current ? selectorRef.current(state) : state),
+            () => {
+              try {
+                // See: https://react-redux.js.org/next/api/hooks#stale-props-and-zombie-children
+                // And reduxes reference impl: https://github.com/reduxjs/react-redux/blob/master/src/hooks/useSelector.js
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -108,7 +108,13 @@ export default function create<TState extends State>(
         ? subscribe(
             // Truthy check because it might be possible to set selectorRef to
             // undefined and call this subscriber before it resubscribes
-            () => (selectorRef.current ? selectorRef.current(state) : state),
+            () => {
+              try {
+                // See: https://react-redux.js.org/next/api/hooks#stale-props-and-zombie-children
+                // And reduxes reference impl: https://github.com/reduxjs/react-redux/blob/master/src/hooks/useSelector.js
+                return selectorRef.current ? selectorRef.current(state) : state
+              } catch (e) {}
```
JeremyRH: You can use a catch block without the `e` param.
https://github.com/tc39/proposal-optional-catch-binding

drcmda: makes sense!

PR #40, opened 2019-06-28
Add IIFE build

nihakue:
Hey! I have a need to drop zustand into a tampermonkey script, but there was no UMD build. I'm not a huge fan of UMD and it seems like you're trying really hard to keep the bytes down so I went with IIFE.

I added 'create' to the exports because `zustand.default(set =>...)` is ick. It increases your bundle sizes, though. I don't mind either way.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/index.js": {
-    "bundled": 2902,
-    "minified": 1064,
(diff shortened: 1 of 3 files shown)


drcmda:
no problem, could you rebase please?

nihakue:
Done. Thanks for the project 👍

drcmda:
its done, released as 1.0.0 :)

Issue #45, opened 2019-07-03
SetState & GetState Typescript declaration in create function arguments (bis)


jfbrazeau:
I opened an issue #32 a few weeks ago. The https://github.com/react-spring/zustand/pull/35 pull request was supposed to fix the problem, but as far as I can see, it doesn't.

jfbrazeau:
The fix works fine with a very basic sample like this one :
![Screen Shot 2019-07-03 at 17 55 54](https://user-images.githubusercontent.com/1252966/60607410-7e9b5680-9dbd-11e9-9b11-31f0536a6c31.png)

But I'm afraid that th fix is not enough because it is very easy to get a non working case. Imagine you have several stores, and imagine you want to do something generic on them. You will create a function that performs the "generic" things and then call that function for each store you have.

This will lead you to create a generic function that encapsulates the zustand create function invokation. But as far as I can see, it doesn't work.

Just take a look at this really trivial sample in which we try to do something generic during store intialization :
![Screen Shot 2019-07-03 at 17 56 45](https://user-images.githubusercontent.com/1252966/60607429-865afb00-9dbd-11e9-8b5f-6ab69c760c59.png)

I hope you will agree with me that this sample code is trivial (10 lines of effective code), and it highlights the problem...

In other words, as soon as you want to add even a thin layer over zustand for your own purpose in Typescript world, you meet that problem.

I have put all this (trivial) code on codesandbox : https://codesandbox.io/s/j2nys
(basic.ts and initStore.ts files)

jfbrazeau:
Another way to present the problem is that today, it is not possible to extend zustand using typescript. Because anybody that would like to extend zustand will quickly have to create a generic function that encapsulates zustand `create`function.

And the `keyof TState extends never` part of the typescript definition prevents from being able to do that:
```ts
export default function create<TState extends State>(createState: keyof TState extends never ? StateCreator<State> : StateCreator<TState>): [UseStore<TState>, StoreApi<TState>];
```

JeremyRH:
I can't find a way to allow extending `create` and keep inference working when not explicitly passing the type of `State`. I'll just allow inference to fail instead of trying to jump through hoops to get it working.

jfbrazeau:
https://github.com/react-spring/zustand/pull/49 fixes this issue a far as I can see. I close the current issue. Thanks a lot.

PR #49, opened 2019-07-20
Remove manual check of passed State generic

JeremyRH:
Should solve #45. We lose inference when not passing in the state though (defaults to `Record<string | number | symbol, any>`).

@jfbrazeau


jfbrazeau:
@JeremyRH : I've updated my zustand dependency to have a check.
As far as I can see, it solves #45 (which can thus be closed).
Thanks a lot.

PR #53, opened 2019-08-15
Incorrect display of messages in devtools redux

SerZubkov:
Incorrect display of messages in devtools redux.

Снимок экрана 2019-08-15 в 15 07 12


drcmda:
good catch, its fixed

PR #56, opened 2019-08-21
Support batched updates

paulshen:
React may sometimes batch updates (e.g. event listeners, maybe more in future). The current implementation of `forceUpdate` does not work when an even number of updates are batched. The added test fails on master.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,6 +1,6 @@
 {
   "dist/index.js": {
-    "bundled": 3082,
+    "bundled": 3081,
     "minified": 1283,
     "gzipped": 668,
     "treeshaked": {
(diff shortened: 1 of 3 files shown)


drcmda:
this is awesome, thanks! @JeremyRH im on vacation right now but you had access to npm to publish patch versions, correct?

JeremyRH:
Yes. I've published 1.0.5.

JeremyRH:
Oops. @drcmda I was not aware you changed the way Zustand is published and I botched 1.0.5. I quickly fixed it in 1.0.6 though 😅

amadeus:
@timkindberg it's been a long time since you asked this - but the answer to your question if you were still wondering - if a store updated more than once before re-rendering, the boolean would be inverted back into it's original value, and then the component would not update: (if it was `true` - an update would come in, inverting it to `false`, but then another update would come in, re-inverting it to `true` again), then react would not update the component since the state did not `change` since last render

PR #58, opened 2019-08-23
Remove optional catch binding

Magellol:
Not sure if you agree to this change.

Optional bindings are currently stage 4 so they will land in the next release, although I've been having issues when running tests with jest since our node version is below 10.

There is also the option of us transpiling zustand prior to running tests.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -24,9 +24,9 @@
     "gzipped": 686
   },
   "dist/middleware.js": {
-    "bundled": 1482,
-    "minified": 759,
-    "gzipped": 425,
+    "bundled": 1491,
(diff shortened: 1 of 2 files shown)


JeremyRH:
So this issue can probably be solved on your end by adding `@babel/plugin-proposal-optional-catch-binding` to your babel config. I assume you're already using babel to transform non-standard language features such as JSX, so it might be as simple as adding this babel plugin. If your testing configuration is different and/or you aren't able to add this babel plugin, I am fine with this change. 

Let me know what your situation is and we can go from there.

drcmda:
Node and testing workout hassle is a good argument imo, I would also approve.

PR #59, opened 2019-09-07
Adjust type inference for store without a selector

dichuvichkin:
This improvement will help with type inference when using `useStore` without selectors.
For example
```
import create from 'zustand'

interface IState {
  count: number;
  increase: () => void;
  reset: () => void;
}

const [useStore] = create<IState>(set => ({
  count: 0,
  increase: () => set(state => ({ count: state.count + 1 })),
  reset: () => set({ count: 0 })
}));

function Counter() {
  const { count, increase, reset } = useStore()
  
  // typeof count === 'number' => true
  // typeof increase === 'function' => true
  // typeof reset === 'function' => true

  return <h1>{count}</h1>
}

```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,6 +20,7 @@ export type StateCreator<T extends State> = (
 ) => T
 export type SetState<T extends State> = (partial: PartialState<T>) => void
 export type GetState<T extends State> = () => T
+export type ReturnState<T extends State, U> = U extends T ? T : U
 export type Subscribe<T extends State> = <U>(
   listener: StateListener<U | void>,
   options?: SubscribeOptions<T, U>
@@ -28,7 +29,7 @@ export type Destroy = () => void
 export type UseStore<T extends State> = <U>(
   selector?: StateSelector<T, U>,
   equalityFn?: EqualityChecker<U>
-) => U
+) => ReturnState<T, U>
 export interface StoreApi<T extends State> {
   setState: SetState<T>
   getState: GetState<T>
@@ -91,7 +92,7 @@ export default function create<TState extends State>(
   const useStore = <StateSlice>(
     selector: StateSelector<TState, StateSlice> = getState,
     equalityFn: EqualityChecker<StateSlice> = Object.is
-  ): StateSlice => {
+  ): ReturnState<TState, StateSlice> => {
     if (Array.isArray(equalityFn)) {
       equalityFn = Object.is
       console.warn(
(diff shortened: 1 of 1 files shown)


JeremyRH:
Thanks for the fix. I noticed this works when using destructuring assignment:
```typescript
const { count, increase, reset } = useStore() // all good
```
but not when using property accessors:
```typescript
const state = useStore()
state.count // error!
```
I think a better change would be to change the UseStore type to an interface with two call signatures:
```typescript
export interface UseStore<T extends State> {
  (): T
}
```
I'll add comments to your changes. Let me know what you think.

src/index.ts
```diff
@@ -91,7 +92,7 @@ export default function create<TState extends State>(
   const useStore = <StateSlice>(
     selector: StateSelector<TState, StateSlice> = getState,
     equalityFn: EqualityChecker<StateSlice> = Object.is
-  ): StateSlice => {
+  ): ReturnState<TState, StateSlice> => {
```
JeremyRH: The `: {TYPE}` can be removed to let TypeScript infer the return type.

src/index.ts
```diff
@@ -126,7 +127,7 @@ export default function create<TState extends State>(
     const forceUpdate = useReducer(forceUpdateReducer, 1)[1]
     useIsoLayoutEffect(() => subscribe(forceUpdate, options), [])
 
-    return options.currentSlice as StateSlice
+    return options.currentSlice as ReturnState<TState, StateSlice>
```
JeremyRH: The `as {TYPE}` can be removed because all possible return types would be covered in the interface.

src/index.ts
```diff
@@ -28,7 +29,7 @@ export type Destroy = () => void
 export type UseStore<T extends State> = <U>(
```
JeremyRH: Replace UseStore with this:
```typescript
export interface UseStore<T extends State> {
  (): T
}
```

src/index.ts
```diff
@@ -20,6 +20,7 @@ export type StateCreator<T extends State> = (
 ) => T
 export type SetState<T extends State> = (partial: PartialState<T>) => void
 export type GetState<T extends State> = () => T
+export type ReturnState<T extends State, U> = U extends T ? T : U
```
JeremyRH: No need for `ReturnState` because the return value for `useStore()` and `useStore(selector)` would be declared in the interface.

dichuvichkin:
> I'll add comments to your changes. Let me know what you think.
Thanks, I think your solution is better. I replaced UseStore type by interface you mentioned above. 
Checked, it works well

src/index.ts
dichuvichkin: Done

src/index.ts
dichuvichkin: Replaced

src/index.ts
dichuvichkin: Done

src/index.ts
dichuvichkin: Done

PR #66, opened 2019-10-11
2.1.x // Type changes and perf/bug fixes.

JeremyRH:
I uncovered a couple issues that lead to these changes. Sparse arrays (arrays with holes) seem to have poor performance when their length grows too large (even if they are empty). I wasn't limiting the subscriber counter so it would potentially assign a large index to the subscribers array and cause the `forEach` loop in `setState` to have performance issues. I fixed that by consolidating the subscribers array during `setState`. This minimally impacts performance for every call to `setState` and greatly improves performance in applications that are long running and/or have many renders per second.

The changes lead to a small refactor and unfortunately, some types changed. Good news is these types are used for internal functionality so our public API hasn't changed.

I also found an issue with concurrent mode. The state slice was being updated during the render phase when a new selector was passed in. If the render is interrupted after the slice updates, the next `setState` call won't update the component in some cases.

Anyway, here is the change list.

- Removed SubscribeOptions and UseStoreSubscribeOptions types.
- Added Subscriber type.
- Improve performance by consolidating subscribers sparse array.
- Fix potential issue with concurrent mode.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/index.js": {
-    "bundled": 3774,
-    "minified": 1232,
-    "gzipped": 589,
+    "bundled": 4474,
+    "minified": 1116,
+    "gzipped": 579,
     "treeshaked": {
       "rollup": {
         "code": 14,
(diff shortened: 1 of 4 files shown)


drcmda:
how did you test concurrent mode? would love to try this out myself. the changes look fine to me, if you like, publish a release.

JeremyRH:
I just used `ReactDOM.unstable_createRoot(document.getElementById("root")).render(<App />);`
Read more here: https://github.com/sw-yx/fresh-concurrent-react/blob/master/apis/roots.md

PR #78, opened 2019-11-06
replace useBuerli with useStore in readme

johnrees:
Not sure if this was a typo or intended? I couldn't find any reference to useBuerli elsewhere in the project


drcmda:
oh my, yes, what an oversight :-/ thanks!

PR #80, opened 2019-11-28
Fix type inference of TypeScript for `shallow` by removing `any`

Cryrivers:
As the title says.

`useStore` also infers the return type via `EqualityChecker<U>`, so the current types of `shallow` would always make TypeScript infer the return type as `any`.

This PR adds more constraints to `obj1` and `obj2` in `shallow`.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -29,8 +29,8 @@
     "gzipped": 256
   },
   "dist/middleware.js": {
-    "bundled": 2126,
-    "minified": 1205,
-    "gzipped": 621
+    "bundled": 2127,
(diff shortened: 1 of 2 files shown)


Cryrivers:
Thanks for the review. By default TypeScript will type all non-inferable types as `{}` or `unknown` (depending on TypeScript versions) instead of `any`. So it would complain `Object.is(objA[keysA[i]], objB[keysA[i]])` because of no index signatures defined.

`extends any` would make the inferance as wide as possible, thus no above-mentioned error.

JeremyRH:
Oh nice, I didn't realize that. Thanks again!

Issue #86, opened 2020-01-28
useStore stops firing on state changes after component remounts


karlpetersson:
When switching between mounting two components that both calls the useStore hook, where one of them conditionally mounts a child that also calls useStore, the parent will stop rendering on state changes after remounting. Here's a minimal example on codesandbox: 

https://codesandbox.io/s/nostalgic-hooks-0x63r

Repro steps are included. Let me know if you want the code pasted here instead.

Could this be related to #85 ?

edit: updated codesandbox link, messed around in the original one I posted without realizing it.

karlpetersson:
Downgrading to 2.2.1 seems to fix this issue. Tested by downgrading the zustand package in the above codesandbox.

karlpetersson:
Might be a dumb question, but why not make the array dense right after adding the subscriber? This seems to solve this issue as well as work with #84, while keeping indices intact. E.g:

```
const subscribe: Subscribe<TState> = <StateSlice>(
    subscriber: Subscriber<TState, StateSlice>
  ) => {
    // ...
    subscribers[subscriber.index] = subscriber;

    subscriberCount = 0;
    subscribers = subscribers.filter(s => {
      subscriber.index = subscriberCount++;
      return true;
    });

    return () => delete subscribers[subscriber.index];
  };
```

Jamikk:
I have the same problem

fzkirablackwhy:
+

fzkirablackwhy:
me too

lolk12:
me too! HELP ZUSTAND

dentrado:
Any news on this? Seems to still be a problem

lolk12:
@dentrado On version 2.2.1, everything is fine

dentrado:
But then I get the bug they fixed in 2.2.2 instead

lolk12:
@dentrado very sadly (

PR #90, opened 2020-02-19
Change so subscribers are added in render phase

karlpetersson:
This fixes an issue with subscriber indices being overriden in some cases, as seen in issue #86. In this scenario, child components will unmount between the render phase of the parent and its layout effect triggering. This causes indices to become invalid since the array is compressed as a result of a component unsubscribing (when it unmounts). 

With this, instead of only assigning the index in the render phase, the subscriber is also added to the array of subscribers, so that there are no pending indices when subsequent unsubscribe functions are called.

This also handles the case in #84.

I'm not sure what the reason was for assigning an index and adding the subscriber in different phases, so please let me know if I missed something.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/index.js": {
-    "bundled": 4021,
-    "minified": 1116,
-    "gzipped": 581,
+    "bundled": 4156,
+    "minified": 1158,
+    "gzipped": 594,
     "treeshaked": {
       "rollup": {
         "code": 14,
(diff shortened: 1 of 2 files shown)


drcmda:
Thanks @karlpetersson ! @JeremyRH can we approve it? These tearing issues have gotten so complex i hardly understand what's going on. ☺️

JeremyRH:
Subscribing during the render phase in not safe in concurrent mode. A component can be called during the render phase but React can bail out before committing it to the DOM. It might never be committed and useEffect/useLayoutEffect will never run and the component won't unsubscribe.

JeremyRH:
Concurrent mode support requires other fixes so maybe we shouldn't worry about it now.

If we don't care about concurrent mode, we can remove the array compression and use a Set() object.

drcmda:
I think Set() makes sense, as long as concurrent worries are still theoretic. Are real solutions even possible?

JeremyRH:
The proposed `useMutableSource` hook solves the tearing issues and provides a safe subscribe/unsubscribe interface. I think we should do some refactoring to make the code more simple and drop whatever partial support for concurrent mode that we have.

JeremyRH:
I submitted a fix in #91 and published it as v2.2.3. Thank you @karlpetersson and I apologize for not working with your PR.

PR #91, opened 2020-02-20
Fix subscribers being overwritten between render and effect

### LINKED ISSUE #86: useStore stops firing on state changes after component remounts

JeremyRH:
Should fix #86 
@karlpetersson Sorry for submitting a separate PR instead of working off yours. I wanted to simplify the logic if we plan on removing the attempt to support concurrent mode. I also added a test.

Thank you @karlpetersson for the help and you did find a fix which I used here.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/index.js": {
-    "bundled": 4021,
-    "minified": 1116,
-    "gzipped": 581,
+    "bundled": 3173,
+    "minified": 1109,
+    "gzipped": 558,
     "treeshaked": {
       "rollup": {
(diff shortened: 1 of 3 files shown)


karlpetersson:
@JeremyRH Thanks! No worries, glad to be of help.

PR #99, opened 2020-04-07
Fix typing of PartialState<T> for immer usage

### LINKED ISSUE #96: type error when using with immer
now the signature of set accepts: `export declare type PartialState<T extends State> = Partial<T> | ((state: T) => Partial<T>);` but with immer it requires `(state: T) => void`, or it cannot pass the type checking.

The complete type should be: `export declare type PartialState<T extends State> = Partial<T> | ((state: T) => Partial<T>) | ((state:T) => void);`

clawoflight:
This fixes #96.


drcmda:
made a path release. thank you! :-)

PR #110, opened 2020-05-14
Added merge flag when using setState to override

richardgorman:


drcmda:
hi @rgorman it looks like @JeremyRH is busy, i think we need to find more people willing to maintain or look over it
the problem for me is that the typescript annotations have become so complex im basically not able to make changes
i think jeremy wanted to change the behaviour altogether, similar to hooks setstate, override instead of merge
people that want to merge then do `set(state => ({ ...state, foo: 1 })`
another thing we wanted to do is assign the api to the hook:

```jsx
const useState = create(...)

useState.getState()
useState.subscribe()

function Foo() {
  const bar = useState(state => state.bar)
```

it would be cleaner than const `[useState, api] = create(...)` and at the same time the api would always be available
if you know your way around TS, do we want to attempt a new major with these changes?

drcmda:
another question i have is, if don't merge, it will wipe out everything, including the actions. are you aware of that? i've tried out these changes here: https://codesandbox.io/s/mystifying-clarke-e4em2?file=/src/index.tsx

richardgorman:
I did not know that. I can fix it but from what I understand a new upcoming major version will improve the API

drcmda:
its not a bug, its just if you have a store like this

```jsx
  foo: 1,
  setFoo: () => set({ foo: 2 })
})
```

if merge wouldn't be the default, then calling setFoo would delete itself and set foo to 2. so i think we could just do it like you suggested, default is merge, the flag can give you the power to overwrite

dai-shi:
We discussed and the flag should be something like `wipeOut`. It's clearer and you don't need `undefined` check.
@richardgorman Would you like to work on it?

dai-shi:
#137 is merged.

PR #117, opened 2020-06-23
fix: devtools not showing actions

MillerGregor:
using devtools did now show my actions as they were dispatched.

While debugging, I found that issuing a manual devtools.send() does something in the redux-devtools-extension to complete the init process, allowing all subsequent actions to be logged.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -29,8 +29,8 @@
     "gzipped": 256
   },
   "dist/middleware.js": {
-    "bundled": 2127,
-    "minified": 1206,
-    "gzipped": 622
+    "bundled": 2187,
+    "minified": 1245,
+    "gzipped": 639
   }
 }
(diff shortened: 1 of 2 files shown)


dkgrieshammer:
For me the following hint made it work; the trick is to add names to the functions in store
https://github.com/react-spring/zustand/issues/77#issuecomment-551435696

MillerGregor:
> For me the following hint made it work; the trick is to add names to the functions in store
Adding names was not enough for me... I didn't see all my actions until I included the "send" in this PR

olavoparno:
Hey, I think this PR is not required as it works without it. There are other problems regarding to logging actions but this is not one. Try opening an issue with your code hosted at codesandbox or something.

MillerGregor:
> There are other problems regarding to logging actions but this is not one

You're right, I got mixed up.  closing

PR #125, opened 2020-07-14
Updates Immer example in readme to prevent "Invalid left-hand side in…

stolinski:
… assignment expression"

Using the code as is in this example causes Invalid left-hand side in assignment expression because we are setting the value rather than doing an increment or something. I ran across this while trying to use this example and have a codesandbox to show this breaking if needed. Please lmk if I am off base here, but this is the code that's actually working for me.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -135,7 +135,9 @@ const [useStore] = create(set => ({
 }))
 
 const set = useStore(state => state.set)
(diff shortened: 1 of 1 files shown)


drcmda:
Void does work but it needs parentheses, prettier would have added them. It also works only for single lines. Your way is more versatile though.

PR #133, opened 2020-08-11
feat: new implementation with useMutableSource

dai-shi:
This is for next major version.

Some notes:
- It will only work with react@experimental, it may change later, and we are not sure when it will be finalized.
- It's based on useMutableSource, and no longer requires useEffect and useRef.
- The render behavior is slightly changed.
- The behavior with exception is changed. The spec is skipped.
- [IMPORTANT] selectors and equalityFn should be stable. Either make them static or wrap with `useCallback`.
  - However, thanks to scoped subscription (unlike Redux), unstable selectors seem to work.
- On the other hand, it needs to cache the selector result `cachedSlices`, which is a bit hacky.
- `functionMap` is hacky too, but can't be helped.
- Needs someone to review the changes in tests, if they are valid or not.
- Bundle size is increased a bit, unfortunately. It could be shaved a bit once we finalize the api.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,10 @@
 {
   "name": "zustand",
   "private": true,
-  "version": "2.2.4",
+  "version": "3.0.0-alpha.0",
+  "publishConfig": {
+    "tag": "alpha"
+  },
   "description": "🐻 Bear necessities for state management in React",
   "main": "index.cjs.js",
   "module": "index.js",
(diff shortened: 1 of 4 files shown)


dai-shi:
```ts
export type StateListener<T> = (state: T | null, error?: Error) => void
```

We should take care of the case `T` includes `null`.

dai-shi:
v3-alpha branch is created, and this PR is based on it.

dai-shi:
@drcmda @stephencorwin Could you review this?

drcmda:
> [IMPORTANT] selectors and equalityFn should be stable. Either make them static or wrap with useCallback.
However, thanks to scoped subscription (unlike Redux), unstable selectors seem to work.

yep, this rolls a little different because it uses references. we actually had a section for that in the docs but i recently removed it because all it would save is a selector run on render. i'll add back a small description mentioning useCallback

dai-shi:
In v2, having `useCallback` is not important, I don't even know if performance gets good or bad with/out it. In the new version with uMS, it's important because uMS optimizes with reference identity.

dai-shi:
@drcmda Any idea how to proceed this PR? It's not well tested, but unless we publish it as alpha, it wound't be easy to evaluate, would it? Do you have your leaner api code ready?

dai-shi:
Let me merge this and I can add some follow up PRs.

PR #136, opened 2020-08-13
[v3-alpha] clean up types

dai-shi:
useMutableSource version doesn't require subscriber, so this simplifies a lot.
Note, this is technically a braking change in the exported types.
I assume people are not extensively using `ApiSubscribe` type in v2.


dai-shi:
Merging this. Feedback is welcome.

PR #137, opened 2020-08-13
Add replace flag in set

dai-shi:
Follow up #110.


dai-shi:
Forgot to modify the test: a97d1db

PR #139, opened 2020-08-13
[v3] eliminate subscriber + unified api

dai-shi:
- eliminate subscriber so that we can extract vanilla version
- unified api with backward compatible tuple (no type)

Important notes:
- It will call selector once more at subscribe due to caching at the both ends
  - This is a bit unfortunate, but it's more consistent with v4-alpha
- The use of `useCallback` for selector and equalityFn is highly recommended
  - Or define the function outside render
  - It will be optimized with ref equality
  - This is also the case with v4-alpha
  - Without `useCallback`, it will still work, can be slow but not noticable in small apps.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/index.js": {
-    "bundled": 3629,
-    "minified": 1239,
-    "gzipped": 602,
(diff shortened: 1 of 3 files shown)


dai-shi:
@drcmda Would please review this?

Issue #114, opened 2020-06-11
Typescript problem with StateListener


rdhox:
Hey!

I use this block of code:

```js
  export interface NetWorkInfo {
    isInternetOn: boolean;
    type: NetInfoStateType;
    isConnectionExpensive: boolean;
    cellularType: NetInfoCellularGeneration | 'null';
  };

  ...

  useEffect(() => {
    const unsubIsInternetOn = apiApp.subscribe<NetWorkInfo>(
      networkInfo => {
        const { isInternetOn } = networkInfo;
        setIsInternetOn(isInternetOn);
      },
      appState => appState.state.networkInfo
    );

    return () => {
      unsubIsInternetOn();
    };
  }, []);
```

And TS is never happy about the first param of subscribe: 
`Property 'isInternetOn' does not exist on type 'NetWorkInfo | null'.`

Because of the `null` [here](https://github.com/react-spring/zustand/blob/d8f8c5f752465065959b3e6782774b316881f72a/src/index.ts#L13).

Is there a reason why the param1 of StateListener should be null ?

Thanks!

dai-shi:
In case of error, it will be `null`.

For now, you need to check it on your end.
```js
      networkInfo => {
        if (networkInfo === null) return; // or report error
        const { isInternetOn } = networkInfo;
        setIsInternetOn(isInternetOn);
      },
```

cc @drcmda We could revisit this in v3. I think throwing an error is reasonable.

rdhox:
Thanks for your answer.
I find a way around to make TS happy:
```typescript
useEffect(() => {
    const unsubIsInternetOn = apiApp.subscribe<NetWorkInfo>(
      networkInfo => {
        const { isInternetOn } = networkInfo as NetWorkInfo;
        setIsInternetOn(isInternetOn);
      },
      appState => appState.state.networkInfo
    );

    return () => {
      unsubIsInternetOn();
    };
  }, []);
```

dai-shi:
That's fine too as long as you are sure that your selector never throws.

> We could revisit this in v3. I think throwing an error is reasonable.

This was my misunderstanding. We need a way to report error via listener.

Closing this.

PR #141, opened 2020-08-14
[v3] add slim subscribe and refactor useStore

dai-shi:
This will refactor subscribe and clean up code quite a bit.

This is technically a breaking change. Exported types are a bit changed.

----

#139 
> It will call selector once more at subscribe due to caching at the both ends

This hack is gone. See the test spec is reverted.

#114
> We could revisit this in v3

Because, we don't have selector in subscribe. No throwing error. Cleaner listener typing.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/index.js": {
-    "bundled": 3116,
-    "minified": 1132,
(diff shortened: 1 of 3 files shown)


drcmda:
I don't use subscribe so often but isn't this a feature regression? Non react users wouldn't have the ability to select state after this?

PR #145, opened 2020-08-16
Uses prefix as devtools instance name

getreworked:
If a prefix is provided, it will be used as the instance name. If one is not provided, the default devtools behaviour will be used.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -38,7 +38,7 @@ const devtools = (fn: any, prefix?: string) => (
   }
   const initialState = fn(namedSet, get, api)
   if (!api.devtools) {
-    api.devtools = extension.connect()
+    api.devtools = extension.connect({ name: prefix })
     api.devtools.prefix = prefix ? `${prefix} > ` : ''
     api.devtools.subscribe((message: any) => {
       if (message.type === 'DISPATCH' && message.state) {


dai-shi:
@olavoparno can confirm if this fixes your issue?

olavoparno:
@dai-shi I was about to check this PR. It solves another problem, indeed, naming the instances names instead of using the project's name to using the second devtools parameter as described in the docs.

My problem I think is more related to my implementation itself because I wanted to centralize all my stores into one unique devtools instance but I think this is rather impossible with the current API.

Thank you anyway @dai-shi and @getreworked

dai-shi:
@olavoparno Thanks for the confirmation!
@getreworked Thanks for the contribution!

Merging this.

getreworked:
@dai-shi Thanks for the quick turnaround.
@olavoparno I think you are right with your assumption. The API isn't designed to merge separate stores into one. Imo, this is a good thing as these stores should remain modular.

olavoparno:
> @dai-shi Thanks for the quick turnaround.
> @olavoparno I think you are right with your assumption. The API isn't designed to merge separate stores into one. Imo, this is a good thing as these stores should remain modular.

It is I who thank you both. And I think so too about them being modular, this maybe the why I love zustand for starters :+1:

PR #152, opened 2020-08-17
Minor README changes

gsimone:
Changed some minor stuff as I was going through the README


drcmda:
all good except the yarn thing, that's 4 extra lines 😝 a yarn user that doesn't know how to do translate npm install ... does that exist?

gsimone:
Yeah, we can drop that 💯

PR #154, opened 2020-08-17
README typo

Liinkiing:
I don't know if it's really a typo or so (it seems that yes, because I don't know where `foo` and `bar` comes from), so it fix the README comments to be more understandable. As far as I understood the shallow diffing, it seems that the comments are wrong.

Great library by the way, loving the usage and the concise API!


drcmda:
thanks! :-)

PR #157, opened 2020-08-18
[v4-experimental] use mutable source

dai-shi:
Now based on v3, the implementation with useMutableSource is much cleaner.


src/index.ts
```diff
@@ -31,86 +34,53 @@ export default function create<TState extends State>(
   const api: StoreApi<TState> =
     typeof createState === 'function' ? createImpl(createState) : createState
 
+  const source = createMutableSource(api, () => api.getState())
+
+  const FUNCTION_SYNBOL = Symbol()
```
timkindberg, line 39: Misspelled SYMBOL

dai-shi: 😓

PR #160, opened 2020-08-18
[v4-experimental] use mutable source (#157)

drcmda:
* feat: new implementation with useMutableSource

* fix: support slice to be null

* [v3-alpha] clean up types

* shave two bytes

* revert type hack for immer

* update size snapshot

* [v3-alpha] unified api

* shave a byte

* v4-alpha

* update size snapshot

* revert skipped test

* fix test for uMS

* fix typo


dai-shi:
I don't think we should merge this into master while experimental. Converted back to draft.

dai-shi:
A counter example:
https://codesandbox.io/s/react-typescript-forked-vkxbo

dai-shi:
closing in favor of #422

PR #169, opened 2020-08-26
fix typing for immer

### LINKED ISSUE #96: type error when using with immer

dai-shi:
Close #96 

The fix in #99 doesn't seem correct. https://github.com/react-spring/zustand/issues/96#issuecomment-673764960

This would be a better fix. Although this reverts #99, I couldn't get response in #96, so let's publish it and see how it works.

Here's immer recipe in TS without middleware.

```ts
interface State {
  lush: {
    forrest: {
      contains: { a: string } | null
    }
  }
  set: (fn: (s: State) => void) => void
}

const useStore = create<State>((set) => ({
  lush: { forrest: { contains: { a: "bear" } } },
  set: (fn) => set(produce(fn)),
}))

const a = useStore((state) => state.lush.forrest.contains?.a)
const set = useStore((state) => state.set)
set(state => {
  state.lush.forrest.contains = null
})
```

Here's the recipe with middleware.

```ts
connst immer = <T>(
  config: StateCreator<T, (fn: (state: T) => void) => void>
): StateCreator<T> => (set, get, api) => config(
  (fn) => set(produce(fn) as (state: T) => T),
  get,
  api
)

const useStore = create<{
  bees: boolean
  setBees: (b: boolean) => void
}>(
  immer((set) => ({
    bees: false,
    setBees: (input) =>
      set((state) => {
        state.bees = input
      }),
  }))
)
```

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,9 +1,7 @@
 export type State = Record<string | number | symbol, any>
-// (state: T) => void is for immer https://github.com/react-spring/zustand/pull/99
 export type PartialState<T extends State> =
   | Partial<T>
   | ((state: T) => Partial<T>)
-  | ((state: T) => void)
 export type StateSelector<T extends State, U> = (state: T) => U
 export type EqualityChecker<T> = (state: T, newState: any) => boolean
 export type StateListener<T> = (state: T) => void
@@ -28,8 +26,8 @@ export interface StoreApi<T extends State> {
   subscribe: Subscribe<T>
   destroy: Destroy
 }
-export type StateCreator<T extends State> = (
-  set: SetState<T>,
+export type StateCreator<T extends State, CustomSetState = SetState<T>> = (
+  set: CustomSetState,
   get: GetState<T>,
   api: StoreApi<T>
 ) => T


julianklumpers:
to get rid of `Type 'T' does not satisfy the constraint 'Record<string | number | symbol, unknown>'`

You can do the following:

```typescript
const immer = <T extends Record<keyof T, T[keyof T]>>(
  config: StateCreator<T, (fn: (state: T) => void) => void>
): StateCreator<T> => (set, get, api) => config(fn => set(produce(fn) as (state: T) => T), get, api);
```

hope this helps.

dai-shi:
@julianklumpers Hey, thanks for the note!

Yeah, since v3.1.0, the base type is stricter. We can use `State` which is exported.

```ts
  config: StateCreator<T, (fn: (state: T) => void) => void>
  config((fn) => set(produce(fn) as (state: T) => T), get, api);
```

I made the working codesandbox: https://codesandbox.io/s/nifty-gould-ik7dy

julianklumpers:
@dai-shi Cool! didn't know that. thank for pointing that out for me!

ivancuric:
> @julianklumpers Hey, thanks for the note!
> 
> Yeah, since v3.1.0, the base type is stricter. We can use `State` which is exported.
> 
> ```ts
> const immer = <T extends State>(
>   config: StateCreator<T, (fn: (state: T) => void) => void>
> ): StateCreator<T> => (set, get, api) =>
>   config((fn) => set(produce(fn) as (state: T) => T), get, api);
> ```
> 
> I made the working codesandbox: https://codesandbox.io/s/nifty-gould-ik7dy

Looking at this solution, I can't seem to extract the asserted type into an interface (https://codesandbox.io/s/festive-sid-w3q88):
```ts
interface IStore {
  bees: boolean;
  setBees: (b: boolean) => void;
}

const useStore = create<IStore>(
  immer((set) => ({
    bees: false,
    setBees: (input) =>
      set((state) => {
        state.bees = input;
      })
  }))
);
```
```
Type 'IStore' does not satisfy the constraint 'Record<string | number | symbol, unknown>'.
  Index signature is missing in type 'IStore'.ts(2344)
```
```
Parameter 'input' implicitly has an 'any' type.ts(7006)
```

dai-shi:
@ivancuric 
Please check this if you haven't.
https://github.com/pmndrs/zustand#typescript

```ts
interface IStore extends State {
  bees: boolean;
  setBees: (b: boolean) => void;
}
```

dai-shi:
> Also, would it be possible to just expose a generic `set` in this store, like in [the example](https://github.com/pmndrs/zustand#sick-of-reducers-and-changing-nested-state-use-immer), alongside the `X, setX` pairs ?

Not sure if I get the question. Isn't it about https://github.com/pmndrs/zustand#middleware ?
Or something different?

ivancuric:
> @ivancuric
> Please check this if you haven't.
> https://github.com/pmndrs/zustand#typescript
> 
> ```ts
> interface IStore extends State {
>   bees: boolean;
>   setBees: (b: boolean) => void;
> }
> ```
Thanks! Not sure why it wasn't extended already when using
```ts
  config: StateCreator<T, (fn: (state: T) => void) => void>
  config((fn) => set(produce(fn) as (state: T) => T), get, api);
```
There's `<T extends State>` in there already.


> Not sure if I get the question. Isn't it about https://github.com/pmndrs/zustand#middleware ?
> Or something different?

Not sure to be honest. I'm still trying to fully understand the zustand + immer pipeline and composition.

What I'm asking is — with the middleware example I have state + actions in the same object. Let's say:
```ts
const useStore = create<{
  x: boolean;
  setX: (b: boolean) => void;
  y: boolean;
  doSomethingWithY: (b: boolean) => void;
}>(
  immer((set) => ({
    x: false,
    setX: (input) =>
      set((state) => {
        state.x = input;
      }),
    y: false,
    doSomethingWithY: (input) =>
      set((state) => {
        state.x = input;
      }),
  }))
);
```

Would it be possible to expose the `set` function in this `useStore` which is already using the typed `immer` middleware?
So that I could, for instance, do:
```ts
const { set }= useStore()
set(state => void state.x = false)
```
Not sure how much of an antipattern this is.

ivancuric:
And another question that is related to immer typing (I'll make a separate issue/discussion if you think it's better):

How would I be able to extract the state and actions into their own objects and merge them [using `combine`](https://github.com/pmndrs/zustand#user-content-typescript:~:text=Or%2C%20use%20combine%20and%20let%20tsc%20infer%20types.) while using this typed immer middleware pattern?

dai-shi:
> Would it be possible to expose the `set` function in this `useStore` which is already using the typed `immer` middleware?
> So that I could, for instance, do:
> 
> ```ts
> const { set }= useStore()
> set(state => void state.x = false)
> ```

I don't think so. You'd need the bare solution: https://github.com/pmndrs/zustand#sick-of-reducers-and-changing-nested-state-use-immer

dai-shi:
> And another question that is related to immer typing (I'll make a separate issue/discussion if you think it's better):
> 
> How would I be able to extract the state and actions into their own objects and merge them [using `combine`](https://github.com/pmndrs/zustand#user-content-typescript:~:text=Or%2C%20use%20combine%20and%20let%20tsc%20infer%20types.) while using this typed immer middleware pattern?

Typing is hard in some cases. Like #216. Yeah, it might be better to open a dedicated issue.

ivancuric:
I actually found a very similar issue: https://github.com/pmndrs/zustand/issues/211

ivancuric:
Any ideas about this? 

> Thanks! Not sure why it wasn't extended already when using
> 
> ```ts
> const immer = <T extends State>(
>   config: StateCreator<T, (fn: (state: T) => void) => void>
> ): StateCreator<T> => (set, get, api) =>
>   config((fn) => set(produce(fn) as (state: T) => T), get, api);
> ```
> 
> There's `<T extends State>` in there already.

dai-shi:
`<T extends State>` is more or less a "requirement". It says "If you wanna use `T` here, it must extends `State`, on your end".

PR #172, opened 2020-08-31
fix: devtools to patch api.setState

### LINKED ISSUE #171: devtools don't always log
devtools log fine as long as you call `set` (inside the `create` function). However, if you use the `useStore.setState` function to set some state, devtools doesn't log it.

Here is a minimal reproduceable example:

https://codesandbox.io/s/beautiful-ellis-17s7g

The `increase` function logs fine, the `setToFive` function does not.

Judging from the code, I think it is because `namedSet` is kept locally here:

https://github.com/react-spring/zustand/blob/a6417ce460a31cec94adc4a32c55d1abc98a00ba/src/middleware.ts#L54-L63

Currently, the middleware can only return the initialState, but since we want to wrap the setState function, we would somehow need to return / expose that function as well. What do you think?

dai-shi:
Close #171. 

~There are a few technical issues:~
~1. we need to fix types in vanilla.ts and index.ts too.~
~2. we can't chain middleware that only patches `set`.~

~Due to 2), I'm a bit hesitant to this approach.~

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 2381,
-    "minified": 1359,
-    "gzipped": 644
+    "bundled": 2409,
+    "minified": 1355,
+    "gzipped": 641
   },
   "vanilla.js": {
     "bundled": 1990,
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 2168,
-    "minified": 1257,
-    "gzipped": 637
+    "bundled": 2352,
+    "minified": 1347,
+    "gzipped": 642
   },
   "vanilla.js": {
     "bundled": 1990,
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 2352,
-    "minified": 1347,
-    "gzipped": 642
+    "bundled": 2381,
+    "minified": 1359,
+    "gzipped": 644
   },
   "vanilla.js": {
     "bundled": 1990,
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 2088,
-    "minified": 1231,
-    "gzipped": 628
+    "bundled": 2168,
+    "minified": 1257,
+    "gzipped": 637
   },
   "vanilla.js": {
     "bundled": 1990,
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 2409,
-    "minified": 1355,
-    "gzipped": 641
+    "bundled": 2381,
+    "minified": 1359,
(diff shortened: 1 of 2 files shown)


dai-shi:
@TkDodo Here's your example with this change. https://codesandbox.io/s/agitated-stonebraker-bg7bi

dai-shi:
Okay, this f4c6239 should be much better.
https://codesandbox.io/s/wonderful-lovelace-rjego

TkDodo:
Thanks! From what I can tell, the code looks good. I understand that we cannot change the interface of  `set`, so there won't be the possibility to give those "actions" a name, but this is an understandable limitation 👍

dai-shi:
@TkDodo Did you try the sandbox?
It would be nice if you could try and find possible bugs.

> we cannot change the interface of `set`, so there won't be the possibility to give those "actions" a name

By `set`, do you mean the first argument of the creator function, or `api.setState`?

TkDodo:
> Did you try the sandbox?
> It would be nice if you could try and find possible bugs.

Yes, I did try the codesandbox and it works fine for me. I made a test case for testing it in combination with the redux middleware: https://codesandbox.io/s/react-forked-xbjg9

From what I can see, the `INCREASE` action is logged twice (once as `action` and once as `INCREASE`). This is the same as the previous behaviour, so this PR didn't introduce this. However, `setState` is now logged as well (wasn't logged before), so that's good. Is this worth a follow up issue maybe?

> By set, do you mean the first argument of the creator function, or api.setState?

I mean `api.setState`. `set` in the creator function is patched / typed so that we have a third argument "name", but that's not the case for `api.setState`

dai-shi:
Thanks!

> From what I can see, the INCREASE action is logged twice

Maybe, I can fix it in this PR.

> I mean api.setState. set in the creator function is patched / typed so that we have a third argument "name", but that's not the case for api.setState

Yeah, totally correct. Let's leave it as a limitation because types would be too messy, if we were to support this. (I think it's technically possible.)

If you are adventurous, it might be able to do like this:

```js
const useStore = create(devtools((set, get, api) => {
  api.setState = set
  return {
    // ...
  }
}))
```

dai-shi:
https://codesandbox.io/s/react-forked-6czzc
Looks good?

TkDodo:
> Yeah, totally correct. Let's leave it as a limitation because types would be too messy, if we were to support this. (I think it's technically possible.)

yes, sounds good.

> If you are adventurous, it might be able to do like this:

Yes, I am actually doing that already as a workaround until this PR lands:

```
let setState: SetState

const useState = create<State>(
    devtools((set) => {
        setState = set
        return { ... }
}))
```

> Oh, but this will stop logging if you use set with redux middleware.

hm, how would you even do that ? In the example, we only pass the `reducer` to `redux`, and `useState.setState` is now correctly logged ... 🤔

dai-shi:
This 99a9f64 is probably better, even thought it's unlikely that people don't use `set` with redux middleware.

dai-shi:
@TkDodo Would you do the final test? If it's OK, we release it.

TkDodo:
using the `redux` example with 99a9f64 now logs `increase` button twice, first as `setState` and then as `INCREASE`

https://codesandbox.io/s/angry-mestorf-qymx1

dai-shi:
Oops!

dai-shi:
Reverted. 😄 

https://codesandbox.io/s/react-forked-661ih

TkDodo:
👍

dai-shi:
Published: https://www.npmjs.com/package/zustand/v/3.0.3

PR #174, opened 2020-09-02
Stricter types

### LINKED ISSUE #95: [Types] Strictly typed store allows invalid values
I might be missing something, but shouldn't following disallow `invalidA` and `invalidB` values to be used in a store? As it is at the moment I don't get any error about them.

```ts
import create from 'zustand';

interface Store {
  count: number;
  increase: () => void;
}

const [useStore] = create<Store>(set => ({
  count: 0,
  invalidA: 0,
  increase: () => set(state => ({ count: state.count + 1 })),
  invalidB: () => set({ count: 0 })
}));

export default useStore;
```

dai-shi:
Close #95 

This adds a new piece of middleware.
This also avoids `any` (which is a breaking change in types, but not in behavior.)

```ts
import create from "zustand";

const useStore = create(combine({
  count: 0,
  text: "hello",
}, (set) => ({
  inc: () => set((state) => ({ count: state.count + 1 })),
  setText: (text: string) => set({ text }),
})));
```

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 2403,
-    "minified": 1359,
-    "gzipped": 644
+    "bundled": 2599,
+    "minified": 1469,
+    "gzipped": 676
   },
   "vanilla.js": {
     "bundled": 1990,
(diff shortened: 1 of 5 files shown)


dai-shi:
Two issues:

- codesandbox doesn't detect type declarations (not even with v3.0.3)
- it would be nice to come up with better middleware name. any suggestions?

dai-shi:
This doesn't break anything for JS users. Let me merge this in.

PR #177, opened 2020-09-06
Fix(tests.tsx) typo in equalityFn definition

javierriveros:
Fix typo when defining the equality function in test `can use exposed types`


dai-shi (approved):
👍
Didn't notice that.

PR #179, opened 2020-09-06
fix: evaluate selector if state is changed

### LINKED ISSUE #123: Selectors defined outside of a render function do not always return the associated store value
I'm having a problem where store values output via a custom hook do not all have the values from the current store. Some have values from the previous store. Here's a brief explanation, I've also pasted more complete code at the end. Appreciate your help!

The crux of the issue is that if I have my selectors in an object useStore(selectors.blah), values from the store are incorrect. If I do this useStore(state => state.blah) there's no problem. More info:

I've grouped my selectors like this (living in a module with the store):

```
export const selectors = {
  status: (state) => state.status,
  profile: (state) => state.profile,
};
```

And use them like this in a custom hook in a separate module

```
  const status = useStore(selectors.status);
  const profile = useStore(selectors.profile);

```

A login and a logout button change store status and profile together from

```
{ status: statusType.LOGGED_OUT, groups: [], profile: null }
```

to

```
{
      status: statusType.LOGGED_IN,
      profile: {
        familyName: 'bob',
      },
    }
```

The values output in the custom hook are not what I was expecting. status has the current value of the store (statusType.LOGGED_IN), profile has the previous value of the store (null). Odd!

If instead I do this in the custom hook

```
  const status = useStore((state) => state.status);
  const profile = useStore(state => state.profile);
```

All is well. status and profile values have the current values of the store on every render

Here's some more complete code:

------

Custom hook

```
import { useStore, selectors, mutators, statusType } from './store';

export function useCustom() {
  const status = useStore(selectors.status);
  const profile = useStore(selectors.profile);

  const login = useStore(mutators.login);
  const logout = useStore(mutators.logout);

  // these do not match the current version of the store on every render
  console.log('profile', profile); // null
  console.log('status', status); // statusType.LOGGED_IN

  switch (status) {
    case statusType.LOGGED_IN:
      return {
        status: statusType.LOGGED_IN,
        isLoggedIn: true,
        profile,
        logout,
      };
    case statusType.LOGGED_OUT:
      return {
        status: statusType.LOGGED_OUT,
        isLoggedIn: false,
        profile,
        login,
      };
    default:
      return {};
  }
}

```
store

```
import create from 'zustand';

export const statusType = {
  LOADING: 'loading',
  LOGGED_IN: 'logged_in',
  LOGGED_OUT: 'logged_out',
};

export const [useStore, store] = create((set) => ({
  status: statusType.LOADING,
  profile: null,
  login: async () => {
    set({
      status: statusType.LOGGED_IN,
      profile: {
        familyName: 'bob',
      },
    });
  },
  logout: async () => {
    set({ status: statusType.LOGGED_OUT, groups: [], profile: null });
  },
}));

if (typeof window !== 'undefined') {
  init();
}

async function init() {
  store.setState({
    status: statusType.LOGGED_IN,
    profile: {
      familyName: 'bob',
    },
  });
}

export const selectors = {
  status: (state) => state.status,
  profile: (state) => state.profile,
};

export const mutators = {
  login: (state) => state.login,
  logout: (state) => state.logout,
};

```

dai-shi:
Close #123 

There was a bug. It ignores state updates when react rerenders before zustand propagates updates.

Thanks  @andrew-me for reporting this.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -39,9 +39,9 @@
     "gzipped": 354
   },
   "index.js": {
-    "bundled": 4173,
-    "minified": 1360,
-    "gzipped": 672,
+    "bundled": 4352,
+    "minified": 1415,
+    "gzipped": 688,
     "treeshaked": {
       "rollup": {
         "code": 14,
@@ -53,14 +53,14 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 5494,
-    "minified": 2072,
-    "gzipped": 886
+    "bundled": 5675,
+    "minified": 2133,
+    "gzipped": 903
   },
   "index.iife.js": {
-    "bundled": 5747,
-    "minified": 1757,
-    "gzipped": 795
+    "bundled": 5936,
+    "minified": 1814,
+    "gzipped": 809
   },
   "shallow.js": {
     "bundled": 644,
--- a/src/index.ts
+++ b/src/index.ts
@@ -36,12 +36,14 @@ export default function create<TState extends State>(
     equalityFn: EqualityChecker<StateSlice> = Object.is
   ) => {
     const forceUpdate = useReducer((c) => c + 1, 0)[1] as () => void
-    const currentSliceRef = useRef<StateSlice>()
+
+    const state = api.getState()
+    const stateRef = useRef(state)
     const selectorRef = useRef(selector)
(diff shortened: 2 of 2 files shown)


dai-shi:
Here's the fixed example: https://codesandbox.io/s/wandering-meadow-ogsgz

dai-shi (commented):
should be fine.

PR #181, opened 2020-09-07
Readme: Add docs for subscribe equality function

timomeh:
I was wondering if the `subscribe` function also supports an equality function as third parameter. Of course it does! But I had to look it up in the code – so I thought adding it to the readme might be helpful.


dai-shi (approved):
Thanks!

PR #188, opened 2020-09-11
make the commit button work in the devtools

### LINKED ISSUE #184: devtools: commit button doesn't work
Not sure if this is `zustand` specific - but clearing the log by pressing `Commit` works in a standard redux-app, but when the store is `zustand`, the button just doesn't do anything. You can try that out here:
https://codesandbox.io/s/affectionate-davinci-kzdte

Click `increase` a bunch of times, look at the redux dev tools and press `Commit`

Not sure if this is a bug or maybe a missing feature that would need to be implemented separately .. ?

TkDodo:
closes #184

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -81,6 +81,11 @@ export const devtools = <S extends State>(
         } else {
           savedSetState(JSON.parse(message.state))
         }
+      } else if (
+        message.type === 'DISPATCH' &&
+        message.payload?.type === 'COMMIT'
+      ) {
+        api.devtools.init(api.getState())
       }
     })
     api.devtools.init(initialState)


src/middleware.ts
```diff
@@ -81,6 +81,11 @@ export const devtools = <S extends State>(
         } else {
           savedSetState(JSON.parse(message.state))
         }
+      } else if (
+        message.type === 'DISPATCH' &&
+        message.payload?.type === 'COMMIT'
+      ) {
+        api.devtools.init(api.getState())
```
TkDodo, line 88: @dai-shi I am not really familiar with the code above, especially what `ignoreState` is doing, since `setState` is called even if `ignoreState` is true ... 🤔 

maybe an explicit switch on the `message.payload.type` like we are seeing [here](https://github.com/aurelia/store/commit/86b7edab8efc025c7fa10491db66bd305f338922#diff-cfb8e44b6757fd740446e9e4486b73c1R337) might be better?

TkDodo:
here is the repro from the issue with this PR, commit button is working fine fore me: https://codesandbox.io/s/stupefied-meninsky-1wldx

src/middleware.ts
dai-shi: @TkDodo I'm not so sure either. You can clean them up as long as you confirm working correctly.

Be careful with `if (!api.dispatch` though, it's to work with the `redux` middleware. You might need to test with it too, if you haven't.

TkDodo: Okay, in that case, I‘d rather leave it as-is 😅

dai-shi: In any case, we'd need to check if this PR works with `redux` middleware.

TkDodo: here is an example of this PR with the redux middleware - works nicely for me, including the commit button: https://codesandbox.io/s/admiring-ardinghelli-28k1c

dai-shi: Thanks!! Confirmed. Working great.

dai-shi (approved):

PR #189, opened 2020-09-11
Fix shallow types and add shallow test

dios-david:
I noticed `shallow` returns `false` when comparing identical arrays of objects, e.g.:

```ts
shallow(
  [
    { foo: 'bar' },
    { asd: 123 }
  ],
  [
    { foo: 'bar' },
    { asd: 123 }
  ],
)
```

This returns `false` using the latest version, but I think this should be treated as equal.

I also added some tests for `shallow` while I was messing around :)


dai-shi:
Thanks for opening this up, but I don't think it's how shallow works. It's like deepequal.

https://www.npmjs.com/package/shallowequal works like this:
```js
> shallow = require('shallowequal');
[Function: shallowEqual]
> shallow(
...   [
...     { foo: 'bar' },
...     { asd: 123 }
...   ],
...   [
...     { foo: 'bar' },
...     { asd: 123 }
...   ],
... )
false
```

btw, I'd like to have your test to be merged in..

dios-david:
Yes you are right, it's cool that we can pass a custom equality function which is super useful!

I just found `shallow` a bit weird, as comparing object works fine:
```ts
expect(shallow(
  { foo: 'bar', asd: 123 },
  { foo: 'bar', asd: 123 }
)).toBe(true);
```

Comparing arrays works fine:
```ts
expect(shallow(
  [1, 2, 3],
  [1, 2, 3]
)).toBe(true);
```

So why comparing arrays of objects does not work?
```ts
// This fails currently
expect(shallow(
  [
    { foo: 'bar' },
    { asd: 123 }
  ],
  [
    { foo: 'bar' },
    { asd: 123 }
   ],
)).toBe(true);
```

But I respect the decision of the maintainers, thought it would be a good addition for zustand users.
Feel free to revert the shallow related changes and merge the tests if you find them useful!

dai-shi:
> But I respect the decision of the maintainers

To be fair, this is not my preference at all. As I understand, it's the definition of shallow equal.

> thought it would be a good addition for zustand users.

What you are suggesting is deep equal.
My preference here is instead of adding deepequal helper, we encourage to use existing libraries.

https://lodash.com/docs/4.17.15#isEqual should work.
```js
import isEqual from 'lodash.isequal'

  useStore(someSelector, isEqual)
```

> Feel free to revert the shallow related changes and merge the tests if you find them useful!

Thanks. Will do.

PR #192, opened 2020-09-15
feat: pass `get` into reducer in redux middleware

narwold:
We recently had a situation where we were storing JS getters in a Zustand store, and when the getters were resolved we wanted to guarantee that they were getting the correct data from the store instead of just the data that was present at the time that the reducer was run.

The core Zustand API supports accessing `get` when a method is called, but the Redux middleware doesn't support this currently. This PR proposes to add that capability.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -8,7 +8,7 @@ import {
 } from './vanilla'
 
 export const redux = <S extends State, A extends { type: unknown }>(
-  reducer: (state: S, action: A) => S,
+  reducer: (state: S, action: A, opts: { get: GetState<S> }) => S,
   initial: S
 ) => (
   set: SetState<S>,
@@ -19,7 +19,7 @@ export const redux = <S extends State, A extends { type: unknown }>(
   }
 ): S & { dispatch: (a: A) => A } => {
   api.dispatch = (action: A) => {
-    set((state: S) => reducer(state, action))
+    set((state: S) => reducer(state, action, { get }))
     if (api.devtools) {
       api.devtools.send(api.devtools.prefix + action.type, get())
     }


dai-shi:
Hi, thanks for the suggestion. However, this is not quite desired, because a reducer should *only* be able to know the previous state.

Could you explain your issue again with a small snippet? Maybe we could suggest something.

narwold:
Sure. I was writing a normalizer for a JSON:API client, and I wanted to store a getter like this in the store:

In the normalizer for the update of the request state for request 1:
```
Object.defineProperty(objThatIsPartOfReq1State, k, {
  get: () => getState().normalized[type][id],
})
```

If the underlying normalized data changes as a result of request 2, the pointer should point to the newest data (the data that's actually in the store)... not the data as it existed when the reducer was run for request 1.

When is it NOT a good idea to be able to imperatively fetch the latest data in the store? Seems like that would always be the most accurate.

dai-shi:
Interesting. So, you run some normalization process after reducer returns a new state?
Anyway, redux reducer is defined to be pure, which is a reasonable restriction.
zustand won't stop you doing as you like.

My suggestion is either do not use redux middleware and you get `get` in `create`,
or create your own middleware like you do in this PR. (We wouldn't call it redux, but it's up to you.)

Closing this now, but feel free to continue discussion in this PR.

dai-shi:
> When is it NOT a good idea to be able to imperatively fetch the latest data in the store? Seems like that would always be the most accurate.

It is basically fine, I guess. Now, I'm not sure if I understand how the code is structured.

narwold:
We actually run a normalization DURING the execution of the reducer, but the way it's normalized is to inject getters (pointers) that won't resolve until something requests that data via a selector or whatever. That data may be requested at more than one time as the requests and the underlying normalized data both change... so it always needs to refer to the latest. Here's a typical example in pseudocode:

```
store: {
  requests: {
    '/widgets': [{ ...widget1 => normalizedWidget1 }, { ...widget2 => normalizedWidget2 }],
    '/widgets/1': widget1 => normalizedWidget1,
    '/widgets/1': widget2 => normalizedWidget2,
  }
  normalized: {
    widgets: {
      1: { ...normalizedWidget1 },
      2: { ...normalizedWidget2 },
    }
  }
}
```

Now let's say that `/widgets/2` refetches and data updates. This means that `normalizedWidget2` changes, which means that in addition to the data for `/widgets/2` updating directly, the pointer in `/widgets` also now resolves to the new data.

In any case, we did end up rolling our own middleware, so we're set. Thought I'd contribute this though if it was worth having. Thanks for your consideration.

dai-shi:
> We actually run a normalization DURING the execution of the reducer

I might still misunderstand it, but if I don't, this would be relevant reducer code.

```js
const reducer = (prevState, action) => {
  let nextState = prevState
  const getState = () => nextState
  // do update nextState immutably
  return nextState
}
```

> That data may be requested at more than one time as the requests and the underlying normalized data both change... so it always needs to refer to the latest.

I have a guess. It's a terminology issue.
- What you are doing and what you are proposing is "middleware" in the redux terminology
- Whereas, the "redux" middleware in zustand is to emulate redux reducer behavior

So, the former middleware and the latter middleware are totally different things.

> In any case, we did end up rolling our own middleware

So, this is totally valid. It's just that the "redux" in zustand is something different.

> Thought I'd contribute this though if it was worth having. Thanks for your consideration.

Thank you too for the proposal anyway!

PR #198, opened 2020-09-24
Change README in relation to the use with immer

caesarsol:
Added a WITHOUT/WITH section, and made the example a bit more concise and real-world.
(Code not tested, please review)

Thanks!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -229,19 +229,37 @@ function Component() {
 
 Reducing nested structures is tiresome. Have you tried [immer](https://github.com/mweststrate/immer)?
 
+Without immer:
+```jsx
+const useStore = create(set => ({
+  lush: { forrest: { contains: { a: "bear" } } },
+  setAnimal: (animal) => set(state => ({ 
+    ...state, 
+    lush: { 
+      ...state.lush, 
+      forrest: { 
+        ...state.lush.forrest, 
+         contains: {
+           ...state.lush.forrest.contains,
+           a: animal,
+         },
+       },
+     },   
+   })),
+}))
+```
+
+With immer:
 ```jsx
 import produce from 'immer'
 
 const useStore = create(set => ({
   lush: { forrest: { contains: { a: "bear" } } },
-  set: fn => set(produce(fn)),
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks for suggestion.
As @drcmda wants to keep README compact, I'm not sure if we really need the without section.
Do we need to promote immer with examples?

readme.md
```diff
@@ -229,20 +229,38 @@ function Component() {
...
+```
+
+With immer:
 ```jsx
 import produce from 'immer'
 
   lush: { forrest: { contains: { a: "bear" } } },
-  set: fn => set(produce(fn)),
+  setAnimal: animal => set(produce(state => { state.lush.forrest.contains.a = animal })),
```
dai-shi, line 258: This is not what we should change, I suppose.

readme.md
```diff
@@ -229,20 +229,38 @@ function Component() {
...
 }))
-
-const set = useStore(state => state.set)
-set(state => {
-  state.lush.forrest.contains = null
-})
 ```
 
+To make it even more comfortable, look at the Middleware paragraph below.
```
dai-shi, line 262: I think it's good have a sentence here, but if we revert the above, what should we say here?

readme.md
```diff
@@ -256,13 +274,14 @@ const log = config => (set, get, api) => config(args => {
...
 // Turn the set method into an immer proxy
+import produce from 'immer'
 const immer = config => (set, get, api) => config(fn => set(produce(fn)), get, api)
 
 const useStore = create(
   log(
     immer((set) => ({
       bees: false,
-      setBees: (input) => set((state) => void (state.bees = input)),
+      setBees: input => set(state => { state.bees = input }),
```
dai-shi, line 284: Some people may get confused. So, I'm ok with this change.

dai-shi (changes requested):
Let's remove the big "Without immer" section, and make small adjustments, then we'd be good to go.

dai-shi:
Closing as stale. Please reopen if modifications are made.

PR #201, opened 2020-09-30
mjs build

### LINKED ISSUE #200: @babel/runtime dependency while consuming Zustand with Typescript
When importing Zustand in a typescript (node CLI) project, I'm getting this error:

```
Error: Cannot find module '@babel/runtime/regenerator'
Require stack:
- /Users/me/projects/project/node_modules/zustand/index.cjs.js
```

There seem to be an `index.js` and an `index.cjs.js`, with Typescript importing the latter. The index.cjs.js requires @babel/runtime. My Node version is 12.x and my tsconfig looks like this:
```
{
    "compilerOptions": {
        "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
        "module": "commonjs",
        "target": "es2019",
        "strict": true,
        "esModuleInterop": true,
        "skipLibCheck": true,
        "forceConsistentCasingInFileNames": true,
        "jsx": "react"
    }
}
```

I rather not change my tsconfig or Node version, unless there is a compelling reason. Is it correct that for my case I need to install `@babel/runtime` as a dependency? I would expect not to, since `tsc` should probably take care of everything?

Thanks!

dai-shi:
close #200 

https://nodejs.org/api/packages.html#packages_determining_module_system


dai-shi:
I added dist/index.mjs which should solve the issue, but it's identical to dist/index.js.
I wonder if we could try `"type": "module"` with which we'd need to rename all cjs build files to "*.cjs".

dai-shi:
Hmmm, I need codesandbox ci to work. Why does it not work?

dai-shi:
OK, here it comes.

@gersomvg Would you try the codesandbox build?

dai-shi:
The new commit is with `"type": "module"`. I hope this won't break things, but 100% certain.

```json
    "zustand": "https://pkg.csb.dev/pmndrs/zustand/commit/58247b6f/zustand"
```

gersomvg:
> The new commit is with `"type": "module"`. I hope this won't break things, but not 100% certain.
> 
> ```json
>     "zustand": "https://pkg.csb.dev/pmndrs/zustand/commit/58247b6f/zustand"
> ```

Thanks for exploring this further!

I still get a
```
Error: Cannot find module '@babel/runtime/regenerator'
Require stack:
- /Users/me/projects/project/node_modules/zustand/index.cjs
```

And if I import literally `import create from 'zustand/index'` to get the non-cjs version, I get this error:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/me/projects/project/node_modules/zustand/index.js
require() of ES modules is not supported.
require() of /Users/me/projects/project/node_modules/zustand/index.js from /Users/me/projects/project/bin/state/router.ts is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/me/projects/project/node_modules/zustand/package.json.
```

dai-shi:
Try:
```js
import create from 'zustand/index.js'
```

We can't omit the extension: https://nodejs.org/api/packages.html#packages_determining_module_system

gersomvg:
Still the same error 😬 . Omitting the extension actually works (at least in the context of typescript)

gersomvg:
Doesn't `type: module` require Node 12 (with an experimental flag) or Node 14? Since I'm consuming this file unbundled, that would also go for the consumer I think (my root project).

dai-shi:
It requires node 12. You are not using node 12?

dai-shi:
Oh, I might have misunderstood. It's the app package.json, not the node_modules package.json...

gersomvg:
I'm not guaranteed of Node 14 or Node 12 (with --experimental-modules) unfortunately. It's fine, I'll just leave `@babel/runtime` in my deps for now.

But I keep wondering how other packages handle this, since I have not encountered this type of issue before.

Thanks a lot for the time that you already put into this! 👏

dai-shi:
You are welcome.
I too am curious how others are doing.

Closing this as this doesn't help other use cases.

dai-shi:
@gersomvg Would you help me as a final trial? (It's actually the same as the first one in this PR.)

```js
  "zustand": "https://pkg.csb.dev/pmndrs/zustand/commit/983e1c62/zustand"
```

Using this build ☝️ , and doing like this 👇 : 

```js
import create from 'zustand/index.mjs'
```

PR #217, opened 2020-10-27
Migrate to @rollup/* packages and make resolve plugin first to run

jcoussard:
See https://github.com/pmndrs/jotai/pull/154 for additional information about these changes.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -58,7 +58,7 @@
     "gzipped": 903
   },
   "index.iife.js": {
-    "bundled": 5974,
+    "bundled": 6280,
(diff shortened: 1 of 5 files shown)


.size-snapshot.json
```diff
@@ -58,7 +58,7 @@
     "gzipped": 903
   },
   "index.iife.js": {
-    "bundled": 5974,
+    "bundled": 6280,
```
dai-shi, line 61: This change should be harmless as we discussed in the chat.

dai-shi (approved):
👍 Thanks!

Issue #212, opened 2020-10-20
Typescript version for zustand redux middleware with dispatch 


Phibee:
Can you give example of your zustand and redux calling dispatch action like this link below using typescripts. I believe there's a dispatch on zustand middleware don't know how to use it.

[Codesandbox link](https://codesandbox.io/s/zustand-immer-redux-f05t3?file=/src/createStore.js:131-138)

dai-shi:
Good point. I don't think we support typing `api.dispatch` currently.
Here's my modification with some type assertions. (Not sure how to type curried produce.)
https://codesandbox.io/s/zustand-immer-redux-forked-q2nkn?file=/src/createStore.ts

We could technically improve store typing with better generics. It can be messy, and I wouldn't feel like doing now with my limited ts skills, unless someone volunteers for it.

PR #222, opened 2020-10-30
improve middleware/combine types

### LINKED ISSUE #216: Enable NamedSet when using devtools and combine
When using devtools with combine, combine's `set` is not of type `NamedSet` so typescript does not permit to give a third parameter to `set` function. Is there a way to work around it?

image

maybe exporting interface from `middleware.ts:31`, I can force it to use.

dai-shi:
close #216 

It uses CustomSetState.
(Note, if we'd support #212, it'd require total overhaul.)

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -98,21 +98,22 @@ export const devtools = <S extends State>(
 
 export const combine = <
   PrimaryState extends State,
-  SecondaryState extends State
+  SecondaryState extends State = State,
+  CustomSetState extends SetState<PrimaryState> = SetState<PrimaryState>
 >(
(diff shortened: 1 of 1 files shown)


dai-shi:
@rosostolato Would you try if this works for both with devtools and without devtools?

rosostolato:
It's working fine! I also created a PR with tests for devtools and combine https://github.com/pmndrs/zustand/pull/230

dai-shi:
Good news. Thanks. Let me merge this.

dai-shi:
This doesn't seem to be a right fix. Reverted: 9b4aadb

PR #225, opened 2020-11-01
add pipe middlewares example

Yhozen:
add ramda's pipe middlewares example as discussed [here](https://github.com/pmndrs/zustand/discussions/224)

```js
import create from "zustand"
import produce from "immer"
import pipe from "ramda/es/pipe"

const log = config => (set, get, api) =>
  config(
    args => {
      console.log("  applying", args)
      set(args)
      console.log("  new state", get())
    },
    get,
    api
  )

const immer = config => (set, get, api) =>
  config(fn => set(produce(fn)), get, api)

const createStore = pipe(log, immer, create)

const useStore = createStore(set => ({
  bears: 1,
  increasePopulation: () => set(state => ({ bears: state.bears + 1 })),
}))

export default useStore
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -269,6 +269,31 @@ const useStore = create(
 )
 ```
 
+
+<overview>
+<summary>How to pipe middlewares</summary>
+<details>
+
+```js
+import create from "zustand"
+import produce from "immer"
+import pipe from "ramda/es/pipe"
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
LGTM

Issue #214, opened 2020-10-23
offline support 


cnx-piyush:
I wanted to know if zustand support offline where the state can be stored in localstorage  . I was hoping a plugin would be super useful in this case

dai-shi:
Do you mean persistence support? Does #76 look relevant?

cnx-piyush:
@dai-shi This is exactly what I was looking for . Thanks and I am closing the issue .

dai-shi:
A note for anyone: We could add some functions to support persistence in middleware.ts. Feel free to open an issue for discussions.

Issue #76, opened 2019-11-02
Persist Middleware


alexluong:
I'm working on a middleware that saves states into LocalStorage. I see that other people also had the same idea (#7).

I have an example working, but I'd love to learn if there is a better approach.

```ts
const isBrowser = typeof window !== "undefined"

const persistedState =
  isBrowser
    ? JSON.parse(localStorage.getItem("sidebarState"));
    : null

const persist = config => (set, get, api) =>
  config(
    args => {
      set(args);
      isBrowser && localStorage.setItem("sidebarState", JSON.stringify(get()));
    },
    get,
    api
  );

const [useSidebar] = create(
  persist(set => ({
    isOpen: persistedState ? persistedState.isOpen : true,
    toggleSidebar: () => {
      set(state => ({ ...state, isOpen: !state.isOpen }));
    },
  }))
);

```

marcoacierno:
Mine was/is a **very** basic app, but that's what I implemented:

```js
import create from 'zustand';

const restoredState = JSON.parse(localStorage.getItem('state'));

const persist = config => (set, get, api) =>
  config(
    args => {
      set(args);
      window.localStorage.setItem('state', JSON.stringify(get()));
    },
    get,
    api,
  );

const [useStore] = create(
  persist(set => ({
    theme: 'light',
    ...restoredState,
    setTheme: theme => set(state => ({ theme })),
  })),
);

export default useStore;
```

where I basically first set the default state of each variable and then overwrite them with what was inside the store.

It won't work well if you change type of information, or anything else really but gets the job done (at least for me)

alexluong:
FWIW, my initial approach is terrible for SSR. It took me 6 hours to figure it out.

Apparently, this:

```js
const isBrowser = typeof window !== "undefined"

const persistedState =
  isBrowser
    ? JSON.parse(localStorage.getItem("sidebarState"));
    : null
```

doesn't work well with SSR.

The approach I need to take is to give the original state inside `create` function the initial state, and then use `useMount` to get `localStorage` data and update state.

Just in case anyone may run into this in the future!

drcmda:
is this something we can add officially to the lib? or is it too specific?

marcoacierno:
What about this?

```
import create from 'zustand';

const persist = config => (set, get, api) => {
  const initialState = config(
    args => {
      set(args);
      window.localStorage.setItem('state', JSON.stringify(get()));
    },
    get,
    api,
  );

  const restoredState =
[...]

PR #229, opened 2020-11-04
feat: add persist middleware

AnatoleLucet:
Hi :wave: I saw [this comment](https://github.com/pmndrs/zustand/issues/214#issuecomment-715895730) so I've decided to open a pr for the middleware I've made and shared [here](https://github.com/pmndrs/zustand/issues/76#issuecomment-670367631).

I've been using it for multiples months on a project at work, and it actually works nicely given its simplicity.
But I still think there's some quite annoying drawbacks that we could probably improve:
- The obligation to give a name to the middleware. It'd be preferable to not hard code the name so the middleware could generate unique names/ids for the user. But I can't come up with a way to do this. 
- Rehydration & merge & update of a stored state that has changed since. I don't think this is possible (?), but it still'd be nice to erase the old (incompatible) stored state by the new one (so there would be some data loss). I have some ideas on how to choose if the stored state needs to be erased or not, but I'd like to have the maintainers' opinions first.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -326,12 +326,6 @@ export const useStore = create(persist(
   {
     name: "food-storage", // unique name
     storage: sessionStorage, // (optional) default is 'localStorage'
-    serialize: (state) => { // (optional) default is 'JSON.stringify'
-      // some stuff to serialize the state and return it
-    }, 
-    deserialize: (str) => { // (optional) default is 'JSON.parse'
-      // some stuff to deserialize the state and return it
-    }
   }
 ))
 ```
--- a/readme.md
+++ b/readme.md
@@ -310,6 +310,32 @@ const immer = <T extends State>(
 </details>
 </overview>
 
+## Persist middleware
+
+You can persist you store's data using any kind of storage.
+
+```jsx
+import create from "zustand"
+import { persist } from "zustand/middleware"
+
+export const useStore = create(persist(
+  (set, get) => ({
+    fish: 0,
+    addAFish: () => set({ fish: get().fish + 1 })
+  }),
+  {
+    name: "food-storage", // unique name
+    storage: sessionStorage, // (optional) default is 'localStorage'
+    serialize: (state) => { // (optional) default is 'JSON.stringify'
+      // some stuff to serialize the state and return it
+    }, 
+    deserialize: (str) => { // (optional) default is 'JSON.parse'
+      // some stuff to deserialize the state and return it
+    }
+  }
+))
+```
+
 ## Can't live without redux-like reducers and action types?
 
 ```jsx
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -118,8 +118,8 @@ export const combine = <
   )
 
 export type StateStorage = {
-  getItem: (key: string) => string | null | Promise<string | null>
-  setItem: (key: string, value: string) => void | Promise<void>
+  getItem: (name: string) => string | null | Promise<string | null>
+  setItem: (name: string, value: string) => void | Promise<void>
 }
 export type PersistOptions<S> = {
   name: string
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,10 +1,10 @@
 import {
-  State,
   GetState,
-  SetState,
-  StoreApi,
   PartialState,
+  SetState,
+  State,
   StateCreator,
+  StoreApi,
 } from './vanilla'
 
 export const redux = <S extends State, A extends { type: unknown }>(
@@ -116,3 +116,48 @@ export const combine = <
       api as StoreApi<PrimaryState>
     )
   )
+
+export type StateStorage = {
+  getItem: (key: string) => string | null | Promise<string | null>
+  setItem: (key: string, value: string) => void | Promise<void>
+}
+export type PersistOptions<S> = {
+  name: string
+  storage?: StateStorage
+  serialize?: (state: S) => string | Promise<string>
+  deserialize?: (str: string) => S | Promise<S>
+}
+
+export const persist = <S extends State>(
+  config: StateCreator<S>,
+  options: PersistOptions<S>
+) => (set: SetState<S>, get: GetState<S>, api: StoreApi<S>): S => {
+  const {
+    name,
+    storage = localStorage,
+    serialize = JSON.stringify,
+    deserialize = JSON.parse,
+  } = options || {}
+
+  const state = config(
+    (payload) => {
+      set(payload)
+      ;(async () => {
+        storage.setItem(name, await serialize(get()))
+      })()
+    },
+    get,
+    api
+  )
+
+  ;(async () => {
+    try {
+      const storedState = await storage.getItem(name)
+      if (storedState) set(await deserialize(storedState))
+    } catch (e) {
+      console.error(new Error(`Unable to get to stored state in "${name}"`))
+    }
+  })()
+
+  return state
+}


dai-shi:
Thanks for opening up this. I'm happy to discuss and I also hope someone to chime in.

First things first, I would like to factor out `localStorage` and make it usable with `AsyncStorage` or any other storage implementations. How does that sound?

AnatoleLucet:
For sure, it sounds great. I've already [responded to someone asking how he could port this to `AsyncStorage`](https://github.com/pmndrs/zustand/issues/76#issuecomment-707198059).
I might need to give a glance at the hook first to understand how it works and how this could be done. But maybe you have something in mind?

dai-shi:
Let me see, it shouldn't have to be a hook.

```js
export const persist = (
  config,
  name,
  storage,
) => (set, get, api) => {
  const store = config(
    (payload) => {
      set(payload) // payload can be a function
      storage.setItem(name, JSON.stringify(get())
    },
    get,
    api
  )

  (async () => {
    const storedState = JSON.parse(await storage.getItem(name)) // needs error handling
    api.setState(storedState);
  })()

  return store
}
```

AnatoleLucet:
That's nice. How should the error handling be done? By logging a message with the `Error` object?

dai-shi:
Hm, don't know. It would be logged anyway. We need to think about the use case how we would recover from errors.

AnatoleLucet:
Well, if an error is thrown during the parsing or the "getting" (`getItem`), it'll be caught (by our try/catch) before setting the state. So it shouldn't crash the user's app nor the store.

dai-shi:
Are you sure? Not sure what is meant by "our try/catch".

Anyway, it can be an edge case. We can start trying our basic idea. How do we proceed? Does my code snippet really work? Would you try? Would you incorporate it in this PR?

AnatoleLucet:
> Are you sure? Not sure what is meant by "our try/catch".

```js
  (async () => {
    try {
      const storedState = JSON.parse(await storage.getItem(name))
      api.setState(storedState);
    } catch (e) {
      // catches JSON.parse and storage.getItem errors
    }
  })()
```
image
image

> Anyway, it can be an edge case. We can start trying our basic idea. How do we proceed? Does my code snippet really work? Would you try? Would you incorporate it in this PR?

I'll try it and if it does, I'll refactor the pr.

AnatoleLucet:
How do you build the package so it can be linked? `yarn prebuild && yarn build && yarn postbuild`?

dai-shi:
Just `yarn build` is fine.
Also, if you push the change into this PR, codesandbox ci will build it for you.

AnatoleLucet:
@dai-shi it seems to work https://codesandbox.io/s/react-typescript-forked-rj85m?file=/src/App.tsx (you can try by commenting / uncommenting the `useEffect`).

dai-shi:
Confirmed. 👍

src/vanilla.ts
```diff
@@ -31,6 +31,12 @@ export type StateCreator<T extends State, CustomSetState = SetState<T>> = (
   get: GetState<T>,
   api: StoreApi<T>
 ) => T
+interface AsyncStorage {
+  getItem: (key: string, ...rest: any[]) => Promise<string | null>
+  setItem: (key: string, value: string, ...rest: any[]) => void
+  [x: string]: any
+}
+export type StateStorage = Storage | AsyncStorage
```
dai-shi, line 39: I don't think we need to define it here, do we?

I prefer more compact type that we really use.
```ts
type StateStorage = {
  getItem: (key: string) => string | null | Promise<string | null>
  setItem: (key: string, value: string) => void | Promise<void>
}
```

src/middleware.ts
```diff
@@ -1,12 +1,15 @@
...
-  StoreApi,
   PartialState,
+  SetState,
+  State,
   StateCreator,
+  StateStorage,
+  StoreApi,
 } from './vanilla'
 
+const isRunningInBrowser = () => typeof window !== 'undefined'
```
dai-shi, line 11: I would like to avoid having this check in our middleware. Let's just leave it to users.

src/middleware.ts
```diff
@@ -116,3 +119,34 @@ export const combine = <
       api as StoreApi<PrimaryState>
     )
   )
+
+export const persist = <S extends State>(
+  name: string,
+  config: StateCreator<S>,
```
dai-shi, line 125:
```suggestion
  config: StateCreator<S>,
  name: string = "zustand-persist-default-key",
```
How about this?

AnatoleLucet: If the user uses multiple persistent stores in the same app/website, it'll create some conflicts.

hmans:
I don't know if this would bust the scope of this change/feature wide open (I do suspect that it does, though), but how do you feel about expressing changes as [JSON Patches](http://jsonpatch.com/) (considering that atomic updates to a zustand store may affect more than 1 key)?

Having written this, I'm pretty sure that this would go beyond what you're trying to do here. I just wanted to point out that JSON Patch et al are a thing, and it would be great to have persistence glue based on it for zustand stores (it would allow some fun implementations of [differential synchronisation](https://github.com/janmonschke/diffsync).) If there is some interest in this stuff, I'd be happy to whip something up, as my current project is sort of heading into its general direction.

src/middleware.ts
dai-shi: Yeah. Do you prefer removing the default value?

AnatoleLucet: Yep, I think that forcing the user to name his store is better (at least if we don't find a solution to manage the names for him).

dai-shi:
@hmans differential synchronization sounds interesting. we can try that road if someone is interested, but it can be another middleware as you implied.

This reminds me of something. How about factoring out JSON.parse/stringify and make them customizable? It should be easy if they are synchronous functions or even async doesn't matter. What do you think? @AnatoleLucet

AnatoleLucet:
@dai-shi sure, why not. Do you think of a custom callback given through the arguments (then we might need to create an option object as the third argument of `persist`) like what we're doing with the storage?

dai-shi:
Maybe parse/stringify only. We don't need to cover 100% use cases.
I might call it serialize/deserialize, but whichever is ok.
option object sounds good. I was thinking about two additional arguments, which can be too many.

AnatoleLucet:
There you go.

src/middleware.ts
```diff
@@ -116,3 +116,53 @@ export const combine = <
...
   )
+
+export type StateStorage = {
+  getItem: (key: string) => string | null | Promise<string | null>
+  setItem: (key: string, value: string) => void | Promise<void>
+}
+type PersistDefaultOptions<S> = {
+  storage: StateStorage
+  serialize: (state: S) => string | Promise<string>
+  deserialize: (state: string) => S | Promise<S>
```
dai-shi, line 127:
```suggestion
  deserialize: (str: string) => S | Promise<S>
```

src/middleware.ts
```diff
@@ -116,3 +116,53 @@ export const combine = <
...
+export type PersistOptions<S> = {
+  name: string
+} & Partial<PersistDefaultOptions<S>>
+
+export const persist = <S extends State>(
+  config: StateCreator<S>,
+  options: PersistOptions<S>
+) => (set: SetState<S>, get: GetState<S>, api: StoreApi<S>): S => {
+  const defaultOptions: PersistDefaultOptions<S> = {
+    storage: localStorage,
```
dai-shi, line 138: This may cause error on non-web environment. Even if we specify `storage` option?

dai-shi (commented):
Looks pretty good.

Feel like we should test it with RN. Any volunteers?

src/middleware.ts
AnatoleLucet: Indeed, but you said [this](https://github.com/pmndrs/zustand/pull/229#discussion_r517953668). Am I missing a point?

dai-shi: Sorry for the confusion. I didn't quite get the problem then.
Now, I still don't like checking `window`, so my suggestion is this:
```suggestion
    storage: typeof localStorage !== 'undefined' ? localStorage : undefined,
```

But, this can raise an error. Do you have better idea using `typeof localStorage` checking?

AnatoleLucet: Do you mean this can raise an error because of `storage` being undefined?

dai-shi: Yeah, I mean raising an error is ok because developers need to know that they need to specify it, but error message won't be very friendly.

dai-shi: I noticed my suggestion doesn't make sense.
```js
typeof localStorage !== 'undefined' ? localStorage : undefined
```
This is just equivalent to `localStorage`. 😄

dai-shi: Testing in chrome console:
image
Okay, it should be good as is.

AnatoleLucet: Well, we could throw an error if `defaultOptions.storage && (typeof storage?.getItem !== "function" || typeof storage?.setItem !== "function")` with this `typeof localStorage !== 'undefined' ? localStorage : undefined` (as in your review above) in the default options.

So this would only throw if the code is running in a browser / browser friendly env and if the user has given a custom storage that doesn't has the `getItem` and `setItem` methods.

dai-shi: No, I don't think it's necessary. Let's release it and see how people report some issues.

dai-shi:
Should we have a small description in readme.md?

src/middleware.ts
dai-shi: On second thought, I think this **is** a problem in RN.
https://github.com/pmndrs/zustand/pull/229/files#r518620500 is working good, but it fails with this.

image

So, we need to change the way to pass default options.

dai-shi: I think this works:
```js
const {
  name,
  storage = localStorage,
  serialize = JSON.stringify,
  deserialize = JSON.parse,
} = options || {}
```

AnatoleLucet:
Done https://github.com/AnatoleLucet/zustand/tree/feat/persit-middleware#persist-middleware

dai-shi:
> And only the parse/stringify or a whole set/get function?

Now, after looking at the readme, I understand what this really means.
If we customize the whole set/get, we don't need `name` and `storage` options, right?
But, maybe it's too much customization, and the current options are fine, I think.
Just want to make it sure if it's aligned with your thought.

AnatoleLucet:
I was just thinking about letting the user do this whole async function :
```jsx
  ;(async () => {
    try {
      const storedState = await storage.getItem(name)
      if (storedState) set(await deserialize(storedState))
    } catch (e) {
      console.error(new Error(`Unable to get to stored state in "${name}"`))
    }
  })()
```

But you're right, this is probably too much customization.

dai-shi:
OK, we are on the same page. I'll think about again if we have a better option. I think our primary target is localStorage in web and AsyncStorage in RN.

Meanwhile, would you check https://github.com/pmndrs/zustand/pull/229#discussion_r518641567 again? It's not resolved.

AnatoleLucet:
> Meanwhile, would you check #229 (comment) again? It's not resolved.

Done

dai-shi (approved):
Looks good!

src/middleware.ts
```diff
@@ -116,3 +116,48 @@ export const combine = <
       api as StoreApi<PrimaryState>
     )
   )
+
+export type StateStorage = {
+  getItem: (key: string) => string | null | Promise<string | null>
+  setItem: (key: string, value: string) => void | Promise<void>
```
dai-shi, line 122:
```suggestion
  getItem: (name: string) => string | null | Promise<string | null>
  setItem: (name: string, value: string) => void | Promise<void>
```

Just minor, but this should be more consistent?

readme.md
```diff
@@ -310,6 +310,32 @@ const immer = <T extends State>(
...
+  }),
+  {
+    name: "food-storage", // unique name
+    storage: sessionStorage, // (optional) default is 'localStorage'
+    serialize: (state) => { // (optional) default is 'JSON.stringify'
+      // some stuff to serialize the state and return it
+    }, 
+    deserialize: (str) => { // (optional) default is 'JSON.parse'
+      // some stuff to deserialize the state and return it
+    }
```
dai-shi, line 334:
```suggestion
```

Let's remove these from readme until we find good use cases.

dai-shi:
I don't know why codesandbox ci doesn't finish until now, but let me merge it since all other tests are passing.

calendee:
Thanks so much for adding this functionality! I had created my own wrapper for doing this in React Native & React Native Web. However, this is much more elegant!

Ke-v:
Hi, this is my store. When I use persist I get an error `TypeError: Object(...) is not a function`. Am I doing anything wrong in the code. When I remove persist it works fine, but I want to use the state on different routes


```
export const useStore = create(persist((set, get) => ({
  vendorData: [],
  menu: [],
  cart: [],
  displayedItems: [],
  total: 0,
  quantityGrouping: {},
  setVendorData: (vendorId)  => {
    Axios.get(`http://localhost:9000/users/vendor/fetch/${vendorId}`)
    .then(response => {
      set(state => ({vendorData: response.data[0]}))
    })
  },
  setMenu: (menuData) => set(state => ({menu: menuData})),
  addToCart: (item) => {
    const currCart = get().cart
    set(state => ({cart: [...currCart, {...item}]}))
  },
  removeFromCart: (index) => {
    const currCart = get().cart
    currCart.splice(index, 1)
    set(state => ({cart: currCart}))
  },
  setTotal: () => {
    const currCart = get().cart
    if (currCart.length !== null){
      const total = (currCart.map(item => item.item_price).reduce(((prev, curr, arr) => prev+curr), 0))
      set({ total: total })
    } else {
      set({ total: 0 })     
    }
  },
  setQuantityGrouping: () => {
    const currCart = get().cart
    const quantityGroup = currCart.reduce((acc, it) => {
      acc[it.item_id] = acc[it.item_id] + 1 || 1;
      return acc;
    }, {})
    set(state => ({ quantityGrouping: quantityGroup}))
  },
  setDisplayedItems: () => {
    const currCart = get().cart
    const cartData = currCart.reduce((acc, it) => {
      const x = acc.find(item => item.item_id === it.item_id);
      if (!x) {
        return acc.concat([it]);
      } else {
        return  acc;
      }
    }, [])
    set({ displayedItems: cartData})
  }

}),
  {
     name: "vendorStore",
     storage: sessionStorage
  }
))
```

PR #234, opened 2020-11-06
fix currentSlice in listeners not being updated when equalityFn returns true

dios-david:
I noticed an issue with the listeners, if the `equalityFn` returns `true` then the internal `currentSlice` doesn't get updated, so the subsequent `equalityFn` calls will receive outdated value in the `previous` param.

**Example**
```js
  const { setState, subscribe } = create(() => ({ value: 0 }))
  const equalityFn = jest.fn(() => true);

  subscribe(
    () => {},
    (state) => state.value,
    equalityFn
  );
  
  setState({ value: 1 })
  expect(equalityFn).toHaveBeenCalledWith(0, 1);
  setState({ value: 2 })
  expect(equalityFn).toHaveBeenCalledWith(1, 2); // this fails as it was called with `0, 2`
  setState({ value: 2 })
  expect(equalityFn).toHaveBeenCalledWith(2, 2); // this fails as it was called with `0, 2`
```

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -61,9 +61,10 @@ export default function create<TState extends State>(
       // the listener from being called.
       // https://github.com/react-spring/zustand/pull/37
       try {
-        const newStateSlice = selector(state)
-        if (!equalityFn(currentSlice, newStateSlice)) {
-          listener((currentSlice = newStateSlice))
+        const previousSlice = currentSlice
+        currentSlice = selector(state)
+        if (!equalityFn(previousSlice, currentSlice)) {
+          listener(currentSlice)
         }
       } catch (error) {
         listener(null, error)
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -438,6 +438,24 @@ it('can subscribe to the store', () => {
   unsub()
 })
 
+it('ensures equalityFn is always called with the correct values', () => {
+  const { setState, subscribe } = create(() => ({ value: 0 }))
+  const equalityFn = jest.fn(() => true);
+
+  subscribe(
+    () => {},
+    (state) => state.value,
+    equalityFn
+  );
+  
+  setState({ value: 1 })
+  expect(equalityFn).toHaveBeenCalledWith(0, 1);
+  setState({ value: 2 })
+  expect(equalityFn).toHaveBeenCalledWith(1, 2);
+  setState({ value: 2 })
+  expect(equalityFn).toHaveBeenCalledWith(2, 2);
+});
+
 it('can destroy the store', () => {
   const { destroy, getState, setState, subscribe } = create(() => ({
     value: 1,


dai-shi:
Thanks for opening this up.

> if the equalityFn returns true then the internal currentSlice doesn't get updated

However, I think it's the expected behavior. equalityFn()=true means we shouldn't replace currentSlice. If we replace it silently, it may introduce other undesired behavior.

> so the subsequent equalityFn calls will receive outdated value in the previous param.

This sounds like you have some issues with the current behavior. Can you share your experience?

dios-david:
In my understanding `equalityFn()=true` means the listener won't get called, as we treat the previous/new state slices equal.

At the point when `equalityFn` gets called the values in the store already changed, right? Do you think there is any valid use cases where I need to compare the new state slice with a slice which was stored several changes ago, in order to decide whether my listener should execute or not? 🤔 Maybe in some weird cases yes, but why is this the default behaviour?

dai-shi:
> At the point when equalityFn gets called the values in the store already changed, right?

No, it shouldn't change the value if equalityFn()=true. The value can be anything and it has nothing to do with the store. It's derived though.

Think about this simple case:
```js
const equalityFn = deepEqual
const state1 = { a: 1 }
const state2 = { a: 1 }
const state3 = { a: 1 }

let state = state1
if (!equalityFn(state, state2)) state = state2
if (!equalityFn(state, state3)) state = state3

state === state1 // this should be true
```

dios-david:
That's correct, my issue is different than what you mentioned above.

Let see this example:
```ts
const { setState, subscribe } = create(() => ({ value: 0 }));

subscribe(
  () => {},
  (state) => state.value,
  (prev, next) => {
    console.log(`prev: ${prev}, next: ${next}`);
    return true;
  }
);
  
setState({ value: 1 });  // equalityFn logs: "prev: 0, next: 1", which is correct
console.log(getState()); // the current state is: { value: 1 }

setState({ value: 2 });  // equalityFn logs: "prev: 0, next: 2", but it should log "prev: 1, next: 2". it receives `prev: 0` because the internal `currentSlice` in the listener did not get updated at the previous change
console.log(getState()); // the current state is: { value: 2 }

```

dai-shi:
Thanks for the example.

> If this is not how you intended to use the subscription logic, then do you have any suggestions how can I implement something similar using zustand?

Yeah, this is a valid question.

Like I said, equalityFn should be used to check equality and keep the previous value unchanged. To be fair, I presume this behavior is not zustand special, for example, it should be the same in redux. So, my suggestion is listen to all changes and run condition on your end.

```js
let prevState;
subscribe((state) => {
  if (condition(prevState, state)) {
    // do something
  }
  prevState = state;
});
```

Hope this helps!

dios-david:
What do you think about passing `prevState` to the listener callback? e.g.:
```ts
subscribe((state, prevState) => {});
```
I'm happy to implement it and open a PR. I think thing would make things a lot more easier for zustand users.

dai-shi:
Unfortunately, that wouldn't help your case because the way prevState is updated won't change. What it means is it should only and always call listeners when we change currentState **and** prevState.

If the previous suggestion doesn't fit wit your use case, you might want to explain your example again. Why do you want to change prevState but not notified?

JeremyRH:
There seems to be unintuitive behavior. The current slice is not updated the same way as state. [State is updated after a === check](https://github.com/pmndrs/zustand/blob/558845d99434c433faf6edc368bb262beea2456e/src/vanilla.ts#L44).

I think this behavior should be consistent, either always update both after a === check or update both after an equalityFn check. The latter seems wrong because you can have different equalityFns per subscriber. I'm in favor of @dios-david 's change but @dai-shi is right, it may introduce other undesired behavior.

dai-shi:
@JeremyRH 

> There seems to be unintuitive behavior. The current slice is not updated the same way as state. [State is updated after a === check](https://github.com/pmndrs/zustand/blob/558845d99434c433faf6edc368bb262beea2456e/src/vanilla.ts#L44).

I disagree. The current behavior is consistent. Both cases update the value if it's not equal (`!==` or `!equalityFn()`), and notify it. If equal, no side effects. This **is** important.

dios-david:
> I think this behavior should be consistent, either always update both after a === check or update both after an equalityFn check. The latter seems wrong because you can have different equalityFns per subscriber.

I totally agree. These inconsistencies results in a really bad developer experience.
To be honest, I still don't see the reason behind the current implementation. I think it's either faulty or counterintuitive.

> it may introduce other undesired behavior

I think it's important to mention here that all of the existing tests are passing with my changes, so technically you should not be afraid of any *undesired behaviour*. Otherwise, how do you guys make sure that the project keeps the *expected behaviour* over the time, if the *expected behaviour* is not covered with tests? 🤔 Most importantly, how the users of this library can make sure?

I know sometimes it can be difficult to maintain the quality of testing while the focus is mainly on improving the product itself, but I suggest you to review your testing practices to make sure you are testing the right things in the right way, so you can avoid being in a position which makes people (even the maintainers) uncertain.

For example, [this test is broken](https://github.com/pmndrs/zustand/blob/558845d99434c433faf6edc368bb262beea2456e/tests/basic.test.tsx#L432) but the test suite doesn't fail because the assertion here doesn't get called at all - this should better be tested with [.toHaveBeenCalled()](https://jestjs.io/docs/en/expect#tohavebeencalled) which would have caught this issue.
I'm not sure what is broken here actually: is the test outdated and did not get updated when the functionality changed at some point in the past? Or is the actual functionality broken here, which did not get caught due to the faulty implementation of the test? None of these options sound promising to be honest.

Personally, I think zustand is one of the greatest state management solutions I have ever seen. In the past few weeks I was working on migrating our project to use zustand and I quite enjoyed the journey. I was amazed how much we could simplify the state management logic in our applications with zustand.

When I ran into this issue I thought it would be cool to fix it and contribute it back to the project. But unfortunately, as I see our opinions do not match about some key concepts, and I found you quite closed for considering changes or adding support for new use cases, as you are not sure what issues could be introduced with these changes. The existing tests don't cover the expected functionalities, or they are broken.

Due to these reasons I'm a bit uncomfortable continuing using zustand in it's current form. I still think this project in amazing and I'm really grateful to you for making this available for us! But I think at this point the best solution is to fork this project and tailor it for our needs. Thanks again, and good luck with v4, I will definitely check that out when it gets released!

dai-shi:
@dios-david I'm so sorry to make you uncomfortable. I guess our tests are not very thorough for vanilla.ts. I will try to add one based on this discussion. It's a bit unfortunate that you leave. Your knowledge around the tests seems far more to exceed mine, and be helpful for the project.

dai-shi:
> What do you think about passing `prevState` to the listener callback? e.g.:
> ```ts
> subscribe((state, prevState) => {});
> ```


My big apologies. I think I misunderstood this suggestion. This has nothing to do with selector or equalityFn. So, my previous comment on this is invalid. This suggestion can be considered as an additional feature we could support.

JeremyRH:
@dios-david I was a maintainer on this project in the past. I like your change and would merge it if @dai-shi was comfortable. Also your suggestion for adding previous state to the listener callback seems safe.

Forking Zustand will probably make things a lot easier for your project. I fork small libraries occasionally and it usually results in learning almost everything about the library. We would appreciate it if you came back after a while with feedback or suggestions.

I'll take a look at the test. I think I wrote that.

dai-shi:
@JeremyRH As you wrote the original code and if you say the new behavior is the intention (and the current behavior is a bug), I should accept it. Especially because @dios-david seems to have some issues with it, which unfortunately I don't quite follow why (I would like to understand the use case.)

But before that, let me create a PR to add some tests why I think the current behavior is consistent. (And, I may find myself wrong while doing it.)

PR #236, opened 2020-11-07
Fix missing import in readme example

Pocket-titan:
Small thingy, the `StateCreator` import was missing in the "How to type immer middleware in TypeScript" example.


dai-shi (approved):
Nice catch! Thanks.

PR #237, opened 2020-11-08
add equality check test

dai-shi:
We had a discussion about equality check in #234.
Unfortunately, two opinions are simply conflicting.

@JeremyRH I made a test to illustrate why I think the current behavior is intuitive for me.
FWIW, react-redux's implementation should be the same behavior.
https://github.com/reduxjs/react-redux/blob/94fcab6002989f1e106d6edd1b41dbd6599e6785/src/hooks/useSelector.js#L60-L64

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -436,6 +436,30 @@ it('can subscribe to the store', () => {
   unsub()
   expect(listener).toHaveBeenCalledTimes(1)
   expect(listener).toHaveBeenCalledWith(initialState.value + 2)
+
+  // Should keep consistent behavior with equality check
+  const isRoughEqual = (x: number, y: number) => Math.abs(x - y) < 1
+  setState({ value: 0 })
+  listener.mockReset()
+  const listener2 = jest.fn()
+  let prevValue = getState().value
+  unsub = subscribe((s) => {
+    if (isRoughEqual(prevValue, s.value)) {
+      // skip assuming values are equal
+      return
+    }
+    listener(s.value)
+    prevValue = s.value
+  })
+  const unsub2 = subscribe(listener2, (s) => s.value, isRoughEqual as any)
+  setState({ value: 0.5 })
+  setState({ value: 1 })
+  unsub()
+  unsub2()
+  expect(listener).toHaveBeenCalledTimes(1)
+  expect(listener).toHaveBeenCalledWith(1)
+  expect(listener2).toHaveBeenCalledTimes(1)
+  expect(listener2).toHaveBeenCalledWith(1)
 })
 
 it('can destroy the store', () => {


dios-david:
What is your opinion about adding a 4th param for `subscribe` and a 3rd one for `useState` which could toggle between these behaviours?
E.g. `compareWithPreviousSlice` / `compareWithLastMatchedSlice` or something like that

dai-shi:
@dios-david That might work with less friction.
But, before going into it, would you help me understand the use case?
The biggest issue for me is to understand when the current behavior is undesired.
I'm getting to understand that there are such cases.
Contradictory to your intuition, the current behavior is more intuitive for me because `equalitfyFn` = true means I should throw away the new value and keep the original.
It feels like your use case would be similar to something like [transient updates](https://github.com/pmndrs/zustand/tree/558845d99434c433faf6edc368bb262beea2456e#transient-updates-for-often-occuring-state-changes)?

Note: changing the current behavior is an option too based on our recent discussion. (in this case, I'd prefer to rename `equalityFn` to something else like `shouldNotNotify`.)

dai-shi:
@JeremyRH Would you also check this please?

JeremyRH:
Updating the current slice after the equalityFn is subjective. Some people might expect the last value to always be updated and if you were to call `api.getState` in the equality function, you would see the state is always updated. It can seem strange that the equalityFn prevents the slice from being updated but not the state.

I don't like the equalityFn and would like to remove it. It seems like we have to make a choice on subjective behavior.

Two middlewares could be created that cover both behaviors. The user would then be able to choose. Unfortunately this would be a large breaking change. I think it's not worth it to change anything unless we plan a next major version.

dai-shi:
I actually tried to remove the scoped subscription from the vanilla api, but that braking change is not acceptable to Paul, because it's useful for some cases.

So, what is your suggestion for now? Would you like to work on for the next major version?

JeremyRH:
I think we can move the scoped subscription and equality function to an optional helper that wraps listeners. [Here is an example of how it might look to write your own listener wrapper.](https://codesandbox.io/s/little-frog-rku2y?file=/src/index.js)

I think this can be done in a more functional way and there needs to be a way to let the React hook use this. I'm not sure how it will all work, this is just an idea.

dai-shi:
Got it. Dropping the scoped subscription was my idea for v3, which was not accepted. But, I didn't propose the optional helper functions. It would reduce the bundle size too.

If we plan to deprecate the scoped subscription in v4, we should provide the helper in v3 and show deprecation notice in vanilla api. Would you like to work on it? @JeremyRH 

It's actually important at this point, as we are making a breaking change in #241 .

Also, we should hear @drcmda comments.

dai-shi:
Okay, even the current behavior were subjective, we should keep it unless it is a bug. Let me merge this in, because I want to check if #241 doesn't break the behavior.

@JeremyRH If you've got time please file a new issue with your idea for discussion. My position is kind of neutral on the scoped subscription (because it's not used for useStore). So, we'd need to discuss more broadly. At least, Paul should have opinions.

Issue #240, opened 2020-11-13
Expose previous value in subscribe


Tirzono:
It would be very useful to also have access to the previous value when subscribing to the store:

```javascript
const unsub2 = useStore.subscribe((paw, previousPaw) => console.log(paw, previousPaw), state => state.paw)
```

dai-shi:
Hi, thanks for opening this up.

@dios-david suggested as mitigation in #234 and I think it's ok: https://github.com/pmndrs/zustand/pull/234#issuecomment-723390078 (at least without selector/equalityFn)

Issue #208, opened 2020-10-15
store.subscribe TS error


lukejagodzinski:
Hey, I know that there were two issues with this problem reported some time ago and both were closed as solved but they are not. Here is reproduction and comments explaining what's wrong:

```ts
import create from "zustand";

type Mode = "FIRST" | "SECOND" | "THIRD";

export type State = {
  mode: Mode | null;
  name: string;
};

export const useStore = create<State>(() => ({
  mode: null,
  name: "name"
}));

useStore.subscribe(
  (mode: State["mode"]) => {
    // When I try to provide type manually then
    // the selector function gives me an error
    console.log(mode);
  },
  (state) => state.mode
);

useStore.subscribe(
  (mode) => {
    // The `mode` variable is `unknown` here and
    // I don't get any benefit of type system
    console.log(mode);
  },
  (state) => state.mode
);

useStore.subscribe<State["mode"]>(
  (mode) => {
    // This works but that could be done automatically.
    console.log(mode);
  },
  (state) => state.mode
);
```

Source: https://codesandbox.io/s/sad-shamir-9ezqu?file=/src/index.ts

Any plants to solve this issue?

dai-shi:
There are two issues to note.
1. subscribe with selector returns `null` when error occurs. So, `null` is special here, which conflicts with `Mode` in your case.
2. With context sensitive function, TS can't infer types.

The first one is only problem when slice value has `null`. See this example without `null`: https://codesandbox.io/s/friendly-curie-txfx1
We might be able to come up with better one for this, but I'm not sure. (at least without breaking change, it wouldn't be possible.)

The second one is not solvable by us. It's TS's behavior. Luckily, you found a solution! I didn't know you can attach slice type in `subscribe`.

lukejagodzinski:
@dai-shi oh I didn't know that `null` is a special case. There is no mention of that in docs. In theory I could use `undefined` but that's kinda hacky to deal with the library problem in that way. I should be able to use whatever type I want in the state.

Also providing error as the second argument is kinda breaking the well established callback structure which always was like:
```ts
(error, result) => {}
```
I guess, it could provide error info in some other way or just don't pass it at all. In the code, I can see that it's only being used when selector or equality function fails: https://github.com/pmndrs/zustand/blob/f32aa02d16f001e659131b9f1f2a27f3680d383d/src/vanilla.ts#L64
[...]

PR #241, opened 2020-11-13
Expose previous value in subscribe

Tirzono:
Decided to go ahead with my proposal of #240. I found the second argument was used for the error in case there's an error, but that's undocumented, so I am not sure how much of a breaking change this would be.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -4,8 +4,8 @@ export type PartialState<T extends State> =
   | ((state: T) => Partial<T>)
 export type StateSelector<T extends State, U> = (state: T) => U
 export type EqualityChecker<T> = (state: T, newState: unknown) => boolean
-export type StateListener<T> = (state: T) => void
-export type StateSliceListener<T> = (state: T | null, error?: Error) => void
+export type StateListener<T> = (state: T, previousState: T) => void
+export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
 export interface Subscribe<T extends State> {
   (listener: StateListener<T>): () => void
   <StateSlice>(
@@ -41,10 +41,11 @@ export default function create<TState extends State>(
   const setState: SetState<TState> = (partial, replace) => {
     const nextState = typeof partial === 'function' ? partial(state) : partial
     if (nextState !== state) {
+      const previousState = state
       state = replace
         ? (nextState as TState)
         : Object.assign({}, state, nextState)
-      listeners.forEach((listener) => listener(state))
+      listeners.forEach((listener) => listener(state, previousState))
     }
   }
 
@@ -57,16 +58,10 @@ export default function create<TState extends State>(
   ) => {
     let currentSlice: StateSlice = selector(state)
     function listenerToAdd() {
-      // Selector or equality function could throw but we don't want to stop
-      // the listener from being called.
-      // https://github.com/react-spring/zustand/pull/37
-      try {
-        const newStateSlice = selector(state)
-        if (!equalityFn(currentSlice, newStateSlice)) {
-          listener((currentSlice = newStateSlice))
-        }
-      } catch (error) {
-        listener(null, error)
+      const nextSlice = selector(state)
+      if (!equalityFn(currentSlice, nextSlice)) {
+        const previousSlice = currentSlice
+        listener((currentSlice = nextSlice), previousSlice)
       }
     }
     listeners.add(listenerToAdd)
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -412,7 +412,10 @@ it('can subscribe to the store', () => {
   setState({ value: initialState.value + 1 })
   unsub()
   expect(listener).toHaveBeenCalledTimes(1)
-  expect(listener).toHaveBeenCalledWith(initialState.value + 1)
+  expect(listener).toHaveBeenCalledWith(
+    initialState.value + 1,
+    initialState.value
+  )
 
   // Should not be called when equality checker returns true
   unsub = subscribe(
@@ -435,7 +438,10 @@ it('can subscribe to the store', () => {
   setState({ value: initialState.value + 2 })
   unsub()
   expect(listener).toHaveBeenCalledTimes(1)
-  expect(listener).toHaveBeenCalledWith(initialState.value + 2)
+  expect(listener).toHaveBeenCalledWith(
+    initialState.value + 2,
+    initialState.value + 2
+  )
 
   // Should keep consistent behavior with equality check
   const isRoughEqual = (x: number, y: number) => Math.abs(x - y) < 1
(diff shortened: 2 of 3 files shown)


dai-shi:
We have some discussions in #208 and I personally think it's ok to make a breaking change (even without major version up as it's an undocumented behavior). I wish original authors respond about this. @drcmda @JeremyRH

src/vanilla.ts
```diff
@@ -5,7 +5,11 @@ export type PartialState<T extends State> =
 export type StateSelector<T extends State, U> = (state: T) => U
 export type EqualityChecker<T> = (state: T, newState: unknown) => boolean
 export type StateListener<T> = (state: T) => void
```
dai-shi, line 7:
```suggestion
export type StateListener<T> = (state: T, previousState: T) => void
```

Would we like to change this too?

src/vanilla.ts
```diff
@@ -5,7 +5,11 @@ export type PartialState<T extends State> =
...
-export type StateSliceListener<T> = (state: T | null, error?: Error) => void
+export type StateSliceListener<T> = (
+  state: T | null,
+  error: Error | null,
+  previousState: T | null
+) => void
```
dai-shi, line 12:
```suggestion
export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
```

If such a breaking change is allowed, I would prefer this.

It doesn't conflict with #37, because we no longer use this in useStore.

JeremyRH: This looks reasonably safe to me. The original issue was selectors could throw an error because they were called at unexpected times. We didn't want to silence all errors in case it was something they need to handle so we exposed it.

I think it's not used very much.

drcmda: i agree, it's probably more useful to have the previous state.

dai-shi: Thanks, guys.
One it's adjusted accordingly, we will release it as a minor update.
Although it's technically a breaking change in the vanilla api, the impact would be tiny.
- it's not documented in readme
- it should not be used very much

dai-shi (changes requested):
@Tirzono Would you work on two change requests?

src/vanilla.ts
Tirzono: I changed it to:

```javascript
export type StateSliceListener<T> = (slice: T | null, previousSlice: T | null) => void
```

Because when there is an error we still return `null, null`. Let me know if you want to return something else instead or remove the try-except completely.

Tirzono:
@dai-shi of course, not a problem. I pushed the changes, let me know if you think it needs more changes.

src/vanilla.ts
```diff
@@ -63,10 +67,11 @@ export default function create<TState extends State>(
...
         const newStateSlice = selector(state)
         if (!equalityFn(currentSlice, newStateSlice)) {
-          listener((currentSlice = newStateSlice))
+          const previousSlice = currentSlice
+          listener((currentSlice = newStateSlice), previousSlice)
         }
-      } catch (error) {
-        listener(null, error)
+      } catch {
+        listener(null, null)
```
JeremyRH, line 74: I don't think we need to call listeners after an error in vanilla. We needed to do it to force React to re-render and that logic still exists in index.ts. The listeners added in index.ts are also wrapped in a try-catch so they won't throw.

My suggestion: remove the listener call in the catch.

Tirzono: Done :)

dai-shi: Yeah, I should have clarified more about the context. Check #208 and we think it's better not to try-catch.

src/vanilla.ts
```diff
@@ -41,10 +41,17 @@ export default function create<TState extends State>(
...
       state = replace
         ? (nextState as TState)
         : Object.assign({}, state, nextState)
-      listeners.forEach((listener) => listener(state))
+      listeners.forEach((listener) => {
+        try {
+          listener(state, previousState)
+        } catch (e) {
+          console.error(e)
+        }
```
dai-shi, line 53:
```suggestion
        listener(state, previousState)
```

We don't need to catch errors in vanilla.js. cc @drcmda

src/vanilla.ts
```diff
@@ -4,8 +4,8 @@ export type PartialState<T extends State> =
   | ((state: T) => Partial<T>)
 export type StateSelector<T extends State, U> = (state: T) => U
 export type EqualityChecker<T> = (state: T, newState: unknown) => boolean
-export type StateListener<T> = (state: T) => void
-export type StateSliceListener<T> = (state: T | null, error?: Error) => void
+export type StateListener<T> = (state: T, previousState: T) => void
+export type StateSliceListener<T> = (state: T, previousState: T) => void
```
dai-shi, line 8:
```suggestion
export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
```

Let's just rename it.

src/vanilla.ts
```diff
@@ -57,16 +64,10 @@ export default function create<TState extends State>(
...
-        const newStateSlice = selector(state)
-        if (!equalityFn(currentSlice, newStateSlice)) {
-          listener((currentSlice = newStateSlice))
-        }
-      } catch (error) {
-        listener(null, error)
+      const newStateSlice = selector(state)
+      if (!equalityFn(currentSlice, newStateSlice)) {
+        const previousSlice = currentSlice
+        listener((currentSlice = newStateSlice), previousSlice)
```
dai-shi, line 70:
```suggestion
      const nextSlice = selector(state)
      if (!equalityFn(currentSlice, nextSlice)) {
        const previousSlice = currentSlice
        listener((currentSlice = nextSlice), previousSlice)
```

Just for naming consistency.

dai-shi (approved):
Thanks!

dai-shi:
@Tirzono I merged a test #237 into master. Would you merge master and check if it passes?

tests/basic.test.tsx
```diff
@@ -459,7 +465,7 @@ it('can subscribe to the store', () => {
   expect(listener).toHaveBeenCalledTimes(1)
   expect(listener).toHaveBeenCalledWith(1)
   expect(listener2).toHaveBeenCalledTimes(1)
-  expect(listener2).toHaveBeenCalledWith(1)
+  expect(listener2).toHaveBeenCalledWith(1, 0)
```
Tirzono, line 468: I rebased my branch and only this test was failing, but I guess that's expected so I changed it.

tests/basic.test.tsx
```diff
@@ -459,7 +465,7 @@ it('can subscribe to the store', () => {
   expect(listener).toHaveBeenCalledTimes(1)
   expect(listener).toHaveBeenCalledWith(1)
```
Tirzono, line 466: I am surprised this test passed, but maybe I am seeing something wrong (@dai-shi)

dai-shi: Ah, nice catch. Could you change L457 to `listener(s.value, prevValue)` please?

Tirzono: Done 🐻

dai-shi:
Thanks! Merging this.

PR #246, opened 2020-11-24
persist middleware to check localStorage

### LINKED ISSUE #245: Persist middleware in server rendered environments
Hi there!

While working with the new persist middleware I figured using it in a server rendered environment (in my case, next.js) it would throw as the default localStorage api expectedly isn't available on the server.

This was easily fixed by passing a dummy storage api but I wondered whenever this should actually be a default within zustand to not confuse users adopting the new middleware within a next.js/nuxt.js project. Not entirely sure whenever this is intended or can be considered a 'bug', but it might help new users if it would not break any server render by default.

here is an example of what I did in my app

```
const dummyStorageApi = {
  getItem: () => null,
  setItem: () => undefined,
}

const useStore = create(
  persist(
    set => ({
      user: null,
      updateUser: user => set({ user }),
    }),
    {
      name: 'someStoragekey',
      storage: typeof window !== 'undefined' ? window.localStorage : dummyStorageApi,
    },
  ),
)

```

dai-shi:
close #245 

@RobbyUitbeijerse How does this work?

@AnatoleLucet Can review?

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -134,7 +134,12 @@ export const persist = <S extends State>(
 ) => (set: SetState<S>, get: GetState<S>, api: StoreApi<S>): S => {
   const {
     name,
-    storage = localStorage,
+    storage = typeof localStorage !== 'undefined'
+      ? localStorage
+      : {
+          getItem: () => null,
+          setItem: () => {},
+        },
     serialize = JSON.stringify,
     deserialize = JSON.parse,
   } = options || {}


AnatoleLucet (approved):

RobbyUitbeijerse:
@dai-shi looks good on my end! Thanks for the swift solution :-)

dai-shi:
Let's merge this, but no plan to release immediately, unless someone suggests it.

dai-shi:
Thanks, guys!

ianstormtaylor:
@dai-shi can you release this? Ran into the same right away when first trying this lib—`persist` errors when using with Next.js because of its server rendering.

dai-shi:
@ianstormtaylor There's an ongoing discussion in #248 which may or may not affect this (hopefully only internally), so give us some more time.
Meanwhile, you can work it around with something like this: https://github.com/pmndrs/zustand/issues/245#issue-749551108

Issue #238, opened 2020-11-09
store.setState() api method doesn't trigger persistance


pauliusef:
I'm trying to use persistance middleware in React, however I'm doing store updates from outside a react component by calling `useStore.setState({ property: value })`. This function call sets the state and appears in Redux devtools without persisting the update in `localStorage`.

dai-shi:
Right, currently it's only patching `set` in store creator.

@AnatoleLucet thoughts?

AnatoleLucet:
I'm not sure. We'll definitely need to modify some bits in `vanilla.ts -> create()`. An option would be to modify the store's internal api object via the middleware (same thing as `set`) and then use it in `create()`. But I'm not sure how this could be done, I'll check later in the day.

dai-shi:
Yeah, I think that's only the option.

`devtools` middleware has this code:
```ts
    const savedSetState = api.setState
    api.setState = (state: PartialState<S>, replace?: boolean) => {
      savedSetState(state, replace)
      api.devtools.send(api.devtools.prefix + 'setState', api.getState())
    }
```

I'm pretty sure if there's just one `persist` middleware alone, it works fine. When we chain multiple middleware, we need to be careful.

AnatoleLucet:
@dai-shi @pauliusef I think this issue can be closed now that #251 is merged.

dai-shi:
Yeah, thanks!

pauliusef:
Thank you, will update and test it out!

dai-shi:
(Sorry, it's not released yet. You can try the codesandbox build in the PR, though.

arryanggaputra:
Is there any way to exclude some key from persist?

dai-shi:
#254 is merged, but we are still discussing the overall design in #248. I would like to hold on releasing before we decide our plan.

arryanggaputra:
@dai-shi  it's weird, store.setState actually persistent.

dai-shi:
Isn't it because you use `set` from the create callback function?

arryanggaputra:
@dai-shi  no, I'm using `store.setState` . This is how I define the store

```typescript
export const _accountDataStore = create(
  persist<TAccountDataStore>(
    (set, get) => ({
      accountData: get()?.accountData || null,
      setAccountData: (data) => set({accountData: data}),
      removeAccountData: () => set({accountData: null}),
    }),
    {
      name: 'sharpmenu-account', // unique name
[...]

PR #251, opened 2020-12-03
Persisting when using the api.setState() method

Qcza:
Allows for persisting the store when using the api.setState method.

Related issue: https://github.com/pmndrs/zustand/issues/238

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -143,20 +143,20 @@ export const persist = <S extends State>(
     serialize = JSON.stringify,
     deserialize = JSON.parse,
   } = options || {}
-  
-  const setItem = async () => storage.setItem(name, await serialize(get()));
 
-  const savedSetState = api.setState;
+  const setItem = async () => storage.setItem(name, await serialize(get()))
+
+  const savedSetState = api.setState
 
   api.setState = (state: PartialState<S>, replace?: boolean) => {
-    savedSetState(state, replace);
-    (async () => await setItem())();
-  };
+    savedSetState(state, replace)
+    ;(async () => await setItem())()
+  }
 
   const state = config(
     (payload) => {
-      set(payload);
-      (async () => await setItem())();
+      set(payload)
+      ;(async () => await setItem())()
     },
     get,
     api
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -148,7 +148,7 @@ export const persist = <S extends State>(
 
   const savedSetState = api.setState
 
-  api.setState = (state: PartialState<S>, replace?: boolean) => {
+  api.setState = (state, replace) => {
     savedSetState(state, replace)
     ;(async () => await setItem())()
   }
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -143,13 +143,20 @@ export const persist = <S extends State>(
     serialize = JSON.stringify,
     deserialize = JSON.parse,
   } = options || {}
+  
+  const setItem = async () => storage.setItem(name, await serialize(get()));
+
+  const savedSetState = api.setState;
+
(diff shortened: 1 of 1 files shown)


dai-shi:
Hi! Can you try if it works with other middlewares?

@AnatoleLucet can review?

src/middleware.ts
```diff
@@ -143,13 +143,20 @@ export const persist = <S extends State>(
...
+  };
 
   const state = config(
     (payload) => {
-      set(payload)
-      ;(async () => {
-        storage.setItem(name, await serialize(get()))
-      })()
+      set(payload);
+      (async () => await setItem())();
```
dai-shi, line 159: Not a full review, but just noticed this and L153 can be:
```suggestion
      setItem()
```

Qcza:
> Hi! Can you try if it works with other middlewares?
> 
> @AnatoleLucet can review?

Just with the devtools, and it works well.

> Hi! Can you try if it works with other middlewares?
> 

Only with devtools and it works well. 

This is the same approach as used in the devtools middleware, so I presume if it is used there and not messing another middleware it may be used in the persist middleware.

src/middleware.ts
```diff
@@ -144,12 +144,19 @@ export const persist = <S extends State>(
     deserialize = JSON.parse,
   } = options || {}
 
+  const setItem = async () => storage.setItem(name, await serialize(get()))
+
+  const savedSetState = api.setState
+
+  api.setState = (state: PartialState<S>, replace?: boolean) => {
```
AnatoleLucet, line 151: The types are redundant
```suggestion
  api.setState = (state, replace) => {
```

AnatoleLucet (commented):
lgtm. We just need to check if it overrides other middlewares setState or not.

dai-shi:
> Tested it only with devtools and it works well.

Would you mind trying persist+devtools+redux, just in case? It might be a tricky combination.

dai-shi:
(Let me remove travis config...

src/middleware.ts
```diff
@@ -144,12 +144,19 @@ export const persist = <S extends State>(
     deserialize = JSON.parse,
   } = options || {}
 
+  const setItem = async () => storage.setItem(name, await serialize(get()))
+
+  const savedSetState = api.setState
+
+  api.setState = (state, replace) => {
+    savedSetState(state, replace)
+    ;(async () => await setItem())()
```
dai-shi, line 153:
```suggestion
    setItem()
```

src/middleware.ts
```diff
@@ -144,12 +144,19 @@ export const persist = <S extends State>(
...
+  }
+
   const state = config(
     (payload) => {
       set(payload)
-      ;(async () => {
-        storage.setItem(name, await serialize(get()))
-      })()
+      ;(async () => await setItem())()
```
dai-shi, line 159:
```suggestion
      setItem()
```

src/middleware.ts
dai-shi, line 153:
```suggestion
    setItem()
```

src/middleware.ts
dai-shi, line 159:
```suggestion
      setItem()
```

dai-shi (commented):
I think this works?

Qcza:
> > Tested it only with devtools and it works well.
> 
> Would you mind trying persist+devtools+redux, just in case? It might be a tricky combination.

Just tested it with devtools+persist+redux and it looks like it works well.

dai-shi:
Thanks for confirming! Nice.

dai-shi (approved):

dai-shi:
Merging this. We will be waiting for #248 before releasing.

PR #254, opened 2020-12-05
feat: blacklist and whitelist for the persist middleware

Qcza:
Useful option if you want to omit some fields from persisting.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -126,6 +126,8 @@ type PersistOptions<S> = {
   storage?: StateStorage
   serialize?: (state: S) => string | Promise<string>
   deserialize?: (str: string) => S | Promise<S>
+  blacklist?: (keyof S)[]
+  whitelist?: (keyof S)[]
 }
 
 export const persist = <S extends State>(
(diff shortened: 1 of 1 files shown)


dai-shi:
Personally, I would avoid adding many options to persist middleware. People would eventually want to add blacklist/whitelist for nested objects.
Wouldn't serialize option work for your case?

```js
persiste({
  name: 'foo',
  serialize: (state) => {
    const stateToStringify = { ...state }
    if (blacklist) {
      blacklist.forEach((key) => delete stateToStringify[key])
    }
    return JSON.stringify(stateToStringify)
  },
})
```

Let's see how @AnatoleLucet wants as he's the original author.

Qcza:
> Wouldn't serialize option work for your case?
> 

This is the approach I took since and it works well. But I think having the option may be more useful and clear to use for people who are familiar with the redux-persist tool.

AnatoleLucet (approved):
I think it's going to confuse users to modify the serializer to do a black/whitelist, but it's still an available option if (as @dai-shi said) someone wants to apply a black/whitelist on nested objects.

dai-shi:
> clear to use for people who are familiar with the redux-persist tool.

Okay, this is convincing. I'm not very familiar with redux-persist, but if we are to only support one-level black/whitelist options, this will be good. Let's recommend customizing serialize for nested ones, as @AnatoleLucet suggested too.

I will wait for some moments and if there aren't any other comments, will merge this.

dai-shi (approved):
Thanks for the work and discussion!

TkDodo:
seems like I’m a bit late to the party, but i think blocklist/allowlist would have been a better naming :/

thgh:
Hi, came here to say the same: black/white are loaded terms and there are alternatives like include/exclude & block/allow.

jacobbuck:
Agreed, especially given that the `master` branch has been recently renamed to `main`.

PR #257, opened 2020-12-12
Use Draft<T> in immer typescript middleware

ckknight:
This was discovered because I defined my state with a bunch of `readonly`s and receiving the error that the property wasn't updated.

See also: https://immerjs.github.io/immer/docs/typescript


dai-shi (approved):
Oh, nice catch. I wasn't aware of it.
Thanks!

PR #266, opened 2020-12-29
fixed an issue in type infering while using eq function in hooks

Kuchasz:
Typescript could not properly infer type of newState in state equality function.

```js
//new is typed as unknown giving type errors
const state = useSomeState(s => s.item, (oldItem, newItem) => oldItem.id === newItem.id)
```
After that change above snippet compiles and works properly.

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -1,6 +1,11 @@
-import React from 'react'
-import ReactDOM from 'react-dom'
-import { act, cleanup, fireEvent, render } from '@testing-library/react'
+import React from "react";
+import ReactDOM from "react-dom";
+import {
+  act,
+  cleanup,
+  fireEvent,
+  render
+  } from "@testing-library/react";
 import create, {
   State,
   StateListener,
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Ah, `unknown` was a bit intentional, but I didn't know the type inference issue. I think this fix is reasonable. (wondering if we can create a small test for this, or use safer types in tests to reveal this issue.)

Kuchasz:
I can tweak one of existing tests or write another one that will give compile-time error in such scenario.

dai-shi:
That would be nice.

Kuchasz:
That one test that verifies equalityFunction will now throw at compile time when eqFunction parameters won't we properly typed.

tests/basic.test.tsx
```diff
@@ -1,6 +1,11 @@
...
-import ReactDOM from 'react-dom'
-import { act, cleanup, fireEvent, render } from '@testing-library/react'
+import React from "react";
+import ReactDOM from "react-dom";
+import {
+  act,
+  cleanup,
+  fireEvent,
+  render
+  } from "@testing-library/react";
```
dai-shi, line 8:
```suggestion
import React from 'react'
import ReactDOM from 'react-dom'
import { act, cleanup, fireEvent, render } from '@testing-library/react'
```

reverting this.

dai-shi:
Thanks for your contribution!

PR #274, opened 2021-01-06
docs: add comment regarding TS State typing

### LINKED ISSUE #213: Type error when using interface instead of type in Typescript
Typescript is throwing error when I use interface instead of type. Take this example:

```ts
export interface DashboardSummary {
  added: number
  updated: number
  active: number
  invite: number
}

const initialState: DashboardSummary = {
  invite: 0,
  active: 0,
  added: 0,
  updated: 0,
}

const useDashboardStore = create(
  combine(initialState, set => ({ // error on `initialState`
    fetch: async () => {
      const res = await axios.get<DashboardSummary>('/api/user/dashboard')

      if (res.status === 200) {
        set(res.data) // error on `res.data`
      }
    },
  }))
)
```

On hover `initialState`, typescript says:
```
Argument of type 'DashboardSummary' is not assignable to parameter of type 'Record<string | number | symbol, unknown>'.
  Index signature is missing in type 'DashboardSummary'.ts(2345)
```

Is it possible to fix it?

3nvi:
## Background

The `State` typing provided to the `create` function must extend `Record`. Thus, the `State` typing that users provide needs to be a `type` and not an `interface`. This PR adds a small comment in the TS Recipe to make that clear.

Closes #213 

## Changes

- Add comment in TS recipe's codee

## Testing

- Github markdown preview

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -379,6 +379,8 @@ devtools takes the store function as its first argument, optionally you can name
 ## TypeScript
 
 ```tsx
+// Make sure that `State` is a `type` and not an `interface`, else the typescript compiler 
+// will complain about a missing index signature
 type State = {
   bears: number
   increase: (by: number) => void
--- a/readme.md
+++ b/readme.md
@@ -379,9 +379,23 @@ devtools takes the store function as its first argument, optionally you can name
 ## TypeScript
 
 ```tsx
-// Make sure that `State` is a `type` and not an `interface`, else the typescript compiler 
-// will complain about a missing index signature
 type State = {
+  bears: number
+  increase: (by: number) => void
+}
+
+const useStore = create<State>(set => ({
+  bears: 0,
+  increase: (by) => set(state => ({ bears: state.bears + by })),
+}))
+```
+
+You can also use an `interface`:
+
+```tsx
+import { State as ZustandState } from 'zustand';
+
+interface State extends ZustandState {
   bears: number
   increase: (by: number) => void
 }
--- a/readme.md
+++ b/readme.md
@@ -399,11 +399,6 @@ interface BearState extends State {
   bears: number
   increase: (by: number) => void
 }
-
-const useStore = create<BearState>(set => ({
-  bears: 0,
-  increase: (by) => set(state => ({ bears: state.bears + by })),
-}))
 ```
 
 Or, use `combine` and let tsc infer types.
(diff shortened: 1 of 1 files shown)


dai-shi:
@3nvi 
Thanks for opening this up.
I don't think we need to force `type`. We can use `interface` as long as it extends the Record.
```ts
import { State } from 'zustand'

export interface MyState extends State {
  text: string
}
```
What do you think?

3nvi:
> @3nvi
> Thanks for opening this up.
> I don't think we need to force `type`. We can use `interface` as long as it extends the Record.
> 
> ```ts
> import { State } from 'zustand'
> 
> export interface MyState extends State {
>   text: string
> }
> ```
> 
> What do you think?

Oh yeah, you're right. Dunno how I missed that. I've updated the PR 😄

readme.md
```diff
@@ -390,6 +390,22 @@ const useStore = create<State>(set => ({
...
+interface State extends ZustandState {
+  bears: number
+  increase: (by: number) => void
+}
+
+const useStore = create<State>(set => ({
+  bears: 0,
+  increase: (by) => set(state => ({ bears: state.bears + by })),
+}))
+```
```
dai-shi, line 407: It looks a bit long, how about this?

```tsx
import { State } from 'zustand'

interface BearState extends State {
  bears: number
  increase: (by: number) => void
}
```

3nvi: I actually had it like that, but thought it would be better to keep naming consistent (since `State` used to be defined by the user, while now it's a named import).

In any case, I've updated the example 😃

dai-shi: It's up to developers to rename things, but in the readme we never add `Zustand` prefix, so it's more consistent.

readme.md
```diff
@@ -390,6 +390,22 @@ const useStore = create<State>(set => ({
...
+
+interface BearState extends State {
+  bears: number
+  increase: (by: number) => void
+}
+
+const useStore = create<BearState>(set => ({
+  bears: 0,
+  increase: (by) => set(state => ({ bears: state.bears + by })),
+}))
```
dai-shi, line 406:
```suggestion
```

Can we remove it? Hops it's clear without it.

dai-shi (approved):
Looks good. Thanks!

PR #275, opened 2021-01-06
feat(persist): improve ssr & add versions & new callback

### LINKED ISSUE #243: Versioned Persist Middleware
Hi, thanks for your great work on this library.

I made an update to the persist middleware to add version property to the stored state.
Would it be a good addition to the core library, or is it out of the scope?

Code here: https://gist.github.com/jvonasek/c6838243f205815077be032b1fa2e521

Works like usual, but with new verison property in the persist middleware options
```javascript
const useStore = create<State>(
  persist(
    (set) => ({
      theme:  Theme.Light,
      setTheme: (theme: Theme) => set(() => ({ theme })),
    }),
    {
      name: 'theme',
      version: 1,
    }
  )
)
```
Produces this state in the storage
```json
{
  "theme": "light",
  "_version": 1
}
```

If I change the shape of my state to something like this
```javascript
type State = {
  themes: [Theme, SubTheme]
  setThemes: (theme: Theme, subTheme: SubTheme) => void
}
```
My  state would normally become this, or if I didn't rename the theme property to themes, it would even break the application
```json
{
  "themes": ["dark","navy"],
  "theme": "light"
}
```
However, if I increase the version property in my persist state creator to 2, the original state will be purged and recreated
```json
{
  "themes": ["dark","navy"],
  "_version": 2
}
```

This can be further iterated on and add migration support, however I feel that's probably better as an external middleware.
Inspired by redux-persist

AnatoleLucet:
- add js doc for option object
- improve ssr support
- add a new version option #243
- refactor postRehydrationMiddleware to onRehydrateStorage

---

This ended up being more of a feature rich pr instead of a full refactor :smile:
Anyway this includes the two breaking changes I intended to add. The callback for the `storage` options, and the new `onRehydrateStorage` option.

I intend to add some more features on this middleware (I'll open a few issues soon), but no breaking changes in mind.

Close #248
Close #243

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -121,14 +121,51 @@ type StateStorage = {
   getItem: (name: string) => string | null | Promise<string | null>
   setItem: (name: string, value: string) => void | Promise<void>
 }
+type StorageValue<S> = { state: S; version: number }
 type PersistOptions<S> = {
+  /** Name of the storage (must be unique) */
   name: string
-  storage?: StateStorage
-  serialize?: (state: S) => string | Promise<string>
-  deserialize?: (str: string) => S | Promise<S>
+  /**
+   * A function returning a storage.
+   * The storage must fit `window.localStorage`'s api (or an async version of it).
+   * For example the storage could be `AsyncStorage` from React Native.
+   *
+   * @default () => localStorage
+   */
+  getStorage?: () => StateStorage
+  /**
+   * Use a custom serializer.
+   * The returned string will be stored in the storage.
+   *
+   * @default JSON.stringify
+   */
+  serialize?: (state: StorageValue<S>) => string | Promise<string>
+  /**
+   * Use a custom deserializer.
+   *
+   * @param str The storage's current value.
+   * @default JSON.parse
+   */
+  deserialize?: (str: string) => StorageValue<S> | Promise<S>
+  /**
+   * Prevent some items from being stored.
+   */
   blacklist?: (keyof S)[]
+  /**
+   * Only store the listed properties.
+   */
   whitelist?: (keyof S)[]
-  postRehydrationMiddleware?: () => void
+  /**
+   * A function returning another (optional) function.
+   * The main function will be called before the storage rehydration.
+   * The returned function will be called after the storage rehydration.
+   */
+  onRehydrateStorage?: (state: S) => ((state: S) => void) | void
+  /**
+   * If the stored state's version mismatch the one specified here, the storage will not be used.
+   * This is useful when adding a breaking change to your store.
+   */
+  version?: number
 }
 
 export const persist = <S extends State>(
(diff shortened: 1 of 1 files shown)


src/middleware.ts
```diff
@@ -122,13 +122,52 @@ type StateStorage = {
   setItem: (name: string, value: string) => void | Promise<void>
 }
 type PersistOptions<S> = {
+  /** Name of the storage (must be unique) */
   name: string
-  storage?: StateStorage
+  /**
+   * A function returning a storage.
+   * The storage must fit `window.localStorage`'s api.
```
dai-shi, line 129: More precisely, it's not localStorage api because it also supports async version of it.

src/middleware.ts
```diff
@@ -122,13 +122,52 @@ type StateStorage = {
...
+   * For example the storage could be `AsyncStorage` from React Native.
+   *
+   * @default () => localStorage
+   */
+  storage?: () => StateStorage
```
dai-shi, line 134:
```suggestion
  getStorage?: () => StateStorage
```
We could name like that. Not a strong opinion, but more precise and less breaking change.

src/middleware.ts
```diff
@@ -122,13 +122,52 @@ type StateStorage = {
...
+   * Use a custom deserializer.
+   * The returned value will be used as the new state.
+   *
+   * @param str The storage's current value.
+   * @default JSON.parse
+   */
   deserialize?: (str: string) => S | Promise<S>
+  /**
+   * Prevent some items from being stored.
+   * Nested properties can be accessed like so: `foo.bar.baz`
```
dai-shi, line 152: Nested properties are not supported, right?

src/middleware.ts
```diff
@@ -137,30 +176,45 @@ export const persist = <S extends State>(
...
     }
     if (blacklist) {
       blacklist.forEach((key) => delete state[key])
     }
-    storage.setItem(name, await serialize(state))
+
+    if (version) {
+      state = {
+        ...state,
+        __version: version,
```
dai-shi, line 213: This is dangerous, it could conflict with the same property name used in the state.
Can we make it safe?

```js
const data = {
  state,
  version,
}
storage.setItem(name, serialize(data))
```

It's another breaking change...

Or something like:
```js
storage.setItem(name, serialize(state))
storage.setItem(`${name}.version`, serialize(version))
```

dai-shi (commented):
Thanks for working on this.
Overall it looks good. Added some comments.

src/middleware.ts
AnatoleLucet: True, what about: "The storage must fit `window.localStorage`'s api (or an async version of it)." ?

src/middleware.ts
AnatoleLucet: Yes, I planned to add it but forgot to remove this comment

src/middleware.ts
AnatoleLucet:
```js
storage.setItem(name, serialize(state))
storage.setItem(`${name}.version`, serialize(version))
```
That's a good idea 👍

AnatoleLucet: After thinking about it, I think it's better to keep everything in one place.

src/middleware.ts
```diff
@@ -137,30 +174,38 @@ export const persist = <S extends State>(
...
+      Object.keys(state).forEach((key) => {
+        !whitelist.includes(key) && delete state[key]
+      })
     }
     if (blacklist) {
       blacklist.forEach((key) => delete state[key])
     }
-    storage.setItem(name, await serialize(state))
+
+    storage?.setItem(name, await serialize({ state, version }))
```
dai-shi, line 208:
```suggestion
    storage.setItem(name, await serialize({ state, version }))
```
Does TS complain without `?`? Maybe, so.

dai-shi (approved):
It looks good! Ready to merge and release this?

src/middleware.ts
AnatoleLucet: Yes, TS complain without it

AnatoleLucet:
Sure

Issue #276, opened 2021-01-07
zustand does not behave properly with useLayoutEffect


leobastiani:
IMO this codesandbox should not blink a red page

https://codesandbox.io/s/sad-currying-wwrm8

Is it an expected behavior?

dai-shi:
Nice catch. I'd say it's not a desired behavior, but it's the expected behavior from the current implementation.

https://github.com/pmndrs/zustand/blob/a4307f32b317a2fae7f7fb602f3e3d6d3f9ca848/src/index.ts#L83
It subscribes in `useEffect` and to mitigate when it's changed before subscription it checks at L105.

To fix this behavior, we'd need to change this useEffect to useLayoutEffect, but I'm not sure if it's worth it.

Note this behavior is only for the mount time, once subscription is made, useLayoutEffect should work properly, I suppose.

leobastiani:
Thank you!
It might be a hard decision to change it or not, but I do think it should be changed or at least to have a caveat
I think that `useIsoLayoutEffect` seems to be more appropriated there.

dai-shi:
Hm, in one of my libs, I used LayoutEffect as there's no other choices.

https://github.com/dai-shi/use-context-selector/blob/65eaf66e27555a78f3f46397f535c8b03f0b7899/src/index.ts#L169

useSubscription from react team uses passive Effect intentionally.

https://github.com/facebook/react/blob/d17086c7c813402a550d15a2f56dc43f1dbd1735/packages/use-subscription/src/useSubscription.js#L70

Maybe because it doesn't know the cost of `subscribe`.

Finally, react-redux uses LayoutEffect.

https://github.com/reduxjs/react-redux/blob/a9235530f4799dd4b2acb3cc65e9caf32efbc44b/src/hooks/useSelector.js#L55

----

As we know subscribe in zustand is pretty lightweight, we can change it to useIsomorphicLayoutEffect. (It would be obsoleted in v4-experimental #160 anyway.)

Anyone wants to open a PR?

leobastiani:
I'm working on it right now 🤗

leobastiani:
https://github.com/pmndrs/zustand/pull/280

PR #280, opened 2021-01-08
feat: works with useLayoutEffect

leobastiani:
close https://github.com/pmndrs/zustand/issues/276

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -148,8 +148,6 @@ it('only re-renders if selected state has changed', async () => {
 })
 
 it('re-renders with useLayoutEffect', async () => {
-  const useEffect = jest.spyOn(React, 'useEffect')
-
   const useStore = create(() => ({ state: false }))
 
   function Test() {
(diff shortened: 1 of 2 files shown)


tests/basic.test.tsx
```diff
@@ -147,6 +147,25 @@ it('only re-renders if selected state has changed', async () => {
...
+    React.useLayoutEffect(() => {
+      useStore.setState({ state: true })
+    }, [])
+    return <>{`${state}`}</>
+  }
+
+  const { container } = render(<Test />)
+
+  expect(container.innerHTML.trim()).toBe('true')
+  expect(useEffect).not.toHaveBeenCalled()
```
dai-shi, line 166: Hm, this is testing the implementation details. In the future we might be using `useEffect` for different purpose.
Can we test the behavior? How about adding useEffect in `Test()` and make sure it's called just once?

dai-shi (commented):
Thanks for your work. A comment on test.

tests/basic.test.tsx
leobastiani: After searching a lot, I actually found how to properly test this behavior. The problem with react-testing-library is that it flushes all the useEffects and useLayoutEffects. Now it is testing the behavior.

dai-shi (approved):
It looks good!

leobastiani:
I just added a cleanup because we aren't using react-testing-library in this test

dai-shi:
Nice.
Unless anyone else says anything, I'll merge this and release a patch version later.

dai-shi:
Published: https://www.npmjs.com/package/zustand/v/3.3.1

PR #284, opened 2021-01-20
feat(persist): add warn if storage is unavailable

### LINKED ISSUE #283: Not available storage fails silently
Hello there 👋 , 

First of all, I really really like zustand as a state management tool and I'm a big fan of it. 
I just noticed one small thing. When using the `persist` middleware to add the state to the `localStorage` or `sessionStorage` and the application doesn't allow to access it for example when used in an SSR application, the error or the fact that is not available is not notified to the user or the developer.

This means it fails silently and when you are not checking the storage yourself you might expect it being saved, but it's not. 

I think by adding some `warning` or information within the console here in the `middleware.ts` file like this could save some debugging time: 
```
try {
  storage = getStorage()
} catch (e) {
 console.warn("You try to persist your state, but the storage is currently not available in your application")
}
```

AnatoleLucet:
Closes #283 

This warning doesn't cover `api.setState`, but I'm not sure if it really needs to.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -191,7 +191,18 @@ export const persist = <S extends State>(
     // prevent error if the storage is not defined (e.g. when server side rendering a page)
   }
 
-  if (!storage) return config(set, get, api)
+  if (!storage) {
+    return config(
+      (...args) => {
+        console.warn(
+          `Persist middleware: unable to update ${name}, the given storage is currently unavailable.`
+        )
+        set(...args)
+      },
+      get,
+      api
+    )
+  }
 
   const setItem = async () => {
     const state = { ...get() }
@@ -240,8 +251,8 @@ export const persist = <S extends State>(
   })()
 
   return config(
-    (payload) => {
-      set(payload)
+    (...args) => {
+      set(...args)
       setItem()
     },
     get,


src/middleware.ts
```diff
@@ -191,7 +191,18 @@ export const persist = <S extends State>(
...
+      (...args) => {
+        console.warn(
+          `Persist middleware: unable to update ${name}, the given storage is currently unavailable.`
+        )
+        set(...args)
+      },
+      get,
+      api
+    )
+  }
```
dai-shi, line 205: Does it need to warn all the time when updating? why not just at the config time?
```js
  if (!storage) {
    console.warn(...)
    return config(get, set, api)
  }
```

AnatoleLucet: As I said in #283, it'll probably be annoying for the user to be spamed if he has a few stores. Especially if he doesn't especially use the store(s) while the storage is unavailable.

dai-shi: I see, but this would be annoying too.
Anyway, if it's intentional, I'm ok with it.

dai-shi (approved):

src/middleware.ts
AnatoleLucet: Well, I guess this warning is always going to be annoying wherever we put it. The question is where it's going to be just annoying enough for the user to notice it without spending hours to debug why the data is not in the storage :smile:

PR #292, opened 2021-01-30
Fix persist error forwarding

sumartoyo:
In current implementation, errors from persist middleware could not be received by the caller. There are chances that errors will happen when reading or writing to storage, so error handling is important. In this PR I changed the way persist works with error forwarding so the caller can catch errors and handle it accordingly.

There are multiple changes I made:

### Tests

Before changing anything, it would be better if we have tests for persist. There are currently another PR with tests for devtools middleware, #233 . I think it would be better if we separate the test file, so I put persist tests in `persist.test.tsx` rather than in `middleware.test.tsx`. Also, to prevent merge conflict with that PR.

### ~~`storage.setItem`~~

~~When the `setItem` function is an async function, such as in AsyncStorage for React Native, the caller would not receive the error because persist middleware call `setItem` without await-ing. Discussion here #287 .~~

~~To solve that, I change all functions that call `storage.setItem` to await or return the promise .~~

### Rehydration

Rehydration is performed asynchronously in store initialization ([middleware.ts#L229-L251](https://github.com/pmndrs/zustand/blob/cf4d1365b66895e6fc776e27ef1f1d25476aed01/src/middleware.ts#L229-L251). Here, the error is thrown, but since it is asynchronous, caller will not receive this exception.

I initially want to change the way `onRehydrateStorage` to be able to send error with the way a callback commonly does: `callback(error, data)`. But it may cause problem (backward incompatibility) because in current implementation the first parameter is the data. If there are already users that utilize this callback, and they expect the first parameter to be the data, it will break their code.

Then I decided to introduce another callback, `onRehydrateError`. This callback will be called if there is any error while performing hydration. The error can be from reading the storage, deserialization, or failed migration (this will be explained below).

### Migration

When studying persist code, I think the way we handle version mismatch is very dangerous. "if versions mismatch, clear storage by storing the new initial state" is basically clear everything and start from new. If I have a store abour user login data, and then I change the storage to a new version, the login data would be cleared and users have to re-login.

Other state persistence libraries (in other state management libraries) commonly use migration to handle this. Migration interface that I introduce here is similar to redux persist. In persist options I introduce `migrate(persistedState, version)`, a function that can be defined by developers to put their migration logic.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -216,14 +216,14 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    storage?.setItem(name, await serialize({ state, version }))
+    return storage?.setItem(name, await serialize({ state, version }))
   }
 
   const savedSetState = api.setState
 
   api.setState = (state, replace) => {
     savedSetState(state, replace)
-    setItem()
+    return setItem()
   }
 
   // rehydrate initial state with existing stored state
@@ -253,7 +253,7 @@ export const persist = <S extends State>(
   return config(
     (...args) => {
       set(...args)
-      setItem()
+      return setItem()
     },
     get,
     api
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -171,6 +171,11 @@ type PersistOptions<S> = {
    * This is useful when adding a breaking change to your store.
    */
   version?: number
+  /**
+   * A function to perform persisted state migration.
+   * This function will be called when persisted state versions mismatch with the one specified here.
+   */
+  migrate?: (persistedState: S, version: number) => S | Promise<S>
 }
 
 export const persist = <S extends State>(
@@ -187,6 +192,7 @@ export const persist = <S extends State>(
     onRehydrateStorage,
     onRehydrateError,
     version = 0,
+    migrate,
   } = options || {}
 
   let storage: StateStorage | undefined
@@ -247,10 +253,17 @@ export const persist = <S extends State>(
     }
 
     if (deserializedStorageValue) {
-      // if versions mismatch, clear storage by storing the new initial state
+      // if versions mismatch, run migration
       if (deserializedStorageValue.version !== version) {
         try {
-          await setItem()
+          const migratedState = await migrate?.(
+            deserializedStorageValue.state,
+            deserializedStorageValue.version
+          )
+          if (migratedState) {
+            set(migratedState)
+            await setItem()
+          }
         } catch (e) {
           onRehydrateError?.(e)
         }
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -162,6 +162,11 @@ type PersistOptions<S> = {
    */
   onRehydrateStorage?: (state: S) => ((state: S) => void) | void
   /**
+   * This callback function will be called when an exception occured in rehydration or migration.
+   * The error can be from `storage.getItem` or when deserialization.
+   */
+  onRehydrateError?: (error: Error) => void
+  /**
    * If the stored state's version mismatch the one specified here, the storage will not be used.
    * This is useful when adding a breaking change to your store.
    */
@@ -180,6 +185,7 @@ export const persist = <S extends State>(
     blacklist,
     whitelist,
     onRehydrateStorage,
+    onRehydrateError,
     version = 0,
   } = options || {}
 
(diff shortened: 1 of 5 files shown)


src/middleware.ts
```diff
@@ -216,44 +228,57 @@ export const persist = <S extends State>(
...
+          const migratedState = await migrate?.(
+            deserializedStorageValue.state,
+            deserializedStorageValue.version
+          )
+          if (migratedState) {
+            set(migratedState)
+            await setItem()
+          }
+        } catch (e) {
+          onRehydrateError?.(e)
```
AnatoleLucet, line 268: Is there a reason for using two try catch here?

AnatoleLucet (commented):
Thanks for your work on this PR! Plenty of good stuff in here 😄

Is there a reason not to do something like `onRehydrateStorage(data, error)` instead of a new callback?

It seems like you also need to alter the return type of `set` and `api.setState` so typescript doesn't complain about a "useless" `await` (see this [codesandbox](https://codesandbox.io/s/react-typescript-forked-ngpu5?file=/src/App.tsx)).

sumartoyo:
>Is there a reason not to do something like onRehydrateStorage(data, error) instead of a new callback?

No reasons other than to follow convention of `cb(error, data)` 😄. Maybe users will be confused because the parameters are reversed.
What do you think? Should I implement it with `onRehydrateStorage(data, error)`? If we document it well we can minimize the confusion.

>It seems like you also need to alter the return type of set and api.setState so typescript doesn't complain about a "useless" await

Yes, I didn't catch that. Thank you for pointing that out. I will fix it.

src/middleware.ts
sumartoyo: Initially they throw two different error object. One error object is instantiation from `RehydrateError` and the second is from `MigrationError`, both are custom class extends from `Error`. I did that to identify where the error occured. And then I decided to pass the original error object rather than instantiating new error object to retain stack trace. So, now there's no need for two try catch blocks. I will combine them together.

AnatoleLucet:
> What do you think? Should I implement it with onRehydrateStorage(data, error)? If we document it well we can minimize the confusion.

Well, it's true that it breaks conventions, but I think it should be ok now that jsdoc, lsp completion and all that as become mainstream in most editors.

sumartoyo:
Hmm I changed return type of `SetState` to `void | Promise<void>` but I'm afraid it may cause problems.

For example, I have to change `act(() => setState({}))` to `act(() => { setState({}) })` because act is confused. It can accept void or Promise but not `void | Promise<void>`.

Is it okay or is there a better way to do it? How do people commonly type a function that can return a promise or not a promise?

AnatoleLucet:
It doesn't work because `void | Promise<void>` doesn't match any of `act()`'s overloads.
One way to fix this would be to define two overloads for `SetState`. One for `void` and another one for `Promise<void>`.
E.g.:
```ts
export type SetState<T extends State> = {
  (partial: PartialState<T>, replace?: boolean): void
  (partial: PartialState<T>, replace?: boolean): Promise<void>
}
```

sumartoyo:
> One way to fix this would be to define two overloads for `SetState`. One for `void` and another one for `Promise<void>`.

I have tried that and it still has problems. When I try  
`const mySetState: SetState<TState> = (partial: PartialState<TState>) => {}`  
it says "Type 'void' is not assignable to type 'Promise<void>'". It has no error if I set the function as async.
`const mySetState: SetState<TState> = async (partial: PartialState<TState>) => {}`

But then, when I try calling it with  
`await mySetState({})`  
it says "'await' has no effect on the type of this expression".

I think we can't have a function that can returns two different types when they have identical parameters. At least that's how understand it in different langsuage (Java, C#), unless we use generics to the function.

It seems that this is more complex than what I thought. How about we take out things related to `storage.setItem` from this PR, so that the scope of this PR is only about rehydration and migration? I will revert the code related to the `setItem` and we continue our discussion in another discussion?

AnatoleLucet:
> It seems that this is more complex than what I thought. How about we take out things related to storage.setItem from this PR, so that the scope of this PR is only about rehydration and migration? I will revert the code related to the setItem and we continue our discussion in another discussion?

Sure. Maybe a new drafted pr will be easier to propose change and discuss the code instead of a discussion.

AnatoleLucet:
@sumartoyo since you haven't left a comment or re-requested a review, I'm not sure if it's ready. Is it?

sumartoyo:
Ah my bad, it is not clear. Yes @AnatoleLucet, it's ready.

src/middleware.ts
```diff
@@ -158,14 +158,19 @@ type PersistOptions<S> = {
...
   /**
    * If the stored state's version mismatch the one specified here, the storage will not be used.
    * This is useful when adding a breaking change to your store.
    */
   version?: number
+  /**
+   * A function to perform persisted state migration.
+   * This function will be called when persisted state versions mismatch with the one specified here.
+   */
+  migrate?: (persistedState: S, version: number) => S | Promise<S>
```
AnatoleLucet, line 173: `persistedState` might not match `S` if the versions doesn't match. You may type this using `any` since we can't really know what the stored state look like?

AnatoleLucet (changes requested):

src/middleware.ts
sumartoyo: You're right. Okay, I have fixed it.

I'm not sure whether there is something wrong with GitHub. The commit isn't shown in this PR even though it has been pushed to my forked repo ([link to commit](https://github.com/sumartoyo/zustand/commit/72cba0820e1f3d357e5bd058fa6991bfa1d5cc96)).

AnatoleLucet: Strange, what appends if you push something new? Can you try (for example) to update your branch and push the changes?

sumartoyo: Yay! That works!

AnatoleLucet: Great!

AnatoleLucet (approved):

AnatoleLucet:
@dai-shi ready to merge :slightly_smiling_face:

dai-shi (commented):
glanced at it. not fully understood, but seems good.

dai-shi:
@AnatoleLucet Do you plan something soon? Or, we should release the current master?

AnatoleLucet:
Nothing else for now, a release is fine for me.

dai-shi:
Published: https://www.npmjs.com/package/zustand/v/3.3.2

PR #293, opened 2021-02-01
Readme: be more explicit about equality function

wkrueger:
Hey!

Although I had already used `shallow`, It took me a while to notice the second argument on the store hook was an equality function, I found it out later reading the typedefs.

I think it would be a good idea (especially for dumb readers like me) to state it more explicitly: "the second argument is an equality function, it has this signature".

Here goes a suggestion. Thank you!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -76,7 +76,16 @@ const nuts = useStore(state => state.nuts)
 const honey = useStore(state => state.honey)
 ```
 
-If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing an alternative equality function.
+For more control over re-rendering, you may provide an alternative equality function on the second argument.
+
+```jsx
+const treats = useStore(
+  state => state.treats,
+  (oldTreats, newTreats) => compare(oldTreats, newTreats)
+)
+```
+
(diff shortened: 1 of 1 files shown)


dai-shi:
Given that zustand is an unopinionated library, it's fine to make it explicit about the equality function.

Personally, I would avoid overusing it. Would you help me learn what's your use case for custom comparator? There should be use cases, but can it be for expert users? I mean, should we be careful not to encourage too much to use custom comparator? The reason for this is, it's sometimes hard to understand the behavior in some edge cases.

```js
const state = useStore(state => state, (a, b) => a.x === a.x)
state.y === useStore.getState().y // is not always true
```
This is an extreme case, but it's a pitfall, when it's hidden in deep code.

wkrueger:
I was working with a list which objects had like 10 properties, but only 3 of them were rendered on a table. The table was rendered on an expensive 3rd party data-table component.

I ended up refactoring that. I agree that it might be uncommon usage, but it might be useful.

dai-shi:
@wkrueger Thanks for the explanation!

@drcmda Any thoughts? Your preference?

dai-shi (approved):
Sorry for taking so much time on this.

As we don't get other feedback until now, let me merge this and see how it goes.
If it misleads someone, we may consider to change some descriptions.

Issue #291, opened 2021-01-29
clunky combine store usage


gerboinger:
I've followed the recommended way to combine stores in #228. It's clunky to work with as inside the stores you have to set state as if you're setting from the root of the state, this makes everything overly verbose. I imagine in many cases sub-stores won't touch other stores state and having one root store would mainly be for simplifying co-dependent state queries outside of the stores.

Take a root store made up of sub-stores like this:
```
const useStore = create((set, get, api) => ({
  users: { ...createUsersStore(set, get, api) },
  restrictedUsers: { ...createRestrictedUsersStore(set, get, api) }
}));
```

Inside the sub-stores, it'd be intuitive to set state like this:
```
const createUsersStore = (set, get, api) => ({
  data: [],
  setUsers: users => set({data: users})
})
```

Example:
```
// expected state after calling useStore.getState().users.setUsers(['user1', 'user2']) :
{
  users: {
    data: ['user1', 'user2']
  },
  restrictedUsers: {...}
}

// instead, it looks like this:
{
  data: ['user1', 'user2'],
  users: {
    data: []
  },
  restrictedUsers: {...}
}
```

To fix this issue, you need to prepend all `set` object params with the sub-store key, and spread the current values:
```
const createUsersStore = (set, get, api) => ({
  data: [],
  setUsers: users => set({users: {...get().users, data: users}})
})
```

This feels like unnecessary ceremony. Immer could be a solution, but I like to have my async logic contained in store _actions_; Immer doesn't like async. The solution I've come to is a helper like this:
```
const _setIn = (set, get, propName) => setObjArg => set({ [propName]: { ...get()[propName], ...setObjArg } });
```

This helper can then be passed down into the stores
```
const useStore = create((set, get, api) => {
  const setIn = curry(_setIn)(set, get);
  return {
    users: { ...createUsersStore(set, get, api, setIn("users")) },
    restrictedUsers: {...createRestrictedUsersStore(set, get, api, setIn("restrictedUsers"))}
  };
});
```

How setting now looks, and works intuitively:
```
const createUsersStore = (set, get, api, setIn) => ({
  data: [],
  setUsers: users => setIn({data: users})
})
```

[...]

Issue #178, opened 2020-09-06
Namespacing / slicing application state alike Redux' combineReducers


kelleyvanevert:
Hey!

I just came across this library, and am quite enthusiastic already 🥳

I've ditched Redux in the past, only to have returned to adding it back later because of a need for a common-sense global application state instance when using various somewhat-interdependent contexts didn't work out better than Redux in the first place.

One think I really dislike about Redux, is the ergonomic pain of having to differentiate between synchronous and asynchronous (or otherwise indirect) operations. This always feels like a "fake" or accidental technically-contrained version of a private/public layered approach. Often, the set of ordinary actions end up kind of like a private store api, and the async actions (or otherwise indirect thunks) end up like the public api.

Zustand seems to nicely eliminate this distinction from it ergonomics, retaining the same general approach but without the "bureaucratic fuss" let's say.

```jsx
const useStore = create((set, get, api) => {
  // set up the state layout
  const initialState = { ... };
  // some internal helpers
  const privateFn = (...) => { ... get() ... set(...) ... };
  // the public api
  const publicFn = (...) => { ... get() ... privateFn(...) ... };
  return {
    ...initialState,
    publicFn,
    ...
  };
});
```

I'm tempted to look into replacing Redux for Zustand in my (medium-scale, web & native) app, but would first need to concoct/add a few tools to be able to transition from what we have now. Up top on that list is the ability to divide the top-level state into somewhat independent "slices", e.g. `combineReducers` in Redux.

Here's a little attempt I made at something that would look similar, but maybe be a bit more flexible than a drop-in `combineCreators` (which was my first idea):

https://codesandbox.io/s/zustand-combinecreators-ojew5?file=/src/App.tsx

```tsx
const useStore = create((get, set, api) => {
  const counterSlice = namespace("counter", createCounter)(get, set,  api);
  ...
  const crossSliceOperation = (...) => { ... };

  return {
    counter: counterSlice,
    crossSliceOperation,
    ...
  };
});
```

(Also, typing all of this is not super trivial, and probably requires some overloading of `SetState` etc.)

What are your thoughts?

dai-shi:
Hi, I've tried something similar: https://github.com/react-spring/zustand/issues/161#issuecomment-678608873
Your `namespace` seems more complete.
I'm not sure if there's a use case to use namespaced `api`. `set` and `get` should be enough for most cases. Anyway, it's understandable to provide consistent signature.

dai-shi:
[...]

PR #301, opened 2021-02-09
Add Combine Store Feature

ragokan:
I made a way to combine unlimited amount of stores to create one base store. I tested it, everything works amazing, even the setState function. Yet, as you can see from the code, it can't totally infer the types if we don't provide an interface. I hope it can be fixed.

```ts
interface IAge extends State {
  age: number
}

interface IName extends State {
  name: string
}

interface IHealth extends State {
  status: boolean
}

const ageState = create<IAge>(() => ({ age: 5 }))
const nameState = create<IName>(() => ({ name: 'zustand' }))
const healthState = create<IHealth>(() => ({ status: true }))

interface combined extends IAge, IName, IHealth {}

const combined = combineStores<combined>(nameState, ageState, healthState)

console.log(combined)
console.log(combined.getState())
console.log(combined.getState().name)
```
Console Results:
image

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,4 +1,4 @@
-import {
+import create, {
   GetState,
   PartialState,
   SetState,
@@ -258,4 +258,17 @@ export const persist = <S extends State>(
     get,
     api
   )
+}
+
+export const combineStores = <T extends State>(
+  ...storesToCombine: StoreApi<any>[]
+): StoreApi<T> => {
+  let values = {}
+
+  storesToCombine.forEach((store) => {
+    values = Object.assign(values, store.getState())
+  })
+
+  let combinedStore: StoreApi<any> = create(() => ({ ...values }))
+  return combinedStore
 }


dai-shi:
Hey, thanks for your interest and your trial.
Non technical issue: We already have something called "combine", so maybe we need to have a different name to avoid confusion.

Looking at the code, it's not very useful as is. What is nice is to accept `set` in the sub stores.

So, it's more like this:
```js
combineStores({ name: nameState, age: ageState, health: healthState })
```

We have some recent discussions in #291. You want to take a look at #178 especially.
(Hm, analogous to `combineReducers`, `combineCreators` may work. Now, I regret naming the bare `combine`...)
Yeah, combining creators sounds better that combining stores.

TypeScript support is very important, I guess. Would you see how `combine` works?

Hope you get some ideas.

ragokan:
> Hey, thanks for your interest and your trial.
> Non technical issue: We already have something called "combine", so maybe we need to have a different name to avoid confusion.
> 
> Looking at the code, it's not very useful as is. What is nice is to accept `set` in the sub stores.
> 
> So, it's more like this:
> 
> ```js
> combineStores({ name: nameState, age: ageState, health: healthState })
> ```
> 
> We have some recent discussions in #291. You want to take a look at #178 especially.
> (Hm, analogous to `combineReducers`, `combineCreators` may work. Now, I regret naming the bare `combine`...)
> Yeah, combining creators sounds better that combining stores.
> 
> TypeScript support is very important, I guess. Would you see how `combine` works?
> 
> Hope you get some ideas.


Thank for your comment, I will try to implement this with a better, other than that, I will try to help others who are trying the same thing, you can close this code if you want.

dai-shi:
Let me close this. Would be better to keep discussion in #178.

PR #303, opened 2021-02-11
docs: add an example for setting the state outside event (#302)

### LINKED ISSUE #302: Zombie child, yet again
There still seems to be the issue of zombie child. It requires some special criteria to reproduce:

1) The state needs to have initial value
2) The delete action, needs to be async called in regards to the event handler (Ex.: wrap the action call inside a setTimeout)
3) This is reproducible only on the elements that were present in the initial state.

Note, that the same scenario is not reproducible while using redux with hooks.

Repo to reproduce: https://github.com/ratoi-crysty/zustand-zombie-child

ratoi-crysty:
close #302.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -365,7 +365,7 @@ const useStore = create(redux(reducer, initialState))
 
 ## Calling actions outside a React event handler
 
-Because React handles `setState` synchronous if it's called outside an event handler. Updating the state outside an event handler will force react to update component synchronously, therefore adding the risk of encountering the zombie-child effect.
+Because React handles `setState` synchronously if it's called outside an event handler. Updating the state outside an event handler will force react to update the components synchronously, therefore adding the risk of encountering the zombie-child effect.
 In order to fix this, the action needs to be wrapped in `unstable_batchedUpdates`
 
 ```jsx
@@ -388,8 +388,6 @@ async function onClickHandler() {
   })
 }
 ```
-
-
 
 More details: https://twitter.com/dan_abramov/status/959507572951797761
 
--- a/readme.md
+++ b/readme.md
@@ -389,7 +389,7 @@ async function onClickHandler() {
 }
 ```
 
-More details: https://twitter.com/dan_abramov/status/959507572951797761
+More details: https://github.com/pmndrs/zustand/issues/302
 
 ## Redux devtools
 
--- a/readme.md
+++ b/readme.md
@@ -369,7 +369,7 @@ Because React handles `setState` synchronously if it's called outside an event h
 In order to fix this, the action needs to be wrapped in `unstable_batchedUpdates`
 
 ```jsx
-import { unstable_batchedUpdates } from 'react-dom'; // or react-native
+import { unstable_batchedUpdates } from 'react-dom' // or 'react-native'
 
 const useStore = create((set) => ({
   fishes: 0,
--- a/readme.md
+++ b/readme.md
@@ -369,22 +369,16 @@ Because React handles `setState` synchronously if it's called outside an event h
 In order to fix this, the action needs to be wrapped in `unstable_batchedUpdates`
 
 ```jsx
-import create from 'zustand'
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Thanks for the suggestion. Would you please make the example code minimal? We'd like to make readme as small as possible. (We could create a wiki page and add a link.)

What do you think about recommending using `batchedUpdates` in store creator, in `addAFish`?

readme.md
```diff
@@ -363,6 +363,34 @@ import { redux } from 'zustand/middleware'
...
+
+async function onClickHandler() {
+  await asyncFunc()
+  unstable_batchedUpdates(() => {
+    addAFish()
+  })
+}
+```
+
+More details: https://twitter.com/dan_abramov/status/959507572951797761
```
dai-shi, line 392: Not sure how this is helpful. Remove it or replace the link to #302 and add the twitter link there?

ratoi-crysty: OK

ratoi-crysty:
@dai-shi I'm not sure what should I delete from the example to make it smaller. As I tried to make it as minimal as possible. Any suggestions? 

Regarding moving the batchFunction inside the action, I did not added there in order to better point out were the issue comes from. I can move it if you think it makes more sense.

readme.md
```diff
@@ -363,6 +363,34 @@ import { redux } from 'zustand/middleware'
...
+))
+
+const addAFish = useStore((state) => state.addAFish)
+
+async function onClickHandler() {
+  await asyncFunc()
+  unstable_batchedUpdates(() => {
+    addAFish()
+  })
+}
```
dai-shi, line 389:
```suggestion
import { unstable_batchedUpdates } from 'react-dom'; // or react-native

const useStore = create((set) => ({
  fishes: 0,
  increaseFishes: () => set((prev) => ({ fished: prev.fished + 1 }))
}))

const nonReactCallback = () => {
  unstable_batchedUpdates(() => {
    useStore.getState().increaseFishes()
  })
}
```

How about this?

dai-shi:
> Regarding moving the batchFunction inside the action, I did not added there in order to better point out were the issue comes from. I can move it if you think it makes more sense.

I understand your point. Let's keep it as is and we will see.

readme.md
```diff
@@ -363,6 +363,28 @@ import { redux } from 'zustand/middleware'
 const useStore = create(redux(reducer, initialState))
 ```
 
+## Calling actions outside a React event handler
+
+Because React handles `setState` synchronously if it's called outside an event handler. Updating the state outside an event handler will force react to update the components synchronously, therefore adding the risk of encountering the zombie-child effect.
+In order to fix this, the action needs to be wrapped in `unstable_batchedUpdates`
+
+```jsx
+import { unstable_batchedUpdates } from 'react-dom'; // or react-native
```
dai-shi, line 372:
```suggestion
import { unstable_batchedUpdates } from 'react-dom' // or 'react-native'
```

readme.md
```diff
@@ -363,6 +363,28 @@ import { redux } from 'zustand/middleware'
...
+
+Because React handles `setState` synchronously if it's called outside an event handler. Updating the state outside an event handler will force react to update the components synchronously, therefore adding the risk of encountering the zombie-child effect.
+In order to fix this, the action needs to be wrapped in `unstable_batchedUpdates`
+
+```jsx
+import { unstable_batchedUpdates } from 'react-dom' // or 'react-native'
+
+const useStore = create((set) => ({
+  fishes: 0,
+  increaseFishes: () => set((prev) => ({ fished: prev.fished + 1 }))
```
dai-shi, line 376:
```suggestion
  increaseFishes: () => set((prev) => ({ fishes: prev.fishes + 1 }))
```

dai-shi (approved):
Merging this.

Tirzono:
Thanks for this addition @ratoi-crysty and @dai-shi, I think this is helpful for more people.

PR #305, opened 2021-02-13
Combine Stores Into One Store

ragokan:
**With this code, you can now create seperate stores and combine them to one single store with all the types and utilities.**

**Almost no change on bundle size, it is like 25 new line which will probably result on 25 more kb**

[Try on Code Sandbox](https://codesandbox.io/s/zustand-combined-store-lrwy3)**

**Create Store**
image

**Get Store**
image

**Set Store**
image

**Subscribe**
image

**Usage In React (totally same as normal functions)**
image

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -68,9 +68,9 @@
     "gzipped": 251
   },
   "middleware.js": {
-    "bundled": 7993,
-    "minified": 3568,
-    "gzipped": 1397
+    "bundled": 8192,
+    "minified": 3730,
+    "gzipped": 1464
   },
   "vanilla.js": {
     "bundled": 1825,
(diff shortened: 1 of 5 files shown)


ragokan:
basic.test.tsx gave error because I added 1 more type to vanilla/index.ts.  Add getRoot: [Function] to line 41, storeApi.getRoot() to line 740.

dai-shi:
wow, this looks so simple and magical. getRoot just returns the store creator, right?

I have two requests, which might be just one, to investigate further.
1. No getRoot. We'd like to avoid new api in vanilla.
2. combine store creators, not stores. There's extra runtime overhead.

I'm not sure if your trick only works with combining stores, but I would really like to see this to combine store creators. How does that sound?

dai-shi:
We need namespacing..
https://github.com/pmndrs/zustand/pull/306#issuecomment-778851649

PR #306, opened 2021-02-14
Combine Store Creators

ragokan:
**[Request To Do This](https://github.com/pmndrs/zustand/pull/305#issuecomment-778702029)
#305 Was the first request but we had to create stores which will use memory which will also result in slow starts.
#178 Asked for namespacing/slicing and there is slicing. Namespacing is actually a little bit hard for now because when you use set in store creator, you actually can access to the whole store, so instead of prev.count +1, it is prev.counter.count +1, for example.
Maybe namespacing feature can also be added in future, the code won't change much, I will try to implement on top of this, yet it is hard.**

**With this code, you can now create seperate stores and combine them to one single store with all the types and utilities.**

**Almost no change on bundle size, it is just 11 new line**

**[Try on Code Sandbox](https://codesandbox.io/s/zustand-combine-state-creators-y0eev)**

**Usage is pretty simple, first create store creators**
image

**Then combine them**
image

**Usage is also same**
image

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -259,3 +259,14 @@ export const persist = <S extends State>(
     api
   )
 }
+
+export const combineStateCreators = <TState extends State>(
+  ...stateCreators: StateCreator<any>[]
+): StateCreator<TState> => (set, get, api) => {
+  let values: any = {}
+
+  stateCreators.forEach((sc) => {
+    values = Object.assign({}, values, sc(set, get, api))
+  })
+  return values
+}


dai-shi:
Unless we support the namespace, it's not very useful.

The naive solution is pretty simple and rather straightforward.

```js
import create from 'zustand'

const createState1 = (set) => ...
const createState2 = (set) => ...
  ...createState1(set),
  ...createState2(set),
}))
```

This might be worth adding in readme if we can't support namespacing.

Issue #317, opened 2021-02-26
weak type safety in setState due to usage of Partial


TkDodo:
reproduction: https://codesandbox.io/s/romantic-moon-xxkzd?file=/src/App.tsx

in this example, our `dec` function sets `count` to `undefined`, even though this is not allowed (it has to be a number). If we try to set it to a string, this will correctly type-error.

`setState` takes a `Partial<State>`, as defined here: https://github.com/pmndrs/zustand/blob/58c286e6cb9d6fc6dfe0142113b12aada0894ec8/src/vanilla.ts#L17-L20

but `Partial` not only lets you omit keys, it also adds lets you explicitly set every key to `undefined`, which is quite dangerous since those partial objects are then usually spread over the existing state, thus setting the value to `undefined` at runtime.

[Here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGxoF4oG8oGMD2BXAdkgE4BcU+uAtgEYREC+AUIzvgM7BQeIRkAK8IsACW8ADYAeODwB8UVBhwFiZAgBMIAM2H4IaqPSA) is a minimal typescript playground repro, so TS is fine with these two lines:

```
type State = { counter: number}
const state: Partial<State> = { counter: undefined }
```

`this.setState` in class-based React components does not have this problem, because they are not using `Partial`, but rather picking the keys manually:

```
setState<K extends keyof S>(
            state: ((prevState: Readonly<S>, props: Readonly<P>) => (Pick<S, K> | S | null)) | (Pick<S, K> | S | null),
```

not sure if we could tweak the types here to do something similar?

dai-shi:
Good point. not sure either.
Anyone interested trying out? A little concern is that we don't have many typing tests, so not sure if it will be a breaking change or not. (But, basically proper types are just good.)

TkDodo:
do you consider changes to the typings to be breaking? I think it would mean better type-safety, and if people get new errors in their code because of it, they were likely doing it wrong before :)

I can maybe take a stab at this on the weekend. do you have a preference on how to test types?

dai-shi:
> do you consider changes to the typings to be breaking?

No, but I'd like to write something in the release note if the change is significant.

> I can maybe take a stab at this on the weekend.

Great!

> do you have a preference on how to test types?

Not much. What I meant was if the current tests are in TS (without `any`), we could have noticed how big the impact would be.
If you have suggestions, they are welcome.

TkDodo:
[...]

PR #320, opened 2021-02-27
(types): improve types for setState

TkDodo:
@dai-shi I think this would be a minimal invasive fix for #317 . I can't know for sure until I write some type tests, which is why this is a draft so far :) What do you think?

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,7 +1,7 @@
 export type State = Record<string | number | symbol, unknown>
-export type PartialState<T extends State> =
-  | Partial<T>
-  | ((state: T) => Partial<T>)
+export type PartialState<T extends State, K extends keyof T = keyof T> =
+  | (Pick<T, K> | T)
+  | ((state: T) => Pick<T, K> | T)
 export type StateSelector<T extends State, U> = (state: T) => U
 export type EqualityChecker<T> = (state: T, newState: T) => boolean
 export type StateListener<T> = (state: T, previousState: T) => void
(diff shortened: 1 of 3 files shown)


src/vanilla.ts
```diff
@@ -1,7 +1,7 @@
 export type State = Record<string | number | symbol, unknown>
-export type PartialState<T extends State> =
-  | Partial<T>
-  | ((state: T) => Partial<T>)
+export type PartialState<T extends State, K extends keyof T = keyof T> =
+  | (Pick<T, K> | T)
+  | ((state: T) => Pick<T, K> | T)
```
dai-shi, line 4: Do you know if we really need `| T`?

dai-shi (commented):
Nice try. Please keep going.

src/vanilla.ts
TkDodo: I do not :)

I got this from the react typings on `setState` of a class component:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c49e45842358ba59a508e13130791989911430d/types/react/v16/index.d.ts#L489-L495

Judging from the comment, it is needed for intelliSense 😅

dai-shi: Ah ha. We need to put the link in the code comment, then.

TkDodo: done here: b142dcda

TkDodo:
@dai-shi I added some tests and tried it out in of my projects where we use zustand, and it worked pretty well. Please have another look

TkDodo:
when I'm trying to run my example in the codesandbox build, I am getting this:

Screenshot 2021-02-27 at 18 19 17

I am not getting this in other PRs 🤔

TkDodo:
okay, this might be something with codesandbox. I installed:

`"zustand": "https://pkg.csb.dev/pmndrs/zustand/commit/5564b272/zustand"`

locally and all was fine, the example now errors on `{ count: undefined }` as expected :)

dai-shi:
(yeah, sometimes I get similar weird behaviors in codesandbox, like caching issue.)

dai-shi:
Confirmed locally that this also shows an error as expected:
```ts
import create from "zustand";

export const useStore = create(
  combine(
    {
      count: 0
    },
    (set) => ({
      inc: () => set({ count: undefined })
    })
  )
);
```

dai-shi (approved):

Issue #334, opened 2021-03-19
Importing in a `type=module` project fails


wolframkriesing:
First: thanks for the simplified version of state management. So glad I found it, so sad it took me so long ;).

I created a small project, where the `package.json` sets it to be a ESM project, by `type=module`.
See https://codeberg.org/wolframkriesing/bug-exploration-zustand-and-esm/src/branch/main/package.json#L4

I ran into two issues, that are somehow related to #201 but I am not sure exactly what's the root cause.

## The Problem(s)
I wanted to get zustand to be testable with mocha in an ESM setup. But I ran into:
1) When using `type=module` and importing `zustand/vanilla` mocha says
   there is no such file. No idea what that might be.
   This is the JS file and the error it throws:
   https://codeberg.org/wolframkriesing/bug-exploration-zustand-and-esm/src/branch/main/src/vanilla.test.js
2) When importing `zustand/vanilla.js` the default export is somehow broken. A change of 
   `exports.default = create;`  
   to   
   `module.exports = create;`  
   in `zustand/vanilla.js` would fix it.
   This sounds like a packager problem in zustand?
   https://codeberg.org/wolframkriesing/bug-exploration-zustand-and-esm/src/branch/main/src/vanilla-js.test.js
   
Is the ESM setup just not there yet, and not recommended? I had hoped to go there gradually.
How do others solve this issue?

wolframkriesing:
One idea I just had to possibly solve it, is to create a pure esm build of zustand, maybe so it can be imported via `import created from "zustand/esm"`. Which would "just" need a stripping away the TS from the original sources, I guess.
Will give it a try and hope to provide a PR.

dai-shi:
As you have already noticed, the issue is `zustand/vanilla` is CJS.
Would you just import `zustand` and see if it works? Or, have you already tried?

wolframkriesing:
Importing `zustand` works, since it is ESM, right. I understood that this is the react-style `create`, right? I want to import the `zustand/vanilla` to be able to write tests that dont need react.
Currently I am doing it like so: 
```
import zustand from 'zustand/vanilla.js';
const create = typeof zustand === 'function' ? zustand : zustand.default;
```

this works, but I guess its not meant this way.
`typeof zustand === 'function'` is `true` when the code was transpiled.

`zustand.default` is needed when running it in a `type=module` package using just nodejs without any transpilation.

dai-shi:
> I understood that this is the react-style `create`, right?

[...]

PR #336, opened 2021-03-19
Create a just ESM version

wolframkriesing:
This PR tries to provide a solution for a pure-ESM package setup.
Solving #334 

This generates a pure ESM version which can be imported like `import create from "zustand/esm"`.
Not sure if this something of interest or if there is a much better and simpler solution.
Looking fwd to feedback.

In this file https://codeberg.org/wolframkriesing/bug-exploration-zustand-and-esm/src/branch/main/src/vanilla-esm.test.js I wrote tests that show this works. Might be a good idea putting them in here!?

Open issues
- [ ] build the middleware.ts

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,29 @@
   "files": [
     "**"
   ],
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./index.d.ts",
+      "import": "./index.js",
+      "default": "./index.cjs.js"
+    },
+    "./vanilla": {
+      "types": "./vanilla.d.ts",
+      "import": "./vanilla.module.js",
+      "default": "./vanilla.js"
+    },
+    "./middleware": {
+      "types": "./middleware.d.ts",
+      "import": "./middleware.module.js",
+      "default": "./middleware.js"
+    },
+    "./shallow": {
+      "types": "./shallow.d.ts",
+      "import": "./shallow.module.js",
+      "default": "./shallow.js"
+    }
+  },
   "sideEffects": false,
   "scripts": {
     "prebuild": "shx rm -rf dist",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -97,9 +97,9 @@ export default [
   createCommonJSConfig('src/vanilla.ts', 'dist/vanilla.js'),
 
   // Create a purely ESM version, can be imported from `zustand/esm`
-  createESMConfig('src/vanilla.ts', 'dist/esm/vanilla.js'),
-  createESMConfig('src/shallow.ts', 'dist/esm/shallow.js'),
-  createESMConfig('src/index.ts', 'dist/esm/index.js'),
-  // I dont get the one below to succeed with rollup :(
-  // createModernESMConfig('src/middleware.ts', 'dist/esm/middleware.js'),
+  createESMConfig('src/vanilla.ts', 'dist/vanilla.module.js'),
+  createESMConfig('src/shallow.ts', 'dist/shallow.module.js'),
+  createESMConfig('src/index.ts', 'dist/index.module.js'),
+  // get this below to succeed with rollup ... fails currently
+  // createModernESMConfig('src/middleware.ts', 'dist/middleware.module.js'),
 ]
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -99,7 +99,6 @@ export default [
   // Create a purely ESM version, can be imported from `zustand/esm`
   createESMConfig('src/vanilla.ts', 'dist/vanilla.module.js'),
   createESMConfig('src/shallow.ts', 'dist/shallow.module.js'),
-  createESMConfig('src/index.ts', 'dist/index.module.js'),
   // get this below to succeed with rollup ... fails currently
   // createModernESMConfig('src/middleware.ts', 'dist/middleware.module.js'),
 ]
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -95,4 +95,11 @@ export default [
   createCommonJSConfig('src/shallow.ts', 'dist/shallow.js'),
   createCommonJSConfig('src/middleware.ts', 'dist/middleware.js'),
   createCommonJSConfig('src/vanilla.ts', 'dist/vanilla.js'),
+
+  // Create a purely ESM version, can be imported from `zustand/esm`
+  createESMConfig('src/vanilla.ts', 'dist/esm/vanilla.js'),
+  createESMConfig('src/shallow.ts', 'dist/esm/shallow.js'),
+  createESMConfig('src/index.ts', 'dist/esm/index.js'),
+  // I dont get the one below to succeed with rollup :(
+  // createModernESMConfig('src/middleware.ts', 'dist/esm/middleware.js'),
 ]


dai-shi:
We've already done something more sophisticated in [jotai](https://github.com/pmndrs/jotai) and [valtio](https://github.com/pmndrs/valtio).

Would you add this in package.json?
```json
  "exports": {
    ".": {
      "types": "./index.d.ts",
      "module": "./index.js",
      "default": "./index.cjs.js"
    },
    "./vanilla": {
      "types": "./vanilla.d.ts",
      "module": "./vanilla.module.js",
      "default": "./vanilla.js"
    },
    "./middleware": {
      "types": "./middleware.d.ts",
      "module": "./middleware.module.js",
      "default": "./middleware.js"
    },
    "./shallow": {
      "types": "./shallow.d.ts",
      "module": "./shallow.module.js",
      "default": "./shallow.js"
    }
  },
```

Oh, but you want to use "import", so it's different from "jotai" and "valtio"

```json
  "exports": {
    ".": {
      "types": "./index.d.ts",
      "import": "./index.js",
      "default": "./index.cjs.js"
    },
    "./vanilla": {
      "types": "./vanilla.d.ts",
      "import": "./vanilla.module.js",
      "default": "./vanilla.js"
    },
    "./middleware": {
      "types": "./middleware.d.ts",
      "import": "./middleware.module.js",
      "default": "./middleware.js"
    },
    "./shallow": {
      "types": "./shallow.d.ts",
      "import": "./shallow.module.js",
      "default": "./shallow.js"
    }
  },
```

I don't know if this would cause issues with other environments.

rollup.config.js
```diff
@@ -95,4 +95,11 @@ export default [
   createCommonJSConfig('src/shallow.ts', 'dist/shallow.js'),
   createCommonJSConfig('src/middleware.ts', 'dist/middleware.js'),
   createCommonJSConfig('src/vanilla.ts', 'dist/vanilla.js'),
+
+  // Create a purely ESM version, can be imported from `zustand/esm`
+  createESMConfig('src/vanilla.ts', 'dist/esm/vanilla.js'),
+  createESMConfig('src/shallow.ts', 'dist/esm/shallow.js'),
+  createESMConfig('src/index.ts', 'dist/esm/index.js'),
```
dai-shi, line 102:
```suggestion
  createESMConfig('src/vanilla.ts', 'dist/vanilla.module.js'),
  createESMConfig('src/shallow.ts', 'dist/shallow.module.js'),
  createESMConfig('src/index.ts', 'dist/index.module.js'),
```
Does this make sense? I'm not sure if "module" and "esm" is the same thing...
Anyway, my point is we want to make it consistent among zustand/jotai/valtio.
(I mean we could change the other two, if that's better.)

rollup.config.js
```diff
@@ -95,4 +95,11 @@ export default [
...
+  // I dont get the one below to succeed with rollup :(
+  // createModernESMConfig('src/middleware.ts', 'dist/esm/middleware.js'),
```
dai-shi, line 104: 😕

wolframkriesing:
I would import in an ESM package like this? `import create from "zustand/vanilla/import"`?

rollup.config.js
wolframkriesing: looks good too. I have no strong opinion. I assume you have a broader overview and insights from users for what suits best.

dai-shi:
If it's node14, it will read "exports" in package.json, so `import create from 'zustand/vanilla'` should work, I expect...

rollup.config.js
dai-shi: As I understand, usually "module" and "esm" refer the same thing, but "mjs" does not.

wolframkriesing:
Changed it to use your `exports` approach in the package.json.

rollup.config.js
```diff
@@ -95,4 +95,11 @@ export default [
   createCommonJSConfig('src/shallow.ts', 'dist/shallow.js'),
   createCommonJSConfig('src/middleware.ts', 'dist/middleware.js'),
   createCommonJSConfig('src/vanilla.ts', 'dist/vanilla.js'),
+
+  // Create a purely ESM version, can be imported from `zustand/esm`
+  createESMConfig('src/vanilla.ts', 'dist/vanilla.module.js'),
+  createESMConfig('src/shallow.ts', 'dist/shallow.module.js'),
+  createESMConfig('src/index.ts', 'dist/index.module.js'),
```
dai-shi, line 102:
```suggestion
```

oops, we don't need this. it's `index.js`. this is inverted for historical reason.

dai-shi:
Looks good. So what's remaining is:
- does it work in your use case?
- fix middleware.module.js
- how can we make sure if this doesn't break other envs...

package.json
```diff
@@ -9,6 +9,29 @@
   "files": [
     "**"
   ],
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./index.d.ts",
+      "import": "./index.js",
```
dai-shi, line 16:
```suggestion
      "module": "./index.js",
```
react doesn't support mjs, so we should only enable "import" for pure libs. let's make it back "module" for webpack5.

dai-shi: hm, I was confused with something else (= type="module"). Should be fine.

rollup.config.js
wolframkriesing: same here. mjs is just the workaround of nodejs that exist(ed) while nodejs was figuring out how to move to real ESM

wolframkriesing:
> how can we make sure if this doesn't break other envs...

seems to be the most challenging.
Using `exports` in package.json (afaiu) we limit import/require to what is listed in `exports`, if we forget one we break it for people.   
And afaik currently there are acouple of ways to import/require even just the `index.*`, e.g. you can import/require
`zustand`, `zustand/index`, `zustand/index.js`, `zustand/index.cjs`, `zustand/index.d.ts`, `zustand/index.iife.js`

Sounds like a couple of tests to ensure this wont break (until intended) might be a good idea.
For that I would be curious, what are the intended usages, starting with the `index.*`? All the above? Would a test ensure all the above keep working? If so I would start with a PR that provides tests for verifying all the current ways zustand can be imported/required. Wdyt?

dai-shi:
we never had tests for `dist`. I'm not even sure if there's some best practice.
*.cjs, *.iife exist for historical reason, which I actually don't know.
Totally unsure if index.iife is used. index.cjs is used in some cases.

AFAIU, once all envs understand `"exports"`, it would be just good.
Meanwhile, we follow some basic conventions: defaults are cjs. `"main"` is cjs. `"module"` is esm.
People can probably use esm explicitly with `import ... from '*.module'`.

Now, jotai and valtio use "module" instead of "import" in "exports" to avoid dual package hazard.
Thinking of zustand, so far everything is pure, "import" would just work. But honestly, we don't want that because we may add some new features that depend on module variable.

Hmmm, I don't think my explanation is not well described.

----

Back to your original issue, type=module and `import create from 'valtio/vanilla'`.
- Does current build work for you? 9a4f7dd
- Does it fail if we modify "import" to "module" in "exports" in package.json? (It's probably same as just removing "import"s)

dai-shi:
Hmm, I feel like I want to change the index.js:
- index.js -> index.module.js
- index.cjs -> index.js

This is a breaking change for users who use `import create from 'zustand/index.cjs'`.

JeremyRH:
The IIFE build is usually used for getting directly from a CDN like unpkg. Unfortunately, we didn't configure it correctly and https://unpkg.com/zustand redirects to the CJS build. I guess it could also be used by some who use a more manual build process like gulp.
Too bad we cant see usage.

dai-shi:
Thanks @JeremyRH . We are going towards ES module age in browser too, so I guess it's time to revisit bundles. It feels like CJS build works for most traditional use cases. My plan is to drop iife in the next minor and see what issues people report.

rollup.config.js
rschristian: Curious what wasn't working? This not only works out of the box for me, but also gets rid of a `regenerator-runtime` dependency which is great. Huge improvement for us.

dai-shi:
v3.4.1 comes with #343 and #350, which should almost do what is proposed in this PR.
Most discussions here are obsoleted, so let me close this and continue discussion in #334.

Issue #119, opened 2020-06-25
Doesn't work in IE11 (even when using core-js)


Jossnaz:
the error is:
>Syntax error

![image](https://user-images.githubusercontent.com/2072058/85792786-acc62480-b6f9-11ea-988b-0c629c1de9c9.png)

I have 

```
import 'core-js';
```
at the top of my file

venturalp:
Hi @Jossnaz, I had the same issue and I found out that I could solve that by using, along with corejs, a different import source for my stores: `import { create } from 'zustand/index.cjs'`

Could you please try that and tell us if it solves?

Jossnaz:
I added, apart from the polyfills, this:
```
import { create } from 'zustand/index.cjs.js
```
and it seems to work.

thanks a lot, would be nice to have it as well in the readme

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
With this solution, How can I resolve type error?

venturalp:
> With this solution, How can I resolve type error?

I guess in the new version of zustand 3+ this problem was solve for good. I didn't update here yet, but I do remember seeing some commit related to cjs modules in this new version. Is it possible for you to update your version?

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@venturalp  What I meant was a typescript error at `import create from 'zustand/index.cjs' code`.
I upgraded to the latest version, but the below error still occurs.

```
Could not find a declaration file for module 'zustand/index.cjs'. '/home/myid/projects/myproject/node_modules/zustand/index.cjs.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/zustand` if it exists or add a new declaration (.d.ts) file containing `declare module 'zustand/index.cjs';`
```

venturalp:
> @venturalp What I meant was a typescript error at `import create from 'zustand/index.cjs'` code on VS Code.
> I upgraded to the latest version, but the below error still occurs.
> 
> ```
> Could not find a declaration file for module 'zustand/index.cjs'. '/home/myid/projects/myproject/node_modules/zustand/index.cjs.js' implicitly has an 'any' type.
>   Try `npm i --save-dev @types/zustand` if it exists or add a new declaration (.d.ts) file containing `declare module 'zustand/index.cjs';`
> ```

Sorry, I forgot to say, I don't think you need to use `/index.cjs` anymore.

[...]

PR #337, opened 2021-03-19
fix: types for index.cjs

dai-shi:
https://github.com/pmndrs/zustand/issues/119#issuecomment-803167696


dai-shi:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs Would you please try the codesandbox build above? ☝️

dai-shi:
With #343, we no longer have `index.cjs`, so this is obsoleted.

PR #343, opened 2021-03-26
build: use esbuild

aulneau:
Hi there!

I made a [PR for jotai](https://github.com/pmndrs/jotai/pull/378) a few days ago updating the build tooling to use esbuild as the typescript transpiler and @dai-shi requested I make PRs for zustand and [valtio](https://github.com/pmndrs/valtio/pull/129), so here we are! The main reason for this change is because of how much faster esbuild is at compiling javascript from typescript:

Before
✨  Done in 25.26s.

After
✨  Done in 6.80s.

There are also some very minor bundle size changes with this update.

### General overview of changes
- added `esbuild` and `rollup-plugin-esbuild`
- replaced the `typescript` plugin in `createESMConfig`, `createCommonJSConfig`, and `createIIFEConfig` to use esbuild
- added `id.includes('@babel/runtime')` to the externals check [per this recommendation](https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers)
- updated the node version used in the codesandbox ci to 12 (due to esbuild requirement)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,16 +3,43 @@
   "private": true,
   "version": "3.3.3",
   "description": "🐻 Bear necessities for state management in React",
-  "main": "index.cjs.js",
-  "module": "index.js",
+  "main": "index.js",
+  "module": "index.module.js",
   "types": "index.d.ts",
   "files": [
     "**"
   ],
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./index.d.ts",
+      "module": "./index.module.js",
+      "default": "./index.js"
+    },
+    "./vanilla": {
+      "types": "./vanilla.d.ts",
+      "module": "./vanilla.module.js",
+      "default": "./vanilla.js"
+    },
+    "./middleware": {
+      "types": "./middleware.d.ts",
+      "module": "./middleware.module.js",
+      "default": "./middleware.js"
+    },
+    "./shallow": {
+      "types": "./shallow.d.ts",
+      "module": "./shallow.module.js",
+      "default": "./shallow.js"
+    }
+  },
   "sideEffects": false,
   "scripts": {
     "prebuild": "shx rm -rf dist",
-    "build": "rollup -c",
+    "build": "concurrently 'yarn:build:*'",
+    "build:base": "rollup -c",
+    "build:vanilla": "rollup -c --config-vanilla",
+    "build:middleware": "rollup -c --config-middleware",
+    "build:shallow": "rollup -c --config-shallow",
     "postbuild": "yarn copy",
     "eslint": "eslint --fix 'src/**/*.{js,ts,jsx,tsx}'",
     "eslint-examples": "eslint --fix 'examples/src/**/*.{js,ts,jsx,tsx}'",
@@ -80,40 +107,46 @@
     ]
   },
   "devDependencies": {
-    "@babel/core": "^7.13.8",
+    "@babel/core": "^7.13.14",
     "@babel/plugin-external-helpers": "^7.12.13",
-    "@babel/plugin-transform-react-jsx": "^7.12.17",
-    "@babel/plugin-transform-runtime": "^7.13.8",
+    "@babel/plugin-transform-react-jsx": "^7.13.12",
     "@babel/plugin-transform-typescript": "^7.13.0",
-    "@babel/preset-env": "^7.13.8",
+    "@babel/preset-env": "^7.13.10",
     "@rollup/plugin-babel": "^5.3.0",
-    "@rollup/plugin-node-resolve": "^11.2.0",
-    "@rollup/plugin-typescript": "^8.2.0",
-    "@testing-library/react": "^11.2.5",
-    "@types/jest": "^26.0.20",
-    "@types/react": "^17.0.2",
-    "@types/react-dom": "^17.0.1",
-    "@typescript-eslint/eslint-plugin": "^4.15.2",
-    "@typescript-eslint/parser": "^4.15.2",
-    "eslint": "^7.21.0",
+    "@rollup/plugin-node-resolve": "^11.2.1",
+    "@rollup/plugin-typescript": "^8.2.1",
+    "@testing-library/react": "^11.2.6",
+    "@types/jest": "^26.0.22",
+    "@types/react": "^17.0.3",
+    "@types/react-dom": "^17.0.3",
+    "@typescript-eslint/eslint-plugin": "^4.20.0",
+    "@typescript-eslint/parser": "^4.20.0",
+    "concurrently": "^6.0.0",
+    "esbuild": "^0.11.2",
+    "eslint": "^7.23.0",
     "eslint-config-prettier": "^8.1.0",
     "eslint-import-resolver-alias": "^1.1.2",
     "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jest": "^24.1.5",
+    "eslint-plugin-jest": "^24.3.2",
     "eslint-plugin-prettier": "^3.3.1",
-    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-react": "^7.23.1",
     "eslint-plugin-react-hooks": "^4.2.0",
     "husky": "^4.3.8",
     "jest": "^26.6.3",
     "json": "^10.0.0",
     "lint-staged": "^10.5.4",
     "prettier": "^2.2.1",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1",
-    "rollup": "^2.40.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "rollup": "^2.44.0",
+    "rollup-plugin-esbuild": "^3.0.2",
     "rollup-plugin-size-snapshot": "^0.12.0",
     "shx": "^0.3.3",
-    "typescript": "^4.2.2"
+    "typescript": "^4.2.3"
+  },
+  "resolutions": {
+    "@types/react": "^17.0.3",
+    "@types/react-dom": "^17.0.3"
   },
   "peerDependencies": {
     "react": ">=16.8"
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -2,25 +2,36 @@ import path from 'path'
 import babel from '@rollup/plugin-babel'
 import resolve from '@rollup/plugin-node-resolve'
 import typescript from '@rollup/plugin-typescript'
+import esbuild from 'rollup-plugin-esbuild'
 import { sizeSnapshot } from 'rollup-plugin-size-snapshot'
 
 const createBabelConfig = require('./babel.config')
-
-const { root } = path.parse(process.cwd())
-const external = (id) => !id.startsWith('.') && !id.startsWith(root)
 const extensions = ['.js', '.ts', '.tsx']
-const getBabelOptions = (targets) => {
+const { root } = path.parse(process.cwd())
+
+function external(id) {
+  return (
+    id.startsWith('./vanilla') || (!id.startsWith('.') && !id.startsWith(root))
+  )
+}
+
+function getEsbuild(target) {
+  return esbuild({
+    minify: false,
+    target,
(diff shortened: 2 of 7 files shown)


aulneau:
@dai-shi as I was comparing the packaging of these three packages, zustand stands out of the convention where you're bundling esm as a separate module, and defaulting to cjs as the main export. After this I saw this PR: https://github.com/pmndrs/zustand/pull/336. Curious how you want to move forward?

dai-shi:
Yes, let's change it.
It's a breaking change for index.cjs users https://github.com/pmndrs/zustand/pull/336#issuecomment-803369768, but i'm going with just minor update with breaking change notice in the release page, as this use case is not very common.

dai-shi:
Same apply here with the comment in valtio: https://github.com/pmndrs/valtio/pull/129#issuecomment-808194781

rollup.config.js
```diff
@@ -2,14 +2,28 @@ import path from 'path'
...
 import { sizeSnapshot } from 'rollup-plugin-size-snapshot'
 
 const createBabelConfig = require('./babel.config')
 
 const { root } = path.parse(process.cwd())
-const external = (id) => !id.startsWith('.') && !id.startsWith(root)
+
+const external = (id) =>
+  (!id.startsWith('.') && !id.startsWith(root)) ||
+  id.startsWith('@babel/runtime')
```
dai-shi, line 14: Though technically a different thing, I'd like to do something more in this build change.

Let's try this, following the valtio setup.
```suggestion
const external = (id) =>
  id.startsWith('./vanilla') ||
  (!id.startsWith('.') && !id.startsWith(root)) ||
  id.startsWith('@babel/runtime')
```

aulneau:
@dai-shi I haven't seen this kind of thing before: `createIIFEConfig`, do you want them in both jotai/valtio?

dai-shi:
Hm, I think we can drop iife build in zustand. I don't know who is using in which case.
Wonder if @drcmda or @JeremyRH have an idea?

package.json
```diff
@@ -3,16 +3,43 @@
...
   "types": "index.d.ts",
   "files": [
     "**"
   ],
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./index.d.ts",
+      "import": "./index.module.js",
+      "default": "./index.cjs.js"
```
dai-shi, line 17:
```suggestion
      "default": "./index.js"
```

aulneau:
There is one new issue that I don't quite understand: https://rollupjs.org/guide/en/#error-this-is-undefined
Screen Shot 2021-03-26 at 8 25 38 AM

aulneau:
@dai-shi I think it's related to this https://github.com/rollup/rollup/issues/1518#issuecomment-374727013

here: https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L218 and https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L241

src/middleware.ts
```diff
@@ -215,7 +215,7 @@ export const persist = <S extends State>(
     )
   }
 
-  const setItem = async () => {
+  const setItem = async function () {
```
aulneau, line 218: @dai-shi this fixes https://github.com/pmndrs/zustand/pull/343#issuecomment-808225863

src/middleware.ts
```diff
@@ -234,7 +234,7 @@ export const persist = <S extends State>(
 
   api.setState = (state, replace) => {
     savedSetState(state, replace)
-    setItem()
+    void setItem()
```
aulneau, line 237: This was added due to:

Screen Shot 2021-03-26 at 8 44 12 AM

dai-shi:
Is it happening only for cjs build (esbuild + babel), or also module build (esbuild only)? Or, am I missing something totally.

aulneau:
> Is it happening only for cjs build (esbuild + babel), or also module build (esbuild only)? Or, am I missing something totally.

Just for the cjs bundle. It seems to be an issue of the conversion down to es5, but https://github.com/pmndrs/zustand/pull/343#discussion_r602305363 fixes it

dai-shi:
@aulneau Thanks for your investigation.
Actually, I would prefer a solution without changing the `src`, because it affects the module bundle.
Would you try `moduleContext` or the custom `onwarn`? I have no idea what to specify for `moduleContext` though, `null`?

aulneau:
> Actually, I would prefer a solution without changing the src, because it affects the module bundle.

I am curious though, I think if we do not make this change, the resulting cjs bundle will not work if I understand what the issue is.

Can I ask why you would prefer to keep arrow functions over non? isn't the result the same?

dai-shi:
> I think if we do not make this change, the resulting cjs bundle will not work if I understand what the issue is.

I didn't take like that... It just seems a potential warning. I will read it again.

> isn't the result the same?

For the module bundle, arrow functions are not transpiled. I think the current code looks just right and don't want to prohibit unbound arrow functions for the future.

aulneau:
> I didn't take like that... It just seems a potential warning. I will read it again.

I see, if it's just a warning then that should be fine.

dai-shi:
https://github.com/rollup/rollup/issues/1518#issuecomment-321875784

> It's a warning because sometimes you can ignore it and sometimes you can't.

I'm assuming this is a warning not for us.
Please try other options without modifying `src`:
- rollup-plugin-async (this is a totally different solution, but it might turn to be good.)
- `moduleContext` to `null`
- custom `onwarn` to ignore `'THIS_IS_UNDEFINED'`

@aulneau once this is done, let's ask someone to  help for trial in discord.

aulneau:
@dai-shi sure thing!

Question for you around the `middleware.ts` file. I see [here that the promise](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L258) is used in `setItem`, however [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L237) and [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L275) the promise is ignored. Is that intentional? If so, do you want to add `void` before the ignored promises?

dai-shi:
> Question for you around the `middleware.ts` file. I see [here that the promise](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L258) is used in `setItem`, however [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L237) and [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L275) the promise is ignored. Is that intentional? If so, do you want to add `void` before the ignored promises?

I think it's intentional. @AnatoleLucet can confirm.

I don't mind adding `void` as you suggested in https://github.com/pmndrs/zustand/pull/343#discussion_r602305572.
So, this is unrelated with `THIS_IS_UNDEFINED` thing, right?

AnatoleLucet:
> > Question for you around the `middleware.ts` file. I see [here that the promise](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L258) is used in `setItem`, however [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L237) and [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L275) the promise is ignored. Is that intentional? If so, do you want to add `void` before the ignored promises?
> 
> I think it's intentional. @AnatoleLucet can confirm.
> 
> I don't mind adding `void` as you suggested in [#343 (comment)](https://github.com/pmndrs/zustand/pull/343#discussion_r602305572).
> So, this is unrelated with `THIS_IS_UNDEFINED` thing, right?

Yes, it's intentional. I don't mind adding a `void` either.

aulneau:
> So, this is unrelated with THIS_IS_UNDEFINED thing, right?

Yep, just something I've noticed.

So everything seems to be working now. I realized we actually don't need to use either `@rollup/plugin-typescript` or `rollup-plugin-esbuild` on the cjs output as we're using babel and it seems to handle it all perfectly fine! Seems to have cut another half a second off the total time :) I think this is ready for folks to test! It'd be good to get some esm based apps (Eg using vite or snowpack) and some cjs apps (maybe a next.js app or something) to ensure nothing is broken!

package.json
```diff
@@ -3,16 +3,43 @@
...
+  "module": "index.module.js",
   "types": "index.d.ts",
   "files": [
     "**"
   ],
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./index.d.ts",
+      "import": "./index.module.js",
```
dai-shi, line 16: The use of "import" may lead dual package hazard issue. We can try that road further but in #336 later.
For this PR, let's keep in sync with jotai/valtio.
```suggestion
      "module": "./index.module.js",
```
Would you change for all four?

rollup.config.js
```diff
@@ -57,42 +67,24 @@ function createCommonJSConfig(input, output) {
     external,
     plugins: [
       resolve({ extensions }),
-      typescript(),
```
dai-shi, line 60: Is this because babel can handle (strip) types?
I assume you tried and check the result doesn't change at all.
(Also, please apply it to jotai and valtio, too.)

aulneau: I did yes, it's a bit hard to track exactly due to the naming changes, and the addition of some extra outputs, but here is the diff: https://github.com/aulneau/zustand/commit/0322e8847be71b10c7a0a1c5657e3a53d0bd4758

dai-shi: Oh, I see > naming changes.
Thanks for confirming.

JeremyRH:
@aulneau Awesome job with this PR. Using rollup-plugin-esbuild to parse TypeScript and concurrently to run the builds in parallel is great. I suspect a lot of the build time improvements are due to running the builds in parallel.

In the future we should look at using esbuild directly in place of rollup. It doesn't support ES5 output but if we plan on removing the iife bundle, do we need to output ES5?

dai-shi:
> It doesn't support ES5 output but if we plan on removing the iife bundle, do we need to output ES5?

Both iife and cjs used to target ie11, and it doesn't change for cjs in this pr. We still use babel for cjs.

dai-shi (approved):
Looks good. Thanks for your work!

rollup.config.js
```diff
@@ -21,6 +36,7 @@ const getBabelOptions = (targets) => {
   }
   return {
     ...config,
+    comments: false,
     extensions,
   }
 }
```
dai-shi, line 42: As we only use babel for cjs, let's make it simple and consistent with jotai/valtio.
```js
const getBabelOptions = (targets) => {
  const config = createBabelConfig({ env: (env) => env === 'build' }, targets)
  return {
    ...config,
    plugins: [
      ...config.plugins,
      ['@babel/plugin-transform-runtime', { helpers: true, regenerator: true }],
    ],
    babelHelpers: 'runtime',
    extensions,
    comments: false,
  }
}
```

I tried removing `'@babel/plugin-transform-regenerator'` and the result is the same. What do you think?

dai-shi: I noticed you put @babel/plugin-transform-runtime in babel.config.js in jotai. Maybe we can do the same in zustand.

rollup.config.js
```diff
@@ -2,25 +2,36 @@ import path from 'path'
...
   const config = createBabelConfig({ env: (env) => env === 'build' }, targets)
   if (targets.ie) {
-    config.plugins = [
-      ...config.plugins,
-      '@babel/plugin-transform-regenerator',
-      ['@babel/plugin-transform-runtime', { helpers: true, regenerator: true }],
-    ]
-    config.babelHelpers = 'runtime'
+    config.plugins = [...config.plugins, '@babel/plugin-transform-regenerator']
+    config.babelHelpers = 'bundled'
```
dai-shi, line 30: Let me check how much this would affect the bundle size.

dai-shi: | | `'runtime'` | `'bundled'` | diff |
|---|---|---|---|
|index.js|3760|3561|-199|
|middleware.js|8720|9147|+427|
|shallow.js|620|620|0|
|vanilla.js|1798|1798|0|

Okay, this is more or less what I expected.
We will probably be refactoring `persist` in middleware.ts without async/await, so the diff is likely to be reduced in the future.
Let's go with bundled as cjs build is kind of fallback.

PR #348, opened 2021-03-31
fix: better ssr detection

### LINKED ISSUE #347: SSR useLayoutEffect warning
I'm using Deno and Aleph and when adding in the Zustand library, I get the following warning:

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. 
This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. 
To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. 
See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

I'm calling out to Zustand here: https://github.com/aadamsx/aleph-error3/blob/master/lib/useStore.ts

And when rendering the component I get the error here: https://github.com/aadamsx/aleph-error3/blob/master/components/People.tsx

This thread traces the problem and possible solutions: https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

And here is where I raised the issue in Aleph: https://github.com/alephjs/aleph.js/issues/221#issuecomment-809024724

Is there any way Zustand can get around this warning for SSRs use cases?

dai-shi:
close #347 

this would work for deno and jsdom with custom userAgent.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,7 @@ const isSSR =
   !window.navigator ||
   /ServerSideRendering/.test(window.navigator.userAgent)
 
-const useIsoLayoutEffect = isSSR ? useEffect : useLayoutEffect
+const useIsomorphicLayoutEffect = isSSR ? useEffect : useLayoutEffect
 
 export interface UseStore<T extends State> {
   (): T
@@ -74,7 +74,7 @@ export default function create<TState extends State>(
     }
 
     // Syncing changes in useEffect.
-    useIsoLayoutEffect(() => {
+    useIsomorphicLayoutEffect(() => {
       if (hasNewStateSlice) {
         currentSliceRef.current = newStateSlice as StateSlice
       }
@@ -85,7 +85,7 @@ export default function create<TState extends State>(
     })
 
     const stateBeforeSubscriptionRef = useRef(state)
-    useIsoLayoutEffect(() => {
+    useIsomorphicLayoutEffect(() => {
       const listener = () => {
         try {
           const nextState = api.getState()
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,7 +16,7 @@ export * from './vanilla'
 // Deno support: https://github.com/pmndrs/zustand/issues/347
 const isSSR =
   typeof window === 'undefined' ||
-  typeof window.navigator === 'undefined' ||
+  !window.navigator ||
   /ServerSideRendering/.test(window.navigator.userAgent)
 
 const useIsoLayoutEffect = isSSR ? useEffect : useLayoutEffect
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,8 +13,13 @@ import createImpl, {
 export * from './vanilla'
 
 // For server-side rendering: https://github.com/react-spring/zustand/pull/34
-const useIsoLayoutEffect =
-  typeof window === 'undefined' ? useEffect : useLayoutEffect
+// Deno support: https://github.com/pmndrs/zustand/issues/347
+const isSSR =
+  typeof window === 'undefined' ||
+  typeof window.navigator === 'undefined' ||
+  /ServerSideRendering/.test(window.navigator.userAgent)
+
+const useIsoLayoutEffect = isSSR ? useEffect : useLayoutEffect
 
 export interface UseStore<T extends State> {
   (): T
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,7 @@ export * from './vanilla'
 const isSSR =
   typeof window === 'undefined' ||
   !window.navigator ||
-  /ServerSideRendering/.test(window.navigator.userAgent)
+  /ServerSideRendering|^Deno\//.test(window.navigator.userAgent)
 
 const useIsomorphicLayoutEffect = isSSR ? useEffect : useLayoutEffect
 


dai-shi:
@aadamsx Would you please try the codesandbox build ☝️ ?

aadamsx:
@dai-shi Here's what I did to test (maybe you can tell me if I did something wrong):

1) Downloaded your build from: https://pkg.csb.dev/pmndrs/zustand/commit/dba40397/zustand
2) Uncompressed it
3) Copied the files over to my project
4) Locally imported the package => index.js

image


5) Reran the project and got the same warning message:

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

image

dai-shi:
@aadamsx the steps seem good to me.
would you do a little debugging? we want to make `isSSR` true and my expectation is `window.navigator` is undefined.

src/index.ts
```diff
@@ -13,8 +13,13 @@ import createImpl, {
 export * from './vanilla'
 
 // For server-side rendering: https://github.com/react-spring/zustand/pull/34
-const useIsoLayoutEffect =
-  typeof window === 'undefined' ? useEffect : useLayoutEffect
+// Deno support: https://github.com/pmndrs/zustand/issues/347
+const isSSR =
+  typeof window === 'undefined' ||
+  typeof window.navigator === 'undefined' ||
```
AnatoleLucet, line 19:
```suggestion
  window.navigator === undefined ||
```
...is the same as: 
```js
  typeof window.navigator === 'undefined' ||
```

dai-shi: nice. this should shave some bytes. how about, `!window.navigator`?

AnatoleLucet: Yes, even better

dai-shi: thanks. let me fix the three prs.

aadamsx:
This was tested before you latest change today

`window.navigator` was defined:

```javascript
typeof window: object
window.navigator: {
        "connection": {
                "downlink": 10,
                "effectiveType": "4g",
                "onchange": null,
                "rtt": 50,
                "saveData": false
        },
        "cookieEnabled": false,
        "language": "en",
        "languages": [
                "en"
        ],
        "onLine": true,
        "platform": "darwin",
        "userAgent": "Deno/1.8.2",
        "vendor": "Deno Land"
}
isWindowUndefined: false
isWindowNavigatorUndefined: false
isSSRUndefined: false
isSSR: false
useIsoLayoutEffect: function useLayoutEffect2(create, deps) {
                var dispatcher = resolveDispatcher();
                return dispatcher.useLayoutEffect(create, deps);
            }
```

I'll test the latest code

dai-shi:
> `window.navigator` was defined

Thanks. Good to know. Now, what we can do currently is to customize userAgent.
Is it possible to do something like this in your code?
```js
window.navigator.userAgent = 'Deno/1.8.2 (ServerSideRendering)';
```

src/index.ts
```diff
@@ -13,8 +13,13 @@ import createImpl, {
 export * from './vanilla'
 
 // For server-side rendering: https://github.com/react-spring/zustand/pull/34
-const useIsoLayoutEffect =
-  typeof window === 'undefined' ? useEffect : useLayoutEffect
+// Deno support: https://github.com/pmndrs/zustand/issues/347
+const isSSR =
+  typeof window === 'undefined' ||
+  !window.navigator ||
```
aadamsx, line 19:
```suggestion
isGlobalThisDeno
```

src/index.ts
```diff
@@ -13,8 +13,13 @@ import createImpl, {
...
+  /ServerSideRendering/.test(window.navigator.userAgent)
```
aadamsx, line 20:
```suggestion
```

src/index.ts
```diff
@@ -13,8 +13,13 @@ import createImpl, {
 export * from './vanilla'
 
 // For server-side rendering: https://github.com/react-spring/zustand/pull/34
-const useIsoLayoutEffect =
-  typeof window === 'undefined' ? useEffect : useLayoutEffect
+// Deno support: https://github.com/pmndrs/zustand/issues/347
```
aadamsx, line 16:
```suggestion
// Deno support: https://github.com/pmndrs/zustand/issues/347
// In the case of Deno, we'll check the globalThis for the existance of the Deno object to see if we are server side
const isGlobalThisDeno = globalThis && globalThis.Deno ? true : false;
```

aadamsx (commented):
Based on this: https://github.com/denoland/deno/discussions/9941

In the case of Deno, think it would be best to check for the existance of globalThis for 100% accuracy.

src/index.ts
AnatoleLucet: > `globalThis &&`

I assume you're doing this just to be sure `globalThis` is defined? But in case `globalThis` doesn't exist (don't even know if it's possible), this code will throw.

Something like this should be safer:
```js
const isGlobalThisDeno = typeof globalThis !== "undefined" && !!globalThis.Deno;
```

src/index.ts
dai-shi: The point is this is not only for deno. This is useful for something else like jsdom.

src/index.ts
dai-shi: tbh, I don't want to check both `window` and `globalThis`. Let me add a commit with the current approach, which should have a small impact on bundle size.

dai-shi:
Here's my suggestion.
The rationale behind it is even if someone created a Deno-based browser,
they would change the userAgent.

```js
const isSSR =
  typeof window === 'undefined' ||
  !window.navigator ||
  /ServerSideRendering|^Deno\//.test(window.navigator.userAgent)
```

----
The alternative could be something like this.
```js
const isSSR =                                                       
  typeof globalThis === 'undefined' ||                                
  'Deno' in globalThis ||                                             
  !globalThis.navigator ||                                              
  /ServerSideRendering/.test(globalThis.navigator.userAgent)          
```
Do people like this better? Looks too deno specific?

aadamsx:
```
const isSSR =                                                       
  typeof globalThis === 'undefined' ||                                
  'Deno' in globalThis ||                                             
  !globalThis.navigator ||                                              
  /ServerSideRendering/.test(globalThis.navigator.userAgent)  
```

We also need to check in the case of Node, so `typeof window === 'undefined'` is needed I think.  Also, to check for `navigator` or `userAgent` is too much, in the case of Deno, all that's need is to check if Deno is defined off globalThis and for Node to check for the existance of `window`.

Which would reduce to this:

```
const isSSR = typeof window === 'undefined' || typeof globalThis === 'undefined' || 'Deno' in globalThis;
````

I can confirm via testing this works in Deno.

**Comment by dai-shi** [comment]
I think I am failing to explain why I want userAgent check.
This is already implemented in some of other libs like https://github.com/pmndrs/valtio/pull/134
It would be useful for testing with jsdom on server.

We support IE11, and `globalThis` can be undefined on client side rendering.

I am also skeptical about `'Deno' in globalThis` condition.
Would someone eventually want to have deno browser and do client side rendering?

---
Basically, what we need to consider as edge cases is:
- IE11
- React Native
- jsdom
- and now Deno

**Comment by dai-shi** [comment]
Let me go with it. It should solve the reported issue and it seems resilient to me.

**Comment by aadamsx** [comment]
@dai-shi hey sorry for not getting back to you on this, got the second Covid vaccine and was down for two days from it. 🥵

Yeah I was reading on it more and you’re right, userAgent is a good way to go. I’ll try to test it out over the next day or so and will report back here. 

Thanks for the PR.

**Comment by aadamsx** [comment]
@dai-shi I noticed you updated master with this PR and more, but didn't publish a new release.  And https://esm.sh/zustand is still on release 3.3.3 from Feb 27th -- I assume this is because a new release was not published?

**Comment by dai-shi** [comment]
@aadamsx Ohh, I was planning to publish a new release after I get a good sign from you.
If you just need to install with npm/yarn, you can install the codesandbox ci build.
https://ci.codesandbox.io/status/pmndrs/zustand/pr/348/builds/115876

actually, this is latest from #350.
https://ci.codesandbox.io/status/pmndrs/zustand/pr/350/builds/116544
```
yarn add https://pkg.csb.dev/pmndrs/zustand/commit/beb348f4/zustand
npm i https://pkg.csb.dev/pmndrs/zustand/commit/beb348f4/zustand
```

can you try this? beb348f4 is identical to the current master.

aadamsx:
@dai-shi the esm code up to this commit https://pkg.csb.dev/pmndrs/zustand/commit/beb348f4/zustand performs ssr detection without error/warning in Deno.

If this was all you needed, please feel free to publish a new release.

And thanks again!

dai-shi:
Thanks! Good. I will publish it soon. This publish will include various changes in build things.

aadamsx:
Nice.  And if there's anything else you'd like for me to test out, please let me know.

PR #359, opened 2021-04-12
fix(persist middleware): incorrect deserialize type

### LINKED ISSUE #358: Deserialize option of PersistOptions has an incorrect type
Option `deserialize` has the following definition in PersistOptions type:
`deserialize?: (str: string) => StorageValue<S> | Promise<S>;`
Because of that asynchronous version demands to return Store value itself and not the object that incapsulates state and version of the state. While code that uses this option always expects `StorageValue<S>` to be returned out of deserialize method, for both sync and async versions.
I think the type should be defined like this:
`deserialize?: (str: string) => StorageValue<S> | Promise<StorageValue<S>>;`

AnatoleLucet:
Closes #358

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -151,7 +151,7 @@ type PersistOptions<S> = {
    * @param str The storage's current value.
    * @default JSON.parse
    */
-  deserialize?: (str: string) => StorageValue<S> | Promise<S>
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
👍

Issue #182, opened 2020-09-07
SSR Example (SSR data is shared)


jd327:
I'm using Next.js and have a separate file `state.js` that has something like this:

```javascript
export const useSessionStore = create(combine({
  session: {},
}, set => ({
  setSession: (session) => set({ session }),
})))
```

And then calling `setSession()` on the server (for example in `_app.js`) after a successful login. When reading `session` on the server + client, it seems to work fine. So far so good.

The problem is: the `session` seems to be saved in Node.js memory and shared with everyone(?). When you login on one browser, the same session data is returned when you load the page on another browser. So I'm assuming when I set session on the server, it persists on Node.js side and returns the same session for all clients.

What am I doing wrong, or is zustand not meant for this?

dai-shi:
The useSessionStore is a store itself as well as a hook. So, it's a singleton. A singleton can't be used for such session data. I'd say it's not meant for this, but other comments from someone are welcome.

jd327:
@dai-shi Thanks for taking a look.

Yeah I see some Next.js example attempts, but nothing official. Can anybody confirm if this is a no-go for SSR? What do others do/use?

dai-shi:
@jd327 Does SSR work if a use case allow to have data shared? What are the examples? Pointers are welcome.

We need to understand how such a library like zustand supports SSR, but personally if there's a way, we would like to add a support for it.

jd327:
@jd327 Thanks for following up.

A bit of context: for this new Next.js app, I was hoping to avoid having one global Redux state. If I could just use Zustand hooks alone throughout the app, that would be a big improvement in simplicity. Previously, I've had to use this: https://github.com/kirill-konshin/next-redux-wrapper 

From what I see so far there's an issue with the way state data is stored/transferred from server to client when using Zustand. When I set data on the server, it seems to persist between requests (another browser will get the same data), which I am guessing means it is stored somewhere in Node.js memory and shared..?

Ideally, I'd just like it to transfer from server to client *per user request*. Example: server reads session cookie > server pulls data from somewhere > server calls `setSession()` > server sends html to the client > client can use `session`.

dai-shi:
> it is stored somewhere in Node.js memory and shared..?

Yes, this is true. But, it must be true for Redux too.
So, probably next-redux-wrapper does some job. I'd take a deeper look later.

We definitely need some more people to help on this issue.

jd327:
[...]

PR #375, opened 2021-05-01
#182 SSR data is shared

Munawwar:


Munawwar:
Pending work & questions:

- there is a FIXME on context.ts, on useStore's type assertion. Seems like an issue with existing type for useStore?
- Update README on how to use the new context APIs
- Should we add a next.js example folder in the repo? or should we just add a link to the next.js canary repo?

I will try to add example to next.js canary repo. However I need someone to help me out in completing the documentation

src/context.ts
```diff
@@ -0,0 +1,50 @@
+import React, { createContext, useContext, useRef } from 'react'
+import create, { UseStore } from './index'
```
dai-shi, line 2:
```suggestion
import create, { UseStore } from 'zustand'
```

We need to alias for development. eslintrc does it. tsconfig has to be modified. check how [jotai](https://github.com/pmndrs/jotai) does it.

src/context.ts
```diff
@@ -0,0 +1,50 @@
...
+  const ZustandContext = createContext<UseStore<TState> | undefined>(undefined)
+
+  const Provider = ({
+    createState,
+    children,
+  }: {
+    createState: StateCreator<TState>
+    children: React.ReactNode
+  }) => {
+    const storeRef = useRef() as React.MutableRefObject<UseStore<TState>>
```
dai-shi, line 15:
```suggestion
    const storeRef = useRef<UseStore<TState>>()
```
I prefer to annotate type instead of assertion here and assert type in L23, but it's not a big deal.

src/context.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    return React.createElement(
+      ZustandContext.Provider,
+      { value: storeRef.current },
+      children
+    )
+  }
+
+  const useZustand = <StateSlice>(
```
dai-shi, line 28:
```suggestion
  const useStore = <StateSlice>(
```
We prefer this, right?

src/context.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  return {
+    Provider,
+    useZustand,
+  }
+}
+
+export default createZustand
+
+export const { Provider, useZustand } = createZustand({})
```
dai-shi, line 50:
```suggestion
```
Let's remove this and provide only one usage. I think it's less confusing.

dai-shi (commented):
Some early feedbacks.

dai-shi:
> Should we add a next.js example folder in the repo? or should we just add a link to the next.js canary repo?

I think just adding a link is fine, assuming we update the example in the nextjs repo.

src/context.ts
Munawwar: yup, missed a spot

Munawwar:
@dai-shi Ah, sorry.. I didn't push my changes properly last time (some last minute force push removed my new changes). Can you check again?

src/context.ts
```diff
@@ -0,0 +1,59 @@
...
+    createState: StateCreator<TState>
+    children: React.ReactNode
+  }) => {
+    const storeRef = useRef<UseStore<TState>>()
+
+    if (!storeRef.current) {
+      storeRef.current = create(createState)
+    }
+
+    return React.createElement(
```
dai-shi, line 27:
```suggestion
    return createElement(
```

Just a minor thing, but please import it directly.

src/context.ts
```diff
@@ -0,0 +1,59 @@
...
+  ) => {
+    // ZustandContext value is guaranteed to be stable.
+    const useProviderStore = useContext(ZustandContext)
+    if (!useProviderStore) {
+      throw new Error(
+        'Seems like you have not used zustand provider as an ancestor.'
+      )
+    }
+    return useProviderStore(
+      // FIXME: this type assertion seems unnecessary, as useStore does accept undefined.
```
dai-shi, line 46: The original useStore type is overloaded, so we'd need to do the same here.
If the type assertion is working correctly, that doesn't look too bad.

tests/context.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const createParam: StateCreator<CounterState> = (set) => ({
+    count: 0,
+    inc: () => set((state) => ({ count: state.count + 1 })),
+  })
+
+  function Counter() {
+    const { count, inc } = useStore()
+    React.useEffect(inc, [inc])
+    return <div>count: {count}</div>
```
dai-shi, line 28: but I wonder if this is properly typed.
```suggestion
    return <div>count: {count * 1}</div>
```
This can detect if `count` is inferred as number.

tests/context.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const { findByText } = render(
+    <Provider createState={createParam}>
+      <Counter />
+    </Provider>
+  )
+
+  await findByText('count: 1')
+})
+
```
dai-shi, line 39: we'd like to have another test with selectors.

```ts
  // like this
  const count = useStore((state) => state.count)
  const inc = useStore((state) => state.inc)
```

dai-shi (commented):
Overall it looks good! Added some comments.

src/context.ts
```diff
@@ -0,0 +1,59 @@
...
+    createState,
+    children,
+  }: {
+    createState: StateCreator<TState>
+    children: React.ReactNode
+  }) => {
+    const storeRef = useRef<UseStore<TState>>()
+
+    if (!storeRef.current) {
+      storeRef.current = create(createState)
```
dai-shi, line 24: It's only important this is initialized once at first, so we could technically accept the store (=useStore).
```js
      storeRef.current = useStore // from Provider's props
```
Is this more handy? or confusing?

src/context.ts
Munawwar: @dai-shi Ah.. createState prop on Provider alone is not going to work for another reason.

Have a look at my last [next.js solution](https://codesandbox.io/s/nextjs-with-zustand-forked-sr6z4?file=/lib/store.js:1266-1450) useHydrate() hook. (It is a working example, with all the bugs fixed I mentioned at #182).

See the logic to merge new states into existing store on client-side? It is unavoidable, and has to be there for next.js to work properly. Our current Provider interface is not doing this, nor does it expose the store.

So I am in favor of passing store. Property name can be either `store` or maybe `initialStore`?

src/context.ts
Munawwar: @dai-shi Ah.. createState prop on Provider alone is not going to work for another reason.

Have a look at my last [next.js solution](https://codesandbox.io/s/nextjs-with-zustand-forked-sr6z4?file=/lib/store.js:1266-1450) useHydrate() hook. (It is a working example, with all the bugs fixed I mentioned at #182).

See the logic to merge new states into existing store on client-side? It is unavoidable, and has to be there for next.js to work properly. Our current Provider interface is not doing this, nor does it expose the store.

So I am in favor of passing store. Property name can be either `store` or maybe `initialStore`?

dai-shi: I'm not sure if I fully understand the issue/solution in nextjs, but passing store is fine. I prefer `initialStore` too, or even `initialUseStore` because it's `useStore` from create, not `store` from vanilla.

Munawwar: I'll call it `initialStore` for simplicity, as.. we also have a `useStore` as return value of `createContext()`. It could get confusing with two useStore's in application code (yes we know internally they are two different things).

dai-shi: Okay. Please check other comments too: https://github.com/pmndrs/zustand/pull/375#pullrequestreview-649818401

src/context.ts
Munawwar: So.. I just removed the FIXME for now.

dai-shi (approved):
Looks good to me. Nice work.

PR #376, opened 2021-05-04
fix: remove mandatory index signature

aigoncharov:
Currently, `type State` in `zustand` forces all users to add an index signature to their state interfaces.

```ts
// We have to add 'extends Record<string, any>'
// which adds an index signature that we do not really want
export interface JediStore extends Record<string, any> {
  jediName: string
  jediPower: number
}
export const useChartComposerDataStore = create<JediStore>(...)
```

With this fix the index signature is no longer mandatory.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -103,7 +103,7 @@ export const devtools = <S extends State>(
           message.payload.nextLiftedState?.computedStates || []
 
         computedStates.forEach(
-          ({ state }: { state: PartialState<S, keyof S> }, index: number) => {
+          ({ state }: { state: PartialState<S> }, index: number) => {
             const action = actions[index] || api.devtools.prefix + 'setState'
 
             if (index === 0) {
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -103,7 +103,7 @@ export const devtools = <S extends State>(
           message.payload.nextLiftedState?.computedStates || []
 
         computedStates.forEach(
-          ({ state }: { state: PartialState<S> }, index: number) => {
+          ({ state }: { state: PartialState<S, keyof S> }, index: number) => {
             const action = actions[index] || api.devtools.prefix + 'setState'
 
             if (index === 0) {
@@ -136,9 +136,9 @@ export const combine = <
     {},
     initialState,
     create(
-      set as SetState<PrimaryState>,
+      (set as unknown) as SetState<PrimaryState>,
       get as GetState<PrimaryState>,
-      api as StoreApi<PrimaryState>
+      (api as unknown) as StoreApi<PrimaryState>
     )
   )
 
@@ -240,7 +240,8 @@ export const persist = <S extends State>(
 
     if (whitelist) {
       Object.keys(state).forEach((key) => {
-        !whitelist.includes(key) && delete state[key]
+        const keyTyped = key as keyof S
+        !whitelist.includes(keyTyped) && delete state[keyTyped]
       })
     }
     if (blacklist) {
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -239,9 +239,8 @@ export const persist = <S extends State>(
     const state = { ...get() }
 
     if (whitelist) {
-      Object.keys(state).forEach((key) => {
-        const keyTyped = key as keyof S
-        !whitelist.includes(keyTyped) && delete state[keyTyped]
+      ;(Object.keys(state) as (keyof S)[]).forEach((key) => {
+        !whitelist.includes(key) && delete state[key]
       })
     }
     if (blacklist) {
--- a/src/shallow.ts
+++ b/src/shallow.ts
@@ -1,9 +1,4 @@
-type Obj = Record<string | number | symbol, unknown>
-
-export default function shallow<T extends any, U extends any>(
-  objA: T,
-  objB: U
-) {
+export default function shallow<T, U>(objA: T, objB: U) {
   if (Object.is(objA, objB)) {
     return true
   }
@@ -15,14 +10,14 @@ export default function shallow<T extends any, U extends any>(
   ) {
     return false
   }
-  const keysA = Object.keys(objA as object)
-  if (keysA.length !== Object.keys(objB as object).length) {
+  const keysA = Object.keys(objA)
+  if (keysA.length !== Object.keys(objB).length) {
     return false
   }
   for (let i = 0; i < keysA.length; i++) {
     if (
       !Object.prototype.hasOwnProperty.call(objB, keysA[i]) ||
-      !Object.is((objA as Obj)[keysA[i]], (objB as Obj)[keysA[i]])
+      !Object.is(objA[keysA[i] as keyof T], objB[keysA[i] as keyof U])
     ) {
       return false
     }
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,4 +1,8 @@
-export type State = Record<string | number | symbol, unknown>
+// https://github.com/microsoft/TypeScript/issues/42825
+// https://github.com/microsoft/TypeScript/issues/41746
+// https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGwIFUB2BrVB7A7qqAvFAEoQDGWATgCYA8AzsJQJaoDmUAPlKgK4C2AIwiUuUeiCFYANgBoovDNjwA+AFChIsBEgCCqEIRLkqdRi3Zi+QkWIlS5UeAfVqKqRlABmqNJlz4RLQAKlAQAB5IqNT02ogoSgEqABTwlGwAXFDBAJSEKk7pblgewN6o+oZBoRFRMXF6LqnpWbn5hWxqaqxIlF7wZNAAUhDUzHDxUADealBzUABWo8wAcvD8EFnmrJ3zi8sACrgiWdbClGoAvsWlUCBY1PBZI2MTSEYze0tjaxtZAOQATQe8H+slm82+zCOOBOUAAjAAGK5dHx+ZSoZL3R45NQ+SpYkE5IA
+// TLDR; Record<string, any> matches any object, while Record<string, unknown> requires an index signature.
+export type State = Record<string | number | symbol, any>
 // types inspired by setState from React, see:
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c49e45842358ba59a508e13130791989911430d/types/react/v16/index.d.ts#L489-L495
 export type PartialState<T extends State, K extends keyof T = keyof T> =
@@ -41,7 +45,9 @@ export default function create<TState extends State>(
   const listeners: Set<StateListener<TState>> = new Set()
 
   const setState: SetState<TState> = (partial, replace) => {
-    const nextState = typeof partial === 'function' ? partial(state) : partial
+    // TODO: Fix me once https://github.com/microsoft/TypeScript/issues/37663 is resolved
+    // https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342
+    const nextState = partial instanceof Function ? partial(state) : partial
     if (nextState !== state) {
       const previousState = state
       state = replace
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,8 +1,4 @@
-// https://github.com/microsoft/TypeScript/issues/42825
-// https://github.com/microsoft/TypeScript/issues/41746
-// https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGwIFUB2BrVB7A7qqAvFAEoQDGWATgCYA8AzsJQJaoDmUAPlKgK4C2AIwiUuUeiCFYANgBoovDNjwA+AFChIsBEgCCqEIRLkqdRi3Zi+QkWIlS5UeAfVqKqRlABmqNJlz4RLQAKlAQAB5IqNT02ogoSgEqABTwlGwAXFDBAJSEKk7pblgewN6o+oZBoRFRMXF6LqnpWbn5hWxqaqxIlF7wZNAAUhDUzHDxUADealBzUABWo8wAcvD8EFnmrJ3zi8sACrgiWdbClGoAvsWlUCBY1PBZI2MTSEYze0tjaxtZAOQATQe8H+slm82+zCOOBOUAAjAAGK5dHx+ZSoZL3R45NQ+SpYkE5IA
-// TLDR; Record<string, any> matches any object, while Record<string, unknown> requires an index signature.
-export type State = Record<string | number | symbol, any>
+export type State = object
 // types inspired by setState from React, see:
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c49e45842358ba59a508e13130791989911430d/types/react/v16/index.d.ts#L489-L495
 export type PartialState<T extends State, K extends keyof T = keyof T> =
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -45,9 +45,12 @@ export default function create<TState extends State>(
   const listeners: Set<StateListener<TState>> = new Set()
 
   const setState: SetState<TState> = (partial, replace) => {
-    // TODO: Fix me once https://github.com/microsoft/TypeScript/issues/37663 is resolved
+    // TODO: Remove type assertion once https://github.com/microsoft/TypeScript/issues/37663 is resolved
     // https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342
-    const nextState = partial instanceof Function ? partial(state) : partial
+    const nextState =
+      typeof partial === 'function'
+        ? (partial as (state: TState) => TState)(state)
+        : partial
     if (nextState !== state) {
       const previousState = state
       state = replace
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -661,7 +661,7 @@ it('ensures a subscriber is not mistakenly overwritten', async () => {
 })
 
 it('can use exposed types', () => {
-  interface ExampleState extends State {
+  interface ExampleState {
     num: number
     numGet: () => number
     numGetState: () => number
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -677,8 +677,11 @@ it('can use exposed types', () => {
     }
   }
   const selector: StateSelector<ExampleState, number> = (state) => state.num
-  const partial: PartialState<ExampleState> = { num: 2, numGet: () => 2 }
-  const partialFn: PartialState<ExampleState> = (state) => ({
+  const partial: PartialState<ExampleState, 'num' | 'numGet'> = {
+    num: 2,
+    numGet: () => 2,
+  }
+  const partialFn: PartialState<ExampleState, 'num' | 'numGet'> = (state) => ({
     ...state,
     num: 2,
   })
@@ -717,7 +720,7 @@ it('can use exposed types', () => {
 
   function checkAllTypes(
     _getState: GetState<ExampleState>,
-    _partialState: PartialState<ExampleState>,
+    _partialState: PartialState<ExampleState, 'num' | 'numGet'>,
     _setState: SetState<ExampleState>,
     _state: State,
     _stateListener: StateListener<ExampleState>,
--- a/readme.md
+++ b/readme.md
@@ -425,6 +425,7 @@ interface BearState {
   increase: (by: number) => void
 }
 
+// And it is going to work for both
 const useStore = create<BearState>(set => ({
   bears: 0,
   increase: (by) => set(state => ({ bears: state.bears + by })),
--- a/readme.md
+++ b/readme.md
@@ -413,26 +413,22 @@ devtools will only log actions from each separated store unlike in a typical *co
 ## TypeScript
 
 ```tsx
-type State = {
+// You can use `type`
+type BearState = {
   bears: number
   increase: (by: number) => void
 }
 
-const useStore = create<State>(set => ({
+// Or `interface`
+interface BearState {
+  bears: number
+  increase: (by: number) => void
+}
+
+const useStore = create<BearState>(set => ({
   bears: 0,
   increase: (by) => set(state => ({ bears: state.bears + by })),
 }))
-```
-
-You can also use an `interface`:
-
-```tsx
-import { State } from 'zustand';
-
-interface BearState extends State {
-  bears: number
-  increase: (by: number) => void
-}
 ```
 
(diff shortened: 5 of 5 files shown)


src/vanilla.ts
```diff
@@ -41,7 +45,9 @@ export default function create<TState extends State>(
   const listeners: Set<StateListener<TState>> = new Set()
 
   const setState: SetState<TState> = (partial, replace) => {
-    const nextState = typeof partial === 'function' ? partial(state) : partial
+    // TODO: Fix me once https://github.com/microsoft/TypeScript/issues/37663 is resolved
+    // https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342
+    const nextState = partial instanceof Function ? partial(state) : partial
```
dai-shi, line 50: I didn't know the 37663 issue, but hit the issue behavior in other projects. If this is only about typing, can we just add a type assertion until the issue is resolved? That's what I do for all other projects.

dai-shi (commented):
This is really nice. Really wanted someone with better TS knowledge to come.

src/vanilla.ts
aigoncharov: I was torn between type assertion and `instanceof` as well. Seems like tomayto/tomahto kind of thing to me. Fixed in https://github.com/pmndrs/zustand/pull/376/commits/9a48fa8dce62fadf75866a4a0569db857e7ab9f0

src/vanilla.ts
```diff
@@ -1,4 +1,8 @@
-export type State = Record<string | number | symbol, unknown>
+// https://github.com/microsoft/TypeScript/issues/42825
+// https://github.com/microsoft/TypeScript/issues/41746
+// https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGwIFUB2BrVB7A7qqAvFAEoQDGWATgCYA8AzsJQJaoDmUAPlKgK4C2AIwiUuUeiCFYANgBoovDNjwA+AFChIsBEgCCqEIRLkqdRi3Zi+QkWIlS5UeAfVqKqRlABmqNJlz4RLQAKlAQAB5IqNT02ogoSgEqABTwlGwAXFDBAJSEKk7pblgewN6o+oZBoRFRMXF6LqnpWbn5hWxqaqxIlF7wZNAAUhDUzHDxUADealBzUABWo8wAcvD8EFnmrJ3zi8sACrgiWdbClGoAvsWlUCBY1PBZI2MTSEYze0tjaxtZAOQATQe8H+slm82+zCOOBOUAAjAAGK5dHx+ZSoZL3R45NQ+SpYkE5IA
+// TLDR; Record<string, any> matches any object, while Record<string, unknown> requires an index signature.
+export type State = Record<string | number | symbol, any>
```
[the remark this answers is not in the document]
aigoncharov: Thank you for creating `zustand` in the first place! Updated the TS-related section - https://github.com/pmndrs/zustand/pull/376/commits/a40c32c261d875014c4d3833cfd726e2dc92ce8f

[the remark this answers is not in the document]
aigoncharov: @dai-shi regarding `object`, I think this one should more or less cover it https://stackoverflow.com/a/61000565/10687034
I created a small example 
https://www.typescriptlang.org/play?strict=true#code/C4TwDgpgBAkgygVwEYBVwQPIDMCiBbMUDJAKwB4UA+KAXihSggA9gIA7AEwGcoBvAXygB+KMABOCaAC4oWAIYAbLhABQoSLESp02fIRAAlCAGMA9mI4VqdBs1aceRsxbJdxASzYBzKAB8obAh4SBBiflBcIMGmCgA0UHJsINQi4pJQMvJKqmroUADCABYmANYAjHK0mshokLoERKRkvEhyYjIATPzUAPQ9UCAQPO7AUO5cuRpFpR2VdPA1OrgNIMTkvFimpp3xre1QAORyB91QfQNDY6Pjk9DTxiUAzHPV2nXL+mtkgcGhvf1cBBiMBicaeLwKECiQpya48RRcUyiCQ5W4FYoPMpIKoLN6YD6gJzmSwtNqdU7nQbDOFo+4lDrY+ZaWr4vSEkzE5qbbZQDq7MmHY4U-pUq5jCbqO4Yp6M14s+r6IkuH4hMT-eiFcZQUxsaBarLKBI8ZiQYysDhAA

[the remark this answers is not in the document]
aigoncharov: I have double-checked `combine`. It seems to keep working with the updated typings https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKMCGNUF44BKwAxtACYA8AzjFAJYB2A5nAD5wMCuAtgEbBQ2cKoj4QANgBo4mBogB8AKEWhIsBMlQAFTLDqZxGbMAoAVOKBwMyVdFhzSA0hZBWbcANbBEEAGZxzfC8ff1N5OFxFOCEACi06Yg8zJ3D2UwBKKNiYmmMALgD0iPD4xOS4R1TC5RVwaHgkFDtjNGBxEhhoMxc3WyMHOABVcPwc+2ACjOKh2rUGzTgAUQBHTgM6JABhAAsSLygzEbgx-IDpBmAAd36JwuneCAlgWVn6jSabgBk6GmALg7CEWOuRwk2kYCgwAAbnQIJwqDdJkVcOEoRA6GRXupGqgbmhxAlgN9fv9DkCcgTiLdTODITC4QjKdTkaj0ZjVG9GDgoL5MFT0JxeFRiPR+N1LH93DdwgBvLIxAkkgQFL4-KwCQ7pAoxFlwNEYrIUPFM+QxLLRRXqqAq8b4wnEq1G20myTm4RtDrQG0tD3EToA6TGwnyV3RaLAVbrJAAMQYAH4Cis1gStrtEhqg1SlNEtcddfrMQBfGoc7ELVowG7i1ySvrjI5y6IUZwS6y2YJ+AKmsC6GD6cQFHR6AxVmkVENwSFgcR84AJuAPJ6yXMFxTF0vzJoAcWAlfG1d6zRwRx101Myg371QABFgDQoBBEOT82ysfAuQJefyMNBgABBMA6APWsj2AWUsioXdEXQKD9zCLImFg04dz3YxDggwVhVFW40EwkU6DFeDojIO9aEfApb3vR81xLOoyw+cZNkhbAunMVspXGaRNnhTpuArG4gX4uD5BPCDdwKbiaAgPikIcBDxLgFDRxDLJMEAm1fwAoD4N1c83zgUgGBoAzpN4Rg8DgCgsi0ehuF0RABPYutjFDGDDLIezHJrNtQMUU0skYDZ+2gmy6DsqAHM4rIRWeUFjjdSCYBVWSTFC8LIuMFSwzgRCksUlKKDSzz61c6I1LoDTIS0wrbOKzKsl1Vp3LqnBFFzG4mNiroioigSADI3IgawWrA6Ycl3aRcukcqWSyAB5XgACsOgAOkwKgqDoJgGDNbKZULUq4ECvsRyi7KYuMXbsvdeB1pg1CcBqsKRqy67cpkWwlP3HqMuPQ6ysAj67E0wCnvS6U3UyHNlGiZRDOM1QOlMBZ8GSU0oQMThmQKAtplxgAGABuOGhuM+FgB-SEgWi0zzKu6IZQXZ4oCoAp8bgA64DdcaYF1GJGcYC7IO1XhEAKLg+AEXVEpiE4cD5xn+F0VnhHGFalZZuAAGoFyfQt0iKfWGovcAkc0J1jE6liDkV5mVYl-goEJo6GCF24YlF8WeEdl8MQ5kSYnJyngEyZQgA

dai-shi:
https://github.com/pmndrs/zustand/pull/376#discussion_r625737319
Thanks for the investigation. You compare `{}` and `Record<string | number | symbol, any>`, but what I mean is to compare with `object`.
https://www.typescriptlang.org/play?ssl=2&ssc=43&pln=2&pc=80#code/C4TwDgpgBAkgygVwEYBVwQPIDMCiBbMUDJAKwB4UA+KAXihSggA9gIA7AEwGcoBvAXygB+KMABOCaAC4oWAIYAbLhABQoSLESp02fIRAAlCAGMA9mI4VqdBs1aceRsxbJdxASzYBzKAB8obAh4SBBiflBcIMGmCgA0UHJsINQi4pJQMvJKqurQ8MhokNjEJCbAVrT0jCzs3FCmpGXCohLSsorKKmroUADCABYmANYAjHKV+dpFuAREpGS8SHJiMgBM-NQA9JtQIBA87sBQ7lzdGgPDq+N0k4WYM-olC1impmvxSytQAORy3xtQba7fbHI4nM7QC7GIYAZmumgKOgec3IgWCoS2Oy4CDEYDEJ08XgUIFE-TkYJ4HVMLUkXQhfUG0JGSAmWjuulmhhM5ksi2WawBQL2Bwp9KhQ1WLJubKRelATh5z1e7ygnxkv3+mOBIuOp1yDOGMKlCKm9zlXOcljRITEWpQ-RO9TY0EdWWUCR4zEgxlYHC6+vFI2MrMR0xKZQWaqg6y1wtBurFjIlwelofu4Z9Sre0Y+-J+f0FOzjhwTAaTMJTJvZWAz5WtGMBO3tjtMzt17WyHuq3t9oleKiAA

If there's zero difference, I might prefer `type State = object`.

src/vanilla.ts
dai-shi: Thanks! It will keep the bundle JS code identical.

src/vanilla.ts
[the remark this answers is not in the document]
dai-shi: Thanks for updating readme.
As for `Record` vs `object`, I replied in the main thread. Making this conversation resolved.

dai-shi:
Reading: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md

> - Avoid the `object` type, as it is currently hard to use due to not being able to assert that keys exist.
>   - See https://github.com/microsoft/TypeScript/issues/21732.

Hmm, I'm not sure what "assert that keys exist" means.

---

At least the original issue is resolved.
https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGwIFUB2BrVB7A7qqAvFAEoQDGWATgCYA8AzsJQJaoDmUAPlKgK4C2AIwiUuUeiCFYANgBoovDNjwA+AFChIsBEgCCqEIRLkqdRi3Zi+QkWIlS5UeAfWbocRBADyggFblgIyw-ALU1ClRGKAAzVDRMXHwiWgAVKAgADyRUanptT3jlVBUACnhKNgAuKBSASkIVJwrwrEjA2P1DZLTM7Nz8vRcyiuq6hqa2FraY1B9-MkDu9KyIHLyPJDmA0vKqmvqCRt2w1iRKaPgyaAApCGpmDegAbzUoN6h-e4A5eH4IavMrEm7w+d2YAAVcCJqtZhJQ1ABfKZREBYajwaq3e6PIwvEGfZg-P7VADkAE00fASbJXu8CZCcNCoABGAAMiLCsUKiRKqPRtTUHQMvMpAtiWwWIv5QA

aigoncharov:
> [#376 (comment)](https://github.com/pmndrs/zustand/pull/376#discussion_r625737319)
> Thanks for the investigation. You compare `{}` and `Record<string | number | symbol, any>`, but what I mean is to compare with `object`.
> https://www.typescriptlang.org/play?ssl=2&ssc=43&pln=2&pc=80#code/C4TwDgpgBAkgygVwEYBVwQPIDMCiBbMUDJAKwB4UA+KAXihSggA9gIA7AEwGcoBvAXygB+KMABOCaAC4oWAIYAbLhABQoSLESp02fIRAAlCAGMA9mI4VqdBs1aceRsxbJdxASzYBzKAB8obAh4SBBiflBcIMGmCgA0UHJsINQi4pJQMvJKqurQ8MhokNjEJCbAVrT0jCzs3FCmpGXCohLSsorKKmroUADCABYmANYAjHKV+dpFuAREpGS8SHJiMgBM-NQA9JtQIBA87sBQ7lzdGgPDq+N0k4WYM-olC1impmvxSytQAORy3xtQba7fbHI4nM7QC7GIYAZmumgKOgec3IgWCoS2Oy4CDEYDEJ08XgUIFE-TkYJ4HVMLUkXQhfUG0JGSAmWjuulmhhM5ksi2WawBQL2Bwp9KhQ1WLJubKRelATh5z1e7ygnxkv3+mOBIuOp1yDOGMKlCKm9zlXOcljRITEWpQ-RO9TY0EdWWUCR4zEgxlYHC6+vFI2MrMR0xKZQWaqg6y1wtBurFjIlwelofu4Z9Sre0Y+-J+f0FOzjhwTAaTMJTJvZWAz5WtGMBO3tjtMzt17WyHuq3t9oleKiAA
> 
> If there's zero difference, I might prefer `type State = object`.

For some reason, I thought `{}` and `object` are aliases to the same thing. You live - you learn :)
I haven't found any evidence that `Record<string | number | symbol, any>` should be preferred over `object`. However, they behave a bit differently. With `object` we'll need a type assertion for `combine`.
https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKMCGNUF44QBGAVsAMYwBQloksCyqACprAJaYA2G2wAPABU4oHADsAJgGd0WHABo4AaWEgxUuAGtgiCADM4Q-Fp36BAPji5KcOAB84ACiZsyGwQsUX7AgJTW7jg6SssAAXAY+lhbOru5KXhHUNODQ8EgoMrxowJzkMNCCKmrSPPJwAKoW+EEh4b5RFcl0aYxwAKIAjgCuXGxIAMIAFuRaUIJVjsG8dQqiwADupWERDYQQELmYok2pDBlLADJswcBzY+aWk7UGCmBQwABubBBdkkt1kbgWDxBs4jv0dKoJZoTguYBHE5ncaXIJgsjLAS3e5PF5veGIz7fX7-Wi7NiiHBQXSYBHoLqESRkKBsQj8IQiU7qJYWADe-gcYKhwCg4UOxzEPPGPnCDixcB+f38fBBGLMDn8Ni5gt5mRwoPBkJVMpCGoRZjkirgkhyeWgfN1pooBSRauAeuABqNwG6vSQADFRAB+cKdHpggbDVxC2Xgsz+EWOcWS-4AXySeMBrWyMCWhUZEhKIQm7JsfGUGfUxj0BnlYFYMA4nHCLHYXDTts8CnuYE4pOAPrgaw2wC2kZjlHjiZaGQA4sBUyF06omVneBMxQ0BNRh3tUAARYDBKAQRCw6M4gHwAlEklkjDQYAAQTAbGnxTtbP8Jsn03QE4b4ZsAHMP9dx6+ODjM+FJUjSdJ8qB1K0vSX5wOIW4wDuiDhJu267oOCYpEm+whP09zYDaRSznaCj9K8+QALYpkslw0VO5gLs+E7hORwQQNRf68IaP4sXAAGfjxcCYLeFqXjed7mJQ4rLkecBkBAojBPJHGEASeBwHw-hMDSlGsIgtGFnO8j+NkCkSPphkzpmdqUPK-gEn0VbvHAOlsHpUAGSEQnUr2OCikaL58lxQFuR5XnzkJvEwOEAlTmFlnZlFwmiTI4m3nwCWeSyEYNGZiniIlvDSRavD4X5BRZRFOBwAAZO+5mFdl2YNEEE4KL+MAKCJbBYv4ADyJB5AAdJgkiSGw36iAqNg2KysbJY5lb1t5Rq+bwM2zTYL7CSUIX8FVLLJdFu38ftmW6UVOBOltNg9adF73BJF3uVdjpGn4Nh+P41DmcptB5AIrT4O48oPFwXSYuEMYNDDAAMADcv2Kcprz2vk9yXP4CmUWpcybXNXa9lAkjhHDcALXARptTA4oOKycAEutJqioQKFwKIXS4zy4ovg4NS8HTDN0qwpPGiEw0iyTcAANRdnusY+JEisRiu4CA4wOplQRGN8MLxNi5z3NQAjjOiMzywOGz4RG3SUAHn8FNmPKaOPcA32UEAA

dai-shi:
Thanks! I'd prefer `object` for no specific reason, if there were no difference. Until we learn something new, let's go with `Record`.

dai-shi (approved):

dai-shi:
Actually, my ideal goal is to make it `State = unknown` and disallow any object, but still want to ensure state to be object (`number` is forbidden, for example).

dai-shi:
2a143c6 👁️

aigoncharov:
@dai-shi I would prefer `object` as well, just for simplicity. 
As we can see from this commit, `object` is still a bit different -https://github.com/pmndrs/zustand/pull/376/commits/2a143c6bdfdb2a352c44a5055717b8118da05312. Not in a bad way necessarily.
BTW, was there a specific reason to use `extends any` for `shallow`? My understanding that everything extends any by default. I tried to remove it in https://github.com/pmndrs/zustand/pull/376/commits/2a143c6bdfdb2a352c44a5055717b8118da05312 and it simplified the typings in `shallow` a bit.
Let me know if we need to revert https://github.com/pmndrs/zustand/pull/376/commits/2a143c6bdfdb2a352c44a5055717b8118da05312

src/middleware.ts
```diff
@@ -103,7 +103,7 @@ export const devtools = <S extends State>(
           message.payload.nextLiftedState?.computedStates || []
 
         computedStates.forEach(
-          ({ state }: { state: PartialState<S> }, index: number) => {
+          ({ state }: { state: PartialState<S, keyof S> }, index: number) => {
```
dai-shi: Just curious why this change is required.

src/middleware.ts
```diff
@@ -136,9 +136,9 @@ export const combine = <
     {},
     initialState,
     create(
-      set as SetState<PrimaryState>,
+      (set as unknown) as SetState<PrimaryState>,
```
dai-shi, line 139: This looks unfortunate, but can be accepted.

src/middleware.ts
```diff
@@ -240,7 +240,8 @@ export const persist = <S extends State>(
 
     if (whitelist) {
       Object.keys(state).forEach((key) => {
-        !whitelist.includes(key) && delete state[key]
+        const keyTyped = key as keyof S
+        !whitelist.includes(keyTyped) && delete state[keyTyped]
```
dai-shi, line 244: I would do this. Does it work?
```ts
      ;(Object.keys(state) as (keyof S)[]).forEach((key) => {
        !whitelist.includes(key) && delete state[key]
```

src/shallow.ts
```diff
@@ -1,9 +1,4 @@
-type Obj = Record<string | number | symbol, unknown>
-
-export default function shallow<T extends any, U extends any>(
-  objA: T,
-  objB: U
-) {
+export default function shallow<T, U>(objA: T, objB: U) {
```
dai-shi, line 1: This looks good. Thanks for the fix. (It was more or less a historical reason.)

tests/basic.test.tsx
```diff
@@ -677,8 +677,11 @@ it('can use exposed types', () => {
     }
   }
   const selector: StateSelector<ExampleState, number> = (state) => state.num
-  const partial: PartialState<ExampleState> = { num: 2, numGet: () => 2 }
-  const partialFn: PartialState<ExampleState> = (state) => ({
+  const partial: PartialState<ExampleState, 'num' | 'numGet'> = {
```
dai-shi, line 680: I see. Looks more precise.
BTW, we should remove `extends State` in L664.

dai-shi (commented):
This looks good.

src/middleware.ts
dai-shi: already resolved.

src/middleware.ts
aigoncharov: Addressed in https://github.com/pmndrs/zustand/pull/376/commits/8473244cce1fa4a10d8a63cc977bae50508d0813

src/middleware.ts
aigoncharov: Agreed. I haven't found a way around it :(

tests/basic.test.tsx
aigoncharov: I wish we didn't have to do it, but without the explicit second argument it does not pass the assertion in `checkAllTypes` after our migration to `object`. I would say it even makes sense :)
At the same time, if we need to make all these changes in the library code because of the migration to `object`, there is a risk it might break code that consumes the library.
I see two ways to go from here to mitigate the risk:
1. Migrate to `object` and make it a minor release, not a patch one.
2. Revert changes related to `object` migration and keep using `Record`.

Personally, I would suggest option #1.

src/middleware.ts
aigoncharov: FYI, it was leftover from my experiment with removing the default value for the second generic.

tests/basic.test.tsx
dai-shi: Yes, this looks good and more like what I wanted. Let's go with minor, and describe breaking changes in types in the release note.

dai-shi (approved):

aigoncharov:
@dai-shi thank for the conversation and for your review! What is the process after your approval?

dai-shi:
It depends. Sometimes, we just wait for others to review.
For this one, I will merge it and make a release soon.

dwelle:
> Avoid the object type, as it is currently hard to use due to not being able to assert that keys exist.
See microsoft/TypeScript#21732.

Just came upon this topic. FYI, what the above refers to is the fact that `object` has no index signature (unlike `Record`), and you cannot use the `in` as a type guard ([playground](https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=13&pc=2#code/MYewdgzgLgBCMF4YG8C+MCGE4CMBWApsFANwBQZAlgGYwAUA5BgzJWHAJQpkwyiQgANgQB0gkAHM6IERg7lUFAPRKYAWg2at2nbr1aK-aHABMiFOiwoMALhgAWE+gA+KHHYCMABi+pyVWkZmVnYQEy5kHj5wCCFRcSkw2XkyVCA)):

```
if ('a' in o) {
	return o.a; // does not type-check for `object` type
}
```

dai-shi:
Thanks. Okay, that's actually what I probably [wanted](https://github.com/pmndrs/zustand/pull/376#issuecomment-832477249). So, this should be a valid use case of `object`.

[playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBCMF4YG8C+MCGE4CMBWApsFANwBQAlgGYwAUA5BvTBWHAJQpkwyiQgAbAgDoBIAOa0QwjO3KoyZPtDgAmRCnRYUGAPwAuGABZVqctTqNmrNZ2Tde4CIJFjJIVTLlkFSp7BAAZg00TGwAJSIQACcAEwAeaGjWcQAaGABXMABrMBAAdzAAPkoaBiYWNiC7B2UXUQkpQK95RWUAoxCtCKi4xKhksDTMMABPEoty6yqjGp46oQb3IxafMiA)

Issue #377, opened 2021-05-05
How to use hook methods getState, setState, and subscribe with the react context hook?


omasback:
Trying out the new context hook. Works great for the base case of selectors, but there doesn't seem to be a way to use the methods `useStore.getState`, `useStore.setState`, and especially `useStore.subscribe`. I can probably work around the lack of get and set, but `subscribe` is important for doing renderless updates with react-spring.

Looking at the `zustand/context` code, it's not obvious to me how this could be implemented. But maybe somebody smarter than me has an idea?

dai-shi:
Ah, nice catch. I haven't thought about the use case, but yeah `subscribe` seems important.
IMO, `useStore.subscribe` is not very well designed for context usage. So, instead of extending `useStore`, I'd suggest adding `useStoreApi`.

```js
const { Provider, useStore, useStoreApi } = createContext()
```

Would you or anyone like to work?

dai-shi:
Fixed by #380. Thanks for your work!

PR #380, opened 2021-05-08
Add useStoreApi context hook

omasback:
Discussed here: https://github.com/pmndrs/zustand/issues/377

Not sure if you want size-snapshot included. I can do another PR without that change if you want.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -50,9 +50,21 @@ function createContext<TState extends State>(_initialState?: TState) {
     )
   }
 
+  const useStoreApi = () => {
+    // ZustandContext value is guaranteed to be stable.
+    const useProviderStore = useContext(ZustandContext)
+    if (!useProviderStore) {
+      throw new Error(
+        'Seems like you have not used zustand provider as an ancestor.'
+      )
+    }
+    return useProviderStore
+  }
+
   return {
     Provider,
     useStore,
+    useStoreApi,
   }
 }
 
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -161,9 +161,9 @@
     }
   },
   "context.js": {
-    "bundled": 815,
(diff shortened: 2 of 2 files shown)


src/context.ts
```diff
@@ -50,9 +50,21 @@ function createContext<TState extends State>(_initialState?: TState) {
...
 
+  const useStoreApi = () => {
+    // ZustandContext value is guaranteed to be stable.
+    const useProviderStore = useContext(ZustandContext)
+    if (!useProviderStore) {
+      throw new Error(
+        'Seems like you have not used zustand provider as an ancestor.'
+      )
+    }
+    return useProviderStore
```
dai-shi, line 61: Oops, sorry, I should have explained better.
This is exactly the pattern we want to avoid. Like, this is the whole point of this utility.
(If this were possible, we wouldn't need `useStore`, or even this utility.)

I hope something like this would work?
```suggestion
    return {
      getState: useProviderStore.getState,
      setState: useProviderStore.setState,
      subscribe: useProviderStore.subscribe,
    }
```

dai-shi (changes requested):

src/context.ts
omasback: Ah okay you want to "de-functionalize" it, I suppose? What about the `destroy` method?

dai-shi: > Ah okay you want to "de-functionalize" it, I suppose?

Yes. It would avoid misusage. (Otherwise, people can violate rules of hooks and eslint can't detect it.)

> What about the `destroy` method?

Oh, I missed it. But, should we add it? I don't see any use cases, but adding it is fine.

dai-shi (approved):
The code looks good. Would you like to add a test for it?

dai-shi (approved):

PR #384, opened 2021-05-09
Allow optional selector in zustand/vanilla

gamedevsam:
This type better reflects the optional nature of the parameter.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -10,14 +10,9 @@ export type StateListener<T> = (state: T, previousState: T) => void
 export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
 export interface Subscribe<T extends State> {
   (listener: StateListener<T>): () => void
-  (
-    listener: StateSliceListener<unknown>,
-    selector?: StateSelector<T, unknown>,
-    equalityFn?: EqualityChecker<unknown>
-  ): () => void
   <StateSlice>(
     listener: StateSliceListener<StateSlice>,
-    selector: StateSelector<T, StateSlice>,
+    selector?: StateSelector<T, StateSlice>,
     equalityFn?: EqualityChecker<StateSlice>
   ): () => void
 }


src/vanilla.ts
```diff
@@ -12,7 +12,7 @@ export interface Subscribe<T extends State> {
   (listener: StateListener<T>): () => void
   <StateSlice>(
     listener: StateSliceListener<StateSlice>,
-    selector: StateSelector<T, StateSlice>,
+    selector?: StateSelector<T, StateSlice>,
```
dai-shi, line 15: Hm, but if this works, we don't need L12?
What will be `StateSlice` inferred if you don't specify the selector?

If there's a better way of typing subscribe, it's very welcomed. Would you explain it in detail?

gamedevsam: The reason for this change is that I wrap the subscribe function and allow users to optionally provide their own selector. So I am passing in a parameter that can be undefined (as a result, line 12 does not apply). 

This typing change enables this type of workflow:

```ts
function subscribeToStore(store: StoreApi<any>, callback: (selectedState) => void, selector?: (state) => any) {
    // this causes an error without the proposed change
    store.subscribe(callback, selector);
}

Comment (gamedevsam) [reply src/vanilla.ts:15]: And now that you mention it, line 12 seems unnecessary.

Comment (gamedevsam) [reply src/vanilla.ts:14]: Ok now I actually understand what this line was doing:

```ts
<StateSlice>(listener: StateSliceListener<StateSlice>, selector: StateSelector<T, StateSlice>, equalityFn?: EqualityChecker<StateSlice>): () => void;
```

It's not very common TS to infer a generic like that, so had to wrap my head around it.

I revised my fix to preserve the type inference while still enabling more dynamic workflows like the one I proposed.

Rather than change the existing types the interface can map to, I've added new one:

```ts
(listener: StateListener<any>, selector?: StateSelector<T, any>, equalityFn?: EqualityChecker<any>): () => void;
```

Basically the "do whatever you like" type. Everything works exactly as it is now, but if you pass a potentially undefined type as the selector, TS compiles, but you give up type assistance.

I wasn't sure if I should use `unknown` or `any`, but decided to be more permissive here to avoid annoying users that "know what they're doing". Let me know if you prefer `unknown` over `any`.

src/vanilla.ts
```diff
@@ -10,6 +10,11 @@ export type StateListener<T> = (state: T, previousState: T) => void
 export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
 export interface Subscribe<T extends State> {
   (listener: StateListener<T>): () => void
+  (
+    listener: StateSliceListener<any>,
+    selector?: StateSelector<T, any>,
+    equalityFn?: EqualityChecker<any>
+  ): () => void
```
dai-shi, line 17: You seem to get the point. (I was replying to your old comment, and deleted.)

```suggestion
  (
    listener: StateSliceListener<T>,
    selector: undefined,
    equalityFn?: EqualityChecker<T>
  ): () => void
```

Does this work?

gamedevsam: It does not sadly:

```
Argument of type '((state: unknown) => AnyObject) | undefined' is not assignable to parameter of type 'StateSelector<S, unknown>'.
      Type 'undefined' is not assignable to type 'StateSelector<S, unknown>
```

gamedevsam: My fix works, and I don't really see a downside to including it. Zustand gives you two ways to subscribe and preserve types, but if you need to do some dynamic untyped stuff you can also do that. If I want to shoot myself in the foot, Zustand shouldn't stand in my way xD

dai-shi:
Unfortunately, this will make some typing looser, which is a downside.

We want to keep the type safety for the normal use case.

https://codesandbox.io/s/react-typescript-forked-jb69r?file=/src/App.tsx
with v3.5.1, it shows type error.

https://codesandbox.io/s/react-typescript-forked-hqdzx?file=/src/App.tsx
with 46f450d, it doesn't because of `any`.

gamedevsam:
> Unfortunately, this will make some typing looser, which is a downside.
> 
> We want to keep the type safety for the normal use case.
> 
> https://codesandbox.io/s/react-typescript-forked-jb69r?file=/src/App.tsx
> with v3.5.1, it shows type error.
> 
> https://codesandbox.io/s/react-typescript-forked-hqdzx?file=/src/App.tsx
> with [46f450d](https://github.com/pmndrs/zustand/commit/46f450d27214a6fc5d03c6f682423a4613f3467c), it doesn't because of `any`.

This example is totally throwing me off.

The first problem is that it shows a type error when the code is perfectly valid at runtime.

To expand on your example, I ran setState to demonstrate the runtime behavior of the code: https://codesandbox.io/s/react-typescript-forked-gmmjs?file=/src/App.tsx

A type error on valid code reveals a typing problem.

The second issue is that the type inference for the slice is failing, since it's the parameter of the listener function is being typed as unknown.

For example, this sandbox expands on this issue: https://codesandbox.io/s/react-typescript-forked-k68cn?file=/src/App.tsx

In here we see the following error:

`Object is of type 'unknown'.ts(2571)`

When the real runtime error actually is: `Cannot read property 'toString' of undefined`

I'm doing some more digging into generic type inference and why this may not be working properly.

gamedevsam:
I did some research and couldn't figure out how to get the type inference to work as we expect it to. As a result I changed the `any` type to `unknown` type to preserve the old behavior. It's not an ideal solution, since perfectly valid runtime code will trigger type errors due to unknown types, but you can easily do a type cast and get on with life.

My proposed solution of adding preserves the current behavior:

```ts
 (
    listener: StateSliceListener<unknown>,
    selector?: StateSelector<T, unknown>,
    equalityFn?: EqualityChecker<unknown>
  ): () => void
```

dai-shi:
Oh, there was a misunderstanding of mine. We want to support this, but even v3.5.1 fails.
```ts
const store = create(() => ({ count: 1 }));
store.subscribe(
  (slice) => {
    console.log(slice * 2);
  },
  (state) => state.count
);
```

https://codesandbox.io/s/react-typescript-forked-umvxc?file=/src/App.tsx

If technically possible, we'd like to support everything including your use case...

dai-shi:
It would have been possible, if selector came before listener.
[playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKMCGNUF44QBGAVsAMYwBQoksCyqACprAJaYA2G2wAPACpxQOAHYATAM7osOADRwA0kJCjJcANbBEEAGZxB+Tdr38AfHFyU4cAD5wAFI1Zl1A+QvN3+ASiu2H9hIywABc+t4W5k4uboqe4dTg0PBIKNI8aMAc5DDQAsqqUtxycACq5viBwWE+kWWJtCkMcACiAI4ArpysSADCABbkmlACFQ5BPDXyIsAA7sWh4XWEEBDZmCINyfRpCwAyrEHAMyNmFuPV+vJgUMAAbqwQHRILNRG45ncQrGJbdKmoBZoDjOYAHI4nUbnQIgsiLfjXW4PJ4vWHw96fb6-GjbVgiHBQHSYOHoDqECRkKCsQh8QTCY5qBbmADefnsIIhwCgYX2h1EXNG3jC9gxcC+Pz8vCBaNM9j81gkWRy0B5wUy2QoeQR6RwwNBplk8rgHP53J1wD1cPBpqlaplhus1mAnW6SAAYiIAPxhdpdEF9QYuAXS-V+IUOUXi34AXz+TV2wBgC3y9PERWCY1Z1l4SlTaiMun0srALBg7A4YWYbE4ye1HnktzAHGJwG9cBWa2AG3DUcosZx-2aAHFE7WCgz0zwxiK6vw4ztUAARYBBKAQRDQyNY+d4glEkkYaDAACCYFYKZUE-NLL8iqTl0y954oz8AHNR5cR0+cC+FWSKVSNI8v+lLUrSph+GIK4wGuiBhMuq7rn2lCUGQEAiEEcBBEeqpHqe57MnAaEdPiYQiB0AC2NJQHA0ZjORHAcHAmBSBsiAANwodhtwAHQSCBgHAHK1hVDwopZo6aEYZ2PEcBAr6iTgPHEficAAFRwAATL41jRoaviUNxwB8QJYHCVhwR1BMchstZwDiUaUkSDJckKXZylPKpGnaX4emUAZRkmeSoE0uZdlWcEHkkTADoOCpMAOY6RHoc52SyfJ9jxepWk6bR+mUEAA)
I wonder if there's any workaround.

gamedevsam:
Haha, KISS principle in action, cleaning up my changes and putting things back the way they were.

gamedevsam:
> It would have been possible, if selector came before listener.
> [playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKMCGNUF44QBGAVsAMYwBQoksCyqACprAJaYA2G2wAPACpxQOAHYATAM7osOADRwA0kJCjJcANbBEEAGZxB+Tdr38AfHFyU4cAD5wAFI1Zl1A+QvN3+ASiu2H9hIywABc+t4W5k4uboqe4dTg0PBIKNI8aMAc5DDQAsqqUtxycACq5viBwWE+kWWJtCkMcACiAI4ArpysSADCABbkmlACFQ5BPDXyIsAA7sWh4XWEEBDZmCINyfRpCwAyrEHAMyNmFuPV+vJgUMAAbqwQHRILNRG45ncQrGJbdKmoBZoDjOYAHI4nUbnQIgsiLfjXW4PJ4vWHw96fb6-GjbVgiHBQHSYOHoDqECRkKCsQh8QTCY5qBbmADefnsIIhwCgYX2h1EXNG3jC9gxcC+Pz8vCBaNM9j81gkWRy0B5wUy2QoeQR6RwwNBplk8rgHP53J1wD1cPBpqlaplhus1mAnW6SAAYiIAPxhdpdEF9QYuAXS-V+IUOUXi34AXz+TV2wBgC3y9PERWCY1Z1l4SlTaiMun0srALBg7A4YWYbE4ye1HnktzAHGJwG9cBWa2AG3DUcosZx-2aAHFE7WCgz0zwxiK6vw4ztUAARYBBKAQRDQyNY+d4glEkkYaDAACCYFYKZUE-NLL8iqTl0y954oz8AHNR5cR0+cC+FWSKVSNI8v+lLUrSph+GIK4wGuiBhMuq7rn2lCUGQEAiEEcBBEeqpHqe57MnAaEdPiYQiB0AC2NJQHA0ZjORHAcHAmBSBsiAANwodhtwAHQSCBgHAHK1hVDwopZo6aEYZ2PEcBAr6iTgPHEficAAFRwAATL41jRoaviUNxwB8QJYHCVhwR1BMchstZwDiUaUkSDJckKXZylPKpGnaX4emUAZRkmeSoE0uZdlWcEHkkTADoOCpMAOY6RHoc52SyfJ9jxepWk6bR+mUEAA)
> I wonder if there's any workaround.

This is strange and very interesting. Why would the order of function definitions alter the behavior of the generic type inference? It kinda seems like a TypeScript compiler bug.

If you're interested in following up, creating an issue with the two playground links you shared above should provide some valuable insights for the TS compiler team.

dai-shi (approved):
Thanks for all those investigations and discussions.

Issue #366, opened 2021-04-18
persist API: store initial value 


kolodziejczak-sz:
Hey guys, first of all, I really like the library! I'd like to extend the PersistOptions API to fit my needs. I'll briefly introduce my use case: I want to set the unique id to the client device, smth like this:

```javascript
create(persist((() => ({
  id: guuid(),
  ...
}), { name: localStorageKey }));
```

Every time I refresh the page, the id changes. I'd love to have that state value saved to the localStorage at the time of a new store creation. I'm aware I can declare extra action that sets the id explicitly, but I think an extra property to PersistOptions, smth like ```{ storeInitialValue: boolean }``` (store only when localStorage.name does not exists)  would be a much cleaner solution.
And of course, maybe I'm missing something, but I'd love to hear your feedback on this.

Mangor1no:
Your case is, basically, calling the "guuid()" function per re-render time, which leads into the changing of your "id" state. I don't think that a problem with the library, it is more likely a way to approach the problem.

axelboc:
As @Mangor1no mentioned, you can approach the problem another way. For instance, you could set the ID in the store from a `useEffect` the first time the app renders:

```ts
const useStore = create(persist((() => ({
  id: undefined,
  setId: (id) => set({ id })
  ...
})));

function App() {
  const id = useStore(state => state.id);
  const setId = useStore(state => state.setId);

  useEffect(() => {
    setId(guuid());
  }, []); // run effect on mount

  if (id === undefined) {
    return null; // app not ready
  }

  return <>...</> // app ready
}
```

kolodziejczak-sz:
thank you for your replies, I'm aware I can get around this problem by using the 'set' fn. I just find it unconvenient that the store value is persisted only after explicit interaction with it.

TheEhsanSarshar:
@dai-shi we can have `{ presistedFields: [ "fieldname" ] }` inside persistoptions object.
if you will I can work on this.

dai-shi:
> @dai-shi we can have `{ presistedFields: [ "fieldname" ] }` inside persistoptions object.
> if you will I can work on this.

I don't think it's that easy. We also need to take care of AsyncStorage.

> (beware, however, that this currently happens asynchronously: #346).

https://github.com/pmndrs/zustand/issues/346#issuecomment-810342287
We should fix this, and then localStorage users can store initial values before calling `create()`.
You want to work on this?
[...]

PR #396, opened 2021-05-22
fix issue of persisting data on initial load.

TheEhsanSarshar:
# WIP:
fixing the issue of #366

- [x] issue
- [ ] test
- [ ] code review

`src/middleware`
`line 277:286`
```diff
         if (migratedState) {
           set(migratedState)
           await setItem()
        }    
+     } else {
+        await storage.setItem(name, await serialize({...get() || {}}) 
+     }
```

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -281,6 +281,8 @@ export const persist = <S extends State>(
         } else {
           set(deserializedStorageValue.state)
         }
+      } else {
+        await storage.setItem(name, await serialize({...get() || {}})
       }
     } catch (e) {
       postRehydrationCallback?.(undefined, e)


dai-shi:
Thanks for your work.
At this point, I'm not sure what issue you are trying to solve.
For #366, my suggestion is firstly to fix the async issue https://github.com/pmndrs/zustand/issues/366#issuecomment-845497855 which allows to fix the issue on lib user end, and if we need better developer experience we can follow up. (I prefer step by step.)

src/middleware.ts
```diff
@@ -281,6 +281,8 @@ export const persist = <S extends State>(
         } else {
           set(deserializedStorageValue.state)
         }
+      } else {
+        await storage.setItem(name, await serialize({...get() || {}})
```
dai-shi, line 285: This seems another issue from what I think we should fix firstly, but I am not sure if this is desired. At least I don't want to write to localStorage without change. I think this should be optional, and if async issues are solved, there's a workaround, and we will see what we want.

TheEhsanSarshar: the initial value of the store shouldn't be written in localStorage? Do you mean this?

for example 
```Javascript
     friends: 10,
     setFriends: () => set({friends: 20})
}, {name: "my-storage"}))
```

so the localStorage will not store the value `friends: 10`, until you update the value by `setFriends`

dai-shi: Yes, in my use case in mind, I wouldn't store the default value.

```js
const useStore = create(persist((set) => ({
     userName: "No Name",
     setUserName: (userName) => set({ userName })
}), { name: "my-user-setting" }))
```

In this case, I don't want to store "No Name" in localStorage.

dai-shi: Hm, and your use case of storing default value seems already covered by `postRehydrationCallback`. So, it's already supported, I'd say.

TheEhsanSarshar: storing is optional right? 
and if someone wants to store, it should use the `postRehydrationCallback` callback right ?

PR #397, opened 2021-05-23
(types): improve types for setState

### LINKED ISSUE #393: v3.3.3 broke some TS types
## Summary

v3.3.3 introduces a breaking change for Typescript users that use `set` with callback with multiple return statements. See repro.

Please advice how to type this properly after v3.3.3.

## Workaround

I don't know.

## Minimal Repro

```ts
import create from "zustand/vanilla";

const TestStore = create(combine({
    locked: false,
    amount: 0
  }, (set) => ({
    setAmount: (newAmount: number) => {
      // Works until 3.3.2, TS-errors on 3.3.3 and later (up to current 3.5.1). Note multiple returns.
      set(({ locked }) => {
        //^^^^^^^^^^^^^^^^^
        // Argument of type '({ locked }: { locked: boolean; amount: number; }) => { amount?: undefined; } | { amount: number; }' is not assignable to parameter of type 'PartialState<{ locked: boolean; amount: number; }, "amount">'.
        // Type '({ locked }: { locked: boolean; amount: number; }) => { amount?: undefined; } | { amount: number; }' is not assignable to type '(state: { locked: boolean; amount: number; }) => { locked: boolean; amount: number; } | Pick<{ locked: boolean; amount: number; }, "amount">'.
        //   Type '{ amount?: undefined; } | { amount: number; }' is not assignable to type '{ locked: boolean; amount: number; } | Pick<{ locked: boolean; amount: number; }, "amount">'.
        //     Type '{ amount?: undefined; }' is not assignable to type '{ locked: boolean; amount: number; } | Pick<{ locked: boolean; amount: number; }, "amount">'.
        //       Type '{ amount?: undefined; }' is not assignable to type 'Pick<{ locked: boolean; amount: number; }, "amount">'.
        //         Types of property 'amount' are incompatible.
        //           Type 'undefined' is not assignable to type 'number'.ts(2345)
        if (locked) return {};
        
        return {
          amount: newAmount
        };
      });
    },
    setAmountForced: (newAmount: number) => {
      // Works always. Note single return.
      set(({ locked }) => {
        return {
          amount: newAmount
        };
      });
    }
  })
));
```

TkDodo:
allow for passing multiple key generics to PartialState; this will make sure that we can have paths in the PartialState setter function that return different keys

closes #393


tests/basic.test.tsx
```diff
@@ -775,10 +775,48 @@ it('should have correct (partial) types for setState', () => {
...
+  })
+
+  // @ts-expect-error type  Type '{ name: boolean; count?: undefined; }' is not assignable to type State.
+  store.setState<'count', 'name'>((previous) => {
+    if (previous.count === 0) {
+      return { count: 1 }
+    }
+    return { name: true }
+  })
+})
```
TkDodo, line 822: i would expect an error here, because `name: true` certainly does not fit. However, there is no error, and I don't really know why. it errors fine the [ts playground](https://www.typescriptlang.org/play?ts=4.2.3#code/KYDwDg9gTgLgBDAnmYcDKMCGNUF44QBGAVsAMYwBQA9NQssAM5wCWAdo2C1MACZyFEcRsBgZsqAGZQIAWzgAlYJgoAaYcGAAuGnQAWMGGEZbaAcxYw9AV0IA6MnOoARYJPaXgAG0QAVBrwubh44Pv4ogYReRNQAbGQALACcwAkArAAcCQBMAMyZhJhpSUUADBnAAIy51aUA7EmVSRlJjZUJuaWBSCiM1DwqMNQAbpWx1Oy8oHa8djCMAMQAMgktALQrSWmUoJCw9ChwAAqYsCyYXuI4ADy+cKA4bLzMV8DqANKV9yCPz3AA1sBEBBJHA7vhAcDQb4Ptlvr9mJCQWC4PhPh9cvDgE9EUDkeC4O9sh8ElicQC8dDUYTcgA+VGUOBwAA+cAAFEcWGR-rcPpV6azOdzeYTsgLjlyeTCaeKhVKSeLfABKRks9lsxhYHBaMFK1H0uUiz6yyVGsVqw3S950i2mq0JRUq3bQeA9VCvNDecgwaC3Ml-V7qACq9PwGq12l1+rgQZ24BdB1QAFEAI7WC6WRAAYT05EBUFuofZmokOulbGAAHdXmW9bh6YQIBAvMo2HG9q6GOgI0sWJrscAC74i+HS2D1GAeMMWBBrIwa1H63BhhAWLx2wm3d2JGgvFzgL3+xWhyPGHuyJHpZPgNPZ-Pz5e6-SV2uN-t2DgoJIVO7bIwyFALCEMAfoPNiAYRvSADeqpsnuR6DjqryHo8g6FkqOpsk+y6ruuTLXB6D60myqpMvBqFQEhEa7vuKEDgWhH7rSqikRoLYUNAAD8VE7l6HFDuojEXsxrHAGmGZIAAYmw3FwKm6Z7kgOZ5mhQnALSqoYey2EvuuAC+lBvp2hyemIEagT84EvJB1Iwfhnz+riUIohClLjqKjkUs5BLojSnlIlSaLEoSpJgeSAUuTKbJgKcMDnF4OonGcFyvEalSwhiCrqDwYBeD+smNs2rZabplAGc6+xbgA4qIqV3GFEESCO2G+EZiZwK4moyEIYY6bhbUfoO34Xt20DAAAglwFkItuODQaqIhmWOpl1RpTJmLVEY6jVS03MOC1-gBQGRmgh2AcBhaqlMXUQIgOqdTA3VlW1W6vFmAw+kOnmBnAWZzj6sgrRG1JAxIhYjgtog6n9mpyKDOAsetUNwDtq2I3AmBcFRY2TSwl0tYZlCOBw8Cwzw1IAES8NYsiyIgFMY8w1hsP8bAQJWbCM6NPC49cUFwJITY6l17BmAA3AIpw6mwNPAVAcB6RpZPAHYi2pQA5ILEDq+o6uFFA6vEdFU4znO2F2awoLGzepuMHY+uorg+CVHqFtMjwMDWFAnP81rOoU1rDN6WLqoGe7ohez7kuUQgUDWKgwdlUqIeUEAA)

@dai-shi do you see why?

TkDodo: ahm okay, found it. seems to because of `"name"` ? If I rename the prop to something different, it errors correctly :/

cd1a1fa7

dai-shi: so, "name" is something special in ts??

dai-shi (approved):
The code looks good. Thanks for your work!

tests/basic.test.tsx
TkDodo: no, not that I know of. Have no idea tbh

TkDodo:
thx. noticed another tiny thing in middleware.ts: was using `State` instead of `S`, which extends State: 82b26ea3

tests/basic.test.tsx
bozdoz: FYI
"name" is a global: https://developer.mozilla.org/en-US/docs/Web/API/Window/name
makes for many annoying false-positives when using TypeScript 😄

dai-shi:
@Akuukis can try this?

Akuukis:
@dai-shi thanks :+1: I will try this on my real usecase in upcoming days.

dai-shi:
https://ci.codesandbox.io/status/pmndrs/zustand/pr/397/builds/132694
image
Note that you can just install it with yarn/npm.

Akuukis:
Thanks @dai-shi , the command didn't work but replacing the version number in the `package.json` did.

In my real usecase:

1. `set<"myProp">((state) => { ... if(bail) return state; ... })` doesn't show TS errors (:+1:)
2. `set((state) => { ... if(bail) return {} as Record<never, never>; ... })` doesn't show TS errors (:+1:)

I'm fine either way, but would like to read in the readme whenever `state` or `{}` is preferable and use that one.

Sadly, both allows extra arbitrary props or existing props with wrong type in the real returned object (:-1:). Ironically, one of my stores is using a `persist` where I'm not using an old prop anymore, and thus can pass `myObseleteProp: undefined` to let zustand clear it from storage whenever it will update.

dai-shi:
Thanks for confirming.

`return {}` doesn't bail early, so `return state` is recommended in this case.
It's been an undocumented behavior. Yeah, it would probably be good to document it.

> extra arbitrary props or existing props with wrong type in the real returned object

Not sure if I understand this type issue.

TkDodo:
> extra arbitrary props or existing props with wrong type in the real returned object
Not sure if I understand this type issue.

me neither, as we now have a dedicated test that expects a ts error if a non existing prop is returned

TkDodo:
cool, thanks for merging @dai-shi 🙏

dai-shi:
will release it with other changes. Thanks for your work.

drcmda:
@TkDodo could you add a short description as a comment (jsdoc? so perhaps intellisense can pick it up if that's even possible). we updated from 2 to 3 at work and it's not so easy to understand what k1 k2 k3 k4 does and why there are four keys.

TkDodo:
@dai-shi came up with the initial proposal here: https://github.com/pmndrs/zustand/issues/393#issuecomment-844546143

it basically lets TypeScript infer the types of the keys we want to set, which is important if we have different branches in the same setter that set different keys. The limit to 4 is arbitrary I think 😅

Issue #346, opened 2021-03-30
Is it possible to immediately load persisted data?


aidan-keay:
I'm using the persist middleware and loading state from local storage. As far as I am aware, there is a brief period where the store state is the default state and then the store is rehydrated with the persisted state. Is there any way to load this immediately if the storage access method is synchronous? It works if I load from localStorage manually at the top of the file and set this as the store state although I would rather use the middleware.

AnatoleLucet:
I thought about conditionally making the middleware's hydration sync or async. But I don't know if people would find this confusing or not.
Any thoughts?

sandren:
> I thought about conditionally making the middleware's hydration sync or async. But I don't know if people would find this confusing or not.
> Any thoughts?

@AnatoleLucet We're addressing the same concern [here](https://github.com/pmndrs/jotai/pull/394#issue-603874729) with Jotai as well. Right now it's set up to update the state when it is first mounted to accommodate SSR in Next.js and Gatsby, but perhaps the persistence implementation in both libraries could be made to support both sync and async with an extra parameter similar to how both will have a `getStorage` parameter with a default function of `() => window.localStorage`. Maybe we could both add a fourth parameter `ssrMode`, which defaults to `true` or `false`? I'm curious what @dai-shi thinks as well.

dai-shi:
> I thought about conditionally making the middleware's hydration sync or async.

I think it's possible if we check the return value of `getItem` is `instanceof Promise` and otherwise process synchronously. That's how jotai works internally.

aidan-keay:
> I think it's possible if we check the return value of `getItem` is `instanceof Promise` and otherwise process synchronously. That's how jotai works internally.

Will this catch async deserialization and migration functions?

AnatoleLucet:
@aidan-keay yes, we would just need to check each of these three.

ChrisKG32:
I have noticed this problem too, as the SSR will render a page with the default persisted state, and not trigger a re-render with the actual persisted state.

axelboc:
I bumped into this after I started injecting a zustand store through context as shown in https://github.com/pmndrs/zustand/issues/182#issuecomment-803172558.

Somewhere in the children of the store's provider, I have a `useEffect` that sets something in the store on mount. Turns out this `useEffect` is called before the store is rehydrated from `localStorage`.

tkvw:
Thanks everyone!
I created a small utility function `createPersistentContext`: 

Usage: 
```
[...]

PR #403, opened 2021-05-26
Migrating Async/Await to Promise

TheEhsanSarshar:
# WIP
`middleware.ts` is completely migrated from Async/Await to Promise.
and also sync apis will be called in a sync manner.
- [x] a general API created to handle both sync/async peacefully
- [x] async tests passed
- [x] sync tests added
- [x] sync test passed
- [x] types added

fixing #346


src/middleware.ts
```diff
@@ -248,7 +248,10 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    return storage?.setItem(name, await serialize({ state, version }))
+    let serializeResult = serialize({ state, version })
+    return serializeResult instanceof Promise
+      ? serializeResult.then((value) => storage?.setItem(name, value))
+      : storage?.setItem(name, serializeResult)
```
dai-shi, line 254: To keep the behavior, we don't invoke serialize, if storage is undefined.
```suggestion
    if (storage) {
      const serializeResult = serialize({ state, version })
      return serializeResult instanceof Promise
        ? serializeResult.then((value) => storage.setItem(name, value))
        : storage.setItem(name, serializeResult)
    }
```

src/middleware.ts
```diff
@@ -259,35 +262,55 @@ export const persist = <S extends State>(
...
-
-        // if versions mismatch, run migration
-        if (deserializedStorageValue.version !== version) {
-          const migratedState = await migrate?.(
-            deserializedStorageValue.state,
-            deserializedStorageValue.version
-          )
+    if (storageValueUnknown instanceof Promise) {
+      storageValueUnknown
+        .then((storageValue) => storageValue && deserialize(storageValue))
```
dai-shi, line 272: The original code doesn't seem to check `storageValue`.
```suggestion
        .then((storageValue) => deserialize(storageValue))
```

src/middleware.ts
```diff
@@ -259,35 +262,55 @@ export const persist = <S extends State>(
...
+        .catch((e) => {
+          postRehydrationCallback?.(undefined, e)
+        })
+    } else {
+      try {
+        let deserializedValueUnknown = deserialize(
+          storageValueUnknown as string
+        )
+        if (deserializedValueUnknown instanceof Promise) {
+          deserializedValueUnknown.then((result) => set(result.state))
```
dai-shi, line 305: We need to run postRehydrationCallback after this is resolved or rejected.

dai-shi: Actually, we lose `migrate` in L300 condition branch.

Hmm, this looks hard. You might want to create some functions and compose them.

TheEhsanSarshar: Ok, thank you I will fix that.

src/middleware.ts
TheEhsanSarshar: dear @dai-shi, the original source code also checks the storage value before serializing.
```Javascript
const storageValue = await storage.getItem(name)
if (storageValue) {        
   const deserializedStorageValue = await deserialize(storageValue)
```

src/middleware.ts
TheEhsanSarshar: pretty neat.

src/middleware.ts
dai-shi: oops, sorry. looks like i missed it.

src/middleware.ts
```diff
@@ -248,7 +248,12 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    return storage?.setItem(name, await serialize({ state, version }))
+    if (storage) {
+      const serializeResult = serialize({ state, version })
+      return serializeResult instanceof Promise
+        ? serializeResult.then((value) => storage.setItem(name, value))
+        : storage.setItem(name, serializeResult)
```
dai-shi, line 255: I think your idea of `toThenable()` is super neat.
```js
const toThenable = (fn, input) => {
  try {
    const result = fn(input)
    if (result instanceof Promise) {
      return result
    }
    return { then: (onFulfilled) => onFulfilled(result) }
  } catch (e) {
    return { then: (onFulfilled, onRejected) => onRejected(e) }
  }
}
```

```suggestion
      toThenable(serialize, { state, version }).then(
        (value) => storage.setItem(name, value),
        (err) => { throw e }
      )
```

Something like this? @TheEhsanSarshar

TheEhsanSarshar: Yeah, you did it. I was thinking on how to implement it. Thank you @dai-shi

dai-shi: Good. Still more to do, including to make it TypeScript.

src/middleware.ts
```diff
@@ -258,36 +263,63 @@ export const persist = <S extends State>(
...
+        }
+      })
+      .then((deserializedStorageValue) => {
+        if (deserializedStorageValue) {
+          if (deserializedStorageValue.version !== version) {
+            return makeThenable(
+              migrate?.(
+                deserializedStorageValue.state,
+                deserializedStorageValue.version
+              )
```
barbogast, line 297: I think this needs to be
```suggestion
            
            return makeThenable(
              migrate?.(
                deserializedStorageValue.state,
                deserializedStorageValue.version
              )
              .then((migratedState) => set(migratedState))
              .then(setItem)
```

barbogast:
@dai-shi https://gist.github.com/barbogast/5679eabc25b761da0f500f53ca3ea0c2 contains an idea for a fix of the remaining issue with the rehydration. Is this approach correct, or do you think it should be fixed in an entirely different way?

src/middleware.ts
```diff
@@ -258,36 +263,63 @@ export const persist = <S extends State>(
...
+        } else {
+          return makeThenable(null)
+        }
+      })
+      .then(() => {
+        postRehydrationCallback?.(get(), undefined)
+        return makeThenable(null)
+      })
+      // this handler will only catch promise related errors
+      .catch((e) => {
```
barbogast, line 320: This `catch` handles only errors thrown in async mode. For sync mode the old `try...catch` needs to be reintroduced, meaning we need both `catch()` and `try...catch`.

src/middleware.ts
```diff
@@ -258,36 +263,63 @@ export const persist = <S extends State>(
     void setItem()
   }
 
+  const makeThenable = (value: any) => {
+    if (value instanceof Promise) {
+      return value
+    } else {
+      return {
+        then: (callback) => callback(value),
```
barbogast, line 271:
```suggestion
        then: (callback) => makeThenable(callback(value)),
```
Not strictly necessary, but maybe a good idea. I had a bug where I forgot to manually wrap the result of `then()` with `makeThenable()`. Also it would resemble real promises a more.

barbogast:
Just updated both gists. The tests should be finished, and are passing against the updated `middleware.ts` gist. How do you want me to submit the code? A PR against https://github.com/TheEhsanSarshar/zustand/tree/async-await-fix? Or do you just want to grab the gists and add them to PR directly, @TheEhsanSarshar?

barbogast:
I'm wondering... is it possible to correctly type `makeThenable()`? I had a quick try and didn't manage, especially since the returned object needs to be compatible with real Promises for TS (I guess?). 

That together with the subtle but important differences between sync mode and async mode within the same code block make me wonder if it's worth it to handle both cases with the same code. Maybe it would be better to duplicate the code and handle both cases separately. Not sure, though, just sharing the thought.

src/middleware.ts
```diff
@@ -258,36 +263,63 @@ export const persist = <S extends State>(
     void setItem()
   }
 
+  const makeThenable = (value: any) => {
```
dai-shi, line 266: Does it work for error handling.
It has to be something like this? not 100% sure.
```ts
type Thenable<Result> = {
  then: (onFulfilled: ..., onRejected: ...) => ...
}
const makeFunctionThatReturnsThenable = <Input, Result>(
  fn: ((input: Input) => Result | Promise<Result>)
): (input: Input) => Thenable<Result> => {
 ...
}
```
Also, we might want to make types correctly to avoid misunderstanding.

barbogast: 👍 to proper typing

I guess this would be the optimal solution if it weren't for the bundle size: https://github.com/fluffynuts/synchronous-promise. Might still serve as reference point, though.

dai-shi: I'd like to avoid dependencies not only because of the size, but also other various bundling issues.

TheEhsanSarshar: it works for async error handling.
but for non async it won't.

TheEhsanSarshar: I will work on types after these changes become final.

TheEhsanSarshar: Ok, I am going to check how promise internally work. maybe I can improve `toThenable()` to exactly handle `then` and `catch` the same way as promise do.

TheEhsanSarshar: maybe `mongoose` source code has something to help.

TheEhsanSarshar: `then: (onResolve, onReject)`
will not work on a custom object

TheEhsanSarshar: I am working on a new solution that will handle catch on both async and sync. just a little debugging remained

TheEhsanSarshar:
> Just updated both gists. The tests should be finished, and are passing against the updated `middleware.ts` gist. How do you want me to submit the code? A PR against https://github.com/TheEhsanSarshar/zustand/tree/async-await-fix? Or do you just want to grab the gists and add them to PR directly, @TheEhsanSarshar?

I am going to make some changes to add support for sync error handling. then you can make a PR on the repo

src/middleware.ts
TheEhsanSarshar: good idea.

TheEhsanSarshar:
Hi dear @dai-shi.
in this gist, I just implemented a new `toThroable` that is able to handle errors in both syn/async. it also handles thrown exception inside `then` and `catch`
can you please have a look at that?
https://gist.github.com/TheEhsanSarshar/d9852a7cc102dae0f07186e1dc4fe700

TheEhsanSarshar:
the async tests are failing. dear @barbogast would you mind please check that.

dai-shi:
https://gist.github.com/TheEhsanSarshar/d9852a7cc102dae0f07186e1dc4fe700#gistcomment-3761417
The conversations are scattered.
We could just paste the code here.

barbogast:
FWIW, I created a PR with my results agains @TheEhsanSarshar 's branch: https://github.com/TheEhsanSarshar/zustand/pull/1. The changes are separated by commit, so you can cherry-pick. It contains the tests for the sync storage and code adjustments to `middleware.ts` so that all the tests are passing.

barbogast:
>in this gist, I just implemented a new toThenable that is able to handle errors in both syn/async. it also handles thrown exception inside then and catch
can you please have a look at that?
https://gist.github.com/TheEhsanSarshar/d9852a7cc102dae0f07186e1dc4fe700

Hmm... just realised, that it might be worth it to have separate tests `toThenable`, then. There are a lot of ways in which this function can be used, and without tests it might be tricky to guarantee that all of them are handled correctly. Especially thinking about all the different ways in which errors could occur.

dai-shi:
Thanks guys for working on this.
I thought too that tests for `toThenable` would be nice.
But, we haven't agreed on the api yet.
Let me put my proposal here:
```js
const toThenable = (fn) => (input) => {
  try {
    const result = fn(input)
    if (result instanceof Promise) {
      return result
    }
    return {
      then(onFulfilled) { return toThenable(onFulfilled)(result) },
      catch(onRejected) { return this },
    }
  } catch (e) {
    return {
      then(onFulfilled) { return this },
      catch(onRejected) { return toThenable(onRejected)(e) },
    }
  }
}
```

TheEhsanSarshar:
thanks @dai-shi I am going to playaround with your proposal

TheEhsanSarshar:
> > in this gist, I just implemented a new toThenable that is able to handle errors in both syn/async. it also handles thrown exception inside then and catch
> > can you please have a look at that?
> > https://gist.github.com/TheEhsanSarshar/d9852a7cc102dae0f07186e1dc4fe700
> 
> Hmm... just realised, that it might be worth it to have separate tests `toThenable`, then. There are a lot of ways in which this function can be used, and without tests it might be tricky to guarantee that all of them are handled correctly. Especially thinking about all the different ways in which errors could occur.

@barbogast are agree to finalize the api first and then write tests ?

barbogast:
Sure, no worries. The only "hard" rule I suggest is having tests before merging the code. This might help during the implementation, so it could make sense to write tests next.

barbogast:
Although... I've still got the nagging feeling that this approach might not be worth the effort and the complexity. This feels like a partial re-implementation of Promises. And yeah, if we wanna use `catch()` that's necessary for the code within `then()` to be executed synchronously. But to me it's not trivial to implement that correctly, get the edge cases right, writing tests, and later, when using it, being aware of the (probably subtle) differences to real promises.

Compared to all that the `try... catch` solution that works with the existing `makeThenable()` seems quite appealing, even if it looks awkward and requires one line of code duplication.

The reason that might make it necessary is if `makeThenable` cannot be typed correctly, though.

And, don't get me wrong, I'm totally on-board with the `toThenable()` approach from the gist. Just wanted to share my thoughts.

dai-shi:
I'm kind of enjoying this effort. If it works as expected, it might be able to apply to other projects, namely jotai.

TheEhsanSarshar:
> Although... I've still got the nagging feeling that this approach might not be worth the effort and the complexity. This feels like a partial re-implementation of Promises. And yeah, if we wanna use `catch()` that's necessary for the code within `then()` to be executed synchronously. But to me it's not trivial to implement that correctly, get the edge cases right, writing tests, and later, when using it, being aware of the (probably subtle) differences to real promises.
> 
> Compared to all that the `try... catch` solution that works with the existing `makeThenable()` seems quite appealing, even if it looks awkward and requires one line of code duplication.
> 
> The reason that might make it necessary is if `makeThenable` cannot be typed correctly, though.
> 
> And, don't get me wrong, I'm totally on-board with the `toThenable()` approach from the gist. Just wanted to share my thoughts.

just to make it clear dear @barbogast. the `try catch` solution will only work if all three values (storage, deserialize, migrate) are `sync`.
if one of them becomes async then we will lose the `try catch` context.

TheEhsanSarshar:
> Thanks guys for working on this.
> I thought too that tests for `toThenable` would be nice.
> But, we haven't agreed on the api yet.
> Let me put my proposal here:
> 
> ```js
> const toThenable = (fn) => (input) => {
>   try {
>     const result = fn(input)
>     if (result instanceof Promise) {
>       return result
>     }
>     return {
>       then(onFulfilled) { return toThenable(onFulfilled)(result) },
>       catch(onRejected) { return this },
>     }
>   } catch (e) {
>     return {
>       then(onFulfilled) { return this },
>       catch(onRejected) { return toThenable(onRejected)(e) },
>     }
>   }
> }
> ```

thank you @dai-shi. I agree with this API. it's working like a charm 
todo list:
- [x] finalize API
- [ ] cherry-pick test provided by @barbogast 
- [ ] add types
- [ ] merge

TheEhsanSarshar:
Good job. everybody.
now we need to make the sync tests pass.

barbogast:
Will have a look later or tomorrow.

barbogast:
With https://github.com/TheEhsanSarshar/zustand/pull/2 the tests are all green 🎉.

The solution feels a bit like a hack though, maybe there is a better way. One way to also fix it could be to later only override the state that was set by middlewares if the state is still empty. Might be a breaking change, though.

dai-shi:
@TheEhsanSarshar wrt. https://gist.github.com/TheEhsanSarshar/d9852a7cc102dae0f07186e1dc4fe700#gistcomment-3762845

How about this? [playground](https://www.typescriptlang.org/play?#code/ATCWDsBcFMCcDMCGBjawAqALa5ECMAbaAHgDVECBXaAPmAG8AoEESbcMmgCgHtwAxSgXigCRACYAuYFwBuFatPJVoASmABeOqWAAfDO3xFOq6VhxGSpGsxbJEkZJk69wAJWgAraMhhSZsNCIAM580pTgANbgPADu4OpawDr65riEVjSmBhYZnLYAvoy2yHzBkMAAHprA4NCxwAAKsDwAtqDBJOCUrXhw3Fywqlr0BarFpeDlwJA8aZY1xB7BQpAANMAAkuAADpSQ3PDg0lwQe5DS2+eJdMurek0t7Z1L0CsEBw-zeXcfWZp0U67faXYGQbLfYy-T5JJgsGawACeDFs8Mm00C7wqGmARyB11RLFA8ACb3uEHKiHAqB4JOabQ6ahR8PhgUglFg4GAmPuIRy6ShZL+hJARRZ3Og7M5zPFM3YrkEwlEEnUcNlIDZHK5s0h0AVQhEYmg4lUgyFFT50PG6oKaxFdgcTlcHm8vmNqvtrMlWrlHWAfKpyIA9EHgPxNgANACyAFFgPYubE0IhZDxQOJ-eBEQB+T3AW32sUsArxx2YGRMtVeqVcqsstg4fVKo0mmXqzXSth+gNZ4AhsOR2OlxPJ1PpzM5vMF9X2RyYZ1eHx+D3qjXeztzQwZBeu5dcNRTu3iouiwpAA)

Also, please define `toThenable` outside `persist` and we can define some of functions in advance.

TheEhsanSarshar:
@dai-shi please have a final review on this.

src/middleware.ts
```diff
@@ -259,43 +296,60 @@ export const persist = <S extends State>(
...
-            deserializedStorageValue.version
-          )
-          if (migratedState) {
-            set(migratedState)
-            await setItem()
+  // The state that was loaded from the storage and is set via set() within this
+  // middleware would later be overridden with the initial state by create(). To
+  // avoid this we merge the state from localStorage into the initial state. This
+  // way the state gets overridden with the state loaded from storage.
+  // TODO: find a better solution for this
```
dai-shi, line 304: Hmm... this is only for sync. Not typed. It seems really TODO.

src/middleware.ts
```diff
@@ -259,43 +296,60 @@ export const persist = <S extends State>(
...
+      })
+      .then((deserializedStorageValue) => {
+        if (deserializedStorageValue) {
+          if (deserializedStorageValue.version !== version) {
+            if (!migrate) return
+            stateFromStorage = migrate(
+              deserializedStorageValue.state,
+              deserializedStorageValue.version
+            )
+            return stateFromStorage
```
dai-shi, line 324:
```suggestion
            if (migrate) {
              stateFromStorage = migrate(
                deserializedStorageValue.state,
                deserializedStorageValue.version
              )
              return stateFromStorage
            }
```
Seems more consistent with other code.

src/middleware.ts
```diff
@@ -259,43 +296,60 @@ export const persist = <S extends State>(
...
+  const configResult = config(
     (...args) => {
       set(...args)
       void setItem()
     },
     get,
     api
   )
+
+  return { ...configResult, ...(stateFromStorage || {}) }
```
dai-shi, line 354: This is the workaround, right? I'm not sure if I fully understand it.

```suggestion
  return stateFromStorage ? configResult : { ...configResult, ...stateFromStorage }
```
should be better.

dai-shi (changes requested):
Commented. I might have some other feedback in the second iteration.

dai-shi:
@AnatoleLucet Can you carefully review this?

src/middleware.ts
```diff
@@ -248,7 +280,12 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    return storage?.setItem(name, await serialize({ state, version }))
+    if (storage) {
+      const serializeResult = serialize({ state, version })
+      return serializeResult instanceof Promise
+        ? serializeResult.then((value) => storage?.setItem(name, value))
+        : storage.setItem(name, serializeResult)
```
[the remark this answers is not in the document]
TheEhsanSarshar: yes, we can.

src/middleware.ts
barbogast: Would need to be the other way around:
```suggestion
  return stateFromStorage ? { ...configResult, ...stateFromStorage } : configResult
```

It's the continuation of what's described in the commend above:
```js
  // The state that was loaded from the storage and is set via set() within this
  // middleware would later be overridden with the initial state by create(). To
  // avoid this we merge the state from localStorage into the initial state. This
  // way the state gets overridden with the state loaded from storage.
```

src/middleware.ts
barbogast: True, the type is missing. I guess it would be
```js
  let stateFromStorage: S | void
```

src/middleware.ts
```diff
@@ -259,43 +296,60 @@ export const persist = <S extends State>(
...
-          if (migratedState) {
-            set(migratedState)
-            await setItem()
+  // The state that was loaded from the storage and is set via set() within this
+  // middleware would later be overridden with the initial state by create(). To
+  // avoid this we merge the state from localStorage into the initial state. This
+  // way the state gets overridden with the state loaded from storage.
+  // TODO: find a better solution for this
+  let stateFromStorage
+  ;(() => {
```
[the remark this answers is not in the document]
barbogast: Good catch, looks like it's not needed, the tests still pass when the code is directly in `persist()`.

src/middleware.ts
TheEhsanSarshar: yeah, it's a workaround.
I think there is something wrong with provided snipped.

src/middleware.ts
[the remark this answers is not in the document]
TheEhsanSarshar: I think we need another PR to fix this issue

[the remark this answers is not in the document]
TheEhsanSarshar: oh 😆. 
I misunderstood it.

src/middleware.ts
```diff
@@ -248,7 +281,11 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    return storage?.setItem(name, await serialize({ state, version }))
+    if (storage) {
+      toThenable(serialize)({ state, version }).then((serializedValue) => {
```
AnatoleLucet, line 285:
```suggestion
      return toThenable(serialize)({ state, version }).then((serializedValue) => {
```
Not sure if the `storage?.setItem` should also be returned since it's possibly async? 🤔

src/middleware.ts
```diff
@@ -259,43 +296,59 @@ export const persist = <S extends State>(
...
+            )
+            return stateFromStorage
           }
         } else {
+          stateFromStorage = deserializedStorageValue.state
           set(deserializedStorageValue.state)
         }
       }
-    } catch (e) {
+    })
```
AnatoleLucet, line 328: Not sure if it'd work, but to solve the above `todo`: maybe we could return instead of assigning to `stateFromStorage`, do a little condition to return the final state, or else, continue the `.then` chain after that?

tests/persistSync.test.tsx
```diff
@@ -0,0 +1,173 @@
+import create from '../src/index'
+import { persist } from '../src/middleware'
+
+const consoleError = console.error
+afterEach(() => {
+  console.error = consoleError
```
AnatoleLucet, line 6: What is this for? I'm not sure to understand since we're not resigning console.error, and you're neither mocking it.

tests/persistSync.test.tsx
```diff
@@ -0,0 +1,173 @@
+import create from '../src/index'
+import { persist } from '../src/middleware'
+
```
AnatoleLucet, line 3: maybe a little `describe` over this whole file could structure this a bit more?

tests/persistSync.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+      expect(value).toBe(
+        JSON.stringify({
+          state: { count: 42 },
+          version: 0,
+        })
+      )
+    },
+  }
+
+  const useStore = create<any>(
```
AnatoleLucet, line 85: Maybe using real interfaces instead of `any` could help to test the typing integration? (I'm speaking of every test in this file, not this `create` specifically)

AnatoleLucet (changes requested):
I'm a bit late for the party, but thanks a lot to you all for your work 🙂 
There's some small things here and there, but the overall seems great!

src/middleware.ts
dai-shi: In addition to it, my suggestion is to define `const thenabledSerialize = toThenable(serialize)` outside `setItem`, and

```js
if (storage) {
  return thenabledSerialize({ state, version })
    .then((serializedValue) => storage.setItem(name, serializedValue))
    .catch((e) => { throw e })
}
```

src/middleware.ts
```diff
@@ -259,43 +296,59 @@ export const persist = <S extends State>(
...
-      const storageValue = await storage.getItem(name)
 
+  // a workaround to solve the issue of not storing rehydrated state in sync storage
+  // the set(state) value would be later overridden with initial state by create()
+  // to avoid this, we merge the state from localStorage into the initial state.
+  // TODO: find a better solution for this
+  let stateFromStorage: any
+  const postRehydrationCallback = onRehydrateStorage?.(get()) || undefined
+  // bind is used to avoid `TypeError: Illegal invocation` error
+  toThenable(storage.getItem.bind(storage))(name)
```
dai-shi, line 307: Is `storage` always defined? Oh, L259 checks it.

src/middleware.ts
dai-shi: As L259 checks `storage`, we don't need `if`.

src/middleware.ts
```diff
@@ -259,43 +296,59 @@ export const persist = <S extends State>(
...
         if (deserializedStorageValue.version !== version) {
-          const migratedState = await migrate?.(
-            deserializedStorageValue.state,
-            deserializedStorageValue.version
-          )
-          if (migratedState) {
-            set(migratedState)
-            await setItem()
+          if (migrate) {
+            stateFromStorage = migrate(
```
dai-shi, line 317: This is wrong. `migrate` can return a promise. `toThenable(migrate)`

src/middleware.ts
```diff
@@ -259,43 +296,59 @@ export const persist = <S extends State>(
...
-    const postRehydrationCallback = onRehydrateStorage?.(get()) || undefined
-
-    try {
-      const storageValue = await storage.getItem(name)
 
+  // a workaround to solve the issue of not storing rehydrated state in sync storage
+  // the set(state) value would be later overridden with initial state by create()
+  // to avoid this, we merge the state from localStorage into the initial state.
+  // TODO: find a better solution for this
+  let stateFromStorage: any
```
dai-shi, line 304: Okay, I understand why this is required. I think it's the right solution.
Let's rename it like:
```suggestion
  let stateFromStorageInSync: S | undefined
```

The type `S` should probably be `Partial<S>` not only here, but all over the place. Maybe we should tackle it in a separate PR.

src/middleware.ts
TheEhsanSarshar: returning some value from inside the then is a little challenging. specially it's a combination of sync/async `then` chain.

tests/persistSync.test.tsx
barbogast: I did copy that from `persistent.test.tsx` and not verify that it's needed, good catch. I can be removed, right? Might make sense to remove it from `persistent.test.tsx` as well, I guess.

tests/persistSync.test.tsx
TheEhsanSarshar: would you mind please check that @barbogast

barbogast: 👍 will add

tests/persistSync.test.tsx
barbogast: Just removing the `<any>` will make TS automatically figure out the type correctly. Would that be sufficient?

src/middleware.ts
barbogast: Ou, right, I wasn't aware. Then we need to also adapt [this](https://github.com/pmndrs/zustand/blob/master/tests/persist.test.tsx#L154) to use a `async` migrate so that both versions are tested properly. Will do.

TheEhsanSarshar: so stateFromStorage should be inside `L239` `then` handler. right?
I think there is something wrong with `types` the 'handlers' input values are never `promise`. but it detects it as a promise.

src/middleware.ts
barbogast: The goal of assigning the state to `stateFromStorage` is to have the value available outside the closures of promise callbacks. This way it can be returned as initial state: https://github.com/pmndrs/zustand/pull/403/files#diff-0626369278c89781505d2c04865ecce1302e979271bcfb471e029c19addc13a0R352.

Returning it from within a `then` callback would just get the value wrapped into another `thenable`, it wouldn't be returned to the `persist` function. Right?

src/middleware.ts
TheEhsanSarshar: Ok, I did it. is there any reason to define it outside setItem?

src/middleware.ts
TheEhsanSarshar: yeah, it will also make the `toThenable()` function non general

src/middleware.ts
TheEhsanSarshar: @barbogast I just move it to the next `then()` I think that would solve the issue.

src/middleware.ts
TheEhsanSarshar: I think we still need `if (storage) ` because the function not immediately run after `L259`.
and there may be someone make it `undefined`. before calling setItem(). 
at least typescript will not complain about it.

tests/persistSync.test.tsx
barbogast: Actually this concerns all test files. It's not apparent to me why `console.error` gets mocked. Does zustand write to `console.error`, and the mocks where once verifying that? Address in a separate PR?

src/middleware.ts
TheEhsanSarshar: and yeah, we should move `if (storage)` inside `then` handler

src/middleware.ts
AnatoleLucet: Ok I see the issue now. Indeed, it wouldn't solve it :confused:

tests/persistSync.test.tsx
AnatoleLucet: cc @dai-shi

tests/persistSync.test.tsx
AnatoleLucet: sure :+1:

tests/persistSync.test.tsx
dai-shi: @barbogast

> Actually this concerns all test files.

Totally. They are not well maintained, I'd say... 😅  Happy to take PRs for improvements. Please address in a separate PR.

barbogast:
Right now the code doesn't behave correctly if the `version` in localStorage doesn't match the `version` in the code _and_ there is no `migrate` function. There is also no test covering this. 

Is this case supposed to be handled correctly? Thinking about it, a `console.error` warning might also be helpful in this case.

TheEhsanSarshar:
> Right now the code doesn't behave correctly if the `version` in localStorage doesn't match the `version` in the code _and_ there is no `migrate` function. There is also no test covering this.
> 
> Is this case supposed to be handled correctly? Thinking about it, a `console.error` warning might also be helpful in this case.
> 
> --- edit
> PR: [TheEhsanSarshar#4](https://github.com/TheEhsanSarshar/zustand/pull/4). Also, the code handling the migration can be simplified. Would do that later.

Yeah, as you said a single console.warn would be enough.

src/middleware.ts
```diff
@@ -236,7 +269,9 @@ export const persist = <S extends State>(
     )
   }
 
-  const setItem = async () => {
+  const thenableSerialize = toThenable(serialize)
+
+  const setItem = (): Thenable<void> | void => {
```
dai-shi, line 274: Does it return`void`? ~I'm fine with inferring return type (= no explicit return type), because we don't use it anywhere.~

src/middleware.ts
```diff
@@ -259,43 +302,59 @@ export const persist = <S extends State>(
...
-  ;(async () => {
-    const postRehydrationCallback = onRehydrateStorage?.(get()) || undefined
-
-    try {
-      const storageValue = await storage.getItem(name)
 
+  // a workaround to solve the issue of not storing rehydrated state in sync storage
+  // the set(state) value would be later overridden with initial state by create()
+  // to avoid this, we merge the state from localStorage into the initial state.
+  // TODO: find a better solution for this
```
dai-shi, line 309: Do we need this TODO line? It looks to me that this is a right solution.
What do you guys think?

src/middleware.ts
```diff
@@ -248,7 +283,15 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    return storage?.setItem(name, await serialize({ state, version }))
+    return thenableSerialize({ state, version }).then((serializedValue) => {
+      if (storage) {
+        storage.setItem(name, serializedValue)
```
dai-shi, line 288: As we don't return the `storage.setItem` result, `return` in L286 is useless.
We don't use the return value of setItem anywhere (or did I miss it?), could we remove `return` in L286? (and that will change other code, if it works.)

src/middleware.ts
dai-shi: ~Per my another comment, this may become just `void` (or no explicit typing).~

src/middleware.ts
```diff
@@ -259,43 +302,59 @@ export const persist = <S extends State>(
...
           set(deserializedStorageValue.state)
         }
       }
-    } catch (e) {
+    })
+    .then((migratedState) => {
+      if (migratedState) {
+        stateFromStorageInSync = migratedState as S
+        set(migratedState as PartialState<S, keyof S>)
+        return setItem()
```
dai-shi, line 338: Oh, here I missed it, it returns thenable.

src/middleware.ts
dai-shi: L338 uses it.
So, we should return thenable.
Please define `thenableStorageSetItem` in advance, and return here. keep `return` in L286.

src/middleware.ts
dai-shi: Then this should simply be:
```suggestion
  const setItem = (): Thenable<void> => {
```

src/middleware.ts
```diff
@@ -248,7 +283,15 @@ export const persist = <S extends State>(
...
-    return storage?.setItem(name, await serialize({ state, version }))
+    return thenableSerialize({ state, version }).then((serializedValue) => {
+      if (storage) {
+        storage.setItem(name, serializedValue)
+      } else {
+        console.warn(
+          `Persist middleware: unable to update ${name}, the given storage is currently unavailable.`
+        )
+      }
+    })
```
dai-shi, line 294:
```suggestion
    }).catch((e) => { throw e })
```
I'm pretty sure we want this, no?

src/middleware.ts
```diff
@@ -248,7 +283,15 @@ export const persist = <S extends State>(
       blacklist.forEach((key) => delete state[key])
     }
 
-    return storage?.setItem(name, await serialize({ state, version }))
+    return thenableSerialize({ state, version }).then((serializedValue) => {
+      if (storage) {
```
dai-shi, line 287: We are sure `storage` is defined here. Can we remove the `if`? Does TS complain?

dai-shi (changes requested):
Looks great, and getting close. There're a few things to fix for edge cases.
So sorry about the messy comments. If ambiguous, feel free to ask.

src/middleware.ts
TheEhsanSarshar: I think we should remove `Thenable<void>` and `return` in `line:286` as well. what do you think?

src/middleware.ts
TheEhsanSarshar: it's a workaround. 
Yeah, it's working. maybe we can remove `TODO`

src/middleware.ts
TheEhsanSarshar: oh, `line:338` use it.

src/middleware.ts
TheEhsanSarshar: should I define `thenableStorageSetItem` as a function ?

src/middleware.ts
TheEhsanSarshar: I am not sure if it's working in a `sync` manner. let me check that

dai-shi: I think this magically works.

src/middleware.ts
dai-shi: Yeah, like you did with getItem.
```js
const thenableStorageSetItem = toThenable(storage.setItem.bind(storage))
```
Defining it in advance is better because it will be invoked multiple times.

src/middleware.ts
TheEhsanSarshar: the `ts` complain about it. and I agree with `ts`😆 .
because it may be an async call.

dai-shi: Let's either a) assert type or b) define `const` after early return.

TheEhsanSarshar: if we define it as `const` in `line:254` then we lose the variable scope.
type assertion seems like a good option.
```Typescript
 storage!.setItem(name, serializedValue)
```

dai-shi: I meant after L270. But, it increases JS code. Let's do type assertion.
```ts
(storage as StateStorage).setItem(name, serializedValue)
```
I prefer this with no reason. Maybe some eslint rules disliked Non-null Assertion (!)

src/middleware.ts
TheEhsanSarshar: in `line:288` we should also return `storage.setItem(...)` so if it threw an error the `catch` block `in line:347` can handle it.

TheEhsanSarshar: I think `thenableStorageSetItem()` that you mentioned @dai-shi is not possible with current `toThenable` solution. because if we return `thenableStorageSetItem(...)` in `line:288` then the `toThenable` will wrap the `thenableStorageSeteItem` with another `thenable`.

TheEhsanSarshar: dear @dai-shi please have a comment on this.

src/middleware.ts
TheEhsanSarshar: 👍

src/middleware.ts
dai-shi: Sorry, I think this was mymisundersting.
```ts
  return (storage as StateStorage).setItem(name, serializedValue)
```
just works right? (apart from the typing issue)

dai-shi:
Here you go: [playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKMCGNUF44QBGAVsAMYwBQA9NQssAM5wCWAdo2C1MACZyFEcRsBgZsqAGZQIAWzgAlYJgoAaYcGAAuGnQAWMGGEZbaAcxYw9AV0IA6MnOoARYJPaXgAG0QAVBrwubh44Pv4ogYReRNQAbGQALACcwAkArAAcCQBMAMyZhJhpSUUADBnAAIy51aUA7EmVSRlJjZUJuaWBSCiM1DwqMNQAbpWx1Oy8oHa8djCMAMQAMgktALQrSWmUoJCw9ChwAAqYsCyYXuI4ADy+cKA4bLzMV8DqANL3II-PcADWwEQEEkcDu+ABQJBvgAfHBcJQ4HAAD5wAAURxYZD+tw+sJRvgAlAjkWjUYwsDgtKCCXDYRisTi4O88dSduBoPAeqhXmhvOQYNBbl8fi8KW84ABVWH4MliqmE2mStl7TkMOAAUQAjtYLpZEABhPTkAFQW7StHkiTy9RsYAAd1e8ppuFhhAgEC8yjYyo5B25YqWLHJwFtpphcItctB6jAPGGLAg1kYjupiuGEBYvB9+y56DFaC8mOAgeDobNEbJhbI2mjcFjwHjieTVZrCpdcHTmez8HYOCgkhU3NsjDIUBYhGAQoeId+r1hAG9iajC6XgFAqa8S4812aCVTUc7YZ2s4jrjyW9DUcTESvt+u8xIC0WtyGd+ei9DVNeNJ6KNAAPwbvmfJ-mG6jvtWn7fsA2q6kgABibCARqsGFkghrGm++YXsSe5ooeHYZlmAC+lDdn66CiK8U7fDOooSOai6np805PMwELAqCl5gKcMDnF4VInGcFzUb4uLqDwYBeIOyFuh6Xp4celCkbsvq5gA4lRYo0SKD44OaB6Kr45G5q45IyEIMoEUpqn7L2a4DtWebQMAACCXA6XRenAAuxIiGIUa8gFEhmsSZhaVacCacFNwwn5w6juONZoAlY4TqFiJTOZECIFSZkwBZykmWqrz6gMAphsKXmvOo+pJgKshBa8EZNdpMIGX5ohUnV5JyK1Ehfoi4UwFS0WiVBiKYFwQEue5LChQRxmULmPIVZg4URkxcDDQAkjgsj7mwmCyDW5nsGYBFnWwZgkmw1heF4JJHDIshBpOV03Sid0PdCnUwHtwAHWiR0nVSH3qMMFzWKdBXndZRFPS9b3XMev2kSta3hQAalDk5oIxwiBQA3B2a6MAmbBUndsgTlAcDo2qRxk0GMAAPJgHxEAcGejHEtQABU-NwAAcsdqCcVYqC9VA62oKisj1QIqDWGwLDasANL89QxIgzDY7XXzgvXkLrlwJIKsUBTcA8DA1hQKr11wJghPQLLdjG6CRouzLG0K+SZuWHAAAGdqTBAdp2NEZAiZjwBBwA5MwU0sGi0BO2wTuMIgbBkKTUDk1zBAgpYBLu4icBC-BaegMdUmoJL3uy3Ajj3fwE7B65Wc5xgrvhUHZsvYoygUCL2AsMMwBl4i-MewAAlMA73fAhntlHMe98Axva0NWkb8hK+wqtG+GzP5dCxKIhO839VyBoY66gAXmuU8V57qA23btr8B9cChw9Svez4KwDODdpZu2NnPBemAl5wAAFJoFZsLOwH0WCSEQFvTq99CxP33paSkzkfbABxl4aGPNLqw0diiZ6chkYfV+oiAWp9p6Skvs7MgN95BZTXPxFgT8oAvyYa-WePEZbyHMm-Ru4VE7XygDwNg8BIYkMnpAtw0CvDwHgYguwIiRAYMykwbhj9gC4IKmDChF1FQ90IcQ0h+NEY0JEGeWONi8bQnoXARhHtnoNhDPARgchUCeFkMwaQt8JznUAbMPRAhpJYlvPvDiII0AEgANoAF0T4e1ZmwHwgCEBe1vEA2MEAUBnCYAI7ev89CeHifuRJ6BUkZIYUbM+cBTbmxzpzDOH97YRMwGwCAks6aohKV0i4NIOmWy5gIiRshMDsDNhbLpv8WD-3btHB6QCJySBcvkqWYprbAD0IgXgMsukzN8F7HpX9FmdKtn-R66yLien4JgSQfY9mEwkIc45pyx6FzTnaI0Gd+n3FkWnCAZB2GyL4BU4kXMlC-LOcAKxssTGRWSYqVEsoJDITQOoNcMgoDIXVOCqA8NMw0hREpZpgihY7RBKAiqQC8HAGkRPfOVtXqMDmTAMgehPlcyligMgqCWBAKNDwdQTKN4rP-gM+A7ckywo9pcoMrBmDKvNo9IFIYna8F4H0gQAw-gRP5f0jaAo4BAjtoAuFiIOUFyQlTawNM1yZNae0pZVsrWlJ2VAeQpTybBm-gc16ZgzkUwudUkJ3rC4PIARsl5VS9VBpZiyg5jqKbMG5by-lKyrCCttMIEVYqJVrmUWfSp4bkX7zTSGlM-TEAQ2ZlzF1bryWWPsdy1xRVKD2X7ION+R0oiThcb5REks2DXCxpeLm8F7ruE2bwfcijoZUhcQRLG3bkazpJJckMmBR0zuhHhA9I7PQnuJNHPlegT0jLYEoUgFA+D7gGP4ymGoyVbp3Y4vd+JgVHsvbOs9gHj2zr7Y4DgnIIDnqA3gOA1wlCMCXuoHabAwDWBgJeYkkhP2onYJhkacB0NEYIshmBVCkaOIo+olkcHj20ew5QAiBGMNYapKRrDoHD2MaYEvaUE6EBQCEFtREUH-Y8BQ+oiMeG2Nke-KgtEUmYHsEtDnYAnFqE9ppGJ8u1zDnSfgJgZgDHL1MfcYiUi5dDm23tnAPTk7gUPoXV4JdLzdPfhswZgUZngAucXasjzqIVMyZM0PIzRIbNWcGtFm9-KH1Pv5HwTz0X9OiE-vk9V4W-PXCbZZ8uxFYtWeJMRZu2B82og1g578BnHN7LYAFtzQWUs1bS4iHzMbM7Dvg3ltgiACsxa8+V29iXgDPpwLwVL7WfOwbA56MbE2UtVai9For35rP0z7bZeAEn4D1vgPgM8VU2LeRw+Jrm7gzBAQkGVZQFUeaxdGdmwSzNyTsy6YwHmLGsX+Q3BFG4+N1DDVGgD1x6hk4zR4HNMhG5FRbT2218uutivbV3oQisBE16XFjqjkQWDeEIc0UglBaDus4vwai7GuMyGKnBt+LhBOn4RmJ9o04rDmCynvB9AiVOiE06B9+KIKg-i3lR0CmpLNUcIqOSc5FfPUdZsLvgUoqOa0DVKxGZ7XNmBIhRPOUixJPR+NjjdnAfOSQqwXuwPgZFJ0iaR4iMBG18DDT5weTX8WBUrcdx4ug9YJ7yLBUS1gjKvbO8CcwBVcBrc3Kq3YMwdgU0Z3xxy4QmZ34zm4Y7Z2PFwqrcN4iJTqIACEEfpsdYy-ZqDV2rxpdRHYRvpwzCMAIvVi7HB5J2DtKcRrw2bNByZpy-2r0DWeh7zwKkKt4MIAgHAawYBeDfIACTzl1kVz5FhA+SMjzIuRMBcnT8hqs+Ddgg797gKttL-kG9N6gC3q-hXUc7xgM-p2XBvwF7t83HXnJ5sosMWwQQ183-zJEAMJyJGvV-w0H+n2kx3lH-xRiIkE190R1ZU2jgEbwT1EEMkL3LmLwl1CBZgr3LiJlRFZhIH5DsAhEYAp2q3C1RHqWSXSVLn9XVBUD0GxQhDbwvxL0IO8BZjsHYDIBISykYMBBpAADJJCY8+QcAvkcAUkIQmk1tH9Nsi8QRURhc4liDfdERtDRdBC2COCuCJDFQphjd9kJAlDAQ0k1Dv9K87MQF-9eQmd-N5wFDxQlcM5iJS4p1sV8ceEn5eBN14dhsDNZRZVwsj5CFS5-IAZZBUQUc74gi+BN0Nsv8oDoNhBMAJ5eB+p5D8Bk5kEwdv9ij-JmoZRWUJJgApJBweCbNGBci+ACj-NqjDk6jqxH9jwYCEiPcSs+Y6AeAkVvkQh+JPCC0BVQAWYIlpYM0JBv9aAr47RoA-hThEwnhZ9hAPQJ5PkgwUNxYQRo9pYIlhi5cJBQ1RiU9s5c4I9BjPkb9WUaQ11UAVjW4AFpIPkIAOUxwDU9VQ5C0xiLgJjBBm5yp-NVslirVciEYrAgx1A7RUATp7965w8DlQl5BscLdew58G4gTHpWUp5jdPD4IXo+d0M0Abi4cUQrdggv4sj-ZIByREVziul9RnlCgsQtdH1Zc-lzdY5-w7BURhoDwqUaSng6TbcGE6BCBJgNV58RB+BoSeig5whgBSUiVONNkzBgT2B0wb0KZ+5CVoBiQQDeMFsI9sDYDAY7BZSnhIjYiCQkixZH85hnMHTZZQj2x6ti8I9Qj29bNMtGdUiPTqclFH8SsbNfDvw3SQxsVgyjF8jnFcYGjoti8EygCkyN5-SL90yDE3CszrFcY7BvC4AS9cB8BvCSC0ti91ccBqyZsq8M46z-ML80sMzCdCzPTizqi2zosOzgi+cXESzW1vR2tosIy1s+yJMu9jSoA69xzy4g5mpohMAph+AMSd8f9W42B45FUkSWAI0Lj08NM4ABk4AWzbkpkM4e9mBil4x1zz9FzL8L8ytvBL4AyncxRSS+pY4KSbiIwByWjkylFkExQ+yb8gKuywzoYwKJBJzIypyozXT-D5ZDzkUkz4KwiazNCWzML6y9Dy5WUfzGo-y2BKSc4Iw8LmpoiL8b9qKDlwshI+IY4Qp8V-hAROJ8YELAz7N4j9p+ikLCsUL3TUybMmSYAWS+SKZ2SHpOS-hBThScCCR1BaT3Av41DXSvdsUawNToAxLy4JKpLI0uZZKvB5LFK1KbdeACVNLv9Eca9DymMIxHKzAFzERb87Bm9W9sLosb8sDvLJyej+LAZBKhsbNhpUdk5cJv8DNiKySyKKKyBvx-wHNMDG9XKmN1AsD4rfyN5-zKL1C4AqRMr+N1FlIgA)

src/middleware.ts
```diff
@@ -259,43 +302,62 @@ export const persist = <S extends State>(
...
       if (storageValue) {
-        const deserializedStorageValue = await deserialize(storageValue)
-
-        // if versions mismatch, run migration
+        return deserialize(storageValue)
+      }
+    })
+    .then((deserializedStorageValue) => {
+      if (deserializedStorageValue) {
         if (deserializedStorageValue.version !== version) {
```
AnatoleLucet, line 321:
```suggestion
        if ((deserializedStorageValue.version ?? false) && deserializedStorageValue.version !== version) {
```
see #409

dai-shi: removing `?? false` just works?

AnatoleLucet: not if the version is `0`

AnatoleLucet: No, my bad. This doesn't even work if the version is `0`. So maybe something like: `typeof deserializedStorageValue.version === "number" && deserializedStorageValue.version !== version`?

dai-shi: `StorageValue<S>` requires version to be `number` so we can even throw an error if typeof ... !== "number".

AnatoleLucet: Well, that wouldn't permit to store a primitive value as in #409
But maybe there is a reason not to allow that?

dai-shi: well, if we want to allow non-version store, we should also change the `StorageValue<S>` type. I'm fine with either way.

dai-shi (approved):
Looks great to me. Nice work!

TheEhsanSarshar:
Good job everybody. 
Thank you @dai-shi @barbogast @AnatoleLucet for working on this

barbogast:
Like-wise, after years working with open-source software I'm still amazed getting such high-quality software for free. And in addition to that it feels like there's 24/7 support with direct access to the core-developers, no premium-support-subscription needed.
But it's not one-sided since, as seen here, users are happy to contribute and collaborate, and so it's a common effort, rather than provider-consumer. I feel that the open source development is one of the best and highest value-generating way of collaboration that humans have.

Hehe, got all philosophical here 🙈 Anyway, thanks all 😍

src/middleware.ts
AnatoleLucet: Yes, let's merge this first

AnatoleLucet (approved):
Lgtm. Many thanks to everyone for your hard work! 🚀

PR #408, opened 2021-06-03
chore: mark react dependency as optional

triptu:
Discussion - https://github.com/pmndrs/zustand/discussions/407


dai-shi (approved):
Thanks. I think this is reasonable and without this modern bundlers would complain.

dai-shi:
Merged. Should eventually be released with some other fixes. No ETA.

PR #416, opened 2021-06-06
Rename persist.test.tsx to persistAsync.test.tsx and wrap in describe()

barbogast:
... so its consistent with persistSync.test.tsx


dai-shi (approved):
GitHub shows diffs but it's basically just renaming the file, right?

AnatoleLucet (approved):

dai-shi:
I see. There are two commits. 👍

PR #417, opened 2021-06-06
Remove usage of `any` from tests

barbogast:
Notes:
- used `@ts-ignore` twice, makes sense that way IMO ([here](https://github.com/pmndrs/zustand/pull/417/files#diff-c21e24854115b390eccde717da83f91feb2d5927a76c1485e5f0fdd0135c2afaR299) and [here](https://github.com/pmndrs/zustand/pull/417/files#diff-c21e24854115b390eccde717da83f91feb2d5927a76c1485e5f0fdd0135c2afaR343)
- There are 2 TS errors right now which I'm not sure how to resolve as the test seems to violate the type definitions. Remove the test? Ignore the error? One in line 222 and one in 259
- The remaining `any` in the file I couldn't easily remove (L387:404), and the way the test was written looked a bit strange. Should the test really be the way it is, with `getState2()` calling `getState()`which is defined in the outer scope?

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -212,7 +212,7 @@ it('can update the selector', async () => {
 
 it('can update the equality checker', async () => {
   type State = { value: number }
-  type Props = { equalityFn: EqualityChecker<State> }
+  type Props = { equalityFn: EqualityChecker<number> }
   const useStore = create<State>(() => ({ value: 0 }))
   const { setState } = useStore
   const selector: StateSelector<State, number> = (s) => s.value
@@ -256,7 +256,7 @@ it('can call useStore with progressively more arguments', async () => {
 
   let renderCount = 0
   function Component({ selector, equalityFn }: Props) {
-    const value = useStore(selector, equalityFn)
+    const value = useStore(selector as any, equalityFn)
     return (
       <div>
         renderCount: {++renderCount}, value: {JSON.stringify(value)}
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -69,9 +69,9 @@ it('uses the store with no args', async () => {
 })
 
 it('uses the store with selectors', async () => {
-  const useStore = create<any>((set) => ({
+  const useStore = create<CounterState>((set) => ({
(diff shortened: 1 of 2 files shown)


dai-shi:
> used `@ts-ignore` twice, makes sense that way IMO

You want to use `@ts-expect-error` instead.

dai-shi:
> There are 2 TS errors right now

The first one is fixable. The second one is really dynamic, and not sure if it's possible.
[patch.txt](https://github.com/pmndrs/zustand/files/6605211/patch.txt)

dai-shi:
> The remaining any in the file I couldn't easily remove

Here's the one I fixed:
```tsx
  it('can get the store', () => {
    type State = { 
      value: number         
      getState1: () => State
      getState2: () => State
    }
    const { getState } = create<State>((_, get) => ({
      value: 1,
      getState1: () => get(),
      getState2: (): State => getState(),
    }))
                                                                               
    expect(getState().getState1().value).toBe(1)                               
    expect(getState().getState2().value).toBe(1)                        
  })                                                                       
```

You can try the rest. Let me know if not successful.

> looked a bit strange.

But, they are tests to ensure the strange usage. At least, it's valid JS code.

barbogast:
All comments addressed except for one last TS error. I'll have another look tomorrow.

barbogast:
Okay, all comments addressed. PR is ready for all I know.

dai-shi (approved):
Overall looks good. Just one comment if it works.
Thanks for your work!

dai-shi:
So, do you mean you have more work coming?
Anyway, let's merge this as this is clean with only type changes.

barbogast:
@dai-shi 
> So, do you mean you have more work coming?
Not sure. What I did notice is that some of the test stores seemed unnecessarily different. Having one, or a more limited amount of store configurations and respective types might bring more order and clarity.
But that's a more invasive change, and maybe having a individual store per test is even better. I don't have a clear enough opinion here yet.

> Anyway, let's merge this as this is clean with only type changes.
Yeah, totally. Thanks for merging :-)

PR #420, opened 2021-06-08
Remove unnecessary console.error mocks from tests

barbogast:
... and move one necessary one into the describe() block


dai-shi (approved):
Looks good. Thanks!

PR #421, opened 2021-06-08
export NamedSet type for combining with immer

zgid123:
From PR#397, we changed the type of `PartialState`, so when combining `devtools` with `immer`, we have type error.

**Before**

```ts
export type PartialState<T extends State, K extends keyof T = keyof T> =
  | (Pick<T, K> | T)
  | ((state: T) => Pick<T, K> | T)
```

**Now**
```ts
export type PartialState<
  T extends State,
  K1 extends keyof T = keyof T,
  K2 extends keyof T = K1,
  K3 extends keyof T = K2,
  K4 extends keyof T = K3
> =
  | (Pick<T, K1> | Pick<T, K2> | Pick<T, K3> | Pick<T, K4> | T)
  | ((state: T) => Pick<T, K1> | Pick<T, K2> | Pick<T, K3> | Pick<T, K4> | T)
```

If we define `immer` like `README`

```ts
  config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>
  config((fn) => set(produce<T>(fn)), get, api)
```

it will raise error like this

image

So, if we export this type, we can fix this error

```ts
import produce, { Draft } from 'immer';
import { devtools } from 'zustand/middleware';
import create, { PartialState, State, StateCreator, UseStore } from 'zustand';

type NamedSet<T extends State> = {
  <K extends keyof T>(
    partial: PartialState<T, K>,
    replace?: boolean,
    name?: string,
  ): void;
};

type TImmerConfigFn<T extends State> = (fn: (draft: Draft<T>) => void) => void;
type TImmerConfig<T extends State> = StateCreator<T, TImmerConfigFn<T>>;

  config: TImmerConfig<T>,
): StateCreator<T, NamedSet<T>> => {
  return (set, get, api) => {
    return config((fn) => set(produce<T>(fn)), get, api);
  };
};

const createZustandStore = <T extends State>(
  createState: TImmerConfig<T>,
  prefix = 'project name',
): UseStore<T> => {
  return create(devtools(immer(createState), prefix));
};

export default createZustandStore;
```

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -30,7 +30,7 @@ export const redux =
     return { dispatch: api.dispatch, ...initial }
   }
 
-type NamedSet<T extends State> = {
+export type NamedSet<T extends State> = {
   <K extends keyof T>(
     partial: PartialState<T, K>,
     replace?: boolean,


dai-shi (approved):
Thanks for the investigation. Didn't know this.
If we export this and you can fix it with it, let's export it.
(Typing middleware is tough.)

dai-shi:
cc @TkDodo

Don-Cryptus:
> ```ts
> const immer = <T extends State>(
>   config: TImmerConfig<T>,
> ): StateCreator<T, NamedSet<T>> => {
>   return (set, get, api) => {
>     return config((fn) => set(produce<T>(fn)), get, api);
>   };
> };
 
I am getting an error `produce<T>(fn)`

error

is there a way to solve this?

dai-shi:
@sandren said there's an issue with `devtools`. Do you mean it's even without immer?

Don-Cryptus:
> ```ts
> const immer = <T extends State>(
>   config: TImmerConfig<T>,
> ): StateCreator<T, NamedSet<T>> => {
>   return (set, get, api) => {
>     return config((fn) => set(produce<T>(fn)), get, api);
>   };
> };
> ```

even after exporting the type I am still getting a typescript error on ` set(produce<T>(fn)`

error

what am I missing?

zgid123:
I just created a quick sandbox, there is no error

https://codesandbox.io/s/type-for-zustand-mrl95?file=/src/types.ts

zgid123:
could you put the code sandbox here and we can investigate together?

PR #423, opened 2021-06-09
Use jest.fn() to test callbacks

barbogast:
Using `jest.fn()` not only is more concise but more importantly is less prone to be used incorrectly.

As an example, `expect()` calls within the `onRehydrateStorage` callbacks would not make a test fail at all.

Also in `persistAsync.test.tsx` add some checks verifying that `onRehydrateStorage` was not called with an error. (In `persistSync.test.tsx` those checks had been added previously.)

### DIFF
--- a/tests/persistAsync.test.tsx
+++ b/tests/persistAsync.test.tsx
@@ -5,8 +5,6 @@ import { persist } from '../src/middleware'
 
 describe('persist middleware with async configuration', () => {
   it('can rehydrate state', async () => {
-    let postRehydrationCallbackCallCount = 0
-
     const storage = {
       getItem: async (name: string) =>
         JSON.stringify({
@@ -16,6 +14,7 @@ describe('persist middleware with async configuration', () => {
       setItem: () => {},
     }
 
+    const spy = jest.fn()
     const useStore = create(
       persist(
         () => ({
@@ -25,11 +24,7 @@ describe('persist middleware with async configuration', () => {
         {
           name: 'test-storage',
           getStorage: () => storage,
-          onRehydrateStorage: () => (state) => {
-            postRehydrationCallbackCallCount++
-            expect(state?.count).toBe(42)
-            expect(state?.name).toBe('test-storage')
-          },
+          onRehydrateStorage: () => spy,
         }
       )
     )
@@ -47,7 +42,7 @@ describe('persist middleware with async configuration', () => {
 
     await findByText('count: 0, name: empty')
     await findByText('count: 42, name: test-storage')
-    expect(postRehydrationCallbackCallCount).toBe(1)
+    expect(spy).toBeCalledWith({ count: 42, name: 'test-storage' }, undefined)
   })
 
   it('can throw rehydrate error', async () => {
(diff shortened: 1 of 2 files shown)


tests/persistAsync.test.tsx
```diff
@@ -25,11 +24,7 @@ describe('persist middleware with async configuration', () => {
         {
           name: 'test-storage',
           getStorage: () => storage,
-          onRehydrateStorage: () => (state) => {
-            postRehydrationCallbackCallCount++
-            expect(state?.count).toBe(42)
```
barbogast, line 30: As written in the description, those `expect()` calls never worked. They resulted in an error being thrown [here](https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L363) which only a second `catch()` would catch (I actually tried).

So I'd even go so far as to say that this way of testing callbacks is an anti pattern that should be avoided.

tests/persistAsync.test.tsx
```diff
@@ -5,8 +5,6 @@ import { persist } from '../src/middleware'
 
 describe('persist middleware with async configuration', () => {
   it('can rehydrate state', async () => {
-    let postRehydrationCallbackCallCount = 0
```
barbogast, line 8: Those counts are no longer necessary as jest will fail the test if the callback wasn't called. Nice.

tests/persistSync.test.tsx
```diff
@@ -28,118 +27,96 @@ describe('persist middleware with sync configuration', () => {
...
+    )
+    expect(onRehydrateStorageSpy).toBeCalledWith(undefined, undefined)
   })
 
   it('can migrate persisted state', () => {
-    let migrateCallCount = 0
-    let setItemCallCount = 0
+    const setItemSpy = jest.fn()
+    const onRehydrateStorageSpy = jest.fn()
+    const migrateSpy = jest.fn(() => ({ count: 99 }))
```
barbogast, line 89: Note that in this case the spy is not simply to verify that it has been called, it needs to return a value and is actually is a crucial part of the test itself. 

Which, one could argue, makes the test less expressive. Do you still want to use `jest.fn()`, here, or should those cases be left as they are?

barbogast:
This PR is not finished yet, but I'd still like you to have a look. I left a couple of notes to point out a couple of specifics.

Would be great if you could briefly give feedback whether you are fine with the pattern. Once we agreed I'll apply it to the rest of the tests.

So far I'm clearly in favour of using `jest.fn()`. Seems like an important improvement.

tests/persistSync.test.tsx
AnatoleLucet: I think the `jest.fn` is fine. I've always used mocked functions to return static values, so I don't really find it disturbing

AnatoleLucet (approved):
Definitely looks better with mocking :+1:
Thanks!

dai-shi (approved):
Yeah, looks good with jest.fn().

barbogast:
Ready for review / merge.

AnatoleLucet (approved):

barbogast:
Is there anything left to be done here? Merging this would unblock more of the test improvements, as I don't want to run into merge conflicts.

dai-shi:
Sorry, looks like I just missed the ready comment.

PR #424, opened 2021-06-09
Move some tests out of basic.test.tsx into separate files

barbogast:


barbogast:
This PR just moves code, no other changes. What you see where the only tests where I could see a distinct pattern. Don't know how the rest should be organised.

Also, with [this one](https://github.com/pmndrs/zustand/pull/424/files#diff-cd66dff526a0357b5999d7724f7eb1829885532641f281710bad1abfc77c5829R140) I'm not actually sure if it is type related. If not it should go back to `basic.test.tsx`.

The long test function for `subscribe` can probably broken down into multiple tests. Since this is way more invasive I'd do a follow-up PR.

dai-shi (approved):
Thanks for your work!

As this is just moving around, let merge it. If something should be changed/reverted, let's do follow-ups.

PR #427, opened 2021-06-11
Split test for subscribe() into independent it() functions

barbogast:


barbogast:
Looks like this is a decent improvement. The tests are now decoupled, and it's easier to read and debug.

The only issue is that previously the unsubscribe functionality was implicitly tested, which is not the case anymore. I'll add a separate test case for that.

In a follow-up I'll replace callbacks with spies as well. (I found another instance where a test would not actually fail if the code was broken. Boy, I didn't expect callbacks being such a bad idea in tests...)

dai-shi:
Sounds good, looks good.
Is it ready for merge?

barbogast:
Yeah, should be ready.

PR #428, opened 2021-06-12
Improve type definition of `create` to accept `CustomSetState`

hachibeeDI:
## Motivation

I was creating a middleware which returns `StateCreator<State, CustomSetState>` but it was denied because `create` is not accept `CustomSetState` as a second type argument.

## Bonus

As you noticed, I modified a test "can set the store without merging" in `basic.test.tsx`.

Previously `StateCreator` couldn't infer the state correctly so that store was inferred as object.

```typescript
// const setState: <never, never, never, never>(partial: PartialState<object, never, never, never, never>, replace?: boolean | undefined) => void
setState({ b: 2 }, true)

// const getState: () => object
getState()
```

But now it is as it should so we need let TypeScript compiler knows that keys `a` and `b` are possibly undefined.

Thanks!


dai-shi:
Hi, thanks for your work! This looks like a good improvement. Does this solve some existing issues?

I haven't looked into details yet, but I think we should keep the JS bundle size. Can you change it to make changes only in TS?

hachibeeDI:
Hello! Yes, my issue is going to be solved.
Also, I guess the issue https://github.com/pmndrs/zustand/pull/421 is related to this patch. That error says "Types of parameter 'set' and 'set' are incompatible.

Sure thing! Some of my extra nit pick makes size up. I'll fix it.

hachibeeDI:
I confirmed now the JS bundle size is as it was. 👍

src/vanilla.ts
```diff
@@ -48,19 +60,17 @@ export type StateCreator<T extends State, CustomSetState = SetState<T>> = (
...
   const listeners: Set<StateListener<TState>> = new Set()
 
   const setState: SetState<TState> = (partial, replace) => {
     // TODO: Remove type assertion once https://github.com/microsoft/TypeScript/issues/37663 is resolved
     // https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342
-    const nextState =
-      typeof partial === 'function'
-        ? (partial as (state: TState) => TState)(state)
-        : partial
+    const nextState = typeof partial === 'function' ? partial(state) : partial
```
dai-shi, line 73: Hm, so we can remove the type assertion because of this PR, or was it already possible with the previous one?
Either way, let's remove L71-L72 comments.

src/vanilla.ts
```diff
@@ -109,6 +119,7 @@ export default function create<TState extends State>(
 
   const destroy: Destroy = () => listeners.clear()
   const api = { setState, getState, subscribe, destroy }
-  state = createState(setState, getState, api)
+  // SetState<TState> includes CustomSetState so it should be ok
```
dai-shi, line 122: Can we do something like this?
```diff
diff --git a/src/vanilla.ts b/src/vanilla.ts
index ed33bc2..38d1fd7 100644
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -62,7 +62,7 @@ export type StateCreator<T extends State, CustomSetState = SetState<T>> = (
 
 export default function create<
   TState extends State,
-  CustomSetState = SetState<TState>
+  CustomSetState extends SetState<TState> = SetState<TState>
 >(createState: StateCreator<TState, CustomSetState>): StoreApi<TState> {
   let state: TState
   const listeners: Set<StateListener<TState>> = new Set()
@@ -119,7 +119,6 @@ export default function create<
 
   const destroy: Destroy = () => listeners.clear()
   const api = { setState, getState, subscribe, destroy }
-  // SetState<TState> includes CustomSetState so it should be ok
-  state = createState(setState as any as CustomSetState, getState, api)
+  state = createState(setState as CustomSetState, getState, api)
   return api
 }
```

dai-shi (commented):
Bundle size is good now. Please check two comments.

src/vanilla.ts
hachibeeDI: Ah, sorry my fix was sloppy it wasn't revert correctly but... it work okay. Sure, gonna remove comments.

src/vanilla.ts
hachibeeDI: Ahh,I should have figure it out before using `any`... Thank you for good catch!

hachibeeDI: Hm, like a test I added, I'd like to apply a `CustomSetState` that set a different type of state.
In a example, `set` cannot update `readA/B` because the second argument is `SetState<Pick<ExampleState, 'a' | 'b'>>`. However, `Pick<ExampleState, 'a' | 'b'>` couldn't satisfies `TState`...

```typescript
  CustomSetState extends SetState<any> = SetState<TState>
```

is going to fix `setState as any as CustomSetState` but test passes. I'll send a patch for your check.

dai-shi: Yeah, that's something I was concerned. Maybe this can't be helped.

dai-shi:
Not related, but linking #441.

src/vanilla.ts
```diff
@@ -48,19 +60,15 @@ export type StateCreator<T extends State, CustomSetState = SetState<T>> = (
   api: StoreApi<T>
 ) => T
 
-export default function create<TState extends State>(
-  createState: StateCreator<TState>
-): StoreApi<TState> {
+export default function create<
+  TState extends State,
+  CustomSetState extends SetState<any> = SetState<TState>
```
dai-shi, line 65: I'm not sure what this `any` would cause. It could cause runtime error without compile error, technically?
@TkDodo can you review this PR please?

dai-shi:
Hmm, after merging master, the problem seems gone. Let me add a commit.

dai-shi:
@hachibeeDI 
How is this 9b65883 ?
Does it accomplish what you would like to do?

src/index.ts
```diff
@@ -21,18 +21,21 @@ const isSSR =
 
 const useIsomorphicLayoutEffect = isSSR ? useEffect : useLayoutEffect
 
-export interface UseStore<T extends State> {
+export interface UseStore<T extends State, CustomSetState = SetState<T>> {
```
TkDodo, line 24: does `CustomSetState` not need an upper bound here as well?


```suggestion
export interface UseStore<T extends State, CustomSetState extends SetState<T> = SetState<T>> {
```

src/vanilla.ts
```diff
@@ -109,6 +120,6 @@ export default function create<TState extends State>(
 
   const destroy: Destroy = () => listeners.clear()
   const api = { setState, getState, subscribe, destroy }
-  state = createState(setState, getState, api)
+  state = createState(setState as CustomSetState, getState, api)
```
TkDodo, line 123: why is this type assertion necessary  now please?

src/index.ts
dai-shi: Actually, this was my misunderstanding. Let me revert it.

src/vanilla.ts
dai-shi: TS complains. I wonder if you have a better idea...

dai-shi: Now, I realized this doesn't solve anything...

dai-shi:
Let me hold on this PR. The type improvement seems good, so I'll create a separate PR.

src/vanilla.ts
dai-shi:
```suggestion
  state = createState(setState, getState, api)
```
reverting this for now.

dai-shi:
My big apologies about adding some commits without any success. We should probably start over.
At this point, what we have tried seems to add `any` in the library code to avoid `any` in the app code, which doesn't make a lot of sense.
Or, I may misunderstand the original issue. Anyway, let's close this and hope someone to open a new discussion. I'm generally willing to improve types around `CustomSetState`.

PR #429, opened 2021-06-12
Change tests for subscribe() to use jest.fn()

barbogast:
This commit also fixes the test  `should be called if new state identity is different` which was broken by 4632b74 (and unfortunately didn't fail).


barbogast:
Is ready for merging as far as I can tell. 

The only note I've got is that I've changed the existing spies to use the name "spy" to be consistent, also with the persist*.test.tsx files. But if you prefer to use "listener" I can change it back, no problem.

dai-shi (approved):
Looks good to me.

PR #430, opened 2021-06-12
Add test with missing migrate function to persistAsync.test.tsx

barbogast:


barbogast:
Ready for merging.

**And important**: I just noticed that the `onRehydrateStorage` callback gets called with different arguments when called sync vs async in the case of a missing migrate function.

[Here](https://github.com/pmndrs/zustand/pull/430/files#diff-a97f9733f2a6ac5d1855b9fc154caea95e3dd0908876f93bfc0b5deeba04ec16R197) you can see that the first argument is `{count: 0}` for the async version, while [this](https://github.com/pmndrs/zustand/blob/master/tests/persistSync.test.tsx#L148) shows that it is `undefined` when sync.

I'd consider that not in the scope of this PR, so if you'd like this to be addressed I'd have a look in a follow-up.

AnatoleLucet (approved):
It's looking good

> Ready for merging.
> 
> **And important**: I just noticed that the `onRehydrateStorage` callback gets called with different arguments when called sync vs async in the case of a missing migrate function.
> 
> [Here](https://github.com/pmndrs/zustand/pull/430/files#diff-a97f9733f2a6ac5d1855b9fc154caea95e3dd0908876f93bfc0b5deeba04ec16R197) you can see that the first argument is `{count: 0}` for the async version, while [this](https://github.com/pmndrs/zustand/blob/master/tests/persistSync.test.tsx#L148) shows that it is `undefined` when sync.

Well, that's a bit concerning

> I'd consider that not in the scope of this PR, so if you'd like this to be addressed I'd have a look in a follow-up.

Sure, feel free to open a pr 🙂

PR #431, opened 2021-06-12
Extend test case for perist middleware

barbogast:
This commit extends the test 'can persist state' to not only write state to the storage but to also reinitialise the store afterwards to check if the value would be rehydrated correctly.

### DIFF
--- a/tests/persistSync.test.tsx
+++ b/tests/persistSync.test.tsx
@@ -1,6 +1,28 @@
 import create from '../src/index'
 import { persist } from '../src/middleware'
 
+const createPersistantStore = (initialValue: string | null) => {
+  let state = initialValue
+
+  const getItem = (): string | null => {
+    getItemSpy()
+    return state
+  }
+  const setItem = (name: string, newState: string) => {
+    setItemSpy(name, newState)
+    state = newState
+  }
+
+  const getItemSpy = jest.fn()
+  const setItemSpy = jest.fn()
+
+  return {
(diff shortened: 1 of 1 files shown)


barbogast:
This feels a bit more like an integration test which simulates the whole persist-rehydrate scenario. It adds quite some setup-code, though, and maybe it should go to a new test-case.

And now that I think about it... maybe [this](https://github.com/pmndrs/zustand/blob/master/tests/persistSync.test.tsx#L82) is proof enough. So maybe this one isn't necessary.

If you think it is I'd also add an async version. Other I'd just close this one unmerged.

AnatoleLucet (approved):
It might be a bit redundant, but I think it's more expressive that way.
Looks fine to me. Thanks again  for your work 🚀

barbogast:
Oups, this one wasn't actually finished, and I forgot to flag it as such. Doesn't matter, I can do a follow up.

So far only the test for the sync storage is extended. Shall I do the same for the async version as well?

Thinking about it I realise that it might be worth it as there might be timing issues with the async version that the sync version doesn't have.

Doing this would mean that `createPersistantStore()` would either live in both `persistSync.test.jsx` and `persistAsync.test.jsx`, or it could live in a new file `tests/utils.ts` (which would be my preference).

@AnatoleLucet Your thoughts?

AnatoleLucet:
If you have the will to write the async tests, I'd prefer if the `createPersistantStore` is shared between the sync and async tests :+1:

Issue #432, opened 2021-06-13
createContext not working as expected


karthikeyan676:
Hey everyone, I just tried to use the createContext API but it seems saving the state in a global and updating changes to all the contexts instead of the particular single context. 

Here is the reproducible code sandbox link - [here](https://codesandbox.io/s/hopeful-davinci-29sxs?file=/src/App.js)

This is the expected result [here](https://codesandbox.io/s/context-api-forked-o8bdz?file=/src/index.js)

karthikeyan676:
It seems like working now. The mistake was I initiated the store outside from the component. So, it was using the same store for all the components instead of creating a new store. 

overall code
```jsx
import "./styles.css";
import create from "zustand";
import createContext from "zustand/context";

const { Provider, useStore } = createContext();

// Before store = create(...)
const store = () =>
  create((set) => ({
    bears: 0,
    increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
    removeAllBears: () => set({ bears: 0 })
  }));

const Button = () => {
  return (
    <Provider initialStore={store()}>
      {/** Before - initialStore={store} */}
      <ButtonChild />
    </Provider>
  );
};

const ButtonChild = () => {
  const state = useStore();
  return (
    <div>
      {state.bears}
      <button
        onClick={() => {
          state.increasePopulation();
        }}
      >
        +
      </button>
    </div>
  );
};

export default function App() {
  return (
    <div className="App">
      <Button></Button>
      <Button></Button>
    </div>
  );
}

```

dai-shi:
I originally suggested to pass `createStore={create}` to the Provider to avoid this misusing, but then we discussed and agreed that `initialStore={create()}` is more flexible. Maybe a doc issue?

karthikeyan676:
Yes, the documentation is incorrect.

dai-shi:
Mind opening a PR?

karthikeyan676:
Sure. I'm on it.

PR #433, opened 2021-06-13
docs: zustand/context correct doc, added snippets

karthikeyan676:
Corrected the wrong docs and added snippets to understand the createContext better. For more info check [here](https://github.com/pmndrs/zustand/issues/432#issuecomment-860151450)


readme.md
```diff
@@ -456,10 +456,10 @@ import createContext from 'zustand/context'
 
 const { Provider, useStore } = createContext()
 
-const store = create(...)
+const store = () => create(...)
```
dai-shi, line 459:
```suggestion
const createStore = () => create(...)
```
Maybe this seems better?

dai-shi (changes requested):
Thanks for your suggestion!
We have docs only in readme & wiki pages for now.
Would you move the snippet either in a) `<details>` in readme.md, b) a new wiki page?
I prefer a wiki page so that you and others can add more snippets.

readme.md
karthikeyan676: I would prefer the snippets to be in the repo. What if we added wiki snippets reference in the README.md and that person edit it later. That'll become a problem. What do you mean by <details> in README.md? any ref?

dai-shi (approved):

Issue #409, opened 2021-06-03
Having trouble getting deserialize to work


faroutchris:
Sorry for the noob question, I don't know if there's a better forum for these types of issues

I'm trying to persist my accessToken to localstorage. It works great. But when reloading the page, the deserialize function doesn't seem to hydrate the store state at all.

Here's my store
```
// > auth.ts
const useAuthStore = create<AuthStoreState>(
  devtools(
    persist(
      (set) => ({
        accessToken: '',
        setAccessToken: (accessToken) => set((_state) => ({ accessToken })),
      }),
      {
        name: 'AccessToken',
        serialize: ({ state }) => JSON.stringify(state.accessToken),
        deserialize: (accessToken: string) => JSON.parse(`{ "accessToken": "${accessToken}" }`), 
        // --> logging the JSON.parse statement gives { accessToken: "sometokenvalue" }
      }
    )
  )
);

// > in index.jsx
const { accessToken } = useAuthStore(); // --> <empty string>
```

Any ideas what's going wrong here?

faroutchris:
I realised that I don't need my custom deserializer, the defaults work just as well since I never manually read from localstorage. But still curious about what I'm doing wrong in the above post

dai-shi:
@faroutchris Thanks for reporting.

@barbogast Is this issue related to what you said there? https://github.com/pmndrs/zustand/pull/403#issuecomment-852873280

AnatoleLucet:
It might be because there's no `version` in the stored item.

https://github.com/pmndrs/zustand/blob/master/src/middleware.ts#L272-L280
This condition is true, and since you don't give a `migrate` function, `persist` thinks the stored value is outdated so it doesn't hydrate.

Might be fixed with something like:
```typescript
(deserializedStorageValue?.version ?? false) && deserializedStorageValue.version !== version
```

PR #437, opened 2021-06-18
Fix argument of onRehydrateStorage for missing migrate function

barbogast:
When the state loaded from the storage would need to be migrated but the persist configuration doesn't contain a migrate function `onRehydrateStorage` would be called with the initial state. Now it is called with the state loaded from the storage, which in this case is assumed to be undefined.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -357,7 +357,7 @@ export const persist =
         }
       })
       .then(() => {
-        postRehydrationCallback?.(get(), undefined)
+        postRehydrationCallback?.(stateFromStorageInSync, undefined)
       })
       .catch((e: Error) => {
         postRehydrationCallback?.(undefined, e)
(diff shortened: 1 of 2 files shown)


barbogast:
Strictly speaking this is a breaking change. Might be okay, though, as I guess it's quite edge-casey.

AnatoleLucet (commented):
Just to be sure, this is for #409?

barbogast:
@AnatoleLucet No, it's not related. It's a follow-up to https://github.com/pmndrs/zustand/pull/430#issuecomment-860041259.

AnatoleLucet (approved):
oh ok, my bad 😄

dai-shi:
@barbogast After merging this, the master branch failed on tests. Would you investigate it please?

barbogast:
Sure thing, I'm on it.

barbogast:
@dai-shi @AnatoleLucet Created https://github.com/pmndrs/zustand/pull/442

Issue #434, opened 2021-06-14
[v3.5.2] Type error when using devtools and persist in TypeScript


onlyling:
## Test code

```tsx
import create from 'zustand';
import { devtools, persist } from 'zustand/middleware';

interface TestState {
  t1: string;
  t2: number;
  setT1: (t: string) => void;
  [key: string]: any;
}

const useTest = create<TestState>(
  devtools(
    persist(
      (set) => ({
        t1: null,
        t2: null,
        setT1: (n) => {
          set({
            t1: n,
          });
        },
      }),
      {
        name: 'zustand-state-test',
        getStorage: () => localStorage,
      },
    ),
  ),
);

export default useTest;
```

## What happens?

`zustand: 3.4.2` no type error

`zustand: 3.5.2` show some type error

persist use SetState, but devtools use NamedSet.

## Screenshot

<img width="977" alt="zustand-type-error" src="https://user-images.githubusercontent.com/9999765/121844220-38bfca80-cd16-11eb-81e1-851258a30add.png">

## CodeSandbox

https://codesandbox.io/s/elegant-flower-s0e9c?file=/src/App.tsx

dai-shi:
Can #421 fix this??

onlyling:
> Can #421 fix this??

Not able to.

Maybe there is a problem with the way I use it, I want to use devtools and persist in the same store.

dai-shi:
Maybe #428 is related?

@TkDodo can help?

TkDodo:
I put the example code in a codesandbox: https://codesandbox.io/s/elegant-flower-s0e9c?file=/src/App.tsx

If you put it in the issue description, any linked PR will automatically deploy it, thanks.

I also tested it on #428 but the error remains the same. Need to look closer into it but might take some time.

punit-shah-cko:
I am getting a similar error when trying to use the same `createState` function for `create()` and `create(devtools())` calls

```ts
const createState = (set: SetState<Store>) => ({
  // ...
});

const useStore = isProduction
  ? create<Store>(createState)
  : create<Store>(devtools(createState));
```

It seems `NamedSet` is not compatible with `SetState`?

codesandbox: https://codesandbox.io/s/hopeful-wind-04jph?file=/src/store.ts

FranciscoMendes10866:
> I am getting a similar error when trying to use the same `createState` function for separate `create()` and `create(devtools())` calls
> 
> ```ts
> const createState = (set: SetState<Store>) => ({
[...]

Issue #411, opened 2021-06-04
TypeScript being able to setState by key


Marcsimm:
I have this state below with a bunch of properties. I want `setValue`  function to set a value based on a provided key there won't be a need for a bunch setState methods.

But I am getting the error below 

`
Argument of type '{ [x: string]: ValueOf<PersonStore>; }' is not assignable to parameter of type 'PartialState<SeriesFormStoreTypes, "setValue" | keyof State | "setInitialState">'.
` 

```
type ValueOf<T> = T[keyof T]

type PersonStore = {
    firstName: string
    lastName: string
    age: number
    height: number
    setValue: (key: keyof State, value: ValueOf<PersonStore>) => void
}

export const PersonFormStore = create<PersonStore>(
  (set: SetState<PersonStore>) => ({
    firstName: "",
    lastName: "",
    age 0,
    height: 0,
    setValue: (key, value) => set({ [key]: value }),
  })
```

Any idea of how I can add a function like described above and fix the type error?

> `PersonStore` needs to extend `State` and your `keyof State` should be `keyof PersonStore`:
> 
> ```ts
> type PersonStore = {
>   firstName: string;
>   lastName: string;
>   age: number;
>   height: number;
>   setValue: <K extends keyof Omit<PersonStore, "setValue">>(key: K, value: PersonStore[K]) => void;
> } & State;
> ```
> 
> In the above snippet, I've also omitted `setValue` from the `key` argument since you probably don't want it to be replaced.

I am sorry but the same problem still happens

AnatoleLucet:
`PersonStore` needs to extend `State`:
```ts
type PersonStore = {
  firstName: string;
  lastName: string;
  age: number;
  height: number;
  setValue: <K extends keyof Omit<PersonStore, "setValue">>(key: K, value: PersonStore[K]) => void;
} & State;
```

AnatoleLucet:
> I am sorry but the same problem still happens

My bad, I was on 3.3.1. It would seem like this is a regression (at least with my snippet) introduced in 3.5.0.

dai-shi:
@AnatoleLucet Let me and @TkDodo know if you have trouble fixing your snippet. A typescript playground might help.

dai-shi:
Closing as stale. Let us know if we should re-open.

Issue #331, opened 2021-03-18
useStore.subscribe hook does not correctly infer type from selector


disambiguator:
Demonstrated in this codesandbox: https://codesandbox.io/s/angry-pare-zndc9?file=/src/App.tsx:321-339

When using useStore.subscribe, the argument to the listener function is getting inferred as unknown, even though it should be using the return type from the selector function.

dai-shi:
I'm not sure if I understand the question correctly, because the codesandbox doesn't contain `useStore.subscribe`.

https://codesandbox.io/s/hungry-chatterjee-53zxo
This is what I modified. I think this is because of stricter typing in v3.3.3.

disambiguator:
@dai-shi I'm very sorry - forgot to press save in codesandbox :|. The link should show you the type error now.

dai-shi:
I see.
@TkDodo Can you help?

TkDodo:
@dai-shi I downgraded the example to `3.0.0`, and it also doesn't work for me there, so at least it's not a regression. could it be that this hasn't worked from the beginning?

looking a the types, the correlation seems to be there, and also the type inference looks correct when inspecting:

```
<number>(listener: StateSliceListener<number>, selector: StateSelector<State, number>, equalityFn?: EqualityChecker<number> | undefined) => () => void
```

if you change `(foo)` to `(foo: string)` in the selector, you will also get a type-error, because `number` is expected.

doing `useStore.subscribe<number>` also works, so I have really no clue at the moment where the undefined is coming from :(

dai-shi:
> @dai-shi I downgraded the example to `3.0.0`, and it also doesn't work for me there, so at least it's not a regression. could it be that this hasn't worked from the beginning?

@TkDodo Oooops, my apologies. I naively assumed it's v3.3.3 only.

Okay, if that's the case, it's probably something called context-sensitive function, which can't be solved on our end. It's TS issue/limitation.

If anybody has a real fix or some investigations, please jump in. For now, let me close this issue as a know limitation.

timothyallan:
I'm not a TS guru, but I know that MobX handles this somehow with their `reaction`. It's able to infer the returning property of what you specify in the selector. https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/core/reaction.ts

dai-shi:
Let's reopen this and see someone can tackle it.

dai-shi:
marked it as typing enhancement. (i'm not sure if it's possible.)

[...]

PR #441, opened 2021-06-18
make type NamedSet compatible with type SetState

zgid123:
According to #434, we are having issue with typescript when combining devtool with persist middleware.

The main issue is because NamedSet and SetState are not compatible, so I think we can temporary make them compatible.

I already added some codes to test the type for middleware use case only. And also add `immer` as `devDependency` for testing (not sure this is a way, I will remove if it's not suitable).

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -31,8 +31,13 @@ export const redux =
   }
 
 export type NamedSet<T extends State> = {
-  <K extends keyof T>(
-    partial: PartialState<T, K>,
+  <
+    K1 extends keyof T,
+    K2 extends keyof T = K1,
+    K3 extends keyof T = K2,
+    K4 extends keyof T = K3
+  >(
+    partial: PartialState<T, K1, K2, K3, K4>,
     replace?: boolean,
     name?: string
   ): void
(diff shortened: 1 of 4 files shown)


src/middleware.ts
```diff
@@ -31,8 +31,13 @@ export const redux =
...
 export type NamedSet<T extends State> = {
-  <K extends keyof T>(
-    partial: PartialState<T, K>,
+  <
+    K1 extends keyof T,
+    K2 extends keyof T = K1,
+    K3 extends keyof T = K2,
+    K4 extends keyof T = K3
+  >(
+    partial: PartialState<T, K1, K2, K3, K4>,
```
dai-shi, line 40: This change looks good.

zgid123:
I updated the test file. Not sure if we can combine `devtool`, `persist` and `immer` or not. Can anyone helps to provide a usage?

dai-shi:
Is this related with #216?
#212, #331, #411 are also in the same domain?

I know typing with middleware has been not ideal and, v3.5.* kind of revealed it. #428 is another effort.

I appreciate for your work and hope anyone can jump in.

dai-shi:
@TkDodo Do you think OK to merge this? Can review?

zgid123:
I tried to solve other issues, but still no luck. Will update it if I can solve any issue

dai-shi:
@zgid123 This PR itself is ready for review and merge, right?

zgid123:
Yup, this is for #434

TkDodo (approved):

dai-shi (approved):
Thanks, everyone.

Issue #405, opened 2021-06-01
Add ability to delete, or choose not to rehydrate the store in onRehydrateStorage


mirshko:
This might have overlap with the issue #267, however this isn't fully explained in that. 

I notice there isn't a great way to handle conditional persisted state rehydration. 

In the `onRehydrateStorage` it would be nice to be able to either return `state` or throw an Error / return directly to not rehydrate the storage. 

A little example; here I want to check if a cookie is still present, a JWT that expires after a week, and then choose not to rehydrate, the only option I found is below, to delete the localStorage key.

```ts
onRehydrateStorage: () => {
  if (!hasJWT()) {
    console.log("No JWT Present, Delete State");

    localStorage.removeItem("web3-store");
  }
},
```

An ideal API would be,

```ts
onRehydrateStorage: (state) => {
  if (hasJWT()) {
    return state
  }
  
  return false
}
```

Let me know if I'm missing something!

gabrielmoncea:
Looking forward to it!
https://github.com/pmndrs/zustand/discussions/814

gabrielmoncea:
@barbogast @AnatoleLucet 

## How I managed to make `zustand` accept asynchronous storage.

First of all, I want to acknowledge the simplicity with which this library was created, otherwise I wouldn't have the confidence to look into its source code and try a workaround. Big up team! 🙌🏼 
My hopes are that this is not just a hack, but serves as an inspiration to others or maybe even a good fix for this amazing library.

### Problem
As I said earlier, my issue was that `zustand`'s `persist` middleware was overwriting the storage on hydrate, because there was no "initial" way of making it to or check for a condition before overwriting the storage or wait for a trigger to start hydration, before getting the storage - in case one has an asynchronous storage ( the edge case of an encrypted storage which is available after the user successfully enters the PIN / Password / biometrics ( most cases happen in React Native).

### Solution
1. Because I am using the [`onRehydrateStorage` method](https://github.com/pmndrs/zustand/wiki/Persisting-the-store's-data#how-can-i-check-if-my-store-has-been-hydrated) to know when my hydration ends, I configured my persist store with empty options in order to block any storage overwrite or to solve any type error that comes if my getStorage is an async function.

```javascript
      onRehydrateStorage: () => async () => {},
      getStorage: undefined,
```

[...]

PR #442, opened 2021-06-20
Fix failing test in persistAsync.test.tsx

barbogast:
The test started to fail on master because #436 and #437 were developed in parallel and #436 was merged before #437.

This situation is interesting for 2 reasons:

1. It shows that #436 did add a test that was missing before: It tests an async storage without an initial value.

2. This test shows that the argument of `onRehydrateStorage` not only differ when the migrate function is missing (as was assumed in #437) but when the store is empty. The reason for the difference is that `get()` (see [here](https://github.com/pmndrs/zustand/pull/437/files#diff-0626369278c89781505d2c04865ecce1302e979271bcfb471e029c19addc13a0L360)) returns `undefined` when the storage is synchronous but returns the _default store value_ when asynchronous. The reason for the difference is that in sync mode the default store value hasn't been applied yet as the current function has not returned yet.

All of that means that the change of the argument of onRehydrateStorage in  #437 is more significant, as it doesn't only affect an edge case but a happy path for async storage.

### DIFF
--- a/tests/persistAsync.test.tsx
+++ b/tests/persistAsync.test.tsx
@@ -132,7 +132,7 @@ describe('persist middleware with async configuration', () => {
 
     const { findByText } = render(<Counter />)
     await findByText('count: 0')
-    expect(onRehydrateStorageSpy).toBeCalledWith({ count: 0 }, undefined)
+    expect(onRehydrateStorageSpy).toBeCalledWith(undefined, undefined)
 
     // Write something to the store
     act(() => useStore.setState({ count: 42 }))


barbogast:
I hope that the writeup is understandable and makes sense. Took me a bit to wrap my head around what's happening here :-)

So, to sum up, the fix is trivial, and correct, from what I can tell. But it also reveals the breaking change in #437 to be more significant.

I do think that passing the value of the storage is better than passing "storage-value mixed with default store value". For the use case in #405 it might be crucial. So eventually this change seems good to do.

And as `onRehydrateStorage` is not (yet) documented it might be fine to just change it. But, if you prefer not to do this in a minor upgrade I can revert the changes and have sync and async storages return different values. 

Note that both sync and async to returning the same, mixed, value might be a bit messy to implement. I did had a quick go, and realised that for sync mode it would mean to manually remember if the storage contained a value and if not return the default storage value. Doable, but doesn't feel nice.

dai-shi:
I'm not sure if it's correct, but what would happen if we do something like this?
```ts
        postRehydrationCallback?.(stateFromStorageInSync || get(), undefined)
```

barbogast:
> I'm not sure if it's correct, but what would happen if we do something like this?
> 
> ```ts
>         postRehydrationCallback?.(stateFromStorageInSync || get(), undefined)
> ```

It will result in `onRehydrateStorage` to be called with the default state when the storage is empty for async storage. Doable, but not consistent with the sync storage.

What should the API be? Should `onRehydrateStorage` be called with the `storeContent`, or with with `storeContent || defaultState`?

dai-shi:
Oh, I think we shouldn't use `stateFromStorageInSync` for postRehydrationCallback. It's for "sync" as the name says.
We should received what we want to pass from the promise/thenable value.

```js
      .then((foo) => {
        postRehydrationCallback?.(foo, undefined)
      })
```

> What should the API be?

@AnatoleLucet ?

AnatoleLucet:
I think `onRehydrateStorage` should behave (at the user level) the same way whether it's a sync or async storage

dai-shi:
@AnatoleLucet Should it get the hydrated state from storage, or merged one?

AnatoleLucet:
Not sure what you mean by "the merged one", but it should be the same value as what the state looks like after the hydration
Or maybe I misunderstood the question?

dai-shi:
I meant if the state should contain functions that are not serializable.
Anyway, I read the code again, and what we have now seems good.
Let me merge this to fix the test.

dai-shi (approved):
Thanks!

PR #446, opened 2021-06-21
Compatible set function of immer middleware proxy

meinto:
Until now the immer proxy sample code supports only functions. But the original `set` of zustand also allows to provide an object. The updated immer middleware sample in this pr allows an object too. With this change the immer middleware is fully compatible with the `set` function of zustand.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -322,7 +322,7 @@ const immer = <T extends State>config: StateCreator<T>): StateCreator<T> =>
     const nextState =
       typeof partial === 'function'
         ? produce(partial as (state: T) => T)
-        : partial as Draft<T>
+        : partial as T
     return set(nextState, replace)
   }, get, api)
 ```
--- a/readme.md
+++ b/readme.md
@@ -271,7 +271,12 @@ const log = config => (set, get, api) => config(args => {
 }, get, api)
 
 // Turn the set method into an immer proxy
-const immer = config => (set, get, api) => config(fn => set(produce(fn)), get, api)
+const immer = config => (set, get, api) => config((partial, replace) => {
+  const nextState = typeof partial === 'function'
+      ? produce(partial)
+      : produce(() => partial);
+  return set(nextState, replace)
+}, get, api)
 
 const useStore = create(
   log(
--- a/readme.md
+++ b/readme.md
@@ -274,7 +274,7 @@ const log = config => (set, get, api) => config(args => {
 const immer = config => (set, get, api) => config((partial, replace) => {
   const nextState = typeof partial === 'function'
       ? produce(partial)
-      : produce(() => partial);
+      : produce(() => partial)
   return set(nextState, replace)
 }, get, api)
 
@@ -315,19 +315,16 @@ For a TS example see the following [discussion](https://github.com/pmndrs/zustan
 
 ```ts
 import { State, StateCreator } from 'zustand'
-import produce, { Draft } from 'immer'
+import produce from 'immer'
 
-// Immer V8 or lower
-const immer = <T extends State>(
-  config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>
-): StateCreator<T> => (set, get, api) =>
-  config((fn) => set(produce(fn) as (state: T) => T), get, api)
-
-// Immer V9
-const immer = <T extends State>(
-  config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>
-): StateCreator<T> => (set, get, api) =>
-  config((fn) => set(produce<T>(fn)), get, api)
+const immer = <T extends State>config: StateCreator<T>): StateCreator<T> => 
+  (set, get, api) => config((partial, replace) => {
+    const nextState =
+      typeof partial === 'function'
+        ? produce(partial as (state: T) => T)
+        : produce(() => partial)
+    return set(nextState, replace)
+  }, get, api)
 ```
 
 </details>
--- a/readme.md
+++ b/readme.md
@@ -274,7 +274,7 @@ const log = config => (set, get, api) => config(args => {
 const immer = config => (set, get, api) => config((partial, replace) => {
   const nextState = typeof partial === 'function'
       ? produce(partial)
-      : produce(() => partial)
+      : partial
   return set(nextState, replace)
 }, get, api)
 
@@ -315,14 +315,14 @@ For a TS example see the following [discussion](https://github.com/pmndrs/zustan
 
 ```ts
 import { State, StateCreator } from 'zustand'
-import produce from 'immer'
+import produce, { Draft } from 'immer'
 
 const immer = <T extends State>config: StateCreator<T>): StateCreator<T> => 
   (set, get, api) => config((partial, replace) => {
     const nextState =
       typeof partial === 'function'
         ? produce(partial as (state: T) => T)
-        : produce(() => partial)
+        : partial as Draft<T>
     return set(nextState, replace)
   }, get, api)
 ```


readme.md
```diff
@@ -271,7 +271,12 @@ const log = config => (set, get, api) => config(args => {
 }, get, api)
 
 // Turn the set method into an immer proxy
-const immer = config => (set, get, api) => config(fn => set(produce(fn)), get, api)
+const immer = config => (set, get, api) => config((partial, replace) => {
+  const nextState = typeof partial === 'function'
+      ? produce(partial)
+      : produce(() => partial)
```
dai-shi, line 277: How is it different from this?
```suggestion
      partial
```

dai-shi (commented):
It's a bit unfortunate that the middleware is no longer one-liner.
Anyway, it's good to support better.
Did you have any issue with the previous one? curious when you want to mix fn and obj updates.

readme.md
meinto: Yea you right. There is no difference. I updated the code accordingly.

meinto:
> Did you have any issue with the previous one? curious when you want to mix fn and obj updates.

Yes, for example we want to use both in our project depending wether we need to change the state dependend on other state variables or not. The basic example already uses both variants.

```js
  bears: 0,
  increasePopulation: () => set(state => ({ bears: state.bears + 1 })),
  removeAllBears: () => set({ bears: 0 })
}))
```

readme.md
```diff
@@ -312,17 +317,14 @@ For a TS example see the following [discussion](https://github.com/pmndrs/zustan
...
-const immer = <T extends State>(
-  config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>
-): StateCreator<T> => (set, get, api) =>
-  config((fn) => set(produce<T>(fn)), get, api)
+const immer = <T extends State>config: StateCreator<T>): StateCreator<T> => 
+  (set, get, api) => config((partial, replace) => {
+    const nextState =
+      typeof partial === 'function'
+        ? produce(partial as (state: T) => T)
+        : partial as Draft<T>
```
dai-shi, line 325: Isn't this type just `T` ?
Hm, `config` doesn't seem to be typed properly.
Would you try it with codesandbox? We should still want to have both for v8 and v9?

meinto: Yea, you are right again :) I created a sandbox for this:

https://codesandbox.io/s/laughing-nash-vz53e?file=/src/zustand/middlewares/immer.ts

When I switch from immer v8 to immer v9 I see no difference. The typings should apply for both versions.

readme.md
```diff
@@ -312,17 +317,14 @@ For a TS example see the following [discussion](https://github.com/pmndrs/zustan
...
-  config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>
-): StateCreator<T> => (set, get, api) =>
-  config((fn) => set(produce(fn) as (state: T) => T), get, api)
-
-// Immer V9
-const immer = <T extends State>(
-  config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>
-): StateCreator<T> => (set, get, api) =>
-  config((fn) => set(produce<T>(fn)), get, api)
+const immer = <T extends State>config: StateCreator<T>): StateCreator<T> =>
```
dai-shi, line 320: This is not precise. It's previously `StateCreator<T, (fn: (draft: Draft<T>) => void) => void>`.

```suggestion
const immer = <T extends State>(
  config: StateCreator<T,  (fn: ((draft: Draft<T>) => void) | PartialState<T>, replace?: boolean) => void>
): StateCreator<T> => 
```

But, I can't make it work.
https://codesandbox.io/s/nostalgic-grass-rbdu6?file=/src/zustand/middlewares/immer.ts

BTW, you want to try the immer middleware without `pipe` too.

dai-shi (changes requested):
TS part would require more work. (I'm not sure how easy it is, even if it's possible.)

We can merge only with JS part change, if you want.

dai-shi:
> > Did you have any issue with the previous one? curious when you want to mix fn and obj updates.
> 
> Yes, for example we want to use both in our project depending on wether we need to change the state dependend on other state variables or not. The basic example already uses both variants.
> 
> ```js
> const useStore = create(set => ({
>   bears: 0,
>   increasePopulation: () => set(state => ({ bears: state.bears + 1 })),
>   removeAllBears: () => set({ bears: 0 })
> }))
> ```

Thanks for the use case! This makes sense.
Is it correct that even with the previous immer middleware, the following works?
```js
  removeAllBears: () => set(() => ({ bears: 0 }))
```

readme.md
meinto: Mhh in my point of view the previous typing ist wrong, isn't it? Because `CustomSetState` of the `StateCreator` type doesn't get a `Draft<T>` as first argument. `Draft<T>` is only relevant for the `produce` function of `immer` which is called inside the custom-`set` function.

So the first argument which the `immer`-middleware gets is the `PartialState<T>`. In case that `PartialState<T>` is a function it would be `(state: T) => ...` but not `(state: Draft<T>) => ...`... And inside the middleware the `produce` function is called. Here it has to be cast to `Draft<T>` because the `producer` function of the immer-producer gets a `Draft<T>` as first argument. The `produce` function would return a new immutable state object.

So in case that the first argument (`PartialState<T>`) of the `CustomSetState` is a object we have to return an immutable object too, to satisfy typescript. I found a function exported by the immer library which is called `castImmutable` and seems to do what we need:

```ts
/**
* This function is actually a no-op, but can be used to cast a mutable type
* to an immutable type and make TypeScript happy
* @param value
*/
export declare function castImmutable<T>(value: T): Immutable<T>;
```

So what about the following implementation: https://codesandbox.io/s/laughing-nash-vz53e?file=/src/zustand/middlewares/immer.ts

```ts
import produce, { Draft, castImmutable } from "immer";
import { State, StateCreator } from "zustand";

export const immer = <T extends State>(
  config: StateCreator<T>
  config(
    (partial, replace) => {
      const nextState =
        typeof partial === "function"
          ? produce(partial as (state: Draft<T>) => void)
          : castImmutable(partial);
      return set(nextState, replace);
    },
    get,
    api
  );
```

Let me know if I made thinking mistakes.

The point is, that I intend to write middlewares compatible with previous or subsequent middlewars in the chain. If I don't do that, I can't simply compose other middlewares because I have to struggle with the typings.

I know that I can compose immer without `pipe`, but we have other middlewares which we've put in this middleware-chain and want to use them in our custom `create` & `createPersisted` helpers. And because of dry I did't want to attach them in both helpers.

If you agree with the proposed implementation, I would update the pr accordingly. Let me know what your thoughts are.

meinto:
> Thanks for the use case! This makes sense.
Is it correct that even with the previous immer middleware, the following works?
>  removeAllBears: () => set(() => ({ bears: 0 }))

Yes this is correct. Only the short hand `removeAllBears: () => set({ bears: 0 })` was not working. And the typings were not compatible to other middlewares. So you could'nt simply `pipe` them.

readme.md
dai-shi: Thanks for your thoughts. Please bear with me.
I know your final goal is to support pipe, but I'd like to stick with immer only typing for now.

More precisely, I would like to allow following example.

```ts
const useStore = create<MyState>(
  immer((set) => ({
    bears: 0,
    increasePopulation: () =>
      set((state) => {
        ++state.bears;
      }),
    removeAllBears: () => set({ bears: 0 })
  }))
);
```

----

Oh, I didn't know
```ts
const useStore = create<MyState>((set) => ({
  bears: 0,
  increasePopulation: () =>
    set((state) => {
      ++state.bears;
    }),
  removeAllBears: () => set({ bears: 0 })
}));
```
this doesn't complain.
So, v3.3.3 complains, but v3.4.0 doesn't.

Okay, this is not your fault... My concern has nothing to do with immer.
Let me investigate this. Give me some time.

wip codesandbox: https://codesandbox.io/s/nervous-nightingale-lv2yc?file=/src/App.tsx

dai-shi: Okay, I gave up. Returning `undefined` is actually fine unless `replace=true`.
Types with `replace=true` is not correct at the moment, but fixing it causes too many errors.
As `replace=true` is rare case, let's leave it for now.

In summary, now I understand `PartialState<S>` type is compatible with `(draft: S) => void` since v3.4.0, I will change my mind.

dai-shi (approved):
This is fine for now. Types don't conflict since v3.3.3.
Everything is not very strictly typed, but fairly handy. I don't think we can achieve 100% type safety anyway.

PR #455, opened 2021-06-23
fix(persist): #451

### LINKED ISSUE #451: onRehydrateStorage argument
in v3.5.1 the persist middleware's `onRehydrateStorage` option provided the entire state as the argument.  

in 3.5.4 only the persisted data is returned in the argument.

I was relying on having the whole state, to use these options (as an example):

```javascript
const persistOptions = {
    name: my-storage',
    whitelist: ['persisted'],
    onRehydrateStorage: () => {
        return (state: FullState) => {
            state.setRehydrated(true);
        };
    },
};
```

Is this a deliberate change?  If so, I'll find another way to keep track of rehydration status.

thanks

AnatoleLucet:
Closes #451

This is to follow-up #437 & #442.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -322,6 +322,15 @@ export const persist =
       void setItem()
     }
 
+    const configResult = config(
+      (...args) => {
+        set(...args)
+        void setItem()
+      },
+      get,
+      api
+    )
+
     // rehydrate initial state with existing stored state
 
     // a workaround to solve the issue of not storing rehydrated state in sync storage
@@ -362,22 +371,12 @@ export const persist =
         }
       })
       .then(() => {
+        stateFromStorageInSync = { ...configResult, ...stateFromStorageInSync }
         postRehydrationCallback?.(stateFromStorageInSync, undefined)
       })
       .catch((e: Error) => {
         postRehydrationCallback?.(undefined, e)
       })
 
-    const configResult = config(
-      (...args) => {
-        set(...args)
-        void setItem()
-      },
-      get,
-      api
-    )
-
-    return stateFromStorageInSync
-      ? { ...configResult, ...stateFromStorageInSync }
-      : configResult
+    return stateFromStorageInSync || configResult
   }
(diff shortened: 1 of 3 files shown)


src/middleware.ts
```diff
@@ -362,21 +371,15 @@ export const persist =
         }
       })
       .then(() => {
-        postRehydrationCallback?.(stateFromStorageInSync, undefined)
+        postRehydrationCallback?.(
+          { ...configResult, ...stateFromStorageInSync },
```
dai-shi, line 375: Yeah, this is something I concerned with "merged" state. Looks good.
One caveat is L375 and L384 are different objects. Would be nice to make it referentially identical.

AnatoleLucet: good idea

AnatoleLucet: there you go :slightly_smiling_face:

src/middleware.ts
```diff
@@ -362,22 +371,12 @@ export const persist =
         }
       })
       .then(() => {
+        stateFromStorageInSync = { ...configResult, ...stateFromStorageInSync }
```
dai-shi, line 374:
```suggestion
        stateFromStorageInSync = stateFromStorageInSync && { ...configResult, ...stateFromStorageInSync }
```
We need this, right?

AnatoleLucet: image
:wink:

dai-shi: The point is creating a new object in the case of async: `configResult !== { ...configResult }`. In other words, `|| configResult` does never work.

AnatoleLucet: Yes it does. When using an async storage

dai-shi: Oh, I see. In async, it doesn't reach this line before the return. Then, I'm fine with this.

dai-shi (approved):

dai-shi:
I will merge this later unless @barbogast says something.

dai-shi:
Let's merge this so that we can work on #459 without conflicts.

barbogast:
Sorry, I'm on vacation and missed the notifications.

Post-merge approval from me :-)

PR #459, opened 2021-06-24
Fix/#457

### LINKED ISSUE #457: zustand persist - actions inside nested object are undefined on rehydration
Hi all - I use the pattern of wrapping all my actions in an 'actions' object. I wanted to use persistence for an easter egg state, but I noticed that when I would toggle the easter egg, refresh the page and try again I would get errors with the toggle being undefined.

Is this desired behavior? Moving the actions to a flat object and not in my 'actions' object solved the issue, I only did the actions object because I believe I saw it elsewhere. Are there other shortcomings of using that pattern?

```ts
interface EasterEggState {
    active: boolean;
    actions: {
        set: (value: boolean) => void;
        toggle: () => void;
    };
}

export const useEasterEgg = create<EasterEggState>(
    persist(
        (set, get) => ({
            active: false,
            actions: {
                set: (value) => set({ active: value }),
                toggle: () => set({ active: !get().active }),
            },
        }),
        {
            name: 'app.easteregg',
            version: 1,
        },
    ),
);
```

image

AnatoleLucet:
Closes #457

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -6,6 +6,7 @@ import {
   StateCreator,
   StoreApi,
 } from './vanilla'
+import { deepMerge } from './utils'
 
 export const redux =
   <S extends State, A extends { type: unknown }>(
@@ -371,7 +372,7 @@ export const persist =
         }
       })
       .then(() => {
-        stateFromStorageInSync = { ...configResult, ...stateFromStorageInSync }
+        stateFromStorageInSync = deepMerge(configResult, stateFromStorageInSync)
         postRehydrationCallback?.(stateFromStorageInSync, undefined)
       })
       .catch((e: Error) => {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,0 +1,23 @@
+export const isObject = (o: Object) => o?.toString() === '[object Object]'
+
+export const deepMerge = <T extends Object, S extends Object>(
+  target: T,
+  source?: S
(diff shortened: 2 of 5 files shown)


dai-shi:
@AnatoleLucet zustand api is tuned for shallow merge, at least for now. So, instead of deep merge for all cases, let's add `merge` option in `PersistOptions`, and let users to pass custom merge function.
This doesn't work fully like before for #457, but at least they have a workaround.

AnatoleLucet:
What about an option for conditional deep merge?
```
{
  name: "my-storage",
  deepMergeHydration: true,
}
```

dai-shi:
that might be acceptable, if you would strongly suggest it. but it's not our preference, or at least not mine.

zustand api is minimal, and deep actions are not officially recommended. (it's unopinionated and possible with users' responsibility.) It's very strange to support deep merge (even if it's optional, and I personally don't like to _ease_ deep merge...) in middleware, while core doesn't.

actually for #457, `blacklist: ['actions']` can be an official recommendation, not a workaround, no?

AnatoleLucet:
Yeah, you're right. Haven't though about that.

> actually for #457, blacklist: ['actions'] can be an official recommendation, not a workaround, no?

Definitely. Though if it's the case, I think we must support more advanced paths like `actions.foo`, `actions.0.bar`, etc... so users aren't forced to structure their state in a specific way.

PR #462, opened 2021-06-25
fix: do not use generator to support IE11

### LINKED ISSUE #461: IE 11 not working with > v3.5.2
Hey team,

Looks like the generator jazz that was removed in https://github.com/pmndrs/zustand/releases/tag/v3.5.3 and then added back in https://github.com/pmndrs/zustand/releases/tag/v3.5.4 has stopped IE11 working

I've currently got it set in my webpack config like so:
```javascript
// Zustand in IE land, need to specify that we're using the cjs and not esm
// https://github.com/pmndrs/zustand/issues/354#issuecomment-816534766
{
  resolve: {
    alias: {
      zustand: path.resolve(__dirname, "node_modules/zustand/index.js"),
    },
  },
};
````

And this works in 3.5.2 but not for any of the later versions.

Worth mentioning that I could be missing some key detail in how I'm meant to use this library with IE11 support

Thanks ❤️

dai-shi:
close #461 

https://codesandbox.io/s/react-forked-bqocl?file=/src/App.js works well.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 3858,
-    "minified": 1546,
-    "gzipped": 748,
+    "bundled": 3986,
+    "minified": 1610,
+    "gzipped": 785,
     "treeshaked": {
       "rollup": {
         "code": 124,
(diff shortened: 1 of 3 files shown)


dai-shi:
@adrhumphreys Would you try this with IE11?

See https://ci.codesandbox.io/status/pmndrs/zustand/pr/462 "Local install instructions"

adrhumphreys:
Hey @dai-shi, that fully fixes it for me 🎉 

Thanks so much for being so quick to not only respond but also fix the issue, really appreciate it

PR #463, opened 2021-06-25
feat(persist): add nested paths support for whitelist/blacklist

AnatoleLucet:
Follow up of #459 

This is an initial implementation of nested paths for `persist`'s whitelist and blacklist options. See the tests for an example.

I still need to work a bit on type checks, but the logic is pretty much there. Some bits could be improved, and some vars renamed though. I'd like some feedbacks on this if any :slightly_smiling_face:

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -189,11 +189,11 @@ type PersistOptions<S> = {
   /**
    * Prevent some items from being stored.
    */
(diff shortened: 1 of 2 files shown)


Richard19766 (approved):

dai-shi:
@AnatoleLucet As #575 is merged into main, please merge main here whenever you have a chance.

Issue #465, opened 2021-06-27
persist middleware deserialize typings incorrect.


n1ru4l:
I am trying to use a custom `deserialize` function in order to avoid crashes/mismatches between runtime and typescript types. However, the typings of `persist` seem to be incorrect, as TypeScript complains that the setter function attached to the state is not returned from the `deserialize` function:

[...]

PR #466, opened 2021-06-27
feat(persist): add a merge function option

AnatoleLucet:
Follow up of #459 

This pr adds a new `merge` option for `persist` which can be used to do custom hydration merges.


dai-shi (approved):
Looks great to me. I think tests are just fine too (without deep merge).

dai-shi:
This may affect #465, so merging now.

PR #468, opened 2021-06-27
fix(persist): incorrect deserialize return type

### LINKED ISSUE #465: persist middleware deserialize typings incorrect.

AnatoleLucet:
Closes #465 

@n1ru4l can you try this out with the codesandbox below to see if it fixes your issue? :pray: :slightly_smiling_face:

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -161,7 +161,7 @@ export type StateStorage = {
   setItem: (name: string, value: string) => void | Promise<void>
 }
 type StorageValue<S> = { state: S; version: number }
-type PersistOptions<S> = {
+type PersistOptions<S, PersistedState extends Partial<S> = Partial<S>> = {
   /** Name of the storage (must be unique) */
   name: string
   /**
@@ -185,7 +185,9 @@ type PersistOptions<S> = {
    * @param str The storage's current value.
    * @default JSON.parse
    */
-  deserialize?: (str: string) => StorageValue<S> | Promise<StorageValue<S>>
+  deserialize?: (
+    str: string
+  ) => StorageValue<PersistedState> | Promise<StorageValue<PersistedState>>
   /**
    * Prevent some items from being stored.
    */
@@ -263,7 +265,7 @@ export const persist =
       name,
       getStorage = () => localStorage,
       serialize = JSON.stringify as (state: StorageValue<S>) => string,
-      deserialize = JSON.parse as (str: string) => StorageValue<S>,
+      deserialize = JSON.parse as (str: string) => StorageValue<Partial<S>>,
       blacklist,
       whitelist,
       onRehydrateStorage,


src/middleware.ts
```diff
@@ -156,6 +156,18 @@ export const combine =
       )
     )
 
+type RemoveMethods<T> = Pick<
```
AnatoleLucet, line 159: btw, it's always a joy to get headache while working with ts generics :upside_down_face:

dai-shi (commented):
I wonder if this is a right solution. A custom deserialize may support deserializing methods somehow. We don't deep merge, so recursive RemoveMethods seems incorrect (thus type assertion).

Can we solve it with generics?
```ts
type StateStorage<PersistedState extends Partial<S> = S> = {
  ...
  deserialize?: (str: string) => StorageValue<PersistedState> | Promise<StorageValue<PersistedState>>
```

AnatoleLucet:
Yeah, that's a better idea.
Though I'm not sure to understand why you want to add this generic & method on `StateStorage`. Did you mean `PersistOptions`?

dai-shi:
Ah, that's my mistake. Yes, `PersistOptions`.

AnatoleLucet:
Only thing with a shallow partial is, what about cases like #457? Maybe a deep partial could be ok in this case?

dai-shi:
Sorry if I don't get it correctly, but does it only apply if a user pass custom `merge` option? If the merge function is fully typed, it should just work. Or is it about something other than `merge`?

AnatoleLucet:
I doesn't have anything to do with `merge` :wink: 

Here's an example:
```ts
const unstorableMethod = () => {}

const useStore = create(
  persist(
    () => ({
      count: 0,
      unstorableMethod,
      object: { anotherCount: 0, anotherUnstorableMethod: unstorableMethod },
    }),
    {
      name: 'test-storage',
      deserialize: () => {
        return {
          state: {
            object: {
              anotherCount: 0, // error, `anotherUnstorableMethod` is not optional in this object
            }
          },
          version: 0,
        }
      },
    }
  )
)
```
image

dai-shi:
To this to work with shallow, the state type must be
```ts
type TState = {
  count: number
  unstorableMethod: () => unknown
  object: {
    anotherCount: number
    anotherUnstorableMethod?: () => unknown
  }
}
```
So the type error without it is correct, I suppose.

BTW, `merge` option is only effective with "sync". I guess we would want to have the same behavior with "async".

AnatoleLucet:
I pushed the changes

> BTW, merge option is only effective with "sync". I guess we would want to have the same behavior with "async".

Not sure what you mean. Do you have an example?

dai-shi (approved):
The change looks good.

dai-shi:
> I pushed the changes
> 
> > BTW, merge option is only effective with "sync". I guess we would want to have the same behavior with "async".
> 
> Not sure what you mean. Do you have an example?

In your code example https://github.com/pmndrs/zustand/pull/468#issuecomment-869467669,
if we pass`merge=deepMerge`,
- with sync storage, anotherUnstorableMethod is preserved
- with async storage, anotherUnstorableMethod is not preserved (removed)

Is it correct?

AnatoleLucet:
I don't really understand why it wouldn't, but I tried modifying a test, and it seems to work just like with a sync storage. What makes you think that `anotherUnstorableMethod` wouldn't be preserved?

<details><summary>The test</summary>
<p>

```tsx
  it('gives the merged state to onRehydrateStorage', async () => {
    const onRehydrateStorageSpy = jest.fn()

    const storage = {
      getItem: async () =>
        JSON.stringify({
          state: { count: 1 },
          version: 0,
        }),
      setItem: () => {},
    }

    const unstorableMethod = () => {}

    const isObject = (o: Object) => o?.toString() === '[object Object]'

    const deepMerge = <T extends Object, S extends Object>(
      target: T,
      source?: S
    ): T & S => {
      if (!source) return target as T & S

      const mergedObject: Partial<T & S> = {}
      const mergedEntries: [keyof T & S, (T & S)[keyof T & S]][] =
        Object.entries(target).concat(Object.entries(source)) as any

      mergedEntries.forEach(([key, value]) => {
        if (isObject(target[key]) && isObject(value)) {
          mergedObject[key] = deepMerge(target[key], value)
        } else {
          mergedObject[key] = value
        }
      })

      return mergedObject as T & S
    }

    const useStore = create(
      persist(
        () => ({
          count: 0,
          unstorableMethod,
          object: {
            anotherCount: 0,
            anotherUnstorableMethod: unstorableMethod,
          },
        }),
        {
          name: 'test-storage',
          getStorage: () => storage,
          onRehydrateStorage: () => onRehydrateStorageSpy,
          deserialize: () => {
            return {
              state: {
                count: 1,
                object: {
                  anotherCount: 0,
                },
              },
              version: 0,
            }
          },
          merge: (persistedState, currentState) =>
            deepMerge(currentState, persistedState),
        }
      )
    )

    function Counter() {
      const { count } = useStore()
      return <div>count: {count}</div>
    }

    const { findByText } = render(<Counter />)

    await findByText('count: 0')
    expect(onRehydrateStorageSpy).toBeCalledWith(
      {
        count: 1,
        unstorableMethod,
        object: { anotherCount: 0, anotherUnstorableMethod: unstorableMethod },
      },
      undefined
    )
  })
```

image

</p>
</details>

dai-shi:
https://github.com/pmndrs/zustand/tree/test-persist-merge-nested
@AnatoleLucet A branch is pushed. Hope I don't mistake.

AnatoleLucet:
Oh yeah, you're right. Nice catch! I forgot to make the storage async in the tests :sweat_smile: 
I'll have a look into it.

dai-shi:
please merge master.

AnatoleLucet:
done :slightly_smiling_face:

dai-shi:
The change seems good (harmless), but I wonder if this alone solves @n1ru4l 's issue?

AnatoleLucet:
I'm not familiar at all with fp-ts & io-ts, but it does seem to work by copy pasting @n1ru4l's code in this pr's csb
https://codesandbox.io/s/react-typescript-forked-ubpf8?file=/src/test.ts

dai-shi:
That's great to know. Thanks for confirming!

n1ru4l:
@AnatoleLucet awesome! Thank you for solving this so quick ❤️

PR #469, opened 2021-06-29
fix(persist): merge function not working with async storages

AnatoleLucet:
Context: https://github.com/pmndrs/zustand/pull/468#issuecomment-869501387

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -345,7 +345,7 @@ export const persist =
     // a workaround to solve the issue of not storing rehydrated state in sync storage
     // the set(state) value would be later overridden with initial state by create()
     // to avoid this, we merge the state from localStorage into the initial state.
-    let stateFromStorageInSync: S | undefined
+    let stateFromStorage: S | undefined
     const postRehydrationCallback = onRehydrateStorage?.(get()) || undefined
     // bind is used to avoid `TypeError: Illegal invocation` error
     toThenable(storage.getItem.bind(storage))(name)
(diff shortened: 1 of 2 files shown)


src/middleware.ts
```diff
@@ -368,22 +368,23 @@ export const persist =
...
         }
-      })
-      .then(() => {
+
         stateFromStorageInSync = merge(
           stateFromStorageInSync as S,
           configResult
         )
+
+        set(stateFromStorageInSync as S)
```
dai-shi, line 384:
```suggestion
        set(stateFromStorageInSync as S, true)
```
I feel like we should replace it, but i'm not so sure.
btw, can we call `set` in sync?

src/middleware.ts
```diff
@@ -368,22 +368,23 @@ export const persist =
...
+        return setItem()
+      })
+      .then(() => {
         postRehydrationCallback?.(stateFromStorageInSync, undefined)
```
dai-shi, line 388:
```suggestion
        postRehydrationCallback?.(stateFromStorage, undefined)
```
Now, as it' used for both sync/async, maybe rename it?

src/middleware.ts
```diff
@@ -368,22 +368,23 @@ export const persist =
             )
           } else {
             stateFromStorageInSync = deserializedStorageValue.state
```
dai-shi, line 370:
```suggestion
            return deserializedStorageValue.state
```
Now, it seems we can simplify this and remove L375 if-statement.

src/middleware.ts
dai-shi: hm, it was already called in the previous code too.

dai-shi (commented):
overall looks good. seems there's some room for refactoring. what do you think?

src/middleware.ts
AnatoleLucet: the set in sync doesn't seem to trigger a render if that's what you're worried about

src/middleware.ts
AnatoleLucet: good idea :+1:

AnatoleLucet:
Pushed some changes, let me know what you think

dai-shi (approved):
Looks good. Unless we have any failing tests, we can't notice possible bugs anyway. Let's merge this. Thanks for your work.

src/middleware.ts
dai-shi: if it doesn't error, it's fine. I mistook it was only for async.

PR #484, opened 2021-07-08
Improve createContext's useStore types

### LINKED ISSUE #481: createContext's useStore type are not complete?
Seems like types of `useStore` are not really consistent with the usual `useStore` hook created by `create` function. Due to the fact there is only one call signature it doesn't infer the store's types correctly. Example below:

```ts
const {Provider, useStore} = createContext<SlideshowStore>();
const store = useStore(); // store is of `unknown` type
```

I have worked around it like this:

```ts
const {Provider, useStore, useStoreApi: getStoreApi} = createContext<SlideshowStore>();
const useSlideShow: UseStore<SlideshowStore> = Object.assign(useStore, getStoreApi());
```

Is this an oversight or was this done on purpose? If this is something that could be improved I am willing to work on this and submit a PR.

wadamek65:
Fixes #481 

### Description
I've split up the `UseStore` interface into two interfaces, one which contains the full API and the other one which only contains call signatures. I've reused one of the interfaces to adjust the types of `createContext`'s `useStore` hook which previously didn't infer the state type correctly when it was used without a selector. 

### Other remarks
I was also wondering if there was any reason why the `createContext`'s hooks are split up like that into one that exposes store Api and another one that exposes data? Wouldn't it provide a more consistent API if the returned hook behaved just like `create` one?

Also it might be worth looking why TypeScript didn't report any errors in the tests as there were properties accessed on an `unknown` object. I didn't look too into depth into this one but I might be willing to also check it out in the near future.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -6,8 +6,8 @@ import {
   useMemo,
   useRef,
 } from 'react'
-import { UseStore } from 'zustand'
-import { EqualityChecker, State, StateSelector } from './vanilla'
+import { UseStore, UseStoreData } from 'zustand'
+import { State, StateSelector } from './vanilla'
 
 function createContext<TState extends State>() {
   const ZustandContext = reactCreateContext<UseStore<TState> | undefined>(
@@ -44,9 +44,9 @@ function createContext<TState extends State>() {
     )
   }
 
-  const useStore = <StateSlice>(
+  const useStore: UseStoreData<TState> = <StateSlice>(
     selector?: StateSelector<TState, StateSlice>,
-    equalityFn: EqualityChecker<StateSlice> = Object.is
+    equalityFn = Object.is
   ) => {
     // ZustandContext value is guaranteed to be stable.
     const useProviderStore = useContext(ZustandContext)
--- a/src/context.ts
+++ b/src/context.ts
@@ -9,7 +9,7 @@ import {
 import { EqualityChecker, UseStore } from 'zustand'
 import { State, StateSelector } from './vanilla'
 
-export interface UseStoreData<T extends State> {
+interface UseStoreData<T extends State> {
   (): T
   <U>(selector: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
 }
--- a/src/context.ts
+++ b/src/context.ts
@@ -6,8 +6,13 @@ import {
   useMemo,
   useRef,
 } from 'react'
-import { UseStore, UseStoreData } from 'zustand'
+import { EqualityChecker, UseStore } from 'zustand'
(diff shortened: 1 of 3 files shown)


dai-shi:
This might be just a preference, but let's define the type only in context.ts without changing/reusing index.ts. (Then, it would not export any new types.)

The reason why `createContext` split two hooks is to avoid violating rules of hooks. If we can just return the same hook, it means we wouldn't need to provide special `createContext` at all, and developers can just use context.

If you can look into the tests too, that would be great.

wadamek65:
@dai-shi I've redefined the type in `context.ts`.

I've also taken a deeper look into what was happening before and the behavior was correct after all. The problem was that TypeScript inferred the types based on the type of the returned variable. So if we immediately destructure it into an object with properties, the inferred type of state will be that object.

[More detailed reproduction here](https://www.typescriptlang.org/play?ts=4.3.5#code/JYWwDg9gTgLgBAYygUwIY2QYQgOwwD3gDMoIQ4ByALwFcBnGVHAEwHoFcCYKAoHmAJ5hkcALICAyjGgiAvHADePOHFQAuOAGYeAXx4ccDRXHrIpMuDrjykaDNjzJCAHnHmUAPgAUASj4GjBVUAGjgAI0trEzozaRRfOFZWOABJHCJkKDo4CDCAK2QEeCD1VRwBULCNJgFI1GyGdGR9XDoIABtkADp2iABzL1RKvxbDeAYLeVN3ZASk1PTM7JocAGscCAB3HFUGxgw4ZhoRaTgNuBQYGigcZGY4ADdUKGBUMM64QWFRwNUNBEqGnuVnkExQAG5EskAKJQUhQH4dbq9AYAw4+IA)

The changes in this PR fix that problem and the function is now correctly typed.

wadamek65:
@dai-shi Also, should this new interface also be exported from `context.ts`? Or should I get rid of the export?

dai-shi:
Oh, so it's how TS behaves. A good TIL. Thanks for the investigation.

src/context.ts
```diff
@@ -6,8 +6,13 @@ import {
   useMemo,
   useRef,
 } from 'react'
-import { UseStore } from 'zustand'
-import { EqualityChecker, State, StateSelector } from './vanilla'
+import { EqualityChecker, UseStore } from 'zustand'
+import { State, StateSelector } from './vanilla'
+
+export interface UseStoreData<T extends State> {
```
dai-shi, line 12:
```suggestion
interface UseStoreData<T extends State> {
```
Yeah, let remove `export` for this time, because I'm not confident with the name.

dai-shi (approved):
Looks great. Thanks for your work!

PR #486, opened 2021-07-08
Revert Provider initialStore types

wadamek65:
### Description
Recent depracation of `initialStore` prop in `Provider` component causes broken declaration files to be generated. This causes TS compilation error when not using `skipLibCheck: true`.

```
node_modules/zustand/context.d.ts:5:18 - error TS2339: Property 'initialStore' does not exist
 on type '{ createStore: () => UseStore<TState>; children: ReactNode; }'.

5     Provider: ({ initialStore, createStore, children, }: {
                   ~~~~~~~~~~~~
```

You can see this error reported in the `context.d.ts` file after building the package.
```ts
// master version
import { ReactNode } from 'react';
import { UseStore } from 'zustand';
import { EqualityChecker, State, StateSelector } from './vanilla';
declare function createContext<TState extends State>(): {
    Provider: ({ initialStore, createStore, children, }: {
        createStore: () => UseStore<TState>;
        children: ReactNode;
    }) => import("react").FunctionComponentElement<import("react").ProviderProps<UseStore<TState> | undefined>>;
    useStore: <StateSlice>(selector?: StateSelector<TState, StateSlice> | undefined, equalityFn?: EqualityChecker<StateSlice>) => StateSlice;
    useStoreApi: () => {
        getState: import("zustand").GetState<TState>;
        setState: import("zustand").SetState<TState>;
        subscribe: import("zustand").Subscribe<TState>;
        destroy: import("zustand").Destroy;
    };
};
export default createContext;
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -19,7 +19,7 @@ function createContext<TState extends State>() {
     createStore,
     children,
   }: {
-    initialStore?: UseStore<TState>
+    initialStore?: UseStore<TState> /** @deprecated */
     createStore: () => UseStore<TState>
     children: ReactNode
   }) => {
--- a/src/context.ts
+++ b/src/context.ts
@@ -19,7 +19,10 @@ function createContext<TState extends State>() {
     createStore,
     children,
   }: {
-    initialStore?: UseStore<TState> /** @deprecated */
+    /**
+     * @deprecated
+     */
+    initialStore?: UseStore<TState>
     createStore: () => UseStore<TState>
     children: ReactNode
   }) => {
--- a/src/context.ts
+++ b/src/context.ts
@@ -15,11 +15,11 @@ function createContext<TState extends State>() {
   )
 
   const Provider = ({
-    // @ts-expect-error deprecated
     initialStore,
     createStore,
     children,
   }: {
+    initialStore?: UseStore<TState>
     createStore: () => UseStore<TState>
     children: ReactNode
   }) => {


dai-shi:
wow, i didn't know this. but can we mark `initialStore` deprecated so that TS users notice it's deprecated without running the code? How can we annotate a prop `/** @deprecated */`? https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated-comments

wadamek65:
@dai-shi Added the directive, I've checked and it is also present in the type declarations file.
```ts
import { ReactNode } from 'react';
import { UseStore } from 'zustand';
import { EqualityChecker, State, StateSelector } from './vanilla';
declare function createContext<TState extends State>(): {
    Provider: ({ initialStore, createStore, children, }: {
        initialStore?: UseStore<TState> | undefined; /** @deprecated */
        createStore: () => UseStore<TState>;
        children: ReactNode;
    }) => import("react").FunctionComponentElement<import("react").ProviderProps<UseStore<TState> | undefined>>;
    useStore: <StateSlice>(selector?: StateSelector<TState, StateSlice> | undefined, equalityFn?: EqualityChecker<StateSlice>) => StateSlice;
    useStoreApi: () => {
        getState: import("zustand").GetState<TState>;
        setState: import("zustand").SetState<TState>;
        subscribe: import("zustand").Subscribe<TState>;
        destroy: import("zustand").Destroy;
    };
};
export default createContext;
```

dai-shi:
https://codesandbox.io/s/react-typescript-forked-t34q8?file=/src/App.tsx
Hm, it doesn't show any deprecated messages. Is it because of codesandbox?

wadamek65:
@dai-shi Hmm, unfortunately, I do not know. I built the package locally and the directive is definitely there. I also tried using this locally but at least WebStorm doesn't show anything about deprecation. Are we sure this directive works with object properties? I played around and it actually worked when I added it to the whole function but couldn't get it working with an object property.

At the very least the [documentation](https://jsdoc.app/tags-deprecated.html) also doesn't mention that it can be used for anything else than an entire symbol so this might not be possible I think?

wadamek65:
@dai-shi I also checked on VSCode and it seems to work there but it has to be formatted in a certain way. Sorry for the confusion.
image

dai-shi:
https://codesandbox.io/s/react-typescript-forked-j4phl?file=/src/App.tsx

image

Looks good now.

dai-shi (approved):
Thanks for the fix!!

PR #491, opened 2021-07-12
better immer help

tonypee:
The docs example for immer could embed the immer code in the common 'action' style setter as in the other examples. This might give a better example of the common usage of immer in zustand. Otherwise, it encourages putting the setter code in the component

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -240,13 +240,13 @@ import produce from 'immer'
 
 const useStore = create(set => ({
   lush: { forest: { contains: { a: "bear" } } },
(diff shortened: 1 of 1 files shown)


dai-shi:
I assumed there was an intention by @drcmda or someone?

tonypee:
Yah, even if it shows a different approach, this aligns more with the first examples. Maybe have both if you really want it

dai-shi (approved):
Sounds fair. We may modify or revert it for unknown reason, but I think this can be merged now.

PR #492, opened 2021-07-12
fix(context): Export UseContextStore interface

wadamek65:
The changes from #484 cause an error when building declaration files from code using the `createContext` API. This was an oversight and the interface has to be exported or the declaration files cannot be emitted properly. I suppose TypeScript is not able to infer an interface that only contains call signatures with it not being exported.

```
error TS4023: Exported variable 'useStore'
 has or is using name 'UseStoreData' from external module ".../node_modules/zustand/context" but cannot be named.
```

Exporting the interface fixes this error (tested it locally). I'll also change the name of the interface if needed.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -9,7 +9,7 @@ import {
 import { EqualityChecker, UseStore } from 'zustand'
 import { State, StateSelector } from './vanilla'
 
-export interface UseStoreData<T extends State> {
+export interface UseContextStore<T extends State> {
   (): T
   <U>(selector: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
 }
@@ -52,7 +52,7 @@ function createContext<TState extends State>() {
     )
   }
 
-  const useStore: UseStoreData<TState> = <StateSlice>(
+  const useStore: UseContextStore<TState> = <StateSlice>(
     selector?: StateSelector<TState, StateSlice>,
     equalityFn = Object.is
   ) => {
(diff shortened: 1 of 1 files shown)


src/context.ts
```diff
@@ -9,7 +9,7 @@ import {
 import { EqualityChecker, UseStore } from 'zustand'
 import { State, StateSelector } from './vanilla'
 
-interface UseStoreData<T extends State> {
+export interface UseStoreData<T extends State> {
```
dai-shi, line 12:
```suggestion
export interface UseStore<T extends State> {
```
`Data` looks strange to me. Is this too confusing?

wadamek65: On the other hand, `UseStore` is already declared in the main file and it also contains the store API
```ts
export interface UseStore<T extends State> {
  (): T
  setState: SetState<T>
  getState: GetState<T>
  subscribe: Subscribe<T>
  destroy: Destroy
}
```

I will be fine with renaming it to anything that you would like to, but just wanted to bring this up.

dai-shi: Yeah, that's why I wondered if it's too confusing.
How about `UseStoreInContext`? This is for TS and users wouldn't typically write it directly, right?

dai-shi: Or, `UseStoreWithoutApi`

wadamek65: I like your first suggestion better I think, but to make it a little more readable how about `UseContextStore`?

dai-shi: I'm ok with it.

dai-shi (approved):
Looks good. Does anyone else have a comment?

dai-shi:
Publish: https://www.npmjs.com/package/zustand/v/3.5.7

Issue #497, opened 2021-07-19
Split store into multiple files 


getzelus:
Hello, thanks for Zustand, magnificient library simple and efficient for React state management.

I would like to be able to use a single store with create but split it into multiple files like combineReducers in Redux. 
Because in a big project, i may have a lot of data and functions for different components that will interact with each other in the same store. The js file will be very big and difficult to maintain.

It would be cool to be able to import data and functions (with get and set state) inside the main store from different js files.
Regards,
Getzel

dai-shi:
How about this?
```js
const createSliceA = (set) => ({
  countA: 0,
  incA: () => set((prev) => ({ countA: prev.countA + 1 })),
}))

const createSliceB = (set) => ({
  countB: 0,
  incB: () => set((prev) => ({ countB: prev.countB + 1 })),
}))

const useStore = create((set) => ({
  ...createSliceA(set),
  ...createSliceB(set),
}))
```

Related: #13

getzelus:
Thanks, it seems to work perfectly so I proposed a new example in the documentation.

dai-shi:
In terms of conceptual documentation, I would like to somehow show this pattern.
```js
const useStore = create((set) => ({
  storeSliceA: ...,
  storeSliceB: ...,
  storeSliceC: ...,
  dispatchX: () => set(...),
  dispatchY: () => set(...),
}))
```

zbeyens:
[zustood](https://github.com/udecode/zustood) is proposing a pattern to have a global store combining all the stores

PR #498, opened 2021-07-20
Update readme.md - tutorial on splitting the store

getzelus:
Hello, i propose my little stone to the documentation with a little tutorial on how to split the store. 
The purpose if to clarify it for new users of Zustand. 
Thanks to the all, the creators and dai-shi who helped me : https://github.com/pmndrs/zustand/issues/497

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -570,18 +570,18 @@ So, like combineReducers in Redux, you can split your store in different functio
 
 import create from 'zustand'
 
-const bearStore = (set, get) => ({
+const createBearSlice = (set, get) => ({
     bears: 1,
     eatsFish: () => set((prev) => ({ fishes: prev.fishes > 1 ? prev.fishes - 1 : 0}))
   })
 
-  const fishStore = (set, get) => ({
+  const createFishSlice = (set, get) => ({
     fishes: 10
   })
 
 const useStore = create( (set, get) => ({
-    ...bearStore(set, get),
-    ...fishStore(set, get)
+    ...createBearSlice(set, get),
+    ...createFishSlice(set, get)
 }))
 
 function App() {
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Thanks for working on this. It looks good to me. We want to keep readme compact as much as possible (if something is too big, we consider a wiki page), and there might be room for improvement. 

@drcmda Do you have any suggestion?

readme.md
```diff
@@ -559,7 +559,57 @@ const useStore = create(
...
+  
+When your project is big with different kind of datas, for convenient coding you want to keep small files. 
+So, like combineReducers in Redux, you can split your store in different functions that will be able to access each other, keeping track of global state. 
 
+
+```jsx
+
+import create from 'zustand'
+
+const bearStore = (set, get) => ({
```
dai-shi, line 573: Technically, this is not a store, but a store creator. So, I would name it something like `createBearStore`.

getzelus: Yes it is just a slice, i will rename it : createBearStore or bearSlice ?

dai-shi: I would prefer having `create` or `creator`.

Although it's a slice, it can be used alone.
```js
const createBearSlice = (set, get) => ...
const useStore = create(createBearSlice)
```
Now, create, create sounds weird in this case. So, maybe we don't need `create`. 🤔

getzelus: I see no problem, I renamed it createBearSlice.

dai-shi:
I feel like we should put this in a wiki page too.
And then we can add more examples and best practices about organizing a big store.

dai-shi:
Would you create a "Best Practices" section and add a link to the wiki page?

dai-shi:
Closing in favor of #505.

PR #502, opened 2021-07-21
Update readme.md with the multiline set function

getzelus:
I think it is interesting to show this possibility with an example.


dai-shi:
Thanks for the suggestion, but I don't think we should add a new section to show this usage.
This is about very basic JS grammar. If we were to add this level of explanation, there would be many more to add.

Could we edit just one existing example to show `return { ... }`?

Otherwise, let's create a new wiki page for JS beginners and add a link in readme.

dai-shi:
Please follow #505 way.

PR #503, opened 2021-07-21
refactor(example/site): prism renderer, copy button

barelyhuman:
**commit description**
```
added a more maintained prism renderer instead
of react-prism and also a copy button in case
someone wishes to clone the code instead of
enabling pointer events on everything
```

## Why ?
The website renders a code snippet which you cannot copy this has lead a few awkward situations when explaining people how they could use zustand and being unable to copy the code, I understand that the pointer events were disabled to better suit the animated background, so instead a simple button was added for the same. 

## Changes 
- Replaces `react-prism` with `prism-react-renderer` (extra addition, can be removed on request)
- Adds `CodePreview` component (implements the above)
- Adds `CopyButton` component , solves the **Why**

## Additional Comments 

Let me know if anything needs to be changed for the same 😄


dai-shi:
Thanks for your work.
I'm not sure how I'm doing wrong, but opening
https://githubbox.com/barelyhuman/zustand/tree/change-render-and-copy-button/examples
shows something like this.
image

barelyhuman:
> Thanks for your work.
> I'm not sure how I'm doing wrong, but opening
> https://githubbox.com/barelyhuman/zustand/tree/change-render-and-copy-button/examples
> shows something like this.
> image

The rewired setup needs to be checked and updated, I experienced this too but wasn't the scope of the change so I didn't, let me know if you want me to look into it. 

You can run the build and it'll work fine, if you want the HMR to work you can go to the `<CodePreview />` line in `App.js` , comment it out, and uncomment it, the re-render on it fixes the view. 

The final output should look like this 

Screenshot 2021-07-22 at 4 57 26 AM

barelyhuman:
Okay no, it also failed on the build for me this time, I'll fix this

barelyhuman:
> The rewired setup needs to be checked and updated, I experienced this too but wasn't the scope of the change so I didn't, let me know if you want me to look into it.
Not an issue with rewired, it's an issue with `prism-react-renderer`

>You can run the build and it'll work fine, if you want the HMR to work you can go to the <CodePreview /> line in App.js , comment it out, and uncomment it, the re-render on it fixes the view
The solution is a small wait for mount before showing the component, which is done and i'm pushing that. 

Though I think I should dig deeper as to why that happens since it's only happing on this setup and not on a normal CRA App.

dai-shi:
https://githubbox.com/barelyhuman/zustand/tree/change-render-and-copy-button/examples
works good now.

> Though I think I should dig deeper

Yeah, please do if you can.

barelyhuman:
> Yeah, please do if you can. 

`prism-react-renderer`  needs it to be a pre tag to render the preview as intended without the above mount hack, made the needed changes.

dai-shi (approved):
The changes looks good.

dai-shi:
@drcmda Would you be able to deploy this to zustand.surge.sh?

drcmda:
can we use pmnd.rs for this and pu surge to rest?

dai-shi:
Sure!
There are https://valtio-demo.pmnd.rs and https://jotai.pmnd.rs currently.
The former is manual install by me (or any pmndrs members), and the latter is github integration (master/pr branches will be deployed automatically).
Which do you prefer?

Issue #454, opened 2021-06-23
Zustand module bundling problems


jorekai:
Hey, I am building private modules for an application, which are shared through workspaces.
 When using the build library in my ejected react-app i run into difficulties.
 Something with the default export of zustand's create method is not working with my rollup/create-react-app configurations.
 I have no clue if this error is produced by zustand or rollup or CRA/webpack. 
Any help is appreciated.

The error:

![image](https://user-images.githubusercontent.com/15143394/123092908-46b6cf00-d42b-11eb-89a6-962766df3005.png)

The typescript source file which must be bundled (lib/src/stores.ts):

```typescript
import { KitaModel } from "../utils/apiModels";
import create from "zustand";

export class Kita {...}
export interface KitaStore {...}
export const useKitaStore = create<KitaStore>((set) => ({...}));
```

The resulting type declarations (lib/dist/stores.d.ts):

```typescript
import { KitaModel } from "../utils/apiModels";

export declare class Kita {...}
export interface KitaStore {...}
export declare const useKitaStore: import("zustand").UseStore<KitaStore>;
```

The rollup config (lib/rollup.config.js):

```typescript
import pkg from './package.json';
import resolve from '@rollup/plugin-node-resolve';
import { terser } from 'rollup-plugin-terser';
import typescript from 'rollup-plugin-typescript2';

export default {
	preserveSymlinks: true,
	input: 'src/index.ts',
	output: [{
		format: 'esm',
		file: pkg.module,
		sourcemap: false,
	}, {
		format: 'cjs',
		file: pkg.main,
		sourcemap: false,
	},
	{
		name: pkg['umd:name'] || pkg.name,
		format: 'umd',
		file: pkg.unpkg,
		sourcemap: false,
		plugins: [
			terser()
		],
	}],
	external: [
		...require('module').builtinModules,
		...Object.keys(pkg.dependencies || {}),
		...Object.keys(pkg.peerDependencies || {}),
	],
	plugins: [
		resolve(),
		typescript()
	]
}
```

The ts config (lib/tsconfig.json)

```typescript
{
  "compilerOptions": {
    /* Basic Options */
    "target": "es6",                                /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', 'ES2021', or 'ESNEXT'. */
[...]

PR #422, opened 2021-06-09
[v4-alpha] use mutable source

dai-shi:
originally from #157

PR #507, opened 2021-07-22
fix(build): Improved ESM support for most used bundlers with fallbacks for older bundlers

barelyhuman:
commit description
```
separates esm and cjs to different dist folders
forced types of the folder to support other rollup instances
the separation was made to support maximum bundlers , though
the milage may way vary based on the ESM implementation of your
bundler
```

### Why ?
There's a mismatch in implementation when it comes to esModuleInterop between the bundlers 
primarily rollup and webpack , thus causing issues when generating hybrid module libraries. 
The following commit is needed to address #454 where webpack fails to detect the esm files 
this is because webpack depends on the package to be strictly forced to be "type":"module" 
or have `.mjs` files.

### Changes 
- Modifies `rollup.config.js` to direct the built assets into different folders 
- Modifies `package.json#copy` script to handle the strict `"type":"module"` for the esm folder and `"type":"commonjs"` on the main package.json

**Additional Notes**
Please do not merge this unless both bundlers confirm the esm and cjs modules to be working properly, I'll be reporting the tests of both in a few minutes. 

Let me know if something needs to be changed.


barelyhuman:
**Queries**
1.  Do you want me to commit the modified snapshot.json? 
2. Does the snapshot need to be totally replaced since the mentioned `.js` no longer exist and are now `.cjs` and `.mjs` or do you want to keep the history?

dai-shi:
Hey, it's very nice to see you working on this. I have been struggle with this and it requires more knowledge with various bundlers.

- My big concern is if non-`.js` extension works for old bundlers, for example webpack 4.
- Yes, please recreate `.size-snapshot.json`. Delete it and run `yarn build -m 1`.  "-m 1" is important to ensure the order.
- If this is successful, we'd like to apply the same config to [jotai](https://github.com/pmndrs/jotai) and [valtio](https://github.com/pmndrs/valtio). Would you be helping them too?
- I wonder if this may solve this issue. https://github.com/pmndrs/valtio/issues/179

barelyhuman:
> My big concern is if non-.js extension works for old bundlers, for example webpack 4. 

I'll try setting up other bundlers as well, do we want to limit it to webpack 4 or are we planning to go lower?

> Yes, please recreate .size-snapshot.json. Delete it and run yarn build -m 1. "-m 1" is important to ensure the order

Will be done 👍

> If this is successful, we'd like to apply the same config to jotai and valtio. Would you be helping them too?

I wouldn't mind, also I'm looking up ways to test if the esm and cjs are working with the test suite itself, so I don't have to setup multiple repos to check if it's working. I can use https://npmjs.com/esm for the most part but I'll still research a bit more to make sure it's thorough. 

> I wonder if this may solve this issue. Regression with Snowpack from 1.0.6 -> 1.0.7 valtio#179

I've had issues with normal snowpack setups so I'm not sure if it would solve it but I'll try that out as well

dai-shi:
I think webpack 4 is fine without any reason. I'm not so familiar with other bundlers. rollup is already confirmed, right?

If you don't mind, would you please open draft PRs in jotai and valtio?

barelyhuman:
> I think webpack 4 is fine without any reason. I'm not so familiar with other bundlers. rollup is already confirmed, right?

Rollup seems to be working since I tested on the base version from that actual issue and then a custom rollup built library just exporting the `useStore` zustand created hook, so the basic implementation works for now. 

Webpack 4 is possibly already working since CRA and react scripts use webpack 4 and the codesandboxes for both the TS version and JS version worked with the demo code. Let me know if you want me to aggressively check with more complex state hooks. 

Webpack 5 is what needs to be tested, I have a nextjs project running that so I just have to install the new modified zustand bundle on that. 

Setting up templates for the other bundlers on codesandbox (will be doing this post work today)


> If you don't mind, would you please open draft PRs in jotai and valtio?

Sure, will take that up post work as well, or most probably tomorrow

dai-shi:
I assumed some traditional bundlers just ignores `"exports"` in package.json.
In this case, 
```js
import { devtools } from 'zustand/middleware' // or const { devtools } = require('zustand/middleware')
```
will pick `node_modules/zustand/middleware.js`, presumably?
Do you know if there aresuch bundlers?

So, the safe bet might be:
```
- index.js
- middlware.js
- cjs/index.cjs
- cjs/middlware.cjs
- esm/index.mjs
- esm/middlware.mjs
```

barelyhuman:
> I assumed some traditional bundlers just ignores `"exports"` in package.json.
> In this case,
> 
> ```js
> import { devtools } from 'zustand/middleware' // or const { devtools } = require('zustand/middleware')
> ```
> 
> will pick `node_modules/zustand/middleware.js`, presumably?
> Do you know if there aresuch bundlers?
> 
> So, the safe bet might be:
> 
> ```
> - index.js
> - middlware.js
> - cjs/index.cjs
> - cjs/middlware.cjs
> - esm/index.mjs
> - esm/middlware.mjs
> ```

I thought so too , though that shouldn't be an issue since the `main` field points to the cjs file and will be considered as a simple javascript file, this is basically my concern with browserify and maybe parcel but I'll have to test that out before I confirm it.

dai-shi:
> though that shouldn't be an issue since the main field points to the cjs file

Right, `main` field is very traditional and works fine for `from 'zustand'`. My concern is `from 'zustand/middleware'` and other `zustand/*`.

barelyhuman:
> Right, main field is very traditional and works fine for from 'zustand'. My concern is from 'zustand/middleware' and other zustand/*.

Oh yeah, that I'll have to check and basically what I meant by 
> this is basically my concern with browserify and maybe parcel but I'll have to test that out before I confirm it.
I'll be checking all imports and exports in them, or we can create 3 packages, cjs/esm and root all , as you suggested earlier

barelyhuman:
Report

---
`zustand`
| Bundler  | Status |
| ---------| -------|
| CRA / Webpack  | ✅  |
| Rollup  | ✅  |
| Snowpack  | ✅  |
| Browserify  | ✅  |

---

`vanilla`
| Bundler  | Status |
| ---------| -------|
| CRA / Webpack  | Pending|
| Rollup  | Pending|
| Snowpack  | Pending|
| Browserify  | Pending|

---

`context`
| Bundler  | Status |
| ---------| -------|
| CRA / Webpack  | Pending  |
| Rollup  | Pending  |
| Snowpack  | Pending  |
| Browserify  | Pending  |

---

`shallow`
| Bundler  | Status |
| ---------| -------|
| CRA / Webpack  | Pending  |
| Rollup  | Pending  |
| Snowpack  | Pending  |
| Browserify  | Pending  |


---

`middleware`
| Bundler  | Status |
| ---------| -------|
| CRA / Webpack  | Pending  |
| Rollup  | Pending  |
| Snowpack  | Pending  |
| Browserify  | Pending  |

barelyhuman:
@dai-shi While I knew the issue was the named export from the main `index.js` , I left it alone considering that it was made a named export because of a certain reason, but now I feel like that's the only solution out, I can add a babel plugin to add both

```js
exports.default = create
module.exports = create
``` 

to solve the webpack 4's compatibility, but before I do that, I'd like to know if you are fine with that approach. 

On the other hand the above `zustand` one worked in all cases because the `index.js` generated by the splits had a `exports default` line because of the typescript export. Not because I made the splits. 

Also, the only bundler respecting the `exports` right now is snowpack and esbuild, the others still depend on the main so we will have to ask them to use `zustand/esm/[module]` if their bundler doesn't automatically pick it up.

let me know, and I'll push the above babel addition and then update the tests accrodingly.

dai-shi:
> While I knew the issue was the named export from the main `index.js`

Do you mean the "default" export?

> ```js
> exports.default = create
> module.exports = create
> ```
> 
> to solve the webpack 4's compatibility, but before I do that, I'd like to know if you are fine with that approach.

Using a babel plugin sounds fine.

> On the other hand the above `zustand` one worked in all cases because the `index.js` generated by the splits had a `exports default` line because of the typescript export. Not because I made the splits.

I'm not sure if I understand this. because of the typescript export?

> Also, the only bundler respecting the `exports` right now is snowpack and esbuild, the others still depend on the main

webpack 5 doesn't respect "exports"?

I assume webpack 4 understands "module". So, `import create from 'zustand'` picks `zustand/esm/index.js`, and `import ... from 'zustand/middleware'` picks `zustand/middleware.js`. do you know if it's correct?

>  so we will have to ask them to use `zustand/esm/[module]` if their bundler doesn't automatically pick it up.

yeah, that makes sense.

barelyhuman:
> Using a babel plugin sounds fine

Experimenting with that right now and also a few other cases

> webpack 5 doesn't respect "exports"? 

Didn't get time to check so can't assure, but their docs say that they do. 

> so we will have to ask them to use zustand/esm/[module] if their bundler doesn't automatically pick it up.
> > yeah, that makes sense

This isn't working either in most cases so I'm still figuring out what can be done, can get the main index working but everything else basically fails.

---
I'm looking  for a graceful solution so it's easier to handle later on, so still experimenting, will update once I'm done with all the possible solutions

dai-shi:
> I'm looking for a graceful solution so it's easier to handle later on, so still experimenting, will update once I'm done with all the possible solutions

that'd be great. thanks for working on this.

barelyhuman:
> thanks for working on this

I use the library in almost every prototype project I build, I can obviously help with something so small 

BTW, 
if you get time , could you go through the below PR, it's a breaking change on the API so you might wanna review it before I raise the same here.

https://github.com/barelyhuman/zustand/pull/2

dai-shi:
> if you get time , could you go through the below PR, it's a breaking change on the API so you might wanna review it before I raise the same here.
> 
> [barelyhuman#2](https://github.com/barelyhuman/zustand/pull/2)

Oh, I don't think this breaking change is acceptable.
We can't drop default exports and adding named export in addition can still be controversial.
(I mean, I prefer named exports and jotai and valtio are so, but zustand prefers default exports for now.)

We might consider named exports in v4, #422.

barelyhuman:
Yeah, thought so. 

Yeah we can do that in the major update. 

As for now, I've got like one solution left, I'll try that out and then basically decide if we want to drop the issue altogether

Considering the same kind have been open on react and react-scripts for a while now.

dai-shi:
(I'm still not sure if I understand fully. I will wait for your suggestion for v3 after getting feedback in #454.)

barelyhuman:
I mean,

 v4 => Named exports, solve this issue altogether 

For now, v3 => I'm waiting on the feedback on the last build I made on my fork, if that works as expected, that'll be an addition to the current version

package.json
```diff
@@ -3,42 +3,44 @@
...
+  "main": "./index.js",
+  "module": "./esm/index.mjs",
+  "type": "commonjs",
+  "types": "./index.d.ts",
   "files": [
     "**"
   ],
   "exports": {
     "./package.json": "./package.json",
+    "./": "./",
```
dai-shi, line 15: just curious, which bundler or something requires this?

barelyhuman: It's a fallback for the default `zustand/<package>` cjs exports, in case the conditional exports are not working but normal single line exports are supported by the bundler. 

Not specific to any bundler

package.json
```diff
@@ -60,7 +62,7 @@
     "test": "jest",
     "test:dev": "jest --watch --no-coverage",
     "test:coverage:watch": "jest --watch",
-    "copy": "shx cp -r dist/src/* dist/esm && shx mv dist/src/* dist && shx rm -rf dist/{src,tests} && shx cp package.json readme.md LICENSE dist && json -I -f dist/package.json -e \"this.private=false; this.devDependencies=undefined; this.optionalDependencies=undefined; this.scripts=undefined; this.prettier=undefined; this.jest=undefined; this['lint-staged']=undefined;\""
+    "copy": "shx cp -r dist/src/* dist/esm && shx mv dist/src/* dist && shx rm -rf dist/{src,tests} && shx cp package.json readme.md LICENSE dist && json -I -f dist/package.json -e \"this.private=false; this.devDependencies=undefined; this.optionalDependencies=undefined; this.scripts=undefined; this.prettier=undefined; this.jest=undefined; this['lint-staged']=undefined;\" && shx cp package.json dist/esm && shx echo '{\"type\":\"module\"}' > dist/esm/package.json && json -I -f dist/esm/package.json"
```
dai-shi, line 65:
```suggestion
    "copy": "shx cp -r dist/src/* dist/esm && shx mv dist/src/* dist && shx rm -rf dist/{src,tests} && shx cp package.json readme.md LICENSE dist && json -I -f dist/package.json -e \"this.private=false; this.devDependencies=undefined; this.optionalDependencies=undefined; this.scripts=undefined; this.prettier=undefined; this.jest=undefined; this['lint-staged']=undefined;\" && shx cp package.json dist/esm && shx echo '{\"type\":\"module\"}' > dist/esm/package.json"
```
The last command doesn't do anything?

barelyhuman: It's supposed to format but yeah makes no difference since there's just one line. 
I'll make the change.

dai-shi:
So, is this ready for merge?

What should we do with jotai and valtio? They don't have any default exports.

barelyhuman:
I'll go through them once I'm done with Taco (my product) 
And check why the snowpack fails there. 

Yeah you can merge. I've tested, but if you'd like to take a go, you can go ahead and test all the sandboxes

package.json
```diff
@@ -3,42 +3,44 @@
   "private": true,
   "version": "3.5.7",
   "description": "🐻 Bear necessities for state management in React",
-  "main": "index.js",
-  "module": "esm/index.js",
-  "types": "index.d.ts",
+  "main": "./index.js",
+  "module": "./esm/index.mjs",
```
dai-shi, line 7: Do we need this line? Any bundlers make use of it?

package.json
dai-shi: Okay. I didn't know about a bundler that may understand "exports" but not condtional exports.

package.json
dai-shi: I see. It's to format. You can leave it, if you like.

package.json
barelyhuman: Webpack and rollup use it.
pure esm version will pickup the `exports` part , so you can say it's for `backward compat`

package.json
barelyhuman: Oh okay, not sending any more commits for this then

package.json
barelyhuman: Same, saw 
https://github.com/marvinhagemeister/smoldash/pull/10/commits/38ea2735df076e1c60c7e2f8ddacbf46f0ed30ed
do it in the morning and looked up other repos that were doing it and based on a few conversations here and there , people have had issues for libraries with multiple entry points

Ex: I have [one](https://github.com/barelyhuman/wrap) which doesn't fully support conditional exports

barelyhuman: Can remove it, since we are manually declaring every single entry point anyway.

dai-shi: If it helps, let's add it. (and make sure we add it to all zustand/jotai/valtio.)

package.json
dai-shi: Oh, yeah. and those bunders don't complain with `mjs` extension, right?

dai-shi (approved):
The change looks good.

If anyone is interested, please feel free to review and/or try codesandboxes.

package.json
barelyhuman: Webpack reads it as proper esm module but like in https://github.com/pmndrs/zustand/issues/454 if the source is incompatible with esm then the fix provided there will have to be added to webpack 4

And other bundlers that don't support esm or modules will not touch the `/esm` at all so it should be fine. 

Right now there's 3 ways to import 


```js

import {} from "zustand/middleware" // works on both esm and cjs based on the package.json / bundler
import {} from "zustand/esm/middleware" // force esm for certain bundlers (tried on parcel)
import {} from "zustand/esm/middleware.mjs" // shove the esm down the bundler if needed

```

And the `package.json` forces `type:module` in the later two cases so if the bundler doesn't like esm, then it's on the bundler not zustand as a package.

dai-shi:
Would you update the PR title to show the latest change. It will be listed in the release note.

barelyhuman:
The last commit was made to revert to the normal `.js` exports for `esm` and is the only change.

dai-shi:
please recreate size snapshot to see how it's changed.

barelyhuman:
already done in the previous commit

package.json
```diff
@@ -3,14 +3,15 @@
...
-  "types": "index.d.ts",
+  "main": "./index.js",
+  "module": "./esm/index.js",
+  "types": "./index.d.ts",
   "files": [
     "**"
   ],
   "exports": {
     "./package.json": "./package.json",
+    "./": "./",
```
dai-shi, line 14: https://nodejs.org/api/packages.html#packages_subpath_folder_mappings
says it's deprecated. @barelyhuman

barelyhuman: we'll add the glob pattern and test that out as well

dai-shi: Doesn't the glob pattern conflict with conditional exports? Yeah, testing would be good. I just wonder if there are any environments/versions that don't understand conditional exports but does understand glob patterns. `"./": "./"` was for fallback for old ones, right?

barelyhuman: > "./" was for fallback for old ones, right? 

based on reading and referencing a few repos on github, yes. 
We could actually remove it and just wait for someone to bring up the issue instead of over thinking it

dai-shi: >  just wait for someone to bring up the issue

Yeah, I prefer that tbh.

barelyhuman:
> jotai



> @barelyhuman Hi, this is ready for merge, correct?
> Would you please make the same fix for jotai?

Want me to make it *ready for review*?
I'll update jotai for the same

dai-shi:
> Want me to make it _ready for review_?

Did it now.

> I'll update jotai for the same

Thanks. Look forward to it.

dai-shi (approved):
Looks good. Looks safer. Thanks a lot for various investigations.

AnthonyDugarte:
Getting this warning after updating:

```
WARNING in shared module zustand -> /path/to/node_modules/zustand/esm/index.js
No version specified and unable to automatically determine one. No version in description file (usually package.json). Add version to description file, or manually specify version in shared config.
```

barelyhuman:
webpack config, a sandbox, something to help me reproduce it locally?

dai-shi:
Hmmmmmm, this sounds like it's safer to copy `esm/*.js` to `esm/*.mjs` instead of adding `esm/package.json`, I guess.

barelyhuman:
> Hmmmmmm, this sounds like it's safer to copy `esm/*.js` to `esm/*.mjs` instead of adding `esm/package.json`, I guess.

I tested it on nextjs with webpack 5 and it works, so we'll need a reproduction before we do that or we just came full circle

dai-shi:
> we'll need a reproduction before we do that or we just came full circle.

absolutely. 

Meanwhile, please open a draft pr in jotai repo. I want the three to keep in sync.

package.json
```diff
@@ -3,9 +3,9 @@
   "private": true,
   "version": "3.5.7",
   "description": "🐻 Bear necessities for state management in React",
-  "main": "index.js",
-  "module": "esm/index.js",
-  "types": "index.d.ts",
+  "main": "./index.js",
+  "module": "./esm/index.js",
+  "types": "./index.d.ts",
```
dai-shi, line 8: Let's make this change in https://github.com/pmndrs/jotai/pull/657 too.

barelyhuman: Done

MarkLyck:
@dai-shi This update completely broke my project.

https://github.com/pmndrs/zustand/issues/534

I'm on next.js with webpack 5

Issue #249, opened 2020-11-29
devtools: accessing serialize option


JacobJaffe:
Hello! I've just started using zustand, previously having used primarily redux. Big fan so far! 

In setting up the `devtools`, I'm having trouble tying to configure the devtools. Specifically, I'm trying to enable [the `serialize` option](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize), so that I can view a `Map` in the tools nicely.

Example of a store I'm trying to see with devtools:
```ts
const foo: Entity = {
  id: "abc",
  // ...
};

type State = {
  entities: Map<EntityId, Entity>;
};

const store = () => ({
  entities: new Map<EntityId, Entity>([[foo.id, foo]]),
});

const useStore = create<State>(devtools(store));
```

In a redux configuration I would do this with: 
```js
configureStore({
  // ...
  devTools: {
    serialize: {
      options: true
    }
  }
});
```

Is there a similar way to access the devtool options from `zustand/middleware.devtools` ?

----

For anyone also facing this, my current solution is an old hack from redux before the `devTools.serialize` option:
```ts
(Map.prototype as any).toJSON = function () {
  return JSON.parse(JSON.stringify([...this]));
};
```

AnaArsovska:
Hey there Jacob! Great question and even better code! Best of luck sir!

dai-shi:
Thanks for opening this up!
I don't think such an option is supported in the current zustand/middleware.devtools.

Anyone wants to investigate?

cmlarsen:
This is kinda gross and not at all throughly tested, but to display Maps/Sets I used `patch-package` to patch the `devtools` middleware with polyfill for Map and Set

```javascript
// zustand/middlware.js
var devtools = function devtools(fn, prefix) {
  if (process.env.NODE_ENV !== 'production') {
    if (!Map.prototype.toJSON) {
      Map.prototype.toJSON = function() {
        var obj = {};
        this.forEach((value, key) => obj[key] = value);
        return obj;
      }
    }
    if (!Set.prototype.toJSON) {
      Set.prototype.toJSON = function() {
        return Array.from(this);
      };
    }
  }
// ... the rest of the devtools middleware
}

```

dai-shi:
That's a nice hack.
Though, I guess we could simply enable the serialize option and let the work done in the DevTools Extension?

PR #540, opened 2021-08-19
Allow optional REDUX DEVTOOLS "options" object

marcoSven:
## Why is this change needed?
Unable to display Maps/Sets

## How does it address the issue?
optional serialize date, regex, undefined, nan, infinity,
error, symbol, map & set

## breaking change prev "prefix" option
prefix?: string -> options?: { name?: string }

https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize

https://github.com/pmndrs/zustand/issues/249

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -46,7 +46,24 @@ export type NamedSet<T extends State> = {
 export const devtools =
   <S extends State>(
     fn: (set: NamedSet<S>, get: GetState<S>, api: StoreApi<S>) => S,
-    prefix?: string
+    options?: {
+      name?: string
+      serialize?: {
+        options:
+          | boolean
+          | {
+              date?: boolean
+              regex?: boolean
+              undefined?: boolean
+              nan?: boolean
+              infinity?: boolean
+              error?: boolean
+              symbol?: boolean
+              map?: boolean
+              set?: boolean
+            }
+      }
+    }
   ) =>
   (
     set: SetState<S>,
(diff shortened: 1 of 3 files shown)


src/middleware.ts
```diff
@@ -46,7 +46,24 @@ export type NamedSet<T extends State> = {
 export const devtools =
   <S extends State>(
     fn: (set: NamedSet<S>, get: GetState<S>, api: StoreApi<S>) => S,
-    prefix?: string
+    options?: {
```
dai-shi, line 49:
```suggestion
    options?: string | {
```
How about keep supporting the previous behavior so that we can avoid breaking changes?

src/middleware.ts
```diff
@@ -91,8 +108,8 @@ export const devtools =
         savedSetState(state, replace)
         api.devtools.send(api.devtools.prefix + 'setState', api.getState())
       }
-      api.devtools = extension.connect({ name: prefix })
-      api.devtools.prefix = prefix ? `${prefix} > ` : ''
+      api.devtools = extension.connect({ ...options })
```
dai-shi, line 111:
```suggestion
      options = typeof options === 'string' ? { name: options } : options
      api.devtools = extension.connect({ ...options })
```

dai-shi (commented):
Looks like a nice addition! Added one request.

src/middleware.ts
marcoSven: Thank you, yes that is a great suggestion.

dai-shi (approved):
looks good. would you update the PR description?

dai-shi:
Forgot about readme. Thanks!

PR #546, opened 2021-08-27
chore: use rollup copy instead

barelyhuman:
Commit Message
```
added the rollup copy plugin to handle
typescript copy and duplication of .js
files for esm pregenerated files
```

Based on previous discussion about letting rollup copy built files 
instead of regenerating them with the bundler

EDIT:
Drafted for bundler tests

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,17 +1,8 @@
 {
   "index.js": {
-    "bundled": 3993,
-    "minified": 1614,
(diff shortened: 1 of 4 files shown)


dai-shi:
As we have #569 merged in, what's left is only copying d.ts files. Doing it in a npm script is simple enough compared to adding a new rollup plugin. So, let's drop this.

Issue #82, opened 2019-12-04
Initial state from a component?


joelmoss:
So, I need to pass props to the creation of my store, so I can use those props to derive the initial store state. Which means I have to create the store within the component itself, so I have access to its props. But that initialises a new store each time it's called.

Is there a way to create a store from within the component without it creating a new store each time? Or perhaps a better way to set initial state of the store from a component's props?

thx

drcmda:
```jsx
function A(props) {
  const [[useStore]] = useState(() => create(set => ({ a: props.a }))
```

i usually use a callback though, an init() function that i call after components have been created. here's a recent example: https://codesandbox.io/s/react-three-fiber-untitled-game-i2160 (down there where the canvas component it, needed the store to be global, but relies on the camera, which is created later).

tafelito:
@drcmda if you use a callback, your component still rendering twice. Once for the initial value and then after you call the init, right?

Btw, cool game!

drcmda:
yes, that would add one render.

tafelito:
creating the store inside the components state won't allow you to have an initial state. Once the component is unmounted the state is destroyed as does the store. So basically putting the store in state is the same as just using the state by itself

tafelito:
creating the store inside the components state won't allow you to have an initial state. Once the component is unmounted the state is destroyed as does the store. So basically putting the store in state is the same as just using the state by itself

baybara-pavel:
@joelmoss, you can do it in that way too
```js

const [useStore, api] = create(set => ({
  a: "dafaultState",
  set: fn => set(produce(fn))
}));

const ExampleComponent = ({ a }) => {
  const initialState = { a };
  const ref = useRef(initialState);
  const state = ref.current;
  const store = useStore();

  useEffect(() => {
    api.subscribe(
      state => {
        if (state) {
          ref.current = state;
        }
      },
      state => state
    );
  }, []);

  return (
    <div
      onClick={() =>
        store.set(state => {
          state.a = "nextState";
        })
      }
    >
      Test: {state.a}
    </div>
  );
};
```
https://codesandbox.io/s/elastic-sutherland-r43x5

dai-shi:
Look like resolved. Please file a new issue for further discussion.

PR #553, opened 2021-09-02
docs: document that createContext can be used to initialize a store…  

TkDodo:
…depending on props, and how to work with it in TypeScript

refs #552
refs #82


dai-shi (approved):
Looks great. Thanks!

PR #560, opened 2021-09-18
Fix: remove useless internal info in devtools

Aslemammad:
Related to [valtio#229](https://github.com/pmndrs/valtio/issues/229), as they share the same dev tools approach.


dai-shi (approved):
I don't fully understand how this works, but if it works, it seems good.

dai-shi:
Would anyone be willing to review and/or test?

PR #588, opened 2021-10-14
chore: export middleware type definitions

abraaoz:


dai-shi:
> I must use the PersistOptions type from zustand/middleware in order to create my own persist middleware

That sounds reasonable. Can you make the PR to only export `PersistOptions`?

btw, this should be possible too:
```ts
type PersistOptions = Parameters<typeof persist>[1]
```

dai-shi:
> Using Parameters<> looses the Generic in the type definition.

Great point. I missed that.

src/middleware.ts
```diff
@@ -191,7 +191,7 @@ export type StateStorage = {
   setItem: (name: string, value: string) => void | Promise<void>
 }
 type StorageValue<S> = { state: DeepPartial<S>; version?: number }
-type PersistOptions<S, PersistedState extends Partial<S> = Partial<S>> = {
+export type PersistOptions<S, PersistedState extends Partial<S> = Partial<S>> = {
```
dai-shi, line 194:
```suggestion
export type PersistOptions<
  S,
  PersistedState extends Partial<S> = Partial<S>
> = {
```
This should fix it?

PR #601, opened 2021-10-21
feat(types): better middleware support

dai-shi:


src/index.ts
```diff
@@ -21,19 +19,29 @@ const isSSR =
...
   <U>(selector: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
-  setState: SetState<T>
-  getState: GetState<T>
-  subscribe: Subscribe<T>
-  destroy: Destroy
-}
+  setState: CustomStoreApi['setState']
+  getState: CustomStoreApi['getState']
+  subscribe: CustomStoreApi['subscribe']
+  destroy: CustomStoreApi['destroy']
```
AnatoleLucet, line 31: If I'm not wrong, these four fields are already in `CustomStoreApi`, right?
Why not just remove these four lines?

PR #603, opened 2021-10-21
feat(middleware): subscribeWithSelector middleware

### LINKED ISSUE #475: feature request: option for useStore.subscribe fire once on initialisation
Currently `useStore.subscribe` only fires when the subscribed part of the store changes. It would be useful if there was an option to force the function to fire once upon initialisation. Maybe I'm missing something, here's an example:

```javascript
useEffect(() => {
    // I don't want to have to repeat myself up here just for initialisation
    const { position, rotation } = getSceneObject(props.id)(useStore.getState())
    target.current.position.set(position[0], position[1] - 3, position[2])
    plane.current.position.set(...position)
    plane.current.rotation.set(rotation[0], Math.PI, 0)

    return useStore.subscribe(
      ({ position, rotation }: SceneObject) => {
        target.current.position.set(position[0], position[1] - 3, position[2])
        plane.current.position.set(...position)
        plane.current.rotation.set(rotation[0], Math.PI, 0)
      },
      (state) => getSceneObject(props.id)(state)
    )
  }, [props.id])
````

### LINKED ISSUE #555: Missing type inference when calling store.subscribe with a selector
Type inference breaks when using `store.subscribe` with a selector.

```js
import createStore from "zustand/vanilla";

const store = createStore(() => ({
  hello: "world"
}));

store.subscribe((state) => {
  // 👍 state type is inferred here
  console.log(state.hello);
});

store.subscribe(
  (state) => {
    // 💥 state type "unknown" here
    console.log(state.hello);
  },
  ({ hello }) => ({ hello })
);
```

Here's a [Codesandbox](https://codesandbox.io/s/thirsty-kirch-yy5mh).

Is this a known limitation?

dai-shi:
deprecating the core subscribeWithSelector feature.
close #475 
close #555

### DIFF
--- a/tests/context.test.tsx
+++ b/tests/context.test.tsx
@@ -1,7 +1,8 @@
 import { Component as ClassComponent, useEffect, useState } from 'react'
 import { render } from '@testing-library/react'
-import create from 'zustand'
+import create, { StoreApi } from 'zustand'
 import createContext from 'zustand/context'
+import { subscribeWithSelector } from 'zustand/middleware'
 
 const consoleError = console.error
 afterEach(() => {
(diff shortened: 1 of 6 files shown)


tests/context.test.tsx
```diff
@@ -63,22 +64,32 @@ it('uses context store with selectors', async () => {
...
+        inc: () => set((state) => ({ count: state.count + 1 })),
+      }))
+    )
+
+  const useStoreApi = useStoreApiOrig as () => Omit<
+    StoreApi<CounterState>,
+    'subscribe'
+  > & {
+    subscribe: ReturnType<typeof createStore>['subscribe']
+  }
```
grikomsn, line 83: I assume this is a temporary solution to use `subscribe` on context stores, right? Or is there another way?

dai-shi: Good point. #606 for a better way.

Issue #513, opened 2021-07-28
Set Persist API storage `name` with variable


felixhaeberle:
Hey folks, thanks for providing this great module. 🤙

I was wondering on how to set the name of the persistent storage dynamically through a variable.

Something like:

```js
export const useActiveComposition = create<ActiveCompositionState>(persist(
  (set) => ({
    activeCompositionId: null,
    setActiveCompositionId: (composition: Composition) => set({ activeCompositionId: composition ? composition.id : null })
  }),
  {
    name: `melo-active-composition-${useProject.getState().project?.id}`,
    getStorage: () => localStorage,
  }
));
```

But unfortunately, this returns undefined because the hook gets initialized at a time where the value isn't ready. 

How would I approach this problem, and access another stores data or a variable in general to provide it here in a persistent store as a name? Can I pass a variable when calling the hook?

Thanks in advance! 🙏

AnatoleLucet:
There's probably a hacky way to do this, but I can't figure out how.
We've planned to add a way to edit the `persist` config (including the `name`) in #267. But we're kind of stuck with the types.

I'll leave this open in the hope that someone has a way to do this with the current api :crossed_fingers:

Sixzero:
I am also interested in such a solution.
Maybe we could run `create` in a function body where the necessary variables are available? Until that the store is just an empty skeleton with null store?

jonataswalker:
A solution would be fantastic, for not being able to set an unique key (think a user cart) we have to maintain a collection and loop to find the user staff.

AnatoleLucet:
It's now possible to set a dynamic name:
```js
const useStore = create(persist(...))

useStore.persist.setOptions({
  name: "dynamic-name"
})
// to rehydrate the zustand store using the new name
useStore.persist.rehydrate()
```
It'll be released in the next version.

estoldy:
> 

Can you provide an example on how to use inside a store?

estoldy:
> It's now possible to set a dynamic name:
> 
> ```js
> const useStore = create(persist(...))
> 
> useStore.persist.setOptions({
>   name: "dynamic-name"
> })
> // to rehydrate the zustand store using the new name
> useStore.persist.rehydrate()
> ```
> 
> It'll be released in the next version.

Could you provide an example on how to use inside a store?

mieradi:
@ekiwi111 This still feels a little hacky but here's how I am doing it. 

Context: 
- Using Next
- Project Id is a query param 
[...]

PR #624, opened 2021-11-01
feat(middleware): add persist api

### LINKED ISSUE #267: Persist middleware api
Follow up of https://github.com/pmndrs/zustand/pull/248#issuecomment-752130586. The idea is to make an api to interact with `persist`.

#### Change the options
```js
useMyStore.persist.setOptions({
  getStorage: () => sessionStorage,
})
```
This `setOptions` could be useful to, for example, choose a storage at runtime if the use check a box or something. It could also be used to change the name of the storage as needed here: https://github.com/pmndrs/zustand/pull/248#issuecomment-752130586

#### Clear storage data
```js
useMyStore.persist.clearStorage()
```

#### On demand rehydration (see #374)
```js
useMyStore.persist.rehydrate()
```

AnatoleLucet:
Closes #267, #513

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -286,6 +286,7 @@ type DeepPartial<T extends Object> = {
 export type StateStorage = {
   getItem: (name: string) => string | null | Promise<string | null>
   setItem: (name: string, value: string) => void | Promise<void>
+  // Note: This will be required in v4
   removeItem?: (name: string) => void | Promise<void>
 }
 type StorageValue<S> = { state: DeepPartial<S>; version?: number }
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -459,12 +459,12 @@ export const persist =
 
     if (!storage) {
       return config(
-        (...args) => {
+        ((...args) => {
           console.warn(
             `[zustand persist middleware] Unable to update item '${options.name}', the given storage is currently unavailable.`
           )
           set(...args)
-        },
+        }) as CustomSetState,
         get,
         api
       )
@@ -510,10 +510,10 @@ export const persist =
     }
 
     const configResult = config(
-      (...args) => {
+      ((...args) => {
         set(...args)
         void setItem()
-      },
+      }) as CustomSetState,
       get,
       api
     )
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -4,7 +4,6 @@ import {
   PartialState,
   SetState,
   State,
-  StateCreator,
   StateListener,
   StateSelector,
   StateSliceListener,
@@ -287,6 +286,7 @@ type DeepPartial<T extends Object> = {
 export type StateStorage = {
   getItem: (name: string) => string | null | Promise<string | null>
   setItem: (name: string, value: string) => void | Promise<void>
+  removeItem?: (name: string) => void | Promise<void>
 }
 type StorageValue<S> = { state: DeepPartial<S>; version?: number }
 export type PersistOptions<
@@ -360,6 +360,15 @@ export type PersistOptions<
    */
   merge?: (persistedState: any, currentState: S) => S
 }
+export type StoreApiWithPersist<S extends State> = StoreApi<S> & {
+  persist: {
+    setOptions: (options: Partial<PersistOptions<S>>) => void
+    clearStorage: () => void
+    rehydrate: () => Promise<void>
+    hasHydrated: () => boolean
+    onHydrate: (fn: (state: S) => void) => void
+  }
+}
 
 interface Thenable<Value> {
   then<V>(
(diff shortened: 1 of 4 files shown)


src/middleware.ts
```diff
@@ -287,6 +287,7 @@ type DeepPartial<T extends Object> = {
 export type StateStorage = {
   getItem: (name: string) => string | null | Promise<string | null>
   setItem: (name: string, value: string) => void | Promise<void>
+  removeItem: (name: string) => void | Promise<void>
```
AnatoleLucet, line 290: @dai-shi FYI for the changelog: `persist` now require a storage with a `removeItem` method. That's kind of a breaking change, especially if the user gives a custom storage.

dai-shi: Can we have a migration path? In v3, make it optional and, warn if it's not there. And in v4, make it required.

src/middleware.ts
```diff
@@ -360,6 +361,15 @@ export type PersistOptions<
    */
   merge?: (persistedState: any, currentState: S) => S
 }
+interface PersistApi<S extends State> extends StoreApi<S> {
```
dai-shi, line 364:
```suggestion
export type StoreApiWithPersist<S extends State> = StoreApi<S> & {
```

Let's make it consistent with others. (tbh, I'm not sure whether type alias or interface is better in this case, or if it's just a preference thing. maybe, I can open up another PR to make everything type alias.)

src/middleware.ts
```diff
@@ -401,37 +411,37 @@ const toThenable =
...
-      serialize = JSON.stringify as (state: StorageValue<S>) => string,
-      deserialize = JSON.parse as (str: string) => StorageValue<Partial<S>>,
-      blacklist,
-      whitelist,
-      partialize = (state: S) => state,
-      onRehydrateStorage,
-      version = 0,
-      migrate,
-      merge = (persistedState: any, currentState: S) => ({
+  <S extends State>(config: StateCreator<S>, _options: PersistOptions<S>) =>
```
dai-shi, line 414: Not 100% sure about the use case, but we want `CustomSetState`, `CustomGetState`, `CustomStoreApi`, like some other middleware.

dai-shi, line 414: `_options` implies unused variable in TypeScript.
Would you name it `baseOptions` here, or name `mergedOptions` in L416?

src/middleware.ts
```diff
@@ -401,37 +411,37 @@ const toThenable =
...
+  (set: SetState<S>, get: GetState<S>, api: PersistApi<S>): S => {
+    let options = {
```
dai-shi, line 416:
```suggestion
    let options: PersistOptions<S> = {
```
just an extra type checking.

src/middleware.ts
```diff
@@ -449,24 +459,24 @@ export const persist =
...
 
     const setItem = (): Thenable<void> => {
-      const state = partialize({ ...get() })
+      const state = options.partialize({ ...get() })
 
-      if (whitelist) {
+      if (options.whitelist) {
         ;(Object.keys(state) as (keyof S)[]).forEach((key) => {
-          !whitelist.includes(key) && delete state[key]
+          !options.whitelist?.includes(key) && delete state[key]
```
dai-shi, line 469: do we need `?`?

tests/middlewareTypes.test.tsx
```diff
@@ -174,8 +174,8 @@ describe('counter state spec (single middleware)', () => {
   })
 
   it('persist', () => {
-    const useStore = create<CounterState>(
-      persist(
+    const useStore = create(
+      persist<CounterState>(
```
dai-shi, line 178: We would like to keep type annotation in `create`.

dai-shi (changes requested):
It looks overall good. Some minor things and typing issues need to be addressed.

src/middleware.ts
dai-shi: I hope this helps tests/middlewareTypes.test.tsx.

src/middleware.ts
AnatoleLucet: `PersistOptions` keeps optional fields as `<T> | undefined` even if they have a default value. So we need to do `storage = options?.getStorage()` for instance because typescript think `getStorage` is possibly `undefined` even if it has a default value.
Making this change would require to add unneeded conditions just to satisfy typescript.

src/middleware.ts
AnatoleLucet: For some reason, typescript think `options.whitelist` can be `undefined` within the forEach

src/middleware.ts
dai-shi: I didn't notice that. Then, my suggestion is void.

src/middleware.ts
dai-shi: Okay, then let's leave it. Will be removed in #598 anyway.

AnatoleLucet:
Within the tests, [the immer middleware](https://github.com/pmndrs/zustand/blob/d063bf065b38fac31b2a1a0375c9bb693181543c/tests/middlewareTypes.test.tsx#L20) keep the `api` as `StoreApi`. Shall I change this middleware using `CustomStoreApi`?
https://github.com/pmndrs/zustand/blob/d063bf065b38fac31b2a1a0375c9bb693181543c/tests/middlewareTypes.test.tsx#L477

There's still another minor type issue within the middleware with the `create` function, but I'm unable to find how to solve it. Could you check it out?

dai-shi:
> Shall I change this middleware using `CustomStoreApi`?

Yeah, I can't think of other ways.
This may require users to update their immer middleware.

> There's still another minor type issue within the middleware with the `create` function, but I'm unable to find how to solve it. Could you check it out?

Hm, I don't know either. Let's add type assertion.

dai-shi (approved):
The code looks good.

Issue #631, opened 2021-11-03
[Question] [Typescript] StoreApiWithPersist with union type trying to update to 3.6.3


jafin:
I'm attempting to update some code to 3.6.3 (from 3.6).

Some state properties I have are union types:  `number | null`.
Using the test case as an example, I cannot get this to validate, any advice on how to get this compiling?

```ts
type CounterState = {
    count: number | null;
};

const useStore = create<
   CounterState, 
   SetState<CounterState>, 
   GetState<CounterState>, 
   StoreApiWithPersist<CounterState>>(
    persist(
        (set, get) => ({
            count: 1,
        }),
        { name: 'prefix' },
    ),
);
```

The error received on persist:

Argument of type '(set: SetState<CounterState>, get: GetState<CounterState>, api: StoreApiWithPersist<{ count: number; inc: () => void; }>) => { ...; }' is not assignable to parameter of type 'StoreApiWithPersist<CounterState> | StateCreator<CounterState, SetState<CounterState>, GetState<CounterState>, StoreApiWithPersist<...>>'.
  Type '(set: SetState<CounterState>, get: GetState<CounterState>, api: StoreApiWithPersist<{ count: number; inc: () => void; }>) => { ...; }' is not assignable to type 'StateCreator<CounterState, SetState<CounterState>, GetState<CounterState>, StoreApiWithPersist<CounterState>>'.
    Types of parameters 'api' and 'api' are incompatible.
      Type 'StoreApiWithPersist<CounterState>' is not assignable to type 'StoreApiWithPersist<{ count: number; inc: () => void; }>'.
        Type 'StoreApiWithPersist<CounterState>' is not assignable to type 'StoreApi<{ count: number; inc: () => void; }>'.
          Types of property 'setState' are incompatible.
            Type 'SetState<CounterState>' is not assignable to type 'SetState<{ count: number; inc: () => void; }>'.
              Type 'number | null' is not assignable to type 'number'.
                Type 'null' is not assignable to type 'number'.ts(2345)

Changing  `StoreApiWithPersist<CounterState>`  to `StoreApiWithPersist<any>` bypasses the warning

dai-shi:
We have a discussion recently in #619.
This is very unfortunate, and a workaround I found so far is `count: 1 as number | null`.

I wonder if there is any TS expert out there who has other suggestions.

jafin:
Thank you, that's fixed it, if I cast the whole state object it also corrects it (or so it appears)

```ts
[...]

PR #633, opened 2021-11-03
fix(tests): add #632 example and improve middleware types tests

### LINKED ISSUE #632: [TypeScript] v3.6.3 multiple middlewares breaks type check on stores with objects
As per the test example here: https://github.com/pmndrs/zustand/blob/9f47fdc9276c96db99b70ee763e3990b2ec533a7/tests/middlewareTypes.test.tsx#L440-L476

If one of the items stored is an object, type checking reports an error. (Sorry this must be the 5th question/issue about TS being a pain!)

**Minimal example:**

```typescript
type ObjectState = {
  count: 0,
  obj: {
    prop: string;
  };
  inc: () => void;
};

const useStore = create<
  ObjectState,
  SetState<ObjectState>,
  GetState<ObjectState>,
  StoreApiWithSubscribeWithSelector<ObjectState> & StoreApiWithPersist<ObjectState>
>(
  subscribeWithSelector(
    persist(
      (set, get) => ({
        count: 0,
        inc: () => set({ count: get().count + 1 }, false),
      }),
      { name: "count" }
    )
  ),
  { name: "prefix" }
);
```

dai-shi:
close #632
related #631 
related #619

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -609,6 +609,8 @@ const useStore = create(
   ),
 )
 ```
+
+Typing with multiple middleware is a bit of pain. Refer some working examples in [middlewareTypes.test.tsx](./tests/middlewareTypes.test.tsx).
   
 ## Best practices
   
--- a/readme.md
+++ b/readme.md
@@ -318,26 +318,15 @@ const useStore = createStore(set => ({
 
 export default useStore
 ```
+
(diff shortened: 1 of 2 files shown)


istvanp:
Perhaps cast the entire `{ foo: true }` back to `MyState`. The former example works easily for one data item, but anything more complex it's easier to just use the already declared type.

PR #634, opened 2021-11-03
feat(middleware): devtools to patch api.setState with namedSet and fix middleware types

### LINKED ISSUE #517: [Question] Name actions differently in devtools with setState api returned from create
Instead of creating individual fn's to update slices of store, I want to do it with a single function, but I want to have custom action names for each call to `setState`. Something like below.

```
store.setState(callback, false, 'Action1');
store.setState(callback, false, 'Action2');
```
Is there a way I can do this currently? What I notice is `create()` returns `setState(fn: any, replace?: boolean)` only.

I am using vanilla store.

dai-shi:
close #517

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -206,7 +206,7 @@ export const devtools =
 
 export type StoreApiWithSubscribeWithSelector<T extends State> = Omit<
   StoreApi<T>,
-  'subscribe'
+  'subscribe' // FIXME remove omit in v4
 > & {
   subscribe: {
     (listener: StateListener<T>): () => void
@@ -233,7 +233,7 @@ export const subscribeWithSelector =
   (
     set: CustomSetState,
     get: CustomGetState,
-    api: Omit<CustomStoreApi, 'subscribe'> &
+    api: Omit<CustomStoreApi, 'subscribe'> & // FIXME remove omit in v4
       StoreApiWithSubscribeWithSelector<S>
   ): S => {
     const origSubscribe = api.subscribe as Subscribe<S>
@@ -255,7 +255,11 @@ export const subscribeWithSelector =
       }
       return origSubscribe(listener)
     }) as any
-    const initialState = fn(set, get, api as CustomStoreApi)
+    const initialState = fn(
+      set,
+      get,
+      api as CustomStoreApi // FIXME can remove in v4?
+    )
     return initialState
   }
 
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -63,11 +63,10 @@ export type NamedSet<T extends State> = {
   ): void
 }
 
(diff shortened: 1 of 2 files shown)


dai-shi:
[typescript playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGxoF4oG8BQVtQMYC4oA7AVwFsAjCAJwwF8NRIoBBMASylTQY1wHsiAZ2BQIRABbwiuCCy5QAPAGESI-mTacIADyREAJkNYcAfAAosOddQjLbiftULn4HQqvWaOASi6nYBCQMHxc3dg81YA0tKAAydCh4QhFqdiIAcyg6P2QAuEQUAMwcKFtgEmoiKFdff3QrUuxwgDp4BQByCQgAGx7+Dsam8srqmzsHaOpa9h8hhl4BYVFxKRkIACEFFSiYjjE9cSMTdgsh8fsIR2ca8MivLVz8oIgQsPcoT2jvTgS0KAohAo-H4PSu1Ry9QKSGweQapRGVVudThJSazQ4LQoCmA1BIr3R2ERYymEyuUxmc1KCwwfEEIjwkxQSi+e20h0Mxi0Z1KF0mTneEU+ux+T0ChTewoe+zh5h4SWM0hAPgA3LSlgyhABGBS4JnmcxiuV4QgABmyPjmGtEQgATLr9atpLIWAajf8CFBzTlLRg7W06csoEIAMwO8kQcxO9YbN31Y2e72WuahrGBzUAFnDhSjkmdm1zaxdcdlHrNFuTfozAaE1YotKAA)

I guess I understand what was wrong.

barelyhuman:
doesn't this end up adding the initial concern that we had about the function knowing something extra compared to others?

dai-shi:
Do you mean people can add custom store type without using its middleware?
Yes, that's correct. But, I guess it's what #617 did.

At least, if you use `devtools` and you forgot to annotate custom store type, you will get an error. So, I guess it's technically better.
And, if we use `devtools & combine`, we don't need to annotate types, which was really my goal for this.

Most importantly, vanilla.ts doesn't know anything about middleware.

barelyhuman:
The last point makes the change a valid decision. Still a bit on the complicated side since we've got a lot on inner extends that might get hard to control or understand over time. 
Otherwise, looks good.

AnatoleLucet (approved):

dai-shi:
> Still a bit on the complicated side

I don't disagree, it's complicated.
From my perspective, when I did #601, I wasn't comfortable very much and felt something wrong. With some feedback and discussions, I revisited in #617, and it's much cleaner than before, and easier to educate users (it's verbose but straightforward). Now, I have a bit clearer understanding since then, and this PR improves some types. With this change, I feel this is a good balance between the library maintainability and good developer experience.

PR #550, opened 2021-08-31
feat: use-sync-external-store

dai-shi:
## No public API changes

The API is compatible with the latest v3 API

## Major changes

- React part is fully re-implemented with [use-sync-external-store](https://www.npmjs.com/package/use-sync-external-store)

## Better support for context usage

- It exports a new `useStore` hook.
- It re-exports `createStore` from `zustand/vanilla`
- `zustand/context` is fully rewritten with the new `useStore`.

PR #635, opened 2021-11-03
Export created Context value

timoxley:
This enables `zustand` context to be used in apis that expect a `Context` value, e.g. `React.useContext` or `useContextBridge` from `drei`.

Without the export it appears impossible to use `zustand/context` with `useContextBridge`.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -99,6 +99,7 @@ function createContext<
     Provider,
     useStore,
     useStoreApi,
+    Context: ZustandContext,
   }
 }
 


dai-shi:
Hi, I know where you are coming from.
However, not exposing the context value and limiting the usage is the whole point of `zustand/context`. Here's the long history: #182
 
At this point, if you need to bridge the context value, you want to use `React.createContext` and just make sure not to break the rules of hooks (= for example, not change the context value once created.)
This will be improved in v4 with #550.

Closing this, but feel free to have questions about the usage.

Issue #638, opened 2021-11-05
Persist Middleware Typescript Error


andreechristian:
Hello,

I am getting this error bellow when setting my typescript tsconfig strict mode to `true`.

`Argument of type '(set: SetState<{ session: SessionData; user: UserData | null; }>, get: GetState<{ session: SessionData; user: UserData | null; }>, api: StoreApi<...> & { ...; }) => { ...; }' is not assignable to parameter of type 'StoreApi<{ session: SessionData; user: UserData | null; }> | StateCreator<{ session: SessionData; user: UserData | null; }, SetState<{ session: SessionData; user: UserData | null; }>, GetState<...>, StoreApi<...>>'.
  Type '(set: SetState<{ session: SessionData; user: UserData | null; }>, get: GetState<{ session: SessionData; user: UserData | null; }>, api: StoreApi<...> & { ...; }) => { ...; }' is not assignable to type 'StateCreator<{ session: SessionData; user: UserData | null; }, SetState<{ session: SessionData; user: UserData | null; }>, GetState<{ session: SessionData; user: UserData | null; }>, StoreApi<...>>'.
    Types of parameters 'api' and 'api' are incompatible.
      Type 'StoreApi<{ session: SessionData; user: UserData | null; }>' is not assignable to type 'StoreApi<{ session: {}; user: null; }> & { persist: { setOptions: (options: Partial<PersistOptions<{ session: {}; user: null; }, Partial<{ session: {}; user: null; }>>>) => void; clearStorage: () => void; rehydrate: () => Promise<...>; hasHydrated: () => boolean; onHydrate: (fn: (state: { ...; }) => void) => void; }...'.
        Type 'StoreApi<{ session: SessionData; user: UserData | null; }>' is not assignable to type 'StoreApi<{ session: {}; user: null; }>'.
          Types of property 'setState' are incompatible.
            Type 'SetState<{ session: SessionData; user: UserData | null; }>' is not assignable to type 'SetState<{ session: {}; user: null; }>'.
              Type '{ session: SessionData; user: UserData | null; }' is not assignable to type '{ session: {}; user: null; }'.ts(2345)`

I have created a minimum repo [here](https://codesandbox.io/s/typescript-forked-xijc8?file=/src/index.ts).

Thank you in advance.

andreechristian:
Submitted fix for this https://github.com/pmndrs/zustand/pull/639

dai-shi:
We had a similar discussion in #632.
[...]

PR #639, opened 2021-11-05
Fix typescript error, incompatibility with strict mode set to true

andreechristian:
To fix issues described here https://github.com/pmndrs/zustand/issues/638

I am just setting the `persist` option to optional (adding `?`) as seen on line 371 in `src/middleware.ts`
Also to make that possible, we need to exclude `undefined` from typing at line 461

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -368,7 +368,7 @@ export type PersistOptions<
   merge?: (persistedState: any, currentState: S) => S
 }
 export type StoreApiWithPersist<S extends State> = StoreApi<S> & {
-  persist: {
+  persist?: {
     setOptions: (options: Partial<PersistOptions<S>>) => void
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks for the suggestion. We actually want to type `persist` as the required property.
Let's look for another workaround for #638.

This made me realize that there are some persist api tests missing in middlewareTypes.test.tsx.
I will add those later. Thanks again.

andreechristian:
I see, then I guess you would also want to update the documentation? Since it's stated there that it will require nothing to implement the persist middleware. @dai-shi

dai-shi:
Closing in favor of #641.
While I agree there's room for improvement for docs in general (especially typescript usage and persist middleware), #638 is a back compatibility issue we should solve if possible. #641 should do it.

PR #643, opened 2021-11-05
fix(docs): typing with subscribeWithSelector middleware

### LINKED ISSUE #642: [Bug] New subscribe documentation doesn't work
Hi Zustand

Thanks for an awesome library!

I'm running into an issue with the new subscribeWithSelector in the new zustand version

**Zustand Version**
3.6.4

**Problem**
According to your documentation i should be able to use subscribe with the new subscribeWithSelector, but I keep getting the error that I should use the middleware.

My Store:
`export type TControlsStore = {
  publicBargainingAgreement: string;
};`

`const store = (set: NamedSet<TControlsStore>, get: GetState<TControlsStore>) => ({
  publicBargainingAgreement: '',
});`

`const useControlsStore = create<TControlsStore>(devtools(subscribeWithSelector(store)));`

My subscription:
`useControlsStore.subscribe(state => state, console.log("my log"))`

**Steps to reproduce**
Install Zustand 3.6.4 and add subscribe according to the documentation and see that you get the error.

image
image

-----

Am I missing something from the documentation? To me it looks exactly the same and I can seem to find where the error is?

dai-shi:
close #642

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -213,6 +213,30 @@ const unsub4 = useStore.subscribe(state => [state.paw, state.fur], console.log, 
 const unsub5 = useStore.subscribe(state => state.paw, console.log, { fireImmediately: true })
 ```
 
+<details>
+<summary>How to type store with `subscribeWithSelector` in TypeScript</summary>
+
+```ts
+import create, { GetState, SetState } from 'zustand'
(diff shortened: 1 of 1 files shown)


dai-shi:
@iamkimchico can you review this please?

iamkimchico (approved):
Looks good!

Issue #644, opened 2021-11-05
Persisting one of my stores is inconstant on android


haibert:
Hi Guys,

I have fully migrated my app over from redux to zustand and the experience has been wonderful, so much less code.. Thank you for all the hard work. 

I am persisting one of my stores responsible for retaining pictures that failed to upload. I keep the base64 of the picture taken from the camera in an array and present it to the user in a FailedUploads cue screen where they can decide to upload or discard  the photos. This works great in iOS but on android I am seeing inconsistency when reloading the app. Meaning some times it loads the persisted data, sometimes it loads some of it, and sometimes it removes all the persisted data.

I was able to stop the inconsistency by removing the base64 of the photo temporarily and just persisting the photoID and date stamp, so I can say that loading the large base64 could be the issue, or possibly the fact that there is no waiting for store rehydration in my app.js? In Redux for example there is a persist gate that keeps the splash screen showing until the store is rehydrated. But I am not able to find any instructions on that in your documentation for this, if you have such functionality even? Any guidance would be greatly appreciated.

AnatoleLucet:
Hi, you can check if hydration is done using `useStore.persist.hasHydrated()`if that can help (not a reactive value, it's just a regular getter).

Is this related to #562?

haibert:
Thank you for this tip. I was able to resolve the issue by not storing the base64 of the images, as long as its just text being saved it seems to work fine and that is good enough for now. 

can you tell me if useStore.persist.hasHydrated() returns anything ? is it a async function that gives a true or false when the hydration is finished ? If so I can use it to keep the splash screen open until hydration is complete and that would be great

AnatoleLucet:
`useStore.persist.hasHydrated()` returns true if hydration is done, and false if it isn't.

You could reactively check for hydration with the following:
```ts
const [isHydrated, setIsHydrated] = useState(false);

useEffect(() => {
  if (useStore.persist.hasHydrated()) {
    setIsHydrated(true)
  } else {
    useStore.persist.onHydrate(() => {
      setIsHydrated(true)
    })
  }
}, [])
```
`useStore.persist.onHydrate` will be called after hydration. Though this makes me realize `onHydrate` is probably not the most accurate name :thinking:

haibert:
[...]

PR #646, opened 2021-11-07
feat(middleware): add onFinishHydration to persist api

AnatoleLucet:
I noticed in #644 that I kind of misnamed `useStore.persist.onHydrate` :grimacing: 
By the name `onHydrate` we would most likely expect the callback to be called when hydration starts, while in reality it's called when hydration is finished.
I fixed this by actually calling `onHydrate`'s callback when hydration starts, and adding a new `onFinishHydration` method to the persist api.

This pr also adds a new `useStore.persist.useHydration()` hook to make it easier to (mainly?) know if hydration is finished.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,4 +1,3 @@
-import { useEffect, useState } from 'react'
 import {
   EqualityChecker,
   GetState,
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react'
 import {
   EqualityChecker,
   GetState,
@@ -367,13 +368,15 @@ export type PersistOptions<
    */
   merge?: (persistedState: any, currentState: S) => S
 }
+type PersistListener<S> = (state: S) => void
 export type StoreApiWithPersist<S extends State> = StoreApi<S> & {
   persist: {
     setOptions: (options: Partial<PersistOptions<S>>) => void
     clearStorage: () => void
     rehydrate: () => Promise<void>
     hasHydrated: () => boolean
-    onHydrate: (fn: (state: S) => void) => void
+    onHydrate: (fn: PersistListener<S>) => () => void
+    onFinishHydration: (fn: PersistListener<S>) => () => void
   }
 }
 
@@ -457,9 +460,8 @@ export const persist =
     }
 
     let hasHydrated = false
-    const onHydrateCallbacks: Parameters<
-      StoreApiWithPersist<S>['persist']['onHydrate']
-    >[0][] = []
+    const hydrationListeners = new Set<PersistListener<S>>()
+    const finishHydrationListeners = new Set<PersistListener<S>>()
     let storage: StateStorage | undefined
 
     try {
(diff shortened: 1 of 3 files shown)


src/middleware.ts
```diff
@@ -605,6 +614,28 @@ export const persist =
       onHydrate: (cb) => {
         onHydrateCallbacks.push(cb)
       },
+      onFinishHydration: (cb) => {
+        onFinishHydrationCallbacks.push(cb)
+      },
+      useHydration: () => {
```
dai-shi, line 620: I don't think it's a good idea for `zustand/middleware` to depend on `react`. All middlewares including persist should be able to be used in vanilla.

So, I would suggest to not have it in persist api, but in docs as a reference implementation.
Also, if we use a hook, useEffect should be able to unsubscribe the callbacks.

```js
const useStore = create(persist(...))
const useHydration = () => {
  const [hydrated, setHydrated] = useState(useStore.persist.hasHydrated)
  useEffect(() => {
    const unsub1 = useStore.persist.onHydrate(() => setHydrated(false)) // not sure this is required in practice
    const unsub2 = useStore.persist.onFinishHydration(() => setHydrated(true))
    return () => {
      unsub1()
      unsub2()
    }
  }
  return hydrated
}
```

dai-shi (changes requested):
hope it makes sense.

src/middleware.ts
dai-shi: another option is to provide an api so that people can create a hook more easily.
```js
const useHydration = () => {
  const [hydrated, setHydrated] = useState(useStore.persist.hasHydrated)
  useEffect(() => useStore.persist.onChangeHydrated(setHydrated), [])
  return hydrated
}
```

dai-shi: On second thought, why would we need another hook while we already have a store hook.
So, I'd expect this is better practice.
```js
const useStore = create(persist(...))
useStore.persist.onFinishHydration(() => useStore.setState({ hydrated: true }))
```
This may not work if the hydration is sync? Maybe better to have a persist option, instead of using persist api?

Hm, we already have postRehydrationCallback. So:
```js
const useStore = create(persist(
  (set) => ...,
  {
    name: 'name',
    onRehydrateStorage: () => (state) => {
      if (state) { useStore.setState({ hydrated: true }) }
    },
  }
))
```
It looks tricky, though. Oh, and we need to exclude it when serializing... 😕

AnatoleLucet: Surely that does make sens. I wasn't too sure about this and was looking forward to your feedbacks :smile: 

I think we should go with your first example (fully custom `useHydration` hook referenced in the doc). I'll remove the hook and add an unsub function.

> ```useStore.persist.onHydrate(() => setHydrated(false)) // not sure this is required in practice```

FYI this is just in case you want to take into account manual hydrations using `useStore.persist.rehydrate()`

dai-shi: Sounds good.

src/middleware.ts
```diff
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react'
```
dai-shi, line 1:
```suggestion
```

tests/persistSync.test.tsx
```diff
@@ -1,3 +1,4 @@
+import { render } from '@testing-library/react'
```
dai-shi, line 1:
```suggestion
```

dai-shi (approved):
LGTM

PR #660, opened 2021-11-15
wip: phantom api type

dai-shi:

PR #662, opened 2021-11-16
Rewrite with better types

devanshj:
[(Background)](https://twitter.com/devanshj__/status/1460036889990205444) I want to know if yall are interested in me doing a rewrite, it'll improve the user-facing types as well as internal types & code. If yes, I'll first do a free rewrite -- ie without caring about backwards-compatibility, standards, opinions, naming, etc -- then once that's done we'll worry about the other things. You can keep as much code as you want.

Till now I've rewritten `zustand/vanilla` (feel free to ask questions, I'll document some things later). As I continue to rewrite I'll understand what the runtime behavior is (because I haven't used zustand) and hence I'll type accordingly, so it won't be an incremental change, I might change the existing types if required.

Bug fixes till now:
1. [Bug](https://tsplay.dev/w146kW) ([as seen on twitter](https://twitter.com/dai_shi/status/1460023055447592964)). [Fixed](https://tsplay.dev/wOzJlW).
2. [Bug](https://tsplay.dev/w2E41W) ([as seen on twitter](https://twitter.com/dai_shi/status/1460229296178405383)). [Fixed](https://tsplay.dev/NBejdw).

New features till now: (well they are kinda bugs :P)
1. If `replace` flag is set, you'll have to provide the state type not the partial of state type. [Before](https://tsplay.dev/mbGA2W). [After](https://tsplay.dev/mAvj1W).

Progress:
- [x] zustand/vanilla
- [x] zustand/react
- [x] zustand/shallow
- [x] zustand/context
- [x] zustand/middlewares/
  - [x] zustand/middlewares/persist
  - [x] zustand/middlewares/combine
  - [x] zustand/middlewares/subscribeWithSelector
  - [x] zustand/middlewares/devtools
  - [x] zustand/middlewares/redux

Miscellaneous: 
The original version uses some trick (which I've not read yet) to disallow `{ count: undefined }` (but allow `{}`). The new types would disallow `{ count: undefined }` only when `--exactOptionalPropertyTypes` flag is set. I might revert back to original, or use some other trick.

cc @dai-shi


dai-shi:
Yes, interested!

We are planning v4 major some time soon, so having breaking change in types is possible. But, we would like to have a migration path as much as possible. v3.x should have compatible types with `@deprecated` and once it's migrated, safe to upgrade to v4.

We have some confusions around the current hacky `SetState` type, and improving it is welcomed. I hope your rewrite goes well without changing tests, or minimal changes that can easily be reviewed.

#660 is my incomplete. You don't use `unique symbol` in this PR, any drawbacks?

dai-shi:
> If replace flag is set, you'll have to provide the state type not the partial of state type. 

Amazing! 😮 

> The new types would disallow { count: undefined } only when --exactOptionalPropertyTypes flag is set.

I think this sounds reasonable, as long as it won't break with old TS.

Oh, that reminds me, what would be the minimal requirement of TS version for this change? We would like to support CRA. (Can we install new TS with CRA?)
([jotai](https://github.com/pmndrs/jotai) and [valtio](https://github.com/pmndrs/valtio) use [downlevel-dts](https://github.com/sandersn/downlevel-dts).)

devanshj:
> We are planning v4 major some time soon, so having breaking change in types is possible. But, we would like to have a migration path as much as possible. v3.x should have compatible types with `@deprecated` and once it's migrated, safe to upgrade to v4.
> 
> We have some confusions around the current hacky `SetState` type, and improving it is welcomed. I hope your rewrite goes well without changing tests, or minimal changes that can easily be reviewed.

Let's keep this PR for v4 (which will have some breaking changes) and then once this PR is ready I'll make another PR for v3.x which will be fully backward-compatible with all tests passing. Although it's possible that we can't have backward-compatibility for the newer fixes and features in that case we'll have to remove them and v3.x might have no improvements but only `@deprecated` documented for a easier migration to v4.

The breaking changes mostly would be regarding less types being exported, maybe renaming some types and cases where generics are explicitly passed. We'll discuss later how much to break. The migration would be pretty straightforward.

> You don't use unique symbol in this PR

[I do](https://github.com/devanshj/zustand/blob/b84255ff29e111f0d1dd11d3e2cec1da4e38a36a/src/vanilla.ts#L48)

> #660 is my incomplete trial

Let's put that to rest :P

> Amazing! 😮

Haha thanks ;P

> I think this sounds reasonable, as long as it won't break with old TS.

It won't. Although, given very few people have `exactOptionalPropertyTypes` on (nor is it on when doing `tsc --init`), I might keep the existing trick as it's useful. I'll think about it later.

> what would be the minimal requirement of TS version for this change? We would like to support CRA.

I don't think I'll require any fancy newer TS features, I'll have to check what minimum version is required but I think CRA will already fulfil the minimum requirement.

dai-shi:
> Let's keep this PR for v4 (which will have some breaking changes)

I agree. Note that v4 will have other changes, like index.ts -> react.ts.
See: https://github.com/pmndrs/zustand/milestone/1

> Although it's possible that we can't have backward-compatibility for the newer fixes and features in that case we'll have to remove them and v3.x might have no improvements but only `@deprecated` documented for a easier migration to v4.

Makes sense.

> [I do](https://github.com/devanshj/zustand/blob/b84255ff29e111f0d1dd11d3e2cec1da4e38a36a/src/vanilla.ts#L48)

Oh, yeah. I looked at these: https://github.com/devanshj/zustand/blob/b84255ff29e111f0d1dd11d3e2cec1da4e38a36a/src/vanilla.ts#L57-L59
What's the intuition behind these don't need to be unique symbol.

> Let's put that to rest :P

Yeah. If you glance it, you'd see my idea:
```ts
create<MyState>(...) // is for backward compatibility
create<MyState, CustomStore>(...) // is for custom store like for `api.persist`
create(...) // infer both types, works well with `combine()`
```
So, basically I gave up manual state typing and custom store type inference. It is impossible with this PR, isn't it?

Looking forward to this work!

devanshj:
> Note that v4 will have other changes, like index.ts -> react.ts

Should we keep all those changes in this PR? I think #550 should have it's separate PR and once you've merged that, I'll update this branch and rewrite that with new types. The other deprecated removal PRs seem trivial and can be superseded by this PR. Does that sound good? I'm fine with whatever haha.

> What's the intuition behind these don't need to be unique symbol.

Okay so there are two parts to this:
1. **Avoiding working with generics**
    Normally if you were to implement this you'll be working with generics like `T`, `Nt`, `U` in this case and it's harder to work in various ways, examples...
    a. [Subtyping errors in relation to `T` and `Nt`][1]. Although these errors are "correct", they are kinda false positive in our case. 
    b. [Type-parameters are "lossy"][2]
2. **Having stricter types to catch mistakes**
   Technically speaking the type of listener is `<T>(state: T, previousState: Previous<T>) => void` where `Previous` is a kind of newtype (as in Haskell) `Previous<T> = T & { __isPrevious: true }`. We don't want to allow passing the current state in place of previous state, we must statically enforce that. And that's what is already done in this PR, for example if you'll emit the current state instead previous, it won't compile:
   
    Image showing error on emit(state, state)
    

But given the TypeScript users at large aren't so particular at types, don't want to introduce a newtype or other internal stricter types that we might use hence we write separate types for implementation (and separate types for user-facing). And while writing separate types for implementation we make every generic an "existential type" (much like in Haskell) so we can avoid the pitfalls of generics. So `T` becomes `type EState = { __isState: true }`, `U` becomes `type ESelectedState = { __isSelectedSelected: true }`, `Store<T>` becomes `EStore` etc. And they don't have to be unique symbols as we don't export them in the first place.

Had I not written them like this there would be many more assertions (which are bad because you can accidently make a wrong assertion) and less mistakes being caught.

There are also other advantages like you can no longer pass hardcode values, for example you can't call `store.set` with `replace` as `true` because you'll get an error "`true` is not assignable to `EShouldReplaceState`". Had it been only `boolean` or a generic like `R extends boolean`, you can pass hardcode values.

This is actually how things should be typed but the lack of these features without hacks in TypeScript and 99% of TypeScript users not having a very thorough types background, nobody except me writes code like this :P So the "as well as internal types & code" is also true ;)

> ```
> create<MyState>(...) // is for backward compatibility
> ``` 

I did skim read your PR but this won't entirely be backward compatible as `create<T>(persist(...)).persist` won't compile. And I'm yet to see what is `combine` haha.

> It is impossible with this PR, isn't it?

Nope, not possible in general. Although we can export a helper which creates the creator, something like `createWithState<{ count: number }>()(set => ({ count: 0 }))` so the user has to only explicitly type the state and not the store api. And the inference of `state` is probably not possible afaict, although I'll try again at the end.
    
[1]: https://www.typescriptlang.org/play?jsx=0#code/MYewdgzgLgBAZiEMC8MA8AVANDAcrAUwA8oCwATCGABQEMAnKAS1oBtMA+DgCigC4Y2eAO4BKFBzxRxySQG8AUDBisCsIihhQA3EphM4MbhuSn4Y8XIC+eg0ZAAjAFYFgUAJIRjOOBcs2bBVBIWEcXN09NAHlnVygAOiYqWipuTm5aASEHLJlJB0RVWjBRBSA

[2]: https://www.typescriptlang.org/play?jsx=0#code/C4TwDgpgBAygrgIwM4GMBOBLBEA8AVAPigF4oBvKAKCigAoAzALjuGbwEoSiA3AewwAm7Zn0HUoOAKoEGzWqyiTOxHvwEAaKEggAbCCmC80chRy6LhUUQMoBfSpRS8AdkmBR6vXiQmF5bZSIycT13JERUTGwfWgA6eIBDZgAxWIB5ACZYgEE0AHM4AFsIZ2AkHHhkdCxcQgJA8nEaDHo6BNi9ZzzgAAsSYlIMzmCaUahQjx8EgG0ABgBdJtH6WgAiAANJdahsFATiqHW4ZwBrZ14Ad2d11fYl+xp7e0pnfYgkMASUaGTGmggAB5gIzuV7FD5faCZP6jQHAtDuUCQKC5ArFUrlQgkJY0PBQQHAEoCJAwsY0OKJZgYZz0CBoFEARksxzOl2cVDJowpsSSUGptPp2SGzBZ5yuHLJD05UAA-IyoAAfFEZHGjZjOCDcOniJ5AA

dai-shi:
Thanks. (I'm still trying to have better understanding in my head.)

The thing is we need to release v3.7.0 before v4.0.0. So until then, the main branch is v3.x. I can create v4 branch meanwhile with all changes applied, so that you can work on it.

The custom store type is something I introduced in v3.6.0, so `create<MyState>` backward compatibility means for pre v3.6.0.

dai-shi:
https://github.com/pmndrs/zustand/tree/v4 is pushed.
Let me also refactor `index.ts` -> `react.ts` in v3 which should mitigate resolving conflicts.

dai-shi:
#663

devanshj:
> I'm still trying to have better understanding in my head.

Haha take your time. 

> https://github.com/pmndrs/zustand/tree/v4 is pushed.

I'll rebase to this soon.

dai-shi:
If you want to split src/middleware.ts into src/middleware/*.ts, we can do that in v3 without changing the code like #663.

src/middleware/persist.ts
```diff
@@ -0,0 +1,353 @@
...
+interface PersistentStorageValue<T>
+  { state: T
+  , version?: number
+  }
+
+interface PersistStore<T extends UnknownState, U>
+  { persist:
+      { setOptions: (options: O.Partial<PersistOptions<T, U>>) => void
+      , clearStorage: () => void
+      , rehydrate: () => Thenable<void>
```
devanshj, line 98: This previously was `rehydrate: () => Promise<void>` but we were lying in the types and it's actually is `rehydrate: () => Thenable<void>`.
So previously someone writing `rehydrate().finally()` would have compiled and it would have given a runtime error, but now it won't.

src/middleware/persist.ts
```diff
@@ -0,0 +1,353 @@
...
+     */
+    deserialize?:
+      (serializedString: string) => MaybePromise<PersistentStorageValue<T>>
+
+    /**
+     * Filter the persisted value.
+     *
+     * @param state The state's value
+     */
+    partialize?: (state: T) => U
```
devanshj, line 51: Previously this was `(state: T) => DeepPartial<T>` but I don't understand why was it restricted to `DeepPartial<T>`? `partialize` acts like a `selector`/`mapper` the user should be able to map to any shape they wish.

devanshj:
> If you want to split src/middleware.ts into src/middleware/*.ts, we can do that in v3 without changing the code like #663.

I'm not sure what you mean by "without changing" the code, I hope you're aware I'm also rewriting the code and you are also interested in that haha. I'm fine if you only want type changes too.

And I'll reexport all middlewares from `src/middleware/index.ts` so there will be no breaking changes

src/middleware/persist.ts
```diff
@@ -0,0 +1,353 @@
...
+  let hydrationEmitter = emitter(new Set<(state: EState) => void>())
+  let finishHydrationEmitter = emitter(new Set<(state: EState) => void>())
+
+  const hydrate = () => {
+    hasHydrated = false;
+    hydrationEmitter.emit(parentGet());
+    let postRehydrationCallback = options.onRehydrateStorage?.(parentGet()) || undefined
+
+    return thenablify(persistentStorageGetItem)()
+    .then(serialized => serialized !== null ? options.deserialize(serialized) : null)
```
devanshj, line 230: Previously we were checking if `serialized` is falsey instead of null, which is less correct as `serialized` could even be `""`

src/middleware/persist.ts
```diff
@@ -0,0 +1,353 @@
...
+    return stateInitializer(
+      (...a) => {
+        console.warn(messages.unableToUpdateItem(options.name))
+        parentSet(...a);
+      },
+      parentGet, 
+      update(parentStore, "setState", setState => (...a) => {
+        console.warn(messages.unableToUpdateItem(options.name))
+        setState(...a);
+      })
```
devanshj, line 194: Previously this mutation was not made, I think it's good to be consistent and warn on both `setState`.

Also nowhere we're assuming that `set === store.set`, because they could be different functions (say while testing etc), so changes to `set` happen at two places.

dai-shi:
> I'm not sure what you mean by "without changing" the code

It's not your work. I mean to split files in advance. This allows to see the difference more clearly in your PR. See #663.

> And I'll reexport all middlewares from src/middleware/index.ts

I'll re-report from `src/middleware.ts`. Just preference.
I'll let you know when `v4` branch is ready.

devanshj:
> I mean to split files in advance.

Ah gotcha even I was thinking about how can we make view the diff easier, I'll do that, reorder the commits and force push.

> I'll re-report from src/middleware.ts. Just preference.

Yeah no problems.

dai-shi:
> Ah gotcha even I was thinking about how can we make view the diff easier, I'll do that, reorder the commits and force push.

No, I mean #664. I will merge that into main (v3).

devanshj:
> No, I mean #664. I will merge that into main (v3).

Ahhh so now I'll have to only rebase my branch the (newer) main? Cool let me know when you merge

dai-shi:
> Ahhh so now I'll have to only rebase my branch the (newer) main? Cool let me know when you merge

Still no. I want you to rebase this branch from (newer) v4. Hold on a minute.

devanshj:
> Still no. I want you to rebase this branch from (newer) v4. Hold on a minute.

Haha cool cool.

dai-shi:
`v4` branch is ready. Please change the base branch of this PR.

devanshj:
I still will have to remove some commits ig where I did the partial split in this PR itself

devanshj:
Okay looks like I got it right

src/shallow.ts
```diff
@@ -1,26 +1,37 @@
-export default function shallow<T, U>(objA: T, objB: U) {
-  if (Object.is(objA, objB)) {
-    return true
+const shallow = <A, B>(_a: A, _b: B) => {
```
devanshj, line 1: I'm not sure how shallow equals is required here, shouldn't `shallow({ a: undefined }, {})` return `true`? The previous version would return `false` (afaict) but this newer version would return `true`

dai-shi: good question. I believe it's a subtle change, and if we clearly write it in the release note, we can change the behavior.

devanshj: Yeah, I think the newer behavior is better.

src/middleware/devtools.ts
```diff
@@ -1,167 +1,325 @@
...
-          }
-        }
-      }
-      options = typeof options === 'string' ? { name: options } : options
-      const connection = (api.devtools = extension.connect({ ...options }))
-      connection.prefix = options?.name ? `${options.name} > ` : ''
-      connection.subscribe((message: any) => {
-        if (message.type === 'ACTION' && message.payload) {
-          try {
-            api.setState(JSON.parse(message.payload))
```
devanshj, line 116: Here `JSON.parse(message.payload)` would give the action to be dispatched (afaict) and not the state, so this would be buggy, fixed in rewrite

src/middleware/devtools.ts
```diff
@@ -1,167 +1,325 @@
...
+          case 'ROLLBACK': return parseJsonThen(message.state, store.devtools.init);
+          case 'JUMP_TO_STATE':
+          case 'JUMP_TO_ACTION':
+            return parseJsonThen(message.state, silentlySetState);
+          case 'IMPORT_STATE':
+            let { nextLiftedState } = message.payload;
+            let lastComputedState = nextLiftedState.computedStates.at(-1)?.state
+            if (!lastComputedState) return;
+            silentlySetState(lastComputedState);
+            store.devtools.send(null, nextLiftedState);
```
devanshj, line 215: Previously we did something else, which I'm not sure is entirely correct, so this new thing is from [here](https://github.com/zalmoxisus/mobx-remotedev/blob/2d9ea51d3378e0507cfe8295046491e9f25c4a77/src/monitorActions.js#L77-L80).

src/middleware/devtools.ts
```diff
@@ -1,167 +1,325 @@
...
+        if (!isReduxLike(store)) return;
+        if (typeof message.payload !== "string") return;
+        return parseJsonThen(message.payload, store.dispatch);
+
+      case 'DISPATCH':
+        if (!store.devtools) return;
+        switch (message.payload.type) {
+          case 'RESET': return store.devtools.init(initialState);
+          case 'COMMIT': return store.devtools.init(store.getState());
+          case 'ROLLBACK': return parseJsonThen(message.state, store.devtools.init);
```
devanshj, line 206: Now we're also supporting `RESET`, `COMMIT` and `ROLLBACK`

src/middleware/devtools.ts
```diff
@@ -1,167 +1,325 @@
...
+
+  let initialState =
+    storeInitializer(store.setState, parentGet, store);
+
+  store.devtools.subscribe(message => {
+    switch (message.type) {
+      case 'ACTION':
+        if (!isReduxLike(store)) return;
+        if (typeof message.payload !== "string") return;
+        return parseJsonThen(message.payload, store.dispatch);
```
devanshj, line 199: This was previously done in the redux middleware...
https://github.com/pmndrs/zustand/blob/c8ae965435beb2cfb5a8bb8106e4e81712e9de25/src/middleware/redux.ts#L31-L33
So now I moved it here in it's right place.

Also just checking for `dispatch` property is unsafe because the user could have defined their own `dispatch` that is supposed to do something else, so only when the `isReduxLike` flag is set (which the `redux` middleware does) we assume `dispatch` is actual action dispatcher.

src/middleware/combine.ts
```diff
@@ -1,25 +1,60 @@
...
+type Combine =
+  <T extends UnknownState, U extends UnknownState>
+    ( initialState: T
+    , additionalStateCreator:
+        ( set: Store<O.Overwrite<T, U>>['setState']
+        , get: Store<O.Overwrite<T, U>>['getState']
+        , store: Store<O.Overwrite<T, U>>
+        ) => U
+    ) =>
+      StoreInitializer<O.Overwrite<T, U>, Store<O.Overwrite<T, U>>>
```
devanshj, line 15: Previously the combine types were something like this...
```ts
type Combine = 
  <T, U>
    ( controlled: T
    , createController:
        ( set: (state: Partial<T>, replace?: boolean) => void
        , get: () => T
        )
          => U
    ) =>
        ( set: (state: Partial<Overwrite<T, U>, replace?: boolean) => T
        , get: () => Overwrite<T, U>
        )
          => Overwrite<T, U>

type Overwrite<T, U> = Omit<T, keyof U> & U
```

Now these types are lying, the actual types are these...
```diff
  type Combine = 
    <T, U>
      ( controlled: T
      , createController:
-         ( set: (state: Partial<T>, replace?: boolean) => void
+         ( set: (state: Partial<Overwrite<T, U>>, replace?: boolean) => void
-         , get: () => T
+         , get: () => Overwrite<T, U>
          )
            => U
      ) =>
          ( set: (state: Partial<Overwrite<T, U>, replace?: boolean) => T
          , get: () => Overwrite<T, U>
          )
            => Overwrite<T, U>

type Overwrite<T, U> = Omit<T, keyof U> & U
```
...how do I know this? Because that is what the runtime behavior is, `get` would return `Overwrite<T, U>` and not `T` and it matters because I can create many runtime bugs that would compile. Examples...

```ts
let store = create(combine(
  {
    foo: 1,
    bar: 2
  },
  (set, get) => ({
    getSum: () => values(get()).reduce((a, b) => a + b, 0)
  })
))

const values =
  Object.values as (<T>(t: T) => T[keyof T][])

// technically this type of `Object.values` is incorrect
// but still usually people do it when they know what they're doing
// so one could write this thinking get only returns `T`

console.log(store.getState().getSum())
```

Now this compiles because `get()` says it'll return `{ a, b }` but during runtime it'll return `{ a, b, getSum }` and it'll [log `3() => values(get()).reduce((a, b) => a + b, 0)` instead of `3`](https://stackblitz.com/edit/typescript-mvcews?file=tsconfig.json).

This is just one example, it in general sets up a wrong precedence that `set` and `get` only operate with `T` whereas they operate with the whole store ie `Overwrite<T, U>`. Another example is doing `set(foo, true)` thinking it'll only replace the "controlled" `T` state whereas in reality it'll replace the whole state. I can make many more scenarios but you get the point.

So rewrite makes sure we're not lying in the types.

src/middleware/subscribeWithSelector.ts
```diff
@@ -1,62 +1,120 @@
...
+    if (fireImmediately) emit()
+  
+    return parentSubscribe(() => {
+      let nextSelected = selector(parentGet())
+      if (equals(currentSelected, nextSelected)) return
+  
+      previousSelected = E.previous(currentSelected)
+      currentSelected = nextSelected
+      emit()
+    })
```
devanshj, line 83: Previously it was...
```ts
let unsubscribe = parentSubscribe(() => {
  ...
})

if (fireImmediately) emit()

return unsubscribe;
```
And now it's...
```ts
if (fireImmediately) emit()

return parentSubscribe(() => {
  ...
})
```
...so that it looks a little nicer haha, but now it changes the order of events, which I hope is okay because I don't think we guarantee that we'll only emit after subscribing.

src/middleware/devtools.ts
dai-shi: I think it was intentional. We put `state` in the action object. cc @Aslemammad 
We could change the behavior, but I want to have it in a different PR (possibly in advance.)

src/middleware/devtools.ts
dai-shi: This should also be in a separate PR, if it fixes a behavior.
Otherwise, we will not be able to review the PR.

src/middleware/devtools.ts
dai-shi: Is this just refactoring or changing the behavior?

src/middleware/combine.ts
dai-shi: Fixing types without changing the behavior is nice.

src/middleware/subscribeWithSelector.ts
dai-shi: I would prefer the previous one. For now, the behavior is okay, but if something is changed in the future, firing after subscribing seems safer. Anyway, it's the pattern I use everywhere, so keep it consistent.

dai-shi:
I'm not reviewing code while you are free rewriting. Seems like it will require more effort after that.

src/middleware/devtools.ts
devanshj: Ah yes you're write we're sending the state as action:
https://github.com/pmndrs/zustand/blob/c8ae965435beb2cfb5a8bb8106e4e81712e9de25/src/middleware/devtools.ts#L106

In that case I'll keep the behavior same at least in this PR and open a another PR for changing the behavior (will reread the existing code to see if what I'm thinking is possible)

devanshj:
> I'm not reviewing code while you are free rewriting. Seems like it will require more effort after that.

Yeah that's okay let me complete and then once the tests are passing you can later do a final review.

I'll undo the devtools bit and open a seperate PR for that.

src/middleware/devtools.ts
devanshj: Given that I realised we were sending state as the action, this might be changing the behavior, will revert.

src/middleware/subscribeWithSelector.ts
devanshj: Cool will keep it as previous

src/middleware/devtools.ts
devanshj: Actually no, this does not change the behavior afaict, [`send` has signature `(action, state) => void`](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Methods.md#returns), so we're sending `prefix + 'setState'` as the action and not the state itself. But still I'll open a seperate PR for refactoring this.

src/context.tsx
```diff
@@ -0,0 +1,105 @@
...
+    (...a: never[]) => unknown
+
+  export type Call<T extends F.Unknown> =
+    T extends (...a: never[]) => infer R ? R : never
+}
+
+// ============================================================================
+// Exports
+
+export default createContext;
```
dai-shi, line 105: If you rename files in this PR, I can't review the diff easily in github. Would you create a refactoring PR for v3, that addresses only changing file names. Also, not related to this, but please try to minimize the diff in github as much as possible.

src/vanilla.ts
```diff
@@ -1,153 +1,179 @@
...
+
+
+
+
+// ============================================================================
+// Utilities
+
+const objectIs = Object.is as (<T>(a: T, b: T) => boolean)
+
+namespace E {
```
dai-shi, line 138: I was planning to request some small things later, but maybe it's a good timing. For a start, can we avoid using namespace? We strongly prefer explicitly named type aliases.

src/context.tsx
devanshj: Let me use `React.createElement` like before so that I can keep it as `.ts` as it was. And I'll try to minimize diffs.

src/vanilla.ts
```diff
@@ -1,153 +1,179 @@
...
+
+  const store: EStore = {
+    getState: () => state,
+    setState: (nextStateOrUpdater, shouldReplace) => {
+      const nextState =
+        typeof nextStateOrUpdater === 'function'
+          ? nextStateOrUpdater(state)
+          : nextStateOrUpdater
+
+      if (objectIs(nextState, state)) return
```
dai-shi, line 110:
```suggestion
      if (nextState === state) return
```
We don't allow primitive values, so triple equals should just work? If so, we prefer keeping the current logic.

src/vanilla.ts
devanshj: Sure no problems, I'll avoid namespaces.

I usually use them to create inlined modules that work on a specific abstraction or type, you can read more [here](https://twitter.com/devanshj__/status/1427800465169731585). But there's not much utility in this scenario anyway.

src/vanilla.ts
devanshj: There were some typing issues with `===` but now I rechecked and they're apparently gone so I'll revert to `===`.

dai-shi:
can you rebase with the new v4 branch? sorry for the force commits. (it will be so in the future too, until we finally reach v4.)
also, please run prettier. (it might be unfortunate for you, but can't be helped.)

devanshj:
> can you rebase with the new v4 branch? sorry for the force commits. (it will be so in the future too, until we finally reach v4.)

Done

> also, please run prettier. (it might be unfortunate for you, but can't be helped.)

I'll run it soon

devanshj:
Btw is the lint script not windows-friendly? I get this error..
```
PS E:\oss\zustand> yarn eslint:ci
yarn run v1.22.5
$ eslint '*.{js,json}' '{src,tests}/**/*.{ts,tsx}'

Oops! Something went wrong! :(

ESLint: 8.4.1

No files matching the pattern "'*.{js,json}'" were found.
Please check for typing mistakes in the pattern.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I tried removing the `'*.{js,json}'` part and now I get this...
```
PS E:\oss\zustand> yarn eslint:ci
yarn run v1.22.5
$ eslint '{src,tests}/**/*.{ts,tsx}'

Oops! Something went wrong! :(

ESLint: 8.4.1

ESLint couldn't find the config "react-app" to extend from. Please check that the name of the config is correct.

The config "react-app" was referenced from the config file in "E:\oss\zustand\examples\package.json".

If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Are there any workarounds to run eslint locally on windows?

dai-shi:
sorry, i didn't know it doesn't work on windows.
it complains for `'*.{js,json}'`, but not for `'{src,tests}/**/*.{ts,tsx}'`?

> ESLint couldn't find the config "react-app" to extend from.

probably, you might need to run `yarn install` in examples.

src/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from './vanilla'
 export * from './react'
-export { default } from './react'
+export { default, createWithState } from './react'
```
dai-shi, line 3: what's this `createWithState`?

src/middleware.ts
```diff
@@ -3,3 +3,4 @@ export * from './middleware/devtools'
 export * from './middleware/subscribeWithSelector'
 export * from './middleware/combine'
 export * from './middleware/persist'
+export * from './middleware/immer'
```
dai-shi, line 6: This is controversial. I have been hesitating to provide this, but maybe people want it... (hm, I think we already discussed about this somewhere?)

src/react.ts
```diff
@@ -1,81 +1,95 @@
...
+type UseStore = <S extends Store<UnknownState>, U = State<S>>(
+  store: S,
+  selector?: (state: State<S>) => U,
+  equals?: (a: U, b: U) => boolean
+) => U
+
+type Create = {
+  <
+    T extends UnknownState,
+    Mos extends [StoreMutatorIdentifier, unknown][] = []
```
dai-shi, line 23: What's Mos mean?

dai-shi (commented):
Can't read everything. Please keep more backward compatibility in terms of features/naming and also code (basically try to make github diffs smaller, i know it feels unnecessary task for you, but it's necessary to review. Internal refactoring can be separated).

tests/basic.test.tsx
```diff
@@ -7,16 +7,16 @@ import {
 } from 'react'
 import { act, fireEvent, render, waitFor } from '@testing-library/react'
 import ReactDOM from 'react-dom'
-import create, { EqualityChecker, SetState, StateSelector } from 'zustand'
+import { createWithState, Store } from 'zustand'
```
dai-shi, line 10: Please don't change test files, unless we really need to fix them.

src/index.ts
devanshj: If you want to explicitly type the state, you also will have to explicitly type the stores too. But with `createWithState` you can explicitly type the state without having to explicitly type the stores.

```ts
// https://tsplay.dev/NlxqlN

// doesn't compile
create<{ count: number }>(
  withA("a",
    withB("b",
      storeIn => ({ count: 0 })
    )
  )
)

// compiles
create<{ count: number }, [[$$withA, string], [$$withB, string]]>(
  withA("a",
    withB("b",
      storeIn => ({ count: 0 })
    )
  )
)

// compiles
createWithState<{ count: number }>()(
  withA("a",
    withB("b",
      storeIn => ({ count: 0 })
    )
  )
)
```

This would be the recommended way to create stores users. I did mention it in #710 in the others section but didn't explain it.

src/middleware.ts
devanshj: I think some people already copy paste immer middleware from the tests, it'd be hard from them to migrate or even author the middleware themselves so I think this is a fair addition.

src/react.ts
devanshj: `MutationOuts`. The mutations that the initializer "outputs".

`M` is `Mutation`, `Mo` in `MutationOut`, `Mi` in `MutationIn`

tests/basic.test.tsx
devanshj: These are breaking changes. Users will have to migrate from `create` to `createWithState`, or explicitly type mutations in `create` as I mentioned [here](https://github.com/pmndrs/zustand/pull/662#discussion_r774789318). And also as per #715 (which is not finalized yet) we'd be removing those types.

devanshj:
> it complains for `'*.{js,json}'`, but not for `'{src,tests}/**/*.{ts,tsx}'`?

Yep, it seems somehow the command terminates at space and `'{src,tests}/**/*.{ts,tsx}'` is never passed as the argument.

> probably, you might need to run yarn install in examples.

Ah that might do the trick.

dai-shi:
```json
  "eslint:ci": "eslint *.{js,json} src/**/*.ts tests/**/*.tsx",
```
Does this work?

src/middleware.ts
dai-shi: yeah, it's convincing...

dai-shi:
Closing in favor of #725.

PR #664, opened 2021-11-16
refactor(middleware): split into multiple files

dai-shi:
Preparation for #662


src/middleware/combine.ts
```diff
@@ -0,0 +1,25 @@
...
+import { NamedSet } from './devtools'
+
+type Combine<T, U> = Omit<T, keyof U> & U
+
+export const combine =
+  <PrimaryState extends State, SecondaryState extends State>(
+    initialState: PrimaryState,
+    create: (
+      // Note: NamedSet added for convenience
+      set: SetState<PrimaryState> & NamedSet<PrimaryState>,
```
dai-shi, line 11: @devanshj #662 can fix this too??

devanshj: Could you make a typescript playground to demonstrate the current issue and what needs to be fixed?

dai-shi: Well, would you try it in the main branch? If you remove `& NamedSet<PrimaryState>` in this line, a type error shows in `middlewareTypes.test.tsx`.
I didn't know how to solve it before, but I guess the `TagStore` might do.

devanshj: I'm yet to look at combine but I think all problems should get resolved by the rewrite types, I'm done with persist so let me do combine next.

Issue #674, opened 2021-11-22
[middleware/devtools] api.setState doesn't support the third argument


dai-shi:
https://twitter.com/dai_shi/status/1462689869411454979
`useStore.setState` isn't `namedSet`. The type lies (because I didn't know it's not implemented. 😓 )

https://codesandbox.io/s/suspicious-sky-7t3jc?file=/src/App.tsx

<img width="502" alt="image" src="https://user-images.githubusercontent.com/490574/142823594-1258918c-5e03-4698-94a4-d90933b774fd.png">

Having both `action` and `setState` is weird too. So, it's not really implemented correctly.

dai-shi:
@devanshj Are you working on `devtools`? Or, should I just draft a small PR to only fix this issue?

dai-shi:
https://github.com/pmndrs/zustand/blob/833f57ed131e94f3ed48627d4cfbf09cb9c7df03/src/middleware/devtools.ts#L89
I don't know why I did this. This causes double actions.

devanshj:
I'm working on `devtools`, this will get fixed afaict, I'll open a PR in sometime. If it's urgent feel free to open a PR and fix it.

PR #675, opened 2021-11-23
rewrite `devtools`

devanshj:
I'd have documented the behaviour but the code and tests truly speak for themselves ;) How the devtools work can be understood from the code and tests itself, there is no way to verify them because there is no official documentation. I understood how they work by the [mobx intergration](https://github.com/zalmoxisus/mobx-remotedev) and trying a few things in the devtools myself to see what `message` I receive.

Feel free to ask questions.

I've also made some deprecations, take them as suggestions, they are:
1. **No more prefixing:** If you check the git history we used to prefix names/actions because we didn't used to pass the name to `connect` like so `connect({ name })` hence prefixing was needed because all zustand stores would show up in the same store instance. But later we started passing the `name` and that creates separate instances (which means separate action log and state) for each zustand store. But whoever made the change didn't realise that now we no longer need the prefixing, so hence we're removing it now.
Also now given that we `dispatch` action received from the devtools (which we didn't used to do before), we'd have to unprefix them and that would have gotten messy.
Even though we no longer prefix actions, `api.devtool.prefix` would still return an empty string for backward-compatibility. But the user can no longer set it, if they do it'll be ignored, this could be a breaking change but a minor rare one I guess so it's okay make the change in a non-major version. We also a log a deprecation warning if the user gets or sets `prefix` property.

2. **Making the `api.devtools` private:** There is no reason to let the user have access to the devtools instance. Afaict we used to make it a property on `api` so that the `redux` middleware can `api.devtools.send(action, state)` but now the `redux` does `set(state, false, action)` which in turn does the same thing. So there is no reason to expose the devtools instance, which can make things malfuction if the user also starts interacting with it along with us.
For backward-compatibility, in this version we still keep the behavior same and don't make devtools private, we just log a deprecation warning if they get or set it.

Bug fixes:
- If the user dispatches `{ type: "INCREMENT" }` action from the devtools, instead of ignoring it (or dispatching if it's a redux store) we used to update the state with it, so the new state would become `{ ...state, type: "INCREMENT" }` this is fixed. More on the bug [here](https://github.com/pmndrs/zustand/discussions/673#discussioncomment-1676328)
- #674 (there was double overriding of `api.setState` and the latter ignored the third argument or something like that)
- maybe more that I haven't looked at

Improvements:
- If the user dispatches an action from devtools previously we used to do the buggy thing I mentioned above it but now if it the store has a `dispatch` function and has a flag `dispatchFromDevtools` set to `true`, we recognize such a store is capable of dispatching actions. `redux` middleware already implements this requirement, so the users that use `redux` middleware get this feature for free. And the users that have a `dispatch` function they can set `dispatchFromDevtools` to `true` to receive actions from the devtools. 
  The `dispatchFromDevtools` flag is required because we don't want to assume just because the store has a `dispatch` function it actually is redux-like `dispatch` it could be something else too, hence by setting `dispatchFromDevtools` the user confirms that it indeed is a `dispatch` that can take actions from the devtools.
  Suggestions for alternative names of `dispatchFromDevtools` are welcome. Previously is was thinking of naming it `reduxLike: true` but I think this is more generic. Instead of a flag we could also make the user set a property like `dispatcher: "dispatch"` but I think that's a little verbose and it'll make typing a little more difficult because we'd have make sure there is a function `"foo"` if users write `dispatcher: "foo"`.
- User can pass a `anonymousActionType` option to customize the action type logged in the devtools, by default it is "anonymous" previously it was "setState" or "action" (I think it's nice to have this option and also anonymous is a better default name but if wanted can be reverted)
- `JUMP_TO_ACTION` and `JUMP_TO_STATE` now replaces the state, previously used to update it which isn't the correct behavior. Devtools only work with non-functional values (as `JSON.stringify({ f: () => {} })` gives `"{}"`) hence while replacing we preserve the functional values of the store (but only top-level, functions deep is anyway an anti-pattern afaict because zustand does a shallow partial update which can remove them as opposed to a deep partial update).
- Support for resetting state to initial via devtools
- Support for committing state via devtools
- Support for rollbacking state via devtools
- Support for pausing/starting recording via devtools


dai-shi:
@devanshj Thanks for working on this! I will have a look at the code later.

@Aslemammad Meanwhile, may I ask you to review this PR? Both the behavior and the code.

We should check if #674 is resolved.

src/middleware/devtools.ts
```diff
@@ -61,107 +74,190 @@ export const devtools =
...
   ): S => {
-    let extension
-    try {
-      extension =
-        (window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
-        (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
-    } catch {}
-
-    if (!extension) {
+    let devtoolsOptions =
```
dai-shi, line 83: We prefer `const`. I will open a PR to add `prefer-const` rule.

src/middleware/devtools.ts
```diff
@@ -61,107 +74,190 @@ export const devtools =
...
+      options === undefined
+        ? { name: undefined }
+        : typeof options === 'string'
+        ? { name: options }
+        : options
+
+    let extensionConnector =
+      (window as any).__REDUX_DEVTOOLS_EXTENSION__ ??
+      (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
```
dai-shi, line 92: I didn't now `top` is not nullable. 🤯 

I feel like we should fix devtools in jotai and valtio too. cc @Aslemammad

src/middleware/devtools.ts
```diff
@@ -61,107 +74,190 @@ export const devtools =
...
-      delete api.devtools
+
       return fn(set, get, api)
     }
-    const namedSet: NamedSet<S> = (state, replace, name) => {
+
+    let extension = Object.create(extensionConnector.connect(devtoolsOptions))
+    // We're using `Object.defineProperty` to set `prefix`, so if extensionConnector.connect
+    // returns the same reference we'd get cannot redefine property prefix error
+    // hence we `Object.create` to make a new reference
```
dai-shi, line 110: Is this correct that `Object.defineProperty` and `Object.create` are only used for migration (warning messages) and we plan to remove them in v4?

src/middleware/devtools.ts
```diff
@@ -61,107 +74,190 @@ export const devtools =
...
-          if (state !== (newState as any)[DEVTOOLS] && api.devtools) {
-            api.devtools.send(api.devtools.prefix + 'setState', api.getState())
+    const setStateFromDevtools: SetState<S> = (state, replace) => {
+      isRecording = false
+      if (!replace) {
+        set(state)
+      } else {
+        let currentState = get()
+        for (let k in currentState) {
+          if (typeof currentState[k] === 'function') {
```
dai-shi, line 191: I assume this is to recover non-serializable parameters. But, it feels a bit heuristic.
Can we give it up and just do `set(state)`, which is more predictable as the merging is the default behavior?

src/middleware/devtools.ts
```diff
@@ -61,107 +74,190 @@ export const devtools =
...
+  try {
+    parsed = JSON.parse(stringified)
+  } catch (e) {
+    console.error(
+      '[zustand devtools middleware] Could not parse the received json',
+      e
+    )
+    didParse = false
+  }
+  if (didParse) f(parsed as T)
```
dai-shi, line 262: nit: can we do this in the try-clause and remove `didParse`?

dai-shi (commented):
wow, this looks pretty nice. thanks for the migration path too.

jotai and valtio should follow this improvement too (they are originally copied from zustand's.)

Can you please run prettier?

src/middleware/devtools.ts
devanshj: Yep, we'll remove them in v4

src/middleware/devtools.ts
devanshj: I think of `parseJsonThen` as a `JSON.parse` replacement. With `JSON.parse` it's the user's responsibility to catch errors (which they can forget) and log the error message. With `parseJsonThen` the user doesn't have to think about all this, not to mention it makes the code less repetitive. 

`didParse` is need because `stringified` could be `"undefined"` in that case `if (parsed !== undefined) f(parsed)` would not work as `f` should have been called with `undefined`, it's true we probably won't have a case where `stringified` is `"undefined"` but still it's good to be complete.

Moreover in #662 `parseThenJson` would be needed as it's type will be `<T>(stringified: StringifiedJson<T>, f: (parsed: T) => void) => void` where `StringifiedJson<T>` would be an opaque type, which would make parsing type-safe while `JSON.parse` is not.

Let me know if you still would prefer try-catch without `didParse` :)

devanshj:
> wow, this looks pretty nice

Haha thanks! ;)

src/middleware/devtools.ts
dai-shi: if `stringified` is `undefined`, JSON.parse will throw.
But, I realizes that my proposition can hide thrown error from `f`, so let's keep as is.

src/middleware/devtools.ts
devanshj: For almost all operations we need replacement. For example `JUMP_STATE`:

Imagine the store initially is `{ foo: 0, someEffect: () => {} }` then user does `set({ bar: 0 })`. So in the eyes of the devtools the state first was `{ foo: 0 }` and now currently it is `{ foo: 0, bar: 0 }`.

So now when the users time travels to the previous state via the slider we'd receive the message `{ type: "DISPATCH", payload: { type: "JUMP_STATE" }, state: "{ 'foo': 0 }" }` and now if we merge the state ie `set({ foo: 0 })`, `bar` doesn't get removed. So the state should be `{ foo: 0, doEffect: () => {}` after time-travel but it's still the same ie `{ foo: 0, bar: 0, doEffect: () => {} }`

So now given we need to replace the state, we can't replace it directly as `set({ foo: 0 }, true)` would remove `doEffect` hence we do replace the state but we also copy the top-level functions to the new state so that they don't get removed.

I hope I've made myself clear. And yes indeed it's a heuristic, the idealistic solution would be to somehow serialize everything.

src/middleware/devtools.ts
devanshj: > if stringified is undefined, JSON.parse will throw.

Ah yes I mistakenly thought `JSON.stringify(undefined)` is `"undefined"` (just like it is for null, strings, numbers, etc). In that case I'll remove `didParse`.

src/middleware/devtools.ts
dai-shi: #678 is merged.

src/middleware/devtools.ts
dai-shi: I know what you mean and JUMP_STATE doesn't work nice for zustand.

There's no logic to assume there are top-level functions. It's just a convention to add action functions. So, I don't like using the assumption here. It should be more predictable than heuristics.

> to somehow serialize everything.

`Map<SomeSerializableIdentifier, UnserializableObject>`?

Another solution would be to keep the list of (nested) properties that are not serialized, and recover them. But, this is still less predictable, or uneasy.

So, I think it's better to educate to avoid using optional values. Instead of `set({ bar: 0 })`, do `set({ foo: null, bar: 0 })`.

devanshj: > `Map<SomeSerializableIdentifier, UnserializableObject>` ?

Kind of.

I agree that this approach is not the best but I realized we have problems with merging too (along with current approach): Imagine if the state was `{ posts: [] }` then the user did `set(s => ({ posts: [...s.posts, { title: "foo", delete }] })` so for devtools it's from `{ posts: [] }` to `{ posts: [{ title: "foo" }] }`. Now imagine if the user goes one step back, we do `set({ posts: [] })`, eveything good, then the users goes forward and we do `set({ posts: [{ title: "foo" }] })` and now the state no longer has the `delete` function.

So all approaches are only half-good, unless as I said we take the pain and figure out how to somehow serialize everything. I'm not sure if this is even possible, I guess the map approach would work. I think we should give that a try, or maybe pick an half-good approach (like merging) for now and keep his for next versions.

Wdyt? I'm open for everything.

dai-shi: For this PR, I don't want to do fancy staff.
Let's either just merge or simply disable some features.
Let's revisit it after v4.

devanshj: Done we're now merging state

devanshj:
Do we want to keep `eslint/no-non-null-assertion`? If yes then should I the `eslint-disable` comment on every line?

It seems CI's node version doesn't have `Array.prototype.at` and hence it's failing, should I use something else?

dai-shi:
I might change my mind later, but for now can you do type assertion? `(extensionSubscriber as ((message: any) => void))`.
`eslint-disable` is the least thing we'd want.

https://github.com/pmndrs/zustand/blob/30f5505cdb7ecb59f19b2724405f8229cfdaf19b/package.json#L64
Yeah, for now we can't use Array.prototype.at.

devanshj:
> but for now can you do type assertion? (extensionSubscriber as ((message: any) => void))

Sure (`!` is literally a shorthand for that xD)

> Yeah, for now we can't use Array.prototype.at

Cool will use `slice(-1)[0]` instead

Aslemammad:
sorry for my mistake, i forgot to submit my reviews yesterday. i guess now uyou have them.

src/middleware/devtools.ts
```diff
@@ -61,107 +74,188 @@ export const devtools =
...
-        const newState = api.getState()
-        if (state !== newState) {
-          savedSetState(state, replace)
-          if (state !== (newState as any)[DEVTOOLS] && api.devtools) {
-            api.devtools.send(api.devtools.prefix + 'setState', api.getState())
+    const setStateFromDevtools: SetState<S> = (state, replace) => {
+      isRecording = false
+      if (!replace) {
+        set(state)
+      } else {
```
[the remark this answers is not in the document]
devanshj: I did that to make the behavior of `JUMP_STATE` and other operations more correct than they were, as I explained [here](https://github.com/pmndrs/zustand/pull/675#discussion_r756858282), but after discussing we're keeping it as it was before (you're viewing an older commit) and we'll look into it later for v4.

src/middleware/devtools.ts
```diff
@@ -61,107 +74,188 @@ export const devtools =
...
     }
+
+    const initialState = fn(api.setState, get, api)
+    extension!.init(initialState)
+
+    extension!.subscribe((message: any) => {
+      switch (message.type) {
+        case 'ACTION':
+          if (!api.dispatchFromDevtools) return
+          if (typeof api.dispatch !== 'function') return
```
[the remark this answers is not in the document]
devanshj: The previous "dispatching" you're talking about was a bug, I explain it [here](https://github.com/pmndrs/zustand/discussions/673#discussioncomment-1676328). I've also mentioned it in short in the description of the PR:

> If the user dispatches `{ type: "INCREMENT" }` action from the devtools, instead of ignoring it (or dispatching if it's a redux store) we used to update the state with it, so the new state would become `{ ...state, type: "INCREMENT" }` this is fixed.

src/middleware/devtools.ts
```diff
@@ -61,107 +74,188 @@ export const devtools =
   (
     set: CustomSetState,
     get: CustomGetState,
-    api: CustomStoreApi & StoreApiWithDevtools<S> & { dispatch?: unknown }
+    api: CustomStoreApi &
+      StoreApiWithDevtools<S> & {
+        dispatch?: unknown
+        dispatchFromDevtools?: boolean
```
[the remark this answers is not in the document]
devanshj: If you're talking about the types (ie `create(devtools(() => ({ count: 0 })).dispatch` gives no errors) then I'm not changing the types for this PR, I'm keeping it as they are `dispatch?: unknown` was previously there so I just added `dispatchFromDevtools?: boolean` with it. Types will get better with #662

If you're talking about the behavior then, I've explained it in the PR description:
> If the user dispatches an action from devtools previously we used to do the buggy thing I mentioned above it but now if it the store has a `dispatch` function and has a flag `dispatchFromDevtools` set to `true`, we recognize such a store is capable of dispatching actions. `redux` middleware already implements this requirement, so the users that use `redux` middleware get this feature for free. And the users that have a `dispatch` function they can set `dispatchFromDevtools` to `true` to receive actions from the devtools.

src/middleware/devtools.ts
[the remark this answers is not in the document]
Aslemammad: OK. Thanks!

src/middleware/devtools.ts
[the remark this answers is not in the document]
Aslemammad: Aha, could you give an example of how it works (in codesandbox or ...)?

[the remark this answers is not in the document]
devanshj: The [redux middleware](https://github.com/pmndrs/zustand/pull/675/files#diff-b53101e83c3efe5ec91dce4d1c3aeaf1d5752ef078b4bd588ed2a050c8251d5dR31-R35) is an example of that. And [here](https://github.com/pmndrs/zustand/pull/675/files#diff-ed83d433b441e4c067aeb456c22886961dda4161280d2fbe2167d98bccf227b3R431-R457) are the tests to show how it works.

Here's a slightly sophisticated usage:
```ts
// implementation
const withActions =
  (initialState, actionCreators) =>
    (set, get, store) => {
    
  let boundedActions = Omap(
    actionCreators,
    (createAction, type) =>
      createAction((s, r) => set(s, r, { type }), get, store)
  )
    
  store.dispatch = ({ type }) => boundedActions[type]();
  store.dispatchFromDevtools = true;

  Object.assign(store, boundedActions);
  return ({ ...initialState, ...boundedActions })
}

// tests
let store = create(devtools(withActions(
  { count: 0 },
  {
    increment: (set, get) => set({ count: get().count + 1 }),
    decrement: (set, get) => set({ count: get().count - 1 }),
    reset: set => set({ ...s, count: 0 })
  }
)
expect(store.getState()).toMatch({ count: 0 })
store.increment();
expect(extension.send.mock.calls.at(-1)).toMatch([
  { type: "increment" },
  { count: 1 }
])
expect(store.getState()).toMatch({ count: 1 })
extensionSubscriber({ type: "ACTION", payload: `{ "type": "decrement" }` })
expect(store.getState()).toMatch({ count: 0 })

// utils
const Omap = (o, f) =>
  Object.fromEntries(
    Object.entires(o)
    .map(([k, v]) => f(v, k))
  )
```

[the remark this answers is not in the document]
Aslemammad: Thanks, but I meant an example with the `create` hook creator & not the vanilla one.

dai-shi:
CI passing ✅ 👏 

@devanshj If you have any ideas to minimize the diff in #662, feel free to add some commits.

src/middleware/devtools.ts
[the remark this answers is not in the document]
dai-shi: a bit of miscommunication here?

dai-shi:
> * f the user dispatches `{ type: "INCREMENT" }` action from the devtools, instead of ignoring it (or dispatching if it's a redux store) we used to update the state with it, so the new state would become `{ ...state, type: "INCREMENT" }` this is fixed. More on the bug [here](https://github.com/pmndrs/zustand/discussions/673#discussioncomment-1676328)

Finally, I think I get what @Aslemammad and @devanshj are discussing.
So, the question from the usage point of view is how can we set arbitrary state from devtools ext? It seems like you want to disallow it? (because the mental model is different from valtio?)

image

Maybe, it's `IMPORT_STATE` we should use.

dai-shi:
Maybe we can support `{"type": "setState", state: { "count": 99 } }` or something, when `api.dispatch` is not available (= redux middleware is not used).

devanshj:
Ah now it makes sense why we had `set(action)` haha. I'd call this hacking the feature which dispatches an action, to set state. I'd be reluctant to add it but I think `{ type: "setState", state }` is feasible if users find it useful. Should we add this feature? I'm open. If yes, in this PR?

> it's `IMPORT_STATE` we should use

Yeah that would be the official way to do it, although it doesn't 1:1 equivalent it also imports the whole action history and all of that, also it's not very ergonomic you'd have to upload a json file, so I think `IMPORT_STATE` is not for the right choice.

src/middleware/devtools.ts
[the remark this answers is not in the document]
devanshj: > but I meant an example with the create hook creator & not the vanilla one.

Ah okay I forgot there's a `create` that returns a hook too.

But I'm still not sure what you're asking? Can't you pass a store to it? `const useStore = create(store)` So what exactly you want the "example" to do?

dai-shi:
> I'd be reluctant to add it but I think `{ type: "setState", state }` is feasible if users find it useful. Should we add this feature? I'm open. If yes, in this PR?

Yeah, let's do that. It at least keeps the functionality that @Aslemammad made.
I know the previous one is hacky and this one too, but it's harmless. `{ type: "setState", state }` should be fair. It's a little bit of typing though.
Yes, we'd want it in this PR.

devanshj:
Cool, I'll update the PR soon. Would you also like to keep this feature even when there is `dispatch` (like in `redux` middleware)? Meaning `"setState"` would become a sort of reserved action type, if we get this action from the devtools we won't forward it to `dispatch` and set state instead. If user dispatches a `setState` action we'll log a warning that it's reserved.

dai-shi:
I would have this feature only when `api.dispatch` is unavailable.
@Aslemammad thoughts?

Aslemammad:
I guess now the DX of this PR is going to be enough for users. I'm on it. Looks good to me to support that. no harm when https://github.com/pmndrs/zustand/pull/675#issuecomment-981732831

devanshj:
Okay so we're adding a feature to set state from the devtools. But it wouldn't make sense just because the user now uses redux middleware they can no longer use this feature, so I'm keeping when using redux middleware too.

And now given that this would be enabled while using redux middleware too, we're making out `setState` reserved, now that name is too common to be reserved, we don't want to annoy users that already have an action with such name, hence I went with `__setState` I hope that's okay. And how given it's revered we warn users if they are dispatching an action with such name, this requires replacing `api.dispatch`[^1] which makes a new reference and I hope no one is relying on the reference of dispatch.

[^1]: Another option is setting `api.hasDevtools = true` from the devtools middleware and then warning in the redux middleware this preserves the reference but then we're leaking devtools implementation in the redux middleware.

dai-shi:
> so I'm keeping when using redux middleware too.

okay, then `__setState` makes sense. I really don't know how many people use `redux` in what ways, so feels like we just let this go.

devanshj:
My thought process was that if there's a feature `X` (setting an arbitrary state) of middleware `A` (devtools), adding an middleware `B` (redux) to `A` shouldn't remove feature `X` of `A`. In other words, imagine you're used to setting arbitrary states to debug things and then one day you decide to use redux middleware and now you can no longer debug the way you used to, that wouldn't make sense.

But that said I'm open to removing it if you want. I don't mind eitherway.

dai-shi (approved):
Let's ship this and see how it goes. Thanks for your great work!

devanshj:
My pleasure!

Aslemammad:
[csb](https://codesandbox.io/s/react-typescript-forked-0ky8o?file=/src/App.tsx) works really well. And the devtools `dispatch` feature is well made! Thank you @devanshj

devanshj:
Thanks and you're welcome!

JannikWempe:
Hi, this line breaks my gatsby build: 
https://github.com/pmndrs/zustand/compare/v3.6.5...v3.6.6#diff-b027caeba1d5de7ca5a00b1ff24be3dd3dcc3e94d147cc5b3a85ef905f7f359dR90

With the error:

```
 WebpackError: ReferenceError: window is not defined
  
  - middleware.js:29 
    [easydatasolution]/[zustand]/esm/middleware.js:29:1
```

It is accessing `window` right before it is checked for being `undefined`.

dai-shi:
Ooops! Thanks for reporting.

@devanshj That was why we had the try-catch. What's your take?

devanshj:
I assume doing a `window !== undefined` check before will solve this? If yes let's do that, try-catch is going a little overboard imo, but I'm fine with whatever.

Could you take this up @dai-shi? I'm afk for a while today.

Also thanks for reporting @JannikWempe!

dai-shi:
Okay, I'll take it.

devanshj:
Thanks!

dai-shi:
https://github.com/pmndrs/zustand/releases/tag/v3.6.7

PR #693, opened 2021-12-05
fix(middleware): devtools work in non-browser

### LINKED ISSUE #694: ReferenceError when using devtools middleware
Hi, I'm getting `ReferenceError: Cannot access 'useStore' before initialization` when using zustand(3.6.6) with devtools middleware in a Nextjs project created using create-next-app.

Steps to reproduce:
1. `npx create-next-app zustand-test`
2. `cd zustand-test`
3. `npm i zustand`
4. Replace `/pages/index.js` with [index.js] and add `store.js` with [store.js]
5. `npm run dev` or `npm run build`

[index.js]
```js
import { useStore } from '../store';

const Home = () => {
  const value = useStore(s => s.value);
  return <div>{value}</div>;
};

export default Home;
```
[store.js]
```js
import create from 'zustand';
import { devtools } from 'zustand/middleware';

const createStore = () => ({
  value: 10,
});

export const useStore = create(devtools(createStore));
```
ReferenceError

dai-shi:
resolve https://github.com/pmndrs/zustand/pull/675#issuecomment-986086807
close #694

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -88,8 +88,9 @@ export const devtools =
         : options
 
     const extensionConnector =
-      (window as any).__REDUX_DEVTOOLS_EXTENSION__ ??
-      (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
+      typeof window !== 'undefined' &&
+      ((window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
+        (window as any).top.__REDUX_DEVTOOLS_EXTENSION__)
 
     if (!extensionConnector) {
       if (
@@ -174,7 +175,7 @@ export const devtools =
       if (!isRecording) return
       extension.send(
         nameOrAction === undefined
-          ? { type: devtoolsOptions.anonymousActionType ?? 'anonymous' }
+          ? { type: devtoolsOptions.anonymousActionType || 'anonymous' }
           : typeof nameOrAction === 'string'
           ? { type: nameOrAction }
           : nameOrAction,
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -87,21 +87,20 @@ export const devtools =
         ? { name: options }
         : options
 
+    if (typeof window === 'undefined') {
+      return fn(set, get, api)
+    }
+
     const extensionConnector =
-      typeof window !== 'undefined' &&
-      ((window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
-        (window as any).top.__REDUX_DEVTOOLS_EXTENSION__)
+      (window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
+      (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
 
     if (!extensionConnector) {
-      if (
-        process.env.NODE_ENV === 'development' &&
-        typeof window !== 'undefined'
-      ) {
+      if (process.env.NODE_ENV === 'development') {
         console.warn(
           '[zustand devtools middleware] Please install/enable Redux devtools extension'
         )
       }
-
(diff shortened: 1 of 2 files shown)


src/middleware/devtools.ts
```diff
@@ -174,7 +175,7 @@ export const devtools =
       if (!isRecording) return
       extension.send(
         nameOrAction === undefined
-          ? { type: devtoolsOptions.anonymousActionType ?? 'anonymous' }
+          ? { type: devtoolsOptions.anonymousActionType || 'anonymous' }
```
dai-shi, line 178: I suppose falsy check instead of nullish check is enough (can't imagine if anyone want empty string for action type). There's no other code using nullish check in this repo. So, want to make it consistent.

dai-shi:
@JannikWempe Would you be able to test with codesandbox build?
https://ci.codesandbox.io/status/pmndrs/zustand/pr/693
☝️ See "Local Install Instructions"

src/middleware/devtools.ts
```diff
@@ -88,8 +88,9 @@ export const devtools =
         : options
 
     const extensionConnector =
-      (window as any).__REDUX_DEVTOOLS_EXTENSION__ ??
-      (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
+      typeof window !== 'undefined' &&
```
JannikWempe, line 91: Maybe it is also an idea to add an early return in the very beginning (line 83) of the devtools function?
`if (typeof window === 'undefined') return fn(set, get, api)`
It would also make it more obvious that the devtools are only running in a browser environment.

That way you would not have to deal with these window checks throughout the function.

JannikWempe: Your changes resolved the issue.
Thank you for reacting that fast. This is not a matter of course 🙏🏼

JannikWempe:
Your changes resolved the issue.
Thank you for reacting that fast. This is not a matter of course 🙏🏼

src/middleware/devtools.ts
dai-shi: good point. hm, but then we need two returns. let's try...

JannikWempe: Looks better to me 👍🏻

dai-shi:
@devanshj any chance to have a look? As I said, my concern is to prepare for #622 (I mean I want to reduce the diff there, so whatever for that should be in this PR or in another PR for v3.)

devanshj:
LGTM.

Although I wouldn't want to replace `??` with `||` and assume that user wouldn't want to have empty string. I do agree with your consistency point but I'd prefer having all other places (if any) in the codebase where `||` is used to be replaced with `??` (in a different PR ofc).

devanshj:
Also I realized it'd be nice if you could add a test case something like the following...

```ts
it("works in non-browser env", () => {
  const originalWindow = global.window
  global.window = undefined
  
  expect(() => {
    create(devtools(() => ({ count: 0 })))
  }).not.toThrow();
  
  global.window = originalWindow
})
```

dai-shi:
I should have mention that preferring`||` is coming from concerning the bundle size. With the current transpilation config, `??` is transpiled for old browsers, which may increase bundle size. zustand cares bundle size pretty much. In the future version, we may reconsider this, but for now, I want to keep `||` unless there's a specific use case (like a bug report from users). This would apply to #662 too.

devanshj:
In that case I'd prefer to do `anonymousActionType === undefined ? "anonymous" : anonymousActionType` because I personally wouldn't compromise with behavior in any case.

That said I'm fine if you want to keep it as `||` too.

dai-shi:
That's almost how `??` is transpiled. Yeah, let's go with `||`. This is intentional.

PR #706, opened 2021-12-09
Remove devtools prefixing from readme

devanshj:
In accordance with #675


dai-shi (approved):
Ah, thanks. We should have clearly written this in the release note.

Issue #705, opened 2021-12-09
Devtools seems to be broken in 3.6.6 onwards


dsacramone:
ISSUE:
Devtools middleware seems to be broken from 3.6.6+ onwards.

REPRODUCE:
1. Go here: https://codesandbox.io/s/zustand-devtools-oddities-i6rrb?file=/src/chat.store.js
-- currently set at zustand version 3.6.7
-- view devtools. You will see that all actions are "anonymous".

2. Change dependency to 3.6.5, for zustand. (left side under "Dependencies", select zustand and update from dropdown).
-- view devtools, seems to be working with a custom "name" passed. 

3. Change dependency to 3.6.6, for zustand. (left side under "Dependencies", select zustand and update from dropdown).
-- view devtools, You will see that all actions are "anonymous"

3.6.6 seems to have introduced a breaking change.

dai-shi:
@devanshj Can you take a look, please?

devanshj:
This is an intended change, quoting from the PR #675...

> 1. **No more prefixing:** If you check the git history we used to prefix names/actions because we didn't used to pass the name to `connect` like so `connect({ name })` hence prefixing was needed because all zustand stores would show up in the same store instance. But later we started passing the `name` and that creates separate instances (which means separate action log and state) for each zustand store. But whoever made the change didn't realise that now we no longer need the prefixing, so hence we're removing it now.

Okay so let me elaborate on this. Long back all zustand stores had the same devtools instance meaning they shared the same action log. Hence we used to prefix the actions. But after we started creating a different instance (in #145) for each store they no longer shared the same action log, but we forgot to stop prefixing. Before this rewrite prefixing was redundant as all actions in the log have the same prefix. So this rewrite fixing that and we stop prefixing. If you want to see what devtools instance you're working on you can see at the top right dropdown (note it still reads "ACTION-TEST: " in the latest version).

In addition to this we also changed the anonymous action type from "action" to "anonymous", this is also an intended change as stated in the PR

> User can pass a `anonymousActionType` option to customize the action type logged in the devtools, by default it is "anonymous" previously it was "setState" or "action" (I think it's nice to have this option and also anonymous is a better default name but if wanted can be reverted)

So if you prefer some other name you can pass a `anonymousActionType` option.

Because of these two changes the actions change from "ACTION-TEST: > action" to "anonymous", which is intended. I hope I've made myself clear.

omidantilong:
[...]

PR #712, opened 2021-12-13
docs: add note about using set() third param for logging

omidantilong:
Newcomers to zustand may be confused why devtools doesn't show action types by default (I was!), so I've added  a note to the readme following [this comment from @devanshj ](https://github.com/pmndrs/zustand/issues/705#issuecomment-992976936)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -440,15 +440,14 @@ const useStore = create(devtools(redux(reducer, initialState)))
 devtools takes the store function as its first argument, optionally you can name the store or configure [serialize](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize) options with a second argument.  
   
 Name store: `devtools(store, {name: "MyStore"})`, which will create a seperate instance named "MyStore" in the devtools.
-Serialize options: `devtools(store, { serialize: { options: true } })`.  
+
+Serialize options: `devtools(store, { serialize: { options: true } })`. 
   
 #### Logging Actions
 
 devtools will only log actions from each separated store unlike in a typical *combined reducers* redux store. See an approach to combining stores https://github.com/pmndrs/zustand/issues/163
 
-Because of this, actions will be logged as "anonymous" by default. To set a default log action type for all actions from your store: `devtools(store, {name: 'MyStore', anonymousActionType: 'MyStoreAction'})`
-
-Alternatively, `set` functions take a third parameter which can be used to log any action type you choose: 
+You can log a specific action type for each `set` function by passing a third parameter:
 
 ```jsx
 const createBearSlice = (set, get) => ({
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -442,8 +442,25 @@ devtools takes the store function as its first argument, optionally you can name
 Name store: `devtools(store, {name: "MyStore"})`, which will create a seperate instance named "MyStore" in the devtools.
 Serialize options: `devtools(store, { serialize: { options: true } })`.  
   
+#### Logging Actions
+
 devtools will only log actions from each separated store unlike in a typical *combined reducers* redux store. See an approach to combining stores https://github.com/pmndrs/zustand/issues/163
 
+Because of this, actions will be logged as "anonymous" by default. To set a default log action type for all actions from your store: `devtools(store, {name: 'MyStore', anonymousActionType: 'MyStoreAction'})`
```
dai-shi, line 449: This seems an overlap with L442 example. Just a little bit though. If you have an idea to make the docs compact, it's welcomed.

dai-shi (approved):
Looks good to me.
@devanshj can you double check please?

readme.md
devanshj, line 449: The recommended approach is to use the third parameter instead of `anonymousActionType` so let's mention that first.

And then after that let's just write "If the third parameter is not passed, the action type by default becomes `"anonymous"`. You can customize this via the `anonymousActionType` devtools option. Eg `devtool(..., { anonymousActionType: 'unknown', ... })`"

It doesn't make sense to use or encourage the store name as `anonymousActionType` because you can already see the store instance in top right drop-down.

readme.md
```diff
@@ -442,8 +442,25 @@ devtools takes the store function as its first argument, optionally you can name
...
+
+Alternatively, `set` functions take a third parameter which can be used to log any action type you choose:
```
devanshj, line 451: Now that we're writing this first, let's write this: "You can log an action for each `set` by passing an action type to the third parameter. Example:"

devanshj:
With the suggested changes, lgtm.

dai-shi (changes requested):
We are waiting for some changes, right?

devanshj:
> We are waiting for some changes, right?

Yep

omidantilong:
Hey, sorry, I didn't realise you were waiting for me! I will try and make the suggested changes over xmas when I have some free time. I've been a bit busy the last week or so.

omidantilong:
Found the free time :)

devanshj (approved):
Thanks! :) Lgtm.

readme.md
```diff
@@ -440,10 +440,32 @@ const useStore = create(devtools(redux(reducer, initialState)))
 devtools takes the store function as its first argument, optionally you can name the store or configure [serialize](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize) options with a second argument.  
   
 Name store: `devtools(store, {name: "MyStore"})`, which will create a seperate instance named "MyStore" in the devtools.
-Serialize options: `devtools(store, { serialize: { options: true } })`.  
+
+Serialize options: `devtools(store, { serialize: { options: true } })`.
```
dai-shi, line 444:
```suggestion
Serialize options: `devtools(store, { serialize: { options: true } })`.
```

dai-shi (approved):
Thanks for your contribution! Merging.

PR #715, opened 2021-12-17
Deprecations before v4

devanshj:
This PR is to be released as a minor which will contain deprecations that'll make the migration to the next major smoother. This is not to be merged yet, as #662 progresses this will get more commits if required, so this will be ready when #662 is ready.

I'm quite an idealist, I value correctness higher than non-breaking change, so changes are in alignment to that, they can always be negotiated later, I'm not the boss :)

Changes till now:
- **Minimizing Types**.
  - Some types are removed which can be replaced by a existing type. Eg `SetState<T>` to `Store<T>['setState']`. It's best if derived types are actually derived. A type named `SetState<T>` doesn't tell what is it's relationship to `Store<T>` or where did it came from. It's always good to only export the necessary types.
  - Some types are removed which can be inlined. Eg `EqualityChecker<T>` to `(a: T, b: T) => boolean` and `StateSelector<T, U>` to `(state: T) => U`. The beauty of types is that you can tell their implementation by simply looking at them. `StateSelector<T, U>` doesn't tell me anything about the implementation whereas a shorter `(state: T) => U` does. Types should only be extracted if the name is more telling than it's definition or the definition is too long to write, eg `Store<T>`.
  - Some types are only renamed to better names. Eg. `StateCreator` to `StoreInitializer`.
- **Removing the "api" nomenclature**.
  Calling "store" as "storeApi" is like calling "todos" as "todosArray", "Tea" as "Tea beverage", "Devansh" as "Devansh Human". Good part is that this is not visible at a lot of places in the code. When using context, there's a `useStoreApi` so that is renamed to `useStoreRef` (as `useStore` is already for reading the state).
  A more idealist and correct naming would have been calling `MyStoreContext.useStore()` as simply `MyStoreContext.use()`, and `MyStoreContext.useStoreRef()` as `MyStoreContext.useStore()`, but that's too radical change so I avoided it, let me know if you like it anyway.
- `devtools({}, name)` to `devtools({}, { name }`. Ability to pass the name directly is a rather unnecessary feature, more over expecting an object is consistent with `persist`. Good part is the documentation doesn't mention that you can directly pass the name so hopefully many users would not be using it.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,6 +1,18 @@
+/**
+ * @deprecated `State` is renamed to `UnknownState`,
+ * `State` will be removed in next major
+ */
 export type State = object
+
+export type UnknownState = object;
+
 // types inspired by setState from React, see:
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c49e45842358ba59a508e13130791989911430d/types/react/v16/index.d.ts#L489-L495
+/**
+ * @deprecated Use the builtin `Partial<T>` instead of `PartialState<T>`.
+ * Additionally turn on `--exactOptionalPropertyTypes` tsc flag.
+ * `PartialState` will be removed in next major
+ */
 export type PartialState<
   T extends State,
   K1 extends keyof T = keyof T,
@@ -10,10 +22,35 @@ export type PartialState<
 > =
   | (Pick<T, K1> | Pick<T, K2> | Pick<T, K3> | Pick<T, K4> | T)
   | ((state: T) => Pick<T, K1> | Pick<T, K2> | Pick<T, K3> | Pick<T, K4> | T)
+
+/**
+ * @deprecated Use `(t: T) => U` instead of `StateSelector<T, U>`.
+ * `StateSelector` will be removed in next major.
+ */
 export type StateSelector<T extends State, U> = (state: T) => U
+
+/**
+ * @deprecated Use `(a: T, b: T) => boolean` instead of `EqualityChecker<T>.
+ * `EqualityChecker` will be removed in next major.
+ */
 export type EqualityChecker<T> = (state: T, newState: T) => boolean
+
+/** 
+ * @deprecated Use `(state: T, prevState: T) => void` instead of `StateListener<T>`.
+ * `StateListener` will be removed in next major.
+ */
 export type StateListener<T> = (state: T, previousState: T) => void
+
+/** 
+ * @deprecated Use `(slice: T, prevSlice: T) => void` instead of `StateSliceListener<T>`.
+ * `StateSliceListener` will be removed in next major.
+ */
 export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
+
+/**
+ * @deprecated Use `Store<T>['subscribe']` instead of `Subscribe<T>`.
+ * `Subscribe` will be removed in next major.
+ */
 export type Subscribe<T extends State> = {
   (listener: StateListener<T>): () => void
   /**
@@ -26,6 +63,10 @@ export type Subscribe<T extends State> = {
   ): () => void
 }
 
+/**
+ * @deprecated Use `Store<T>['setState']` instead of `SetState<T>`
+ * `SetState` will be removed in next major.
+ */
 export type SetState<T extends State> = {
   <
     K1 extends keyof T,
@@ -37,15 +78,50 @@ export type SetState<T extends State> = {
     replace?: boolean
   ): void
 }
+
+/**
+ * @deprecated Use `() => T` or `Store<T>['getState']` instead of `GetState<T>`
+ * `GetState` will be removed in next major.
+ */
 export type GetState<T extends State> = () => T
+
+/**
+ * @deprecated Use `() => void` or `Store<UnknownState>['destroy']` instead of `Destroy`.
+ * `Destroy` will be removed in next major.
+ */
 export type Destroy = () => void
+
+/**
+ * @deprecated `StoreApi<T>` has been renamed to `Store<T>`
+ * `StoreApi` will be removed in next major.
+ */
 export type StoreApi<T extends State> = {
   setState: SetState<T>
   getState: GetState<T>
   subscribe: Subscribe<T>
   destroy: Destroy
 }
+
+export type Store<T extends State> = {
+  setState: SetState<T>
+  getState: GetState<T>
+  subscribe: Subscribe<T>
+  destroy: Destroy
+}
+
+/**
+ * @deprecated `StateCreator` has been renamed to `StoreInitializer`.
+ * `StateCreator` will be removed in next major.
+ */
 export type StateCreator<
+  T extends State,
+  CustomSetState = SetState<T>,
+  CustomGetState = GetState<T>,
+  CustomStoreApi extends StoreApi<T> = StoreApi<T>
+> = (set: CustomSetState, get: CustomGetState, api: CustomStoreApi) => T
+
+
+export type StoreInitializer<
   T extends State,
   CustomSetState = SetState<T>,
   CustomGetState = GetState<T>,
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -4,7 +4,7 @@
  */
 export type State = object
 
-export type UnknownState = object;
+export type UnknownState = object
 
 // types inspired by setState from React, see:
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c49e45842358ba59a508e13130791989911430d/types/react/v16/index.d.ts#L489-L495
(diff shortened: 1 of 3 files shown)


devanshj:
@dai-shi #662 is almost ready, should we go ahead with these changes? Feel free to disagree.

dai-shi:
will check this later. can you fix ci errors, meanwhile?

dai-shi:
Okay, I looked at the description (not the code yet), but this can't be accepted.
You want to assume that if there are too many breaking changes, people would lose trust and leave for another library.

Here are some overall comments:
- I personally like minimizing types. But, this can be tough for TS beginners, and we will see many complaints. Let's drop some types if they are impossible (or maybe not reasonable) to implement in #662 **and** we can give a migration path easily with `@deprecated` message.
- Same for renaming types. We should keep them as much as possible, unless we have more benefit with renaming, or unless old ones are too confusing.
- Please do not change the JS code in `zustand/context`. We will keep the same functionality in v4 as much as possible. (tbh, I want to drop `zustand/context` entirely, but dropping will be troublesome too.) I understand the naming isn't perfect, but the history (= keeping backward compatibility) is more important.
- `devtools(..., { name })` seems natural and bundle size increase is usually not a big issue for DEV only things. I'm not sure how many uses directly pass `name`. If there are many we might consider keeping the deprecation message in v4 (I think unlikely, though).

I'm still open for suggestions about dropping and renaming types, but generally we need to consider more about backward compatibility.

dai-shi:
The `devtools` signature change/deprecation can be in v3.6.8. What do you think?

devanshj:
> I personally like minimizing types. But, this can be tough for TS beginners, and we will see many complaints. Let's drop some types if they are impossible (or maybe not reasonable) to implement in Rewrite with better types #662 and we can give a migration path easily with @deprecated message.
> Same for renaming types. We should keep them as much as possible, unless we have more benefit with renaming, or unless old ones are too confusing.

I think we should make users, including TS beginners, adapt to the types instead of spoon-feeding them with types as trivial as `EqualityChecker<T>`. My suggestions for the type changes are in the PR, let me know which changes you like and which you don't. If you want to keep types as backward compatible as possbile, that can be done too, as I've done in #725

> Please do not change the JS code in zustand/context. We will keep the same functionality in v4 as much as possible. (tbh, I want to drop zustand/context entirely, but dropping will be troublesome too.) I understand the naming isn't perfect, but the history (= keeping backward compatibility) is more important.

Okay haha I understand, we'll keep it backward compatible.

> devtools(..., { name }) seems natural and bundle size increase is usually not a big issue for DEV only things. I'm not sure how many uses directly pass name. If there are many we might consider keeping the deprecation message in v4 (I think unlikely, though).

> The devtools signature change/deprecation can be in v3.6.8. What do you think?

Yep that sounds good

dai-shi:
Added a comment in #725.

> My suggestions for the type changes are in the PR, let me know which changes you like and which you don't.

I'll review the suggestions again with the new approach in mind.

> > The devtools signature change/deprecation can be in v3.6.8. What do you think?

> Yep that sounds good

Please go ahead and create a PR, then v3.6.8 will be ready.

devanshj:
Done #731

src/vanilla.ts
```diff
@@ -1,6 +1,18 @@
+/**
+ * @deprecated `State` is renamed to `UnknownState`,
+ * `State` will be removed in next major
```
dai-shi, line 3: I would wanted to drop `State` entirely, and not even export `UnknownState`.
But, this will probably be too big change, so let's keep it.

I would even want to make it `unknown`. If it were possible, we could consider dropping.

src/vanilla.ts
```diff
@@ -1,6 +1,18 @@
...
 export type State = object
+
+export type UnknownState = object
+
 // types inspired by setState from React, see:
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c49e45842358ba59a508e13130791989911430d/types/react/v16/index.d.ts#L489-L495
+/**
+ * @deprecated Use the builtin `Partial<T>` instead of `PartialState<T>`.
+ * Additionally turn on `--exactOptionalPropertyTypes` tsc flag.
+ * `PartialState` will be removed in next major
```
dai-shi, line 14: Yes, this has been troublesome and happy to drop it.

src/vanilla.ts
```diff
@@ -10,10 +22,35 @@ export type PartialState<
...
+ * @deprecated Use `(a: T, b: T) => boolean` instead of `EqualityChecker<T>.
+ * `EqualityChecker` will be removed in next major.
+ */
 export type EqualityChecker<T> = (state: T, newState: T) => boolean
+
+/**
+ * @deprecated Use `(state: T, prevState: T) => void` instead of `StateListener<T>`.
+ * `StateListener` will be removed in next major.
+ */
 export type StateListener<T> = (state: T, previousState: T) => void
```
dai-shi, line 42: For `StateSelector`, `EqualityChecker` and `StateListener`, I'm not sure what will happen. I guess many users use them. They might not be necessary nor best named, but not too confusing. So, for now I lean to keep them.

src/vanilla.ts
```diff
@@ -10,10 +22,35 @@ export type PartialState<
...
+
+/**
+ * @deprecated Use `(slice: T, prevSlice: T) => void` instead of `StateSliceListener<T>`.
+ * `StateSliceListener` will be removed in next major.
+ */
 export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
```
dai-shi, line 48: I think my plan was to drop this in #604. I don't know why I didn't. Let's drop this at least from vanilla.ts.

src/vanilla.ts
```diff
@@ -10,10 +22,35 @@ export type PartialState<
...
+
+/**
+ * @deprecated Use `Store<T>['subscribe']` instead of `Subscribe<T>`.
```
dai-shi, line 51: I don't dislike this, but people would get confused. Better to stay. Same for `SetState` and `GetState`.

src/vanilla.ts
```diff
@@ -37,21 +78,55 @@ export type SetState<T extends State> = {
...
+ * @deprecated Use `() => T` or `Store<T>['getState']` instead of `GetState<T>`
+ * `GetState` will be removed in next major.
+ */
 export type GetState<T extends State> = () => T
+
+/**
+ * @deprecated Use `() => void` or `Store<UnknownState>['destroy']` instead of `Destroy`.
+ * `Destroy` will be removed in next major.
+ */
 export type Destroy = () => void
```
dai-shi, line 92: I'm fine to drop this. Not many people should be using.

src/vanilla.ts
```diff
@@ -37,21 +78,55 @@ export type SetState<T extends State> = {
...
+
+/**
+ * @deprecated `StoreApi<T>` has been renamed to `Store<T>`
```
dai-shi, line 95: I know you don't like this, but let's keep it.

src/vanilla.ts
```diff
@@ -37,21 +78,55 @@ export type SetState<T extends State> = {
...
+
+export type Store<T extends State> = {
+  setState: SetState<T>
+  getState: GetState<T>
+  subscribe: Subscribe<T>
+  destroy: Destroy
+}
+
+/**
+ * @deprecated `StateCreator` has been renamed to `StoreInitializer`.
```
dai-shi, line 113: Let's keep this.

dai-shi (commented):
Reviewed vanilla.ts. It turns out, I'd be more conservative.

src/vanilla.ts
dai-shi: I can try there three actually, and see if someone gets confused.

devanshj:
> I would wanted to drop State entirely, and not even export UnknownState.
> But, this will probably be too big change, so let's keep it.

> For StateSelector, EqualityChecker and StateListener, I'm not sure what will happen. I guess many users use them. They might not be necessary nor best named, but not too confusing. So, for now I lean to keep them.

These types are rather simple to replace with, so I don't think it would be troublesome to migrate, but I'm fine if you want to keep them.

dai-shi:
No, it's not about how troublesome they are to migrate. If people need to change something, people would hesitate. If they feel unhappy, they will leave. We are already planning many changes, so from your perspective, please think it as we should keep backward compatibility as much as possible. v3.6.x isn't nice for middleware typing. in this sense, middlware typing should be backward compatible with v3.5.x again as much as possible.

src/vanilla.ts
devanshj: > I can try there three actually, and see if someone gets confused.

Did you thought about this and decided that if you want to keep them or not?

dai-shi: Let's keep them.
I saw some tweets and codesandboxes and know people use them.
If we were to remove those, it would be v5 (taking migration period longer like a few months to a year), which will not happen any time soon.

devanshj:
Ha weird I thought the new branch renaming feature worked across forks xD — will reopen

PR #720, opened 2021-12-21
devtools: minor fixes for pause recording feature

devanshj:
- When setting the state from the devtools we used to temporary turn off recording and then we turn it back. But turning it back isn't correct, what if it was already turned off? Hence we now restore instead of turning it on.
- If the action received isn't a string then instead of parsing it we log a "Unsupported action format"
- We're now overwriting dispatch before subscribing to the devtools just in case of hypothetical sync (eg while testing) dispatching of actions from the devtools

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -182,17 +182,39 @@ export const devtools =
       )
     }
     const setStateFromDevtools: SetState<S> = (...a) => {
+      const originalIsRecording = isRecording
       isRecording = false
       set(...a)
-      isRecording = true
+      isRecording = originalIsRecording
     }
 
     const initialState = fn(api.setState, get, api)
     extension.init(initialState)
 
+    if (api.dispatchFromDevtools && typeof api.dispatch === 'function') {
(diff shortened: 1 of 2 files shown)


dai-shi:
could you write some description for a record?

devanshj:
Sure, done.

dai-shi (approved):
looks okay.

dai-shi:
one question: is this an issue we had before #675 rewrite?

devanshj:
> one question: is this an issue we had before #675 rewrite?

Before the rewrite we didn't have the feature of pausing the recording (ie we didn't handle the `"PAUSE_RECORDING"` action), since there was no pausing feature, there was no way to have this issue in the first place.

So the rewrite introduced this issue or more technically speaking it didn't implement the pausing feature correctly.

dai-shi:
Ah, got it. Thanks for the clarification! and your continuous work!

dai-shi:
@Aslemammad do you get any insights from this for your jotai issue?

PR #722, opened 2021-12-23
fix(react): update reactive listener on api change

### LINKED ISSUE #721: A swapped in store will not update reactively
I'm trying to use `zustand` for a project where a user can specify a `zustand`-backed provider. If the user switches their provider, any changes to the new providers store are not picked up by react. This is because any uses of the `zustand` hook `useStore` will not be updated to listen to the new provider, because the listener is set up once, with no deps.

As a contrived (untested) example:
```
<MyComponent storeBackedProvider={mock} />
```

```
function MyComponent({storeBackedProvider}) {
  const useStore = create(storeBackedProvider.store)
  const value = useStore()

  return <span>{value}</span>
}
```

In this case, if the `mock` is swapped out, `MyComponent` will no longer be reactive to `value`.

---

I propose depending on `api` when registering listeners to fix this.

zzmp:
- Make `useStore` reactive even if the store is swapped after the initial render (fixes #721).

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -147,7 +147,7 @@ function create<
         listener() // state has changed before subscription
       }
       return unsubscribe
-    }, [])
+    }, [api])
 
     const sliceToReturn = hasNewStateSlice
       ? (newStateSlice as StateSlice)
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -7,8 +7,12 @@ import {
 } from 'react'
 import { act, fireEvent, render } from '@testing-library/react'
 import ReactDOM from 'react-dom'
-import create, { EqualityChecker, SetState, StateSelector, UseBoundStore } from 'zustand'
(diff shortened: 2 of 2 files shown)


zzmp:
Hi @dai-shi! We're trying to use zustand, but are blocked on this issue. Please let me know if you need anything else to make this an acceptable PR.

Thanks for jotai/zustand!

dai-shi:
Cool. Would you be able to add a test?

zzmp:
Added a test! The test failed locally without the change in 1cc3035, and passed with it.

dai-shi:
Ah, my apologies. I misunderstood something (I was probably sleeping).
This is actually not what we want to suggest. Let me comment in #721.

dai-shi:
Commented: https://github.com/pmndrs/zustand/issues/721#issuecomment-999983109

Sorry, this is not the right solution for zustand v3.

PR #724, opened 2021-12-23
fix(middleware): revert devtools extension connector with try-catch

### LINKED ISSUE #723: [Devtools middleware] undefined is not an object
After update zustand - "zustand": "^3.6.7"
I get red error 

Screenshot 2021-12-23 at 13 54 47

dai-shi:
close #723

When I reviewed #675, I thought it's clean without try-catch.
Later, an issue is reported and we fix it with #693, but that wasn't enough.
This is to revert to pre #675, with try-catch.

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -87,16 +87,20 @@ export const devtools =
         ? { name: options }
         : options
 
-    if (typeof window === 'undefined') {
-      return fn(set, get, api)
+    let extensionConnector
+    try {
+      extensionConnector =
+        (window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
+        (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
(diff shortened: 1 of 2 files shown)


devanshj:
Could you add a test case?
```ts
it("works in react native env", () => {
  const originalWindow = global.window
  global.window = {}
  
  expect(() => {
    create(devtools(() => ({ count: 0 })))
  }).not.toThrow();
  
  global.window = originalWindow
})
```

I'd still prefer not reaching for try-catch because the empty catch clause is just eating errors without reporting them, had it been try-catch we'd have not got the last two valuable test cases and reports.

But I'm okay with try-catch too, it's not a big deal in this case.

dai-shi:
@devanshj Thanks for the test case. Added.

> the empty catch clause is just eating errors without reporting them

In general, I agree avoiding empty catch. But, this case seems reasonable.

PR #731, opened 2021-12-24
fix(middleware/devtools): deprecate passing name directly instead of an object

devanshj:
From #715:

> - `devtools({}, name)` to `devtools({}, { name }`. Ability to pass the name directly is a rather unnecessary feature, more over expecting an object is consistent with `persist`. Good part is the documentation doesn't mention that you can directly pass the name so hopefully many users would not be using it.


dai-shi (approved):
LGTM!

Issue #716, opened 2021-12-17
Support all devtools features


devanshj:
Redux devtools has many more features that we have not implemented yet. For example this [tracing feature](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/Features/Trace.md). The full list of features is [here](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md). This will require looking into the (actual) redux's middleware/enhancer's source to see what needs to be done to have these features (and if we can even have them in zustand or not).

Putting it out there in case someone's interested, I might look into it when I have time.

---

- [ ] #926

dai-shi:
It feels like we should focus on v4 and work on this after that, unless this introduces any breaking changes. That said, in the meantime, I'm more interested in this for [valtio](https://github.com/pmndrs/valtio)'s `devtools` and [jotai](https://github.com/pmndrs/jotai)'s `useAtomDevtools`.

DanielPower:
The trace property seems to be mostly working already, just that Zustand isn't typed to accept it. I've put up a PR to allow the trace option to be passed.

I have noticed one problem with this, where a callback passed in for `trace` will not receive the action as an argument. I'll look into that and address it in a future PR. But for now, just allowing the option to be passed gets us much of the value.

Aslemammad:
We tried supporting the trace feature in https://github.com/pmndrs/valtio/issues/331 (PR: https://github.com/pmndrs/valtio/pull/380) but as mentioned, there's an issue in redux devtools trace API (https://github.com/reduxjs/redux-devtools/issues/1100), so I guess we can ignore the trace feature for now if possible, and work again on it in case the trace API works again.

dai-shi:
#926 is related, right?

Hideman85:
The trace feature is really great for debugging, I would to see it available 🙂

dai-shi:
Closing this as not actionable.
#1200 covers all configs (and #1991 is a better fix).
The trace feature would be nice, but it's blocking. Let's create a new issue once https://github.com/reduxjs/redux-devtools/pull/1369 is published.

If there are other features to support, please open a new discussion.

PR #752, opened 2022-01-13
Add `trace` property to devtools options

DanielPower:
Relates to #716

From my testing, redux dev tools trace feature is already working without any changes on Zustand's side. All that's necessary is to type the option so that Typescript will allow it to be passed.

There is at least one caveat, and that is that if you try to use a callback for trace, it does not correctly get passed the action as an argument. This functionality exists so you can either customize the trace, or to conditionally trace on some actions like in the example below.
```
trace: (action) => {
  if (shouldTrace(action)) {
    return new Error().trace;
  }
},
```

In this example, `action` will always come back as `undefined`. I intend to look into this and will put up another PR when I can with the necessary fixes. However as an initial step, I'm putting up this PR to type the field, because much of the functionality is already working.

Edit: I've also gone ahead and added `traceLimit` and `features`, as I've tested both of those as working. I'll type out more features in the future as I test them.

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -1,5 +1,39 @@
 import { GetState, PartialState, SetState, State, StoreApi } from '../vanilla'
 
+type DevtoolsOptions = {
+  name?: string
+  anonymousActionType?: string
+  serialize?: {
+    options:
+      | boolean
+      | {
+          date?: boolean
+          regex?: boolean
+          undefined?: boolean
+          nan?: boolean
+          infinity?: boolean
+          error?: boolean
+          symbol?: boolean
+          map?: boolean
+          set?: boolean
+        }
+  }
+  features?: {
+    pause?: boolean
+    lock?: boolean
+    persist?: boolean
+    export?: boolean | 'custom'
+    import?: boolean | 'custom'
+    jump?: boolean
+    skip?: boolean
+    reorder?: boolean
+    dispatch?: boolean
+    test?: boolean
+  }
+  trace?: boolean | (() => string)
+  traceLimit?: number
+}
+
 type DevtoolsType = {
   /**
    * @deprecated along with `api.devtools`, `api.devtools.prefix` is deprecated.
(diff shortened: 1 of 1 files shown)


src/middleware/devtools.ts
```diff
@@ -1,5 +1,39 @@
 import { GetState, PartialState, SetState, State, StoreApi } from '../vanilla'
 
+type DevtoolsOptionsType = {
```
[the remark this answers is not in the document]
DanielPower: I saw `DevtoolsType`, and tried to be consistent with existing code. But I prefer calling it just `DevtoolsOptions` as well. Resolved :)

src/middleware/devtools.ts
```diff
@@ -1,5 +1,39 @@
...
+    lock?: boolean
+    persist?: boolean
+    export?: boolean | 'custom'
+    import?: boolean | 'custom'
+    jump?: boolean
+    skip?: boolean
+    reorder?: boolean
+    dispatch?: boolean
+    test?: boolean
+  }
```
devanshj, line 32: Are you sure we don't need any changes in our implementation to support these flags? Are these entirely handled by the extension? And could point to the type definition of this from the [repo](https://github.com/reduxjs/redux-devtools). We should also include jsdoc for them.

You can opt to remove `features` altogether, no issues, I just don't want to do things half-baked :)

src/middleware/devtools.ts
```diff
@@ -1,5 +1,39 @@
...
+  trace?: boolean | (() => string)
```
devanshj, line 33: Let's remove `(() => string)` if it's not working. (I assume we need to change our implementation to support it?)

src/middleware/devtools.ts
```diff
@@ -1,5 +1,39 @@
...
+  traceLimit?: number
```
devanshj, line 34: The following applies to `trace` and `traceLimit` too

> Are you sure we don't need any changes in our implementation to support these flags? Are these entirely handled by the extension? And could point to the type definition of this from the repo. We should also include jsdoc for them.

src/middleware/devtools.ts
```diff
@@ -1,5 +1,39 @@
 import { GetState, PartialState, SetState, State, StoreApi } from '../vanilla'
 
+type DevtoolsOptions = {
```
devanshj, line 3: If you're interested you can look up the [repo](https://github.com/reduxjs/redux-devtools), find out what other options we're missing that don't require implementation changes and add jsdocs for the them and the existing ones too

dai-shi (changes requested):
Please consider @devanshj's comments.

DanielPower:
Thank you for the comments. Will loop back to this as soon as I can.

dai-shi:
Closing this as stale. Please start over because we have many changes in the main branch. #716 should be the issue for tracking and discussion.

PR #725, opened 2021-12-23
breaking(types): Add higher kinded mutator types

devanshj:
Okay @dai-shi now you have two options:

**A. This PR (not complete yet)**
Makes the change as minimal as possible, just adds the types proposed in #710, zero change in runtime code.

**B. #662**
Not only adds #710 but also rewrites the whole codebase to drastically increase the code quality. Don't mind me but the current code is quite a mess, I like to write code for fun so I thought let's rewrite this whole thing nicely. Now this obviously will produce relatively large and unhelpful diffs, nothing can be done about that. It would be much easier to review if you compare the files locally without the red-green diff.
But now I have a small condition: I'll remove prettier and some lint rules, but don't worry you can add them back but you'll have to do it yourself (you have the edit rights). Consider this a little treat that I'm giving myself :P I hope you won't mind.

Let me know which of the two options you prefer. If you think increasing the code quality is worth taking the pain to review, I'd suggest you go with the second option. I don't mind whichever you pick.

PR #772, opened 2022-02-02
Migrations regarding higher kinded mutators for v4

devanshj:
(#715 branch renamed)


devanshj:
In #715 you reviewed and finally we were going to remove `PartialState`, `StateSliceListener` and `Destroy`. Let's keep `Destroy` just to be consistent as we have types for all other keys of store.

src/context.ts
```diff
@@ -6,12 +6,20 @@ import {
...
+  StateSelector,
+  StoreApi,
+  UseBoundStore,
+} from 'zustand'
 
-export type UseContextStore<T extends State> = {
-  (): T
-  <U>(selector: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
-}
+export type UseContextStore<T extends State | StoreApi<State>> = <
```
dai-shi, line 17: We want to mark the old usage with `@deprecated` with a suggestion message. Would it be possible?

src/vanilla.ts
```diff
@@ -151,3 +160,13 @@ function createStore<
...
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export interface StoreMutators<S, A> {}
+export type StoreMutatorIdentifier = keyof StoreMutators<unknown, unknown>
+
+export type Mutate<S, Ms> = Ms extends []
+  ? S
+  : Ms extends [[infer Mi, infer Ma], ...infer Mrs]
+  ? Mutate<StoreMutators<S, Ma>[Mi & StoreMutatorIdentifier], Mrs>
+  : never
```
dai-shi, line 172: Oh, I see.. This is probably what I wanted to do originally. Incomplete typing, you'd say.
Very nice.

src/middleware/subscribeWithSelector.ts
```diff
@@ -10,6 +10,31 @@ import {
...
+    listener: (selectedState: U, previousSelectedState: U) => void,
+    options?: {
+      equalityFn?: (a: U, b: U) => boolean
+      fireImmediately?: boolean
+    }
+  ) => () => void
+}
+
+/**
+ * @deprecated use `Mutate<StoreApi<T>, ["zustand/subscribeWithSelector"]>`
```
dai-shi, line 36: Can you modify middlewareTypes.test.tsx to show the usage of `Mutate`? Just removing works?

dai-shi (commented):
Looks like a good migration path!

src/context.ts
devanshj: It is possible but then even if they pass `UseContextStore<MyStore>` (which is the correct newer usage) they'll still see the strikethrough, so they won't be able to distinguish between the types that need to migrated and the types that have already been migrated.

src/middleware/subscribeWithSelector.ts
devanshj: Yep sure will do

src/context.ts
dai-shi: that's not nice. ideally, they should see deprecated mark for the old usage and not for the new usage.

devanshj: Yep but we can't have that, the only way to achieve that is have a seperate type but then we'll lose the original name.

dai-shi: That's unfortunate. I don't think people would notice this change.
Fortunately, `zustand/context` wouldn't be much used compared to `zustand` itself,
so we can make a breaking change in v4 with the notice. Or, we create a new name in v4, if we come up with a neat idea.

devanshj: I'm less inclined to a new name, unless we can have a new name that makes sense. Let's go with this and people who will read the deprecation in release notes of v3.7 would be able to migrate and they'll have less things to migrate when v4 comes. And hey that's how users used to come to know about deprecations before the strikethrough feature, so it's not that bad :P

dai-shi: That's fair. Alternatively, what do you think if we drop this type export in v4? I don't know how many people use it. Do you think of its use case? Yeah, I'm actually inclined to it, if possible. We can put it back if too people complain...

devanshj: I'm a bit reluctant but I think we could do that, even if there are use cases, users can use `typeof context.useStore`.

But I have another idea, what if we export the new type with same name from a temporary `zustand/context/forward-compat` and then mark the on exported from `zustand/context` as deprecated. In this way users will be able to migrated to the newer type and then in v4 we'll tell them to use `zustand/context` instead of `zustand/context/forward-compat`.

In fact as you review you'll find many more such types that we could put in a common forward-compat module.

And alternatively instead of deleting the forward-compat module in v4, we could even delete it in V5

Wdyt?

dai-shi: Hmm, unless there's a huge concern for the migration path, I wouldn't go with `forward-compat`. I think what we have in #772 and #725 is mostly okay. It has some breaking changes, but feels reasonable.

if `typeof context.useStore` works, I would like to deprecate `export type UseContextStore` in v3.7.0 and drop(=not export) it in v4.

devanshj: Yeah also on second thoughts I too think it's okay to drop it in v4.

dai-shi:
So, this looks almost ready.
Can you fix remaining issues and CI?

src/middleware/subscribeWithSelector.ts
devanshj: Hmm the module augmentation seems to be not working, looking into it.

devanshj: Oh lol, it should be `Mutate<StoreApi<T>, [["zustand/subscribeWithSelector", never]]>` instead of `Mutate<StoreApi<T>, ["zustand/subscribeWithSelector"]>` fixed it.

Thinking out loud: We could add a sugar to make `Mutator` same as `[Mutator, never]` ie if the payload is not specified it's `never`. But let's not complicate things already, will think about hit later.

tests/middlewareTypes.test.tsx
```diff
@@ -150,7 +151,7 @@ describe('counter state spec (single middleware)', () => {
       CounterState,
       SetState<CounterState>,
       GetState<CounterState>,
-      StoreApiWithSubscribeWithSelector<CounterState>
+      Mutate<StoreApi<CounterState>, [['zustand/subscribeWithSelector', never]]>
```
dai-shi, line 154: Thanks! Looks good. (I'm sure people will be surprised, though. 😄 )

dai-shi (approved):
Okay, I'm merging this for v3.7.0. If we have more ideas for migration paths, we can release patch versions v3.7.x.

devanshj:
> Okay, I'm merging this for v3.7.0. If we have more ideas for migration paths, we can release patch versions v3.7.x.

If you're in a hurry go for it otherwise I was about to deprecate other middleware store types in favor `Mutate` too, we're already deprecating for `subscribeWithSelector` so it'd be nice if we do it for other middlewares too in one release itself.

dai-shi:
> If you're in a hurry go for it otherwise I was about to deprecate other middleware store types in favor `Mutate` too, we're already deprecating for `subscribeWithSelector` so it'd be nice if we do it for other middlewares too in one release itself.

Oh, I thought it's impossible.
It would be nice if you can. I'll wait for it.

devanshj:
Right now the deprecation message is this...

```ts
/**
 * @deprecated Use `Mutate<StoreApi<T>, [["zustand/subscribeWithSelector", never]]>`.
 * If you have multiple middlewares see the documentation for `Mutate` usage.
 */
```

I think we need to educate users how to use `Mutate` for multiple middlewares (see tests), should we do it in the readme? Or in the release notes (in which case I'd replace "documentation" with "release notes for v3.7.0")?

devanshj:
> And it may sound crazy, but in the `@deprecated` message, let's say "See tests/middlewareTypes.test.tsx" and would you modify multiple middleware usage there?

I'd be fine with that, the readme also anyway tells users to refer to the test file regarding middleware types, so it's not a big deal I guess. And I already have migrated the all tests (including ones with multiple middlewares) to `Mutate`.

src/middleware/persist.ts
```diff
@@ -98,6 +115,9 @@ export type StoreApiWithPersist<S extends State> = StoreApi<S> & {
   }
 }
 
+type Write<T extends object, U extends object> = Omit<T, keyof U> & U
+type Cast<T, U> = T extends U ? T : U
```
dai-shi, line 119: move up please.

src/middleware/redux.ts
```diff
@@ -18,6 +22,26 @@ export type StoreApiWithRedux<
...
+type WithRedux<S, A> = Write<Cast<S, object>, StoreRedux<Cast<A, Action>>>
+
+interface Action {
+  type: unknown
+}
+
+interface StoreRedux<A extends Action> {
+  dispatch: (a: A) => A
+  dispatchFromDevtools: true
+}
```
dai-shi, line 40: actually, this one too. okay, i know it's just a preference, but prefer define thing in advance whenever possible. (at least for me it's easier to read in my brain.)

dai-shi (changes requested):
Cool. I'm already pretty happy with this version.

Please follow the coding convention and fix conflicts, and we should be good to go!

src/middleware/subscribeWithSelector.ts
dai-shi: I think I get what you mean. It's a bit complicated, but DX-wise, it should be much better. `, never` is just too confusing.

src/middleware/redux.ts
devanshj: No problems, my style is actually opposite of yours haha.

src/middleware/subscribeWithSelector.ts
devanshj: I kinda agree, but then I wonder if it'll be complicated if a middleware has a payload eg `Mutate<Store<T>, ["zustand/subscribeWithSelector", ["zustand/persist", Partial<T>]]>` I worry this inconsistency in shape would be even more confusing. I think we should keep it as it is.

dai-shi: I see. yeah, that works.

dai-shi (approved):
Looks great! Let's ship it.

PR #781, opened 2022-02-04
fix(devtools): don't show jsdoc deprecation warning when no options are passed

### LINKED ISSUE #780: 'devtools' is deprecated.
Following the library README I saw how to use devtools to debug my application. I have implemented the devtools just like the guide says but from VSC it tells me that the devtools method is deprecated, is there a new way to use the devtools? Can I have some information about it? Thank you.
Below you will find my code.

```import create from "zustand";
import { devtools } from "zustand/middleware";
export const useStore = create(
  devtools((set: any) => ({
    address: "Connect Wallet", // case: Install wallet, Connect Wallet, Address
    setAddress: (newAddress) => set((state) => ({ address: newAddress })),
    removeAddress: () => set({ address: null }),
    chain: null,
    setChain: (chain) => set((state) => ({ chain })),
    removeChain: () => set({ chain: null }),
  }))
);

package.json :
"dependencies ": {
    "next": "12.0.8",
    "react": "17.0.2",
    "react-dom": "17.0.2",
    "zustand": "^ 3.6.9"
  } 

```
Immagine 2022-02-04 172922

devanshj:
Closes #780

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -41,6 +41,22 @@ export type StoreApiWithDevtools<T extends State> = StoreApi<T> & {
   devtools?: DevtoolsType
 }
 
+export function devtools<
+  S extends State,
+  CustomSetState extends SetState<S>,
+  CustomGetState extends GetState<S>,
+  CustomStoreApi extends StoreApi<S>
+>(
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks. The code looks good to me.

PR #787, opened 2022-02-08
match vanilla StateCreator type in middleware/subscribeWithSelector

toddw:
Match
https://github.com/pmndrs/zustand/blob/0ba3c240078cf61792a4b13ff11d774ecca70a0a/src/vanilla.ts#L50-L52
to make subscribeWithSelector more useful with custom states.

Ran into this when trying to use subscribeWithSelector with
custom middleware that uses immer's WritableDraft on SetState

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -47,9 +47,9 @@ export type StoreApiWithDevtools<T extends State> = StoreApi<T> & {
  */
 export function devtools<
   S extends State,
-  CustomSetState extends SetState<S>,
-  CustomGetState extends GetState<S>,
-  CustomStoreApi extends StoreApi<S>
+  CustomSetState extends SetState<S> = SetState<S>,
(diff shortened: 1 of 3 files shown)


dai-shi (commented):
@devanshj Does this conflict with v3.7.0 and v4?

devanshj:
It does “conflict” with v4 but there already is a conflict without this too, and it won't hurdle with v3.7.0 as it's not changing much.

But the issue I find here is it'd make sense for all middlewares to have similar type, so if you think this is a good change the same change should be made to other middlewares too.

dai-shi:
@toddw I think it's a rare case that you need to specify only one type to `subscribeWithSelector`. What's your case?
Do you want to make this change and maybe others, if this is going to live only for a short period?

dai-shi:
I made changes for all middlewares. Let's merge this for v3.7.0.

toddw:
Looks great! Thanks for updating it

PR #817, opened 2022-02-20
fix(context): revert removing overloads in `UseContextStore`

### LINKED ISSUE #812: useCallback selector loses type inference in version 3.7.0
Just updated from 3.6.9 to 3.7.0 and I'm facing typescript errors in my project because useCallback loses type inference.
Here is an example: https://codesandbox.io/s/tender-scott-9902ek?file=/src/App.tsx
hover `valid` and `invalid` variables.

You can change zustand version to 3.6.9 sandbox to see it working.

devanshj:
Fixes #812

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -11,10 +11,10 @@ import { EqualityChecker, State, StateSelector, UseBoundStore } from 'zustand'
 /**
  * @deprecated Use `typeof MyContext.useStore` instead.
  */
-export type UseContextStore<T extends State> = <U = T>(
-  selector?: StateSelector<T, U>,
-  equalityFn?: EqualityChecker<U>
-) => U
+export type UseContextStore<T extends State> = {
+  (): T
+  <U>(selector: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
+}
 
 function createContext<
   TState extends State,
--- a/tests/context.test.tsx
+++ b/tests/context.test.tsx
@@ -157,8 +157,8 @@ it('throws error when not using provider', async () => {
 it('useCallback with useStore infers types correctly', async () => {
   const { useStore } = createContext<CounterState>()
   function _Counter() {
-    let x = useStore(useCallback((state) => state.count, []))
-    expectAreTypesEqual<typeof x, number>().toBe(true)
+    const _x = useStore(useCallback((state) => state.count, []))
+    expectAreTypesEqual<typeof _x, number>().toBe(true)
   }
 })
 
--- a/tests/context.test.tsx
+++ b/tests/context.test.tsx
@@ -1,6 +1,7 @@
 import {
   Component as ClassComponent,
   ReactNode,
+  useCallback,
   useEffect,
   useState,
 } from 'react'
@@ -151,4 +152,20 @@ it('throws error when not using provider', async () => {
     </ErrorBoundary>
   )
   await findByText('errored')
+})
+
+it('useCallback with useStore infers types correctly', async () => {
+  const { useStore } = createContext<CounterState>()
+  function _Counter() {
+    let x = useStore(useCallback((state) => state.count, []))
+    expectAreTypesEqual<typeof x, number>().toBe(true)
+  }
+})
+
+const expectAreTypesEqual = <A, B>() => ({
+  toBe: (
+    _: (<T>() => T extends B ? 1 : 0) extends <T>() => T extends A ? 1 : 0
+      ? true
+      : false
+  ) => {},
 })


dai-shi (commented):
Thanks @devanshj !
Would you be able to add types test to detect this for the future?

tests/context.test.tsx
```diff
@@ -152,3 +153,19 @@ it('throws error when not using provider', async () => {
   )
   await findByText('errored')
 })
+
+it('useCallback with useStore infers types correctly', async () => {
+  const { useStore } = createContext<CounterState>()
+  function _Counter() {
+    let x = useStore(useCallback((state) => state.count, []))
+    expectAreTypesEqual<typeof x, number>().toBe(true)
```
devanshj, line 161: We have to use this because writing `x: number = useStore(..)` won't test the inference as even when the inference fails `x` will be inferred as `any` and it'll still compile.

dai-shi (approved):
LGTM!

tests/context.test.tsx
dai-shi: Interesting. Nice to learn a new technique.

Just wondered why it was inferred as `any` instead of `unknown` before this PR...

GuskiS:
Can we get this merged?

dai-shi:
I have one or two more PRs to work on to make a release. Meanwhile, you can keep using the codesandbox build.

src/context.ts
```diff
@@ -11,10 +11,10 @@ import { EqualityChecker, State, StateSelector, UseBoundStore } from 'zustand'
...
  * @deprecated Use `typeof MyContext.useStore` instead.
  */
-export type UseContextStore<T extends State> = <U = T>(
-  selector?: StateSelector<T, U>,
-  equalityFn?: EqualityChecker<U>
-) => U
+export type UseContextStore<T extends State> = {
+  (): T
+  <U>(selector: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
+}
```
devanshj, line 17: @dai-shi Shouldn't this have been the following?

```ts
export type UseContextStore<T extends State> = {
  (): T
  // note the "?" after selector
  <U>(selector?: StateSelector<T, U>, equalityFn?: EqualityChecker<U>): U
}
```

As `useContextStore(undefined, myEquals)` is also valid.

Should I correct it in #725 or make a new PR?

tests/context.test.tsx
devanshj: > Just wondered why it was inferred as any instead of unknown before this PR...

Oops sorry I missed this. It's because in `useCallback`'s callback's constraint is `(...a: any[]) => any` as you can see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/548b1d876886a4cc4340e33cf6ca3991782bc0d7/types/react/v17/index.d.ts#L1105). Hence [here](https://tsplay.dev/mAyQvm) you will see `x` is `any` in `useFoov37`. But the correct constraint is for an unknown function is `(...a: never[]) => unknown`, in that case the code [wouldn't even compile](https://tsplay.dev/mp9LgW). (I hope this is what you were asking)

src/context.ts
dai-shi: That causes the DX issue in #812.

https://tsplay.dev/WJq5Vm

We don't recommend/allow passing `undefined` for selector as a practice. So, think it like TS types are correct, and JS impl is just naive. (We could use `arguments`, but it's not worth the size&complexity.)

btw, this is the similar discussion for `combine` in https://github.com/pmndrs/zustand/pull/725#discussion_r845632251. TS types are correct for usage, so no need to be too precise about typing JS impl (if it's not the recommended usage. Could be done with `defineProperty`, but...)

devanshj: That's trivial to fix, we can just add another overload — https://tsplay.dev/WGPAJw — we already have 2 overloads so adding another doesn't make a difference.

I think not supporting `undefined` selector is an unnecessary compromise, what if users want to actually use custom equality function without selector then they'd have to do `useStore(x => x, eq)` which looks less expressive and less intuitive than `useStore(undefined, eq)`.

So I insist that we add another overload, I'm doing it in #725, if you don't like it I'll revert.

And in case of `combine` we're being intentionally unprecise with types to make the inference work, just that I think it's important to inform the users about the unsoundness we have produced, even if it is negligible. (Also shallow merge is good, `defineProperty` would make it more confusing and won't even solve the problem.)

dai-shi: I don't like `useStore(undefined, eq)`, and it's almost always a mistake in the usage. But, having another overload is acceptable for an edge case. Will look at #725 and comment there.

PR #819, opened 2022-02-22
breaking(middleware/devtools): use official devtools extension types

dai-shi:
Wouldn't this be more maintainable?


src/middleware/devtools.ts
```diff
@@ -27,21 +29,19 @@ type StoreSetStateWithAction<S> = S extends { getState: () => infer T }
...
+        date?: boolean
+        regex?: boolean
+        undefined?: boolean
+        nan?: boolean
+        infinity?: boolean
+        error?: boolean
+        symbol?: boolean
+        map?: boolean
+        set?: boolean
+      }
```
dai-shi, line 44: So, this is breaking. Why did we have `options` in the first place?

src/middleware/devtools.ts
```diff
@@ -170,16 +170,14 @@ export function devtools<
...
         : typeof options === 'string'
         ? { name: options }
         : options
 
     let extensionConnector
     try {
-      extensionConnector =
-        (window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
-        (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
+      extensionConnector = window.__REDUX_DEVTOOLS_EXTENSION__
```
dai-shi, line 180: Not sure why we need `top`. The official '@redux-devtools/extension' doesn't seem to have one.
This is breaking too.

dai-shi (commented):
@devanshj Hi, can you check this please?

src/middleware/devtools.ts
devanshj: Probably so that the devtools work in codesandbox etc (as `__REDUX_DEVTOOLS_EXTENSION__` would only be in the parent window)

src/middleware/devtools.ts
devanshj: Not sure why there was the `options` thing but this is a good change. We can deprecate the existing shape and support both shaped old and new in 3.x

devanshj:
(can't comment on an unchanged line)

```diff
- let extension = Object.create(extensionConnector.connect(devtoolsOptions))
+ let extension = (Object.create as <T>(t: T) => T)(extensionConnector.connect(devtoolsOptions))
```

`Object.create` returns `any` but `connect` returns the correct type.

src/middleware/devtools.ts
```diff
@@ -27,21 +29,19 @@ type StoreSetStateWithAction<S> = S extends { getState: () => infer T }
 interface DevtoolsOptions {
```
devanshj, line 29: See if we can use the extracted `EnhancerOptions` type too (We'd also get the jsdoc for free). Something like...

```ts
interface DevtoolsOptions extends Pick<EnhancerOptions, "name" | "serialize"> {
  anonymousActionType?: string
}
```

src/middleware/devtools.ts
```diff
@@ -1,3 +1,5 @@
+import '@redux-devtools/extension'
```
devanshj, line 1: Perhaps you'd know this better just make sure this doesn't add any unused runtime code.

src/middleware/devtools.ts
dai-shi: nice idea. i'll try.

src/middleware/devtools.ts
dai-shi: hmm, let me experiment. (jotai/valtio doesn't do this, so we want to make three consistent.)

src/middleware/devtools.ts
dai-shi: yeah, the final app code should be fine, but we should actually remove in dist. I'll see how I can do it.

dai-shi: e9cf15e this should do.

dai-shi:
> `Object.create` returns `any` but `connect` returns the correct type.

Thanks. 06123b1

Now, I get type errors...

src/middleware/devtools.ts
dai-shi: This works, but now TS users have to install some devDependencies (very counter-intuitive), which isn't very handy. Let's keep it as is.

src/middleware/devtools.ts
```diff
@@ -294,66 +294,70 @@ export function devtools<
...
-            )
-            return
-          }
-          return parseJsonThen<{ type: unknown; state?: PartialState<S> }>(
-            message.payload,
-            (action) => {
-              if (action.type === '__setState') {
-                setStateFromDevtools(action.state as PartialState<S>)
-                return
+    ;(extension as any) // FIXME no-any
```
dai-shi, line 297: It would be nice if we could use this type definition: https://github.com/reduxjs/redux-devtools/blob/b82de745928211cd9b7daa7a61b197ad9e11ec36/extension/src/app/api/index.ts#L522-L536
```ts
export interface ConnectResponse {
  init: <S, A extends Action<unknown>>(
    state: S,
    liftedData?: LiftedState<S, A, unknown>
  ) => void;
  subscribe: <S, A extends Action<unknown>>(
    listener: (message: ListenerMessage<S, A>) => void
  ) => (() => void) | undefined;
  unsubscribe: () => void;
  send: <S, A extends Action<unknown>>(
    action: A,
    state: LiftedState<S, A, unknown>
  ) => void;
  error: (payload: string) => void;
}
```

src/middleware/devtools.ts
dai-shi: https://codesandbox.io/s/cool-sun-bnowv5
`window.top` doesn't work anyway in codesandbox.
image

src/middleware/devtools.ts
dai-shi: #828

src/middleware/devtools.ts
devanshj: Yeah that's fair. One could write a postinstall script that fetches the `.d.ts` files from unpkg and writes it to the filesystem but I think that would be going too far.

src/middleware/devtools.ts
devanshj: Yeah. It's a little weird that `ConnectResponse` in `@redux-devtools/extension` only has `init` and `send`, one could consider opening an issue to write the actual type definition too.

dai-shi: filed: https://github.com/reduxjs/redux-devtools/issues/1097

PR #829, opened 2022-02-24
fix: import use-sync-external-store in esm

zzmp:
Imports `use-sync-external-store/shim/with-selector.js` (with a `.js` suffix) so that it is correctly resolved by esm bundlers like nextjs.
Fixes https://github.com/NoahZinsmeister/web3-react/issues/379, which is broken using [4.0.0-beta.1](https://www.npmjs.com/package/zustand/v/4.0.0-beta.1).

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -1,5 +1,5 @@
 import { useDebugValue } from 'react'
-import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector'
+import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector.js'
 import createStore, {
   EqualityChecker,
   GetState,


zzmp:
Context: https://nodejs.org/api/esm.html#mandatory-file-extensions

dai-shi:
Hi, thanks for your suggestion!
I'm aware of mandatory extension in Node.js ESM.
My hope is use-sync-external-store provides correct "exports", and we don't care the actual suffix.
Would you like to open an issue upstream?

Meanwhile, I think we can merge this PR, but there's no guarantee that this is kept. It might be reverted if any other issues occur.

Also, do you think, in web3-react config, it's possible to use CJS instead of ESM? The community will have a hard time transitioning from CJS to ESM for a while...

zzmp:
I think this is important to merge for interim esm compatibility. I can open an issue upstream as well.

I'm working with @NoahZinsmeister, the author of web3-react, on getting the CJS entries updated, but I'm not sure if he'd want to default to CJS (as opposed to supporting both options).

If you do merge this, could you also release a patch via npm so we're able to use it with esm?

dai-shi:
> I think this is important to merge for interim esm compatibility.

Note that this is just a workaround. `use-sync-external-store` doesn't provide ESM.
I still hesitate to merge this PR. It doesn't seem an ideal solution.
At this point, since `use-sync-external-store` is CJS only, we/you should provide/use CJS version of zustand v4. (or maybe with a wrapper...)

> If you do merge this, could you also release a patch via npm so we're able to use it with esm?

I need to merge something else too. No ETA. I really wish codesandbox ci works. (It's broken for these couple of days.)

I would suggest `patch-package` if you want to solve it on your end immediately.

dai-shi:
Let's see how 64536ee4332193c7cafd4c5f47841d28051ff004 goes.

dai-shi:
Please try: https://www.npmjs.com/package/zustand/v/4.0.0-beta.2

PR #845, opened 2022-03-02
Add devtools action name to typings

cortopy:
When devtools is enabled, `setState` function accepts an optional third argument with a name that will be displayed as the action's type in Redux devtools. This is missing in the types and this PR adds it

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -43,7 +43,8 @@ export type SetState<T extends State> = {
     K4 extends keyof T = K3
   >(
     partial: PartialState<T, K1, K2, K3, K4>,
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks for your suggestion.
However, we shouldn't change the vanilla type, which should be correct without middleware.
This will be improved in v4 with #725.
Meanwhile, please see `tests/middlewareTypes.test.tsx` how to type it manually in various cases.

cortopy:
thanks @dai-shi ! I think I'm now getting the structure of zustand codebase. For anyone searching for this, I found `NamedSet` in the devtools middleware

PR #876, opened 2022-03-20
docs: Fix TypeScript docs for using subscribe with selector

### LINKED ISSUE #875: Docs show deprecated syntax for using subscribe with selector in TypeScript
Looks like `StoreApiWithSubscribeWithSelector` was deprecated but the docs still show its usage

disambiguator:
Fixes #875


dai-shi (approved):
Nice catch. Thanks for your contribution!

PR #880, opened 2022-03-22
fix(middleware/devtools): add `enabled` option and show devtools warning only if enabled

### LINKED ISSUE #852: RN Expo [zustand devtools middleware] Please install/enable Redux devtools extension
I use the latest version of zustand 3.7.1
Screenshot 2022-03-05 at 11 31 43

### LINKED ISSUE #874: Zustand DevTools log in production code
My understanding is that the devtools only load in development and when `NODE_ENV === 'production'` they are not included. That being said, I'm looking at my output JS and I'm seeing the `console.warn` still showing in there even though its around a `__DEV__` condition, which I assume is looking at `NODE_ENV === 'development'` from what I can gather looking through the source. Why would this log still be showing (`[zustand devtools middleware] Please install/enable Redux devtools extension`)? Here is a snipped of my production code where you can see the log(s) are still present:

```js
u=void 0===t?{name:void 0,anonymousActionType:void 0}:"string"==typeof t?{name:t}:t;let s;try{s=window.__REDUX_DEVTOOLS_EXTENSION__||window.top.__REDUX_DEVTOOLS_EXTENSION__}catch{}if(!s)return"undefined"!=typeof window&&console.warn("[zustand devtools middleware] Please install/enable Redux devtools extension"),e(n,r,i);let c=Object.create(s.connect(u)),l=!1;Object.defineProperty(i,"devtools",{get:()=>(l||(console.warn("[zustand devtools middleware] `devtools` property on the store is deprecated it will be removed in the next major.\nYou shouldn't interact with the extension directly. But in case you still want to you can patch `window.__REDUX_DEVTOOLS_EXTENSION__` directly"),l=!0),c),set:e=>{l||(console.warn("[zustand devtools middleware] `api.devtools` is deprecated, it will be removed in the next major.\nYou shouldn't interact with the extension directly. But in case you still want to you can patch `window.__REDUX_DEVTOOLS_EXTENSION__` directly"),l=!0),c=e}});
```

This is my hook:
```js
export const useUserStore = create(
    devtools(set => ({
        ...storeStuff
    }))
);
```

### LINKED ISSUE #878: (middleware/devtools) Remove "Please install/enable Redux devtools" warning
That warning is to let developers know when they use `devtools` middleware, but they forgot to use DevTools Extension. It might help some developers, which is a benefit. However, as reported in #852 (and https://github.com/pmndrs/valtio/pull/378), this is troublesome as it's not 100% accurate. It feels like the trouble is more than the benefit.

As the official redux doesn't seem to have such a warning, let's remove the warning altogether.

(We'll do it in zustand first, and make jotai and valtio to follow it.)

Aslemammad:
Resolves #878 
Closes #852 
Also close #874

This PR adds `enabled` option and, if `enabled=true` shows the warning  and then return `fn` immediately! This resolved the concern described in #878.


dai-shi:
Thanks for working on this.
Now I looked at the change, which is 100% what I expected, I'm thinking if we can support disabling it explicitly partially for https://github.com/pmndrs/zustand/issues/874#issuecomment-1073425378. Will get back to you.

dai-shi:
So, 4b7fc77 is an idea. We have three modes: `true` to enable explicitly, `false` to disable explicitly, and `undefined` to auto detect (using `__DEV__`).

src/middleware/devtools.ts
```diff
@@ -194,17 +195,20 @@ export function devtools<
...
       extensionConnector =
-        (window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
-        (window as any).top.__REDUX_DEVTOOLS_EXTENSION__
+        (typeof enabled === 'boolean' ? enabled : __DEV__) &&
+        ((window as any).__REDUX_DEVTOOLS_EXTENSION__ ||
+          (window as any).top.__REDUX_DEVTOOLS_EXTENSION__)
     } catch {
       // ignored
     }
+    console.log(extensionConnector)
```
dai-shi, line 208: Let's remove it.
```suggestion
```

dai-shi (approved):
should be good to go.

PR #888, opened 2022-04-02
fix: Peer dependency of React is uncapped. Same for dev

Starcounter-Jack:
https://github.com/pmndrs/react-three-fiber/issues/2171


package.json
```diff
@@ -165,8 +165,8 @@
     "json": "^11.0.0",
     "lint-staged": "^12.3.4",
     "prettier": "^2.5.1",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": ">=17.0.2",
+    "react-dom": ">=17.0.2",
```
dai-shi, line 169: Does this help your issue?
I wonder if changing peer dependency version somehow would help.

dai-shi: https://github.com/pmndrs/react-three-fiber/issues/2171
Alright, it's a DEV only issue. I don't think it's rational to change the dev dependency with range. Would be nice to find a better solution.

Starcounter-Jack:
16.8 should be supported. Will submit new pull request.

PR #889, opened 2022-04-02
fix(package.json): Aligned React dependencies 

Starcounter-Jack:
https://github.com/pmndrs/react-three-fiber/issues/2171


dai-shi:
Thanks for working on this.
However, I still don't think this is what we really want in the longer term.
How do other libs solve your issue? I've never seen React hook libs to have wide ranges in devDependencies.
If this has been a common issue, we should have more bug reports like when we updated the dev dependency to `17.0.2` because many developers were still using `16.*.*` at that point.

dai-shi:
Hmm, this doesn't make sense at all, because we actually delete `devDependencies` completely.

https://github.com/pmndrs/zustand/blob/ab82143ff085e846d9c1f76bd9d8bb776d11007f/package.json#L61

You can confirm: https://unpkg.com/browse/zustand@3.7.1/package.json

PR #894, opened 2022-04-06
fix(middleware/persist): Merge storage value with the latest store state

zerofirework:
When using persist middleware, it's possible to mutate the store state in an useEffect callback before hydration creates `stateFromStorage`. It should merge the storage value with the latest store state instead of the config result since it can be out of date.

To reproduce:
1. create a storage whose getItem returns the value after a 2sec timeout. The storage is responsible for persisting a partial state. (property a)
2. add a useEffect to update the state right after render. It adds another property to the store. (property b)

Expected: When the storage value is fetched, it merges storage value with the latest store value.
Actual: When the storage value is fetched, it merges storage value with the initial store value.

Error demo: https://codesandbox.io/s/react-forked-oo16v5?file=/src/App.js
Fixed demo: https://codesandbox.io/s/react-forked-hydp10?file=/src/App.js

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -312,7 +312,10 @@ export const persist =
           }
         })
         .then((migratedState) => {
-          stateFromStorage = options.merge(migratedState as S, get())
+          stateFromStorage = options.merge(
+            migratedState as S,
+            get() ?? configResult
+          )
 
           set(stateFromStorage as S, true)
           return setItem()
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -312,7 +312,7 @@ export const persist =
           }
         })
         .then((migratedState) => {
-          stateFromStorage = options.merge(migratedState as S, configResult)
+          stateFromStorage = options.merge(migratedState as S, get())
 
           set(stateFromStorage as S, true)
           return setItem()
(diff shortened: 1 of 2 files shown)


src/middleware/persist.ts
```diff
@@ -312,7 +312,7 @@ export const persist =
           }
         })
         .then((migratedState) => {
-          stateFromStorage = options.merge(migratedState as S, configResult)
+          stateFromStorage = options.merge(migratedState as S, get())
```
AnatoleLucet, line 315: Maybe this could fix your issue and the tests?

```suggestion
          stateFromStorage = options.merge(migratedState as S, configResult.getState())
```

AnatoleLucet (commented):
Thanks for the PR, but unfortunately this change breaks quite a few tests

src/middleware/persist.ts
zerofirework: `configResult` is the initialState object and it doesn't have a `getState()` method.

zerofirework: Fixed tests and added my own test case.

AnatoleLucet (approved):
Looks good, thanks!

dai-shi (approved):
Thanks for your contribution!

PR #912, opened 2022-04-16
feat(context): add `onDestroy` callback to Provider component

nodkz:
### Problem

My zustand `createStore` function create a lot of subscriptions to window.evenListeners to internal eventBus for micro-frontends. BUT I am not able to unsubscribe from them on Provider unmounting. And it leads to memory leaks in my application. 

So this PR adds the ability for housekeeping on Provider unmounting.
  
 ### Example: setting `bears` equal to 0 if the user press the `Esc` key
  
```tsx
  import create from "zustand";
  import createContext from "zustand/context";

  type BearState = {
    bears: number
    increase: () => void
  }

  // pass the type to `createContext` rather than to `create`
  const { Provider, useStore } = createContext<BearState>();

  const createStore = () => 
    create((set) => {
      // Track ESC key press for setting bears equal to 0
      const onKeyup = (e: KeyboardEvent) => {
        if (e.key === 'Escape') {
          set({ bears: 0 })
        }
      };
      window.addEventListener('keyup', onKeyup);

      return {
        bears: initialBears,
        increase: () => set((state) => ({ bears: state.bears + 1 })),
        removeListeners: () => {
          window.removeEventListener('keyup', onKeyup)
        },
      };
    });

  export default function App({ initialBears }: { initialBears: number }) {
    return (
      <Provider
        createStore={createStore}
        onDestroy={(store) => {
          const state = store.getState();

          // removes previously registered event listeners 
          // for preventing memory leaks
          state.removeListeners();
          
          console.log('Last bears value was:', state.bears);
        }}
      >
        <Button />
      </Provider>
  )
}
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -586,6 +586,61 @@ const Component = () => {
   ```
 </details>
 
+<details>
+  <summary>createContext usage with onDestroy callback for store housekeeping before Provider unmounting</summary>
+
+  ```tsx
+  import create from "zustand";
+  import createContext from "zustand/context";
+
+  type BearState = {
+    bears: number
+    increase: () => void
+  }
+
+  // pass the type to `createContext` rather than to `create`
+  const { Provider, useStore } = createContext<BearState>();
+
+  const createStore = () => 
+    create((set) => {
+      // Track ESC key press for setting bears equal to 0
+      const onKeyup = (e: KeyboardEvent) => {
+        if (e.key === 'Escape') {
+          set({ bears: 0 })
+        }
+      };
+      window.addEventListener('keyup', onKeyup);
+
+      return {
+        bears: initialBears,
+        increase: () => set((state) => ({ bears: state.bears + 1 })),
+        removeListeners: () => {
+          window.removeEventListener('keyup', onKeyup)
+        },
+      };
+    });
+
+  export default function App({ initialBears }: { initialBears: number }) {
+    return (
+      <Provider
+        createStore={createStore}
+        onDestroy={(store) => {
+          const state = store.getState();
+
+          // removes previously registered event listeners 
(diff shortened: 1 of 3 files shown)


dai-shi:
Thanks for the real use case. Good point. We are releasing v4 soon which covers more use cases with context. zustand/context isn’t removed for backward compatibility, but no plan to extend it. I think you can try v4-beta. But let me know if it’s not clear. Docs could be improved.

dai-shi:
v4.0.0-rc.0 is released. Please give it a try. Feel free to open a discussion for questions. Closing this as it's not currently planed, but if there is a huge demand, we may re-consider.

nodkz:
> We are releasing v4 soon which covers more use cases with context. zustand/context isn’t removed for backward compatibility, but no plan to extend it. 

@dai-shi I reviewed all changes between `3.7.2` -> `4.0.0-rc.0` and didn't find any notable changes according to `context` (except type definitions changes). And I didn't find any deprecations for using `zustand/context`.

1) Do you plan to deprecate `zustand/context` in the future? (in v5)
2) Should I create my own wrapper on the vanilla store which will provide needed lifecycles (onDestroy phase for the provider) OR I can rebase my changes on the new v4.0.0 codebase?

dai-shi:
> didn't find any notable changes according to context

It's not about `zustand/context`, but `import { useStore } from 'zustand'`.

> Do you plan to deprecate zustand/context in the future? (in v5)

Yes, that will be fairly possible in v5.

> Should I create my own wrapper on the vanilla store

Yes, that's what I'd suggest. https://github.com/pmndrs/zustand#react-context

nodkz:
Thank s for the reply.  I'll do my own Provider.

PR #929, opened 2022-04-24
Added a createImmerStore function 

vandercloak:
This function joins the create and immer function calls shown in this example into one function. Its a small adjustment, but one that just feels cleaner. Would love to hear any feedback!

Before:
```
import create from "zustand"
import { immer } from "zustand/middleware/immer"

const useStore = create(immer((set) => ({
  bees: 0,
  addBees: (by) => set((state) => { state.bees += by }),
})))
```

After:
```
import create from "zustand"
import { createImmerStore } from "zustand/middleware/immer"

const useStore = createImmerStore((set) => ({
  bees: 0,
  addBees: (by) => set((state) => { state.bees += by }),
}))
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -252,6 +252,22 @@ const clearForest = useStore(state => state.clearForest)
 clearForest();
 ```
 
+or use our `createImmerStore` so you can skip wrapping functions in produce:
+
+```jsx
+import { createImmerStore } from "zustand/middleware/immer"
+
+const useStore = createImmerStore(set => ({
+  lush: { forest: { contains: { a: "bear" } } },
+  clearForest: () => set((state => {
+    state.lush.forest.contains = null
+  })
+}))
+
+const clearForest = useStore(state => state.clearForest)
+clearForest();
+```
+
 [Alternatively, there are some other solutions.](https://github.com/pmndrs/zustand/wiki/Updating-nested-state-object-values)
 
 ## Middleware
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -74,11 +74,11 @@ const immerImpl: ImmerImpl = (initializer) => (set, get, store) => {
   return initializer(store.setState, get, store)
 }
 
+export const immer = immerImpl as unknown as Immer
+
 export function createImmerStore<
   T extends State,
   Mos extends [StoreMutatorIdentifier, unknown][] = []
 >(initializer: StateCreator<T, [...Mos, ["zustand/immer", never]]>) {
   return create<T, Mos>(immer<T, Mos>(initializer) as any);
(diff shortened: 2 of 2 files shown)


dai-shi:
Hi, suggestions are appreciated, but I don't think we need to do this. Whether it's good or not, zustand is middleware oriented. The approach in this pr is what I call store creator utils. If this is something valuable to users, let's create a document as one of best practices. For now, wiki pages are available, so feel free to create one. https://github.com/pmndrs/zustand/wiki
I'd also like to hear what is the base problem you want to solve, because I might be missing some difficulties.

vandercloak:
@dai-shi Thanks for your quick feedback! 

I guess my main question here is that isn't the zustand immer middleware always intended to be used with the zustand create store function?  If so, doesn't a simplified means to use the two together make sense?

You do make a good point were this function may cross the boundary of what could be classified as a middleware, but I am left at a place still thinking this has value. You may be right that this could be better left in the wiki, but that doesn't provide the best means for reuse if this does have value.

dai-shi:
> I guess my main question here is that isn't the zustand immer middleware always intended to be used with the zustand create store function?

Well, this is a bit tricky question. My answer is no.

We didn't have the `immer` middleware in v3, it's only one practice to use zustand. V4 includes it for the first time. But, I have a huge hesitation of adding it. I ended up adding it because typing it isn't trivial and we don't want people asking it.

So, in short, while `immer` is provided, it's not considered as a best practice.

Some random thoughts:
- While zustand is quite small, combining immer makes it very big unless you are already using it.
- At poimandres, we have `valtio` which supports mutation style. I generally recommend people who use zustand+immer to give valtio a try.
- Zustand is based on immutable model, and immer is mutation style. In the mental model, they are conflicting. I'd personally prefer some alternatives that use in immutable model if I were to use zustand: https://github.com/pmndrs/zustand/wiki/Updating-nested-state-object-values
- Immer is popular and people like using it with zustand. As zustand is an unopinionated library, it provides it as an option.

vandercloak:
> * While zustand is quite small, combining immer makes it very big unless you are already using it.

✅ this super insightful. I just checked immer on bundlephobia which I should have done prior to bringing it in and it’s like 6 times the size of zustand. I will check out valtio!

wajeshubham:
@dai-shi does `immer` have an effect on anything else besides increasing the bundle size? I assume this is the reason why `immer` is exported separately and not with the other middlewares.

dai-shi:
immer is based on mutable state model. zustand is based on immutable state model. So, there's a mental model mismatch. For mutable state model, we generally suggest trying [valtio](https://github.com/pmndrs/valtio).

PR #892, opened 2022-04-05
breaking(middleware/devtools): remove deprecations and warnings

devanshj:
remove deprecated features for v4.

PR #933, opened 2022-04-25
breaking(middleware/devtools): remove checking old string option

dai-shi:
This is something we forgot in #892.
Also did some refactor to remove our own options in DevtoolsOptions.


dai-shi:
review request @devanshj

devanshj (approved):
> This is something we forgot in #892.

Ah right. LGTM.

PR #940, opened 2022-04-28
Fix TS4023 and other minor changes

devanshj:


devanshj:
> deb6f4d format(docs): add a linebreak after &lt;summary&gt;

<details><summary>Before & After</summary>

image

image

</details>

src/middleware/immer.ts
```diff
@@ -32,22 +33,21 @@ type SkipTwo<T> = T extends []
   ? A
   : never
 
-type WithImmer<S> = S extends {
+type WithImmer<S> = Write<Cast<S, object>, StoreImmer<S>>
```
dai-shi, line 36: Hmm, I don't understand the trick here for better quickinfo...
It's now `extends` is two level. Okay, so many things to learn to be TS friendly.

dai-shi (approved):
Looks good. Thanks for the fix.

src/middleware/immer.ts
devanshj: If `Foo<T>` is an interface then `Foo<T>` will always be shown in the quickinfo as `Foo<T>`. But if `Foo<T>` is an type alias then `Foo<T>` could be shown as `Foo<T>` or it can be inlined and the type it resolves to can be shown, this depends on how much complex `Foo<T>` is. So if it's complex typescript will tend to not inline it as it might become a huge unhelpful type.

I noticed that typescript thinks `WithDevtools` is complex but `WithImmer` is not (even though both do almost the same thing), so I wrote `WithImmer` just like `WithDevtools` (two levels of type alias) and that did the trick.

> Okay, so many things to learn to be TS friendly.

Haha yeah, being generally observant is the key to learning.

ivanbanov: @devanshj Thank you for explaining it, it was the only place on the internet that gave me a hint that actually led to the fix for it in our project at Miro 👌

devanshj: Haha I'm glad it was helpful!

PR #941, opened 2022-04-29
fix(types): avoid any in react.ts

dai-shi:
There was a `TODO` in react.ts.
I originally thought this requires StoreMutator, but it doesn't actually mutate. So, it's rather simple?

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -10,22 +10,31 @@ import createStore, {
   StoreMutatorIdentifier,
 } from './vanilla'
 
-export function useStore<S extends StoreApi<State>>(api: S): ExtractState<S>
-export function useStore<S extends StoreApi<State>, U>(
+type ExtractState<S> = S extends { getState: () => infer T } ? T : never
+
+type WithReact<S extends StoreApi<State>> = S & {
+  getServerState?: () => ExtractState<S>
+}
+
+export function useStore<S extends WithReact<StoreApi<State>>>(
+  api: S
+): ExtractState<S>
+
+export function useStore<S extends WithReact<StoreApi<State>>, U>(
   api: S,
   selector: StateSelector<ExtractState<S>, U>,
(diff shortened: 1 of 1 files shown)


src/react.ts
```diff
@@ -10,32 +10,39 @@ import createStore, {
...
 export function useStore<TState extends State, StateSlice>(
-  api: StoreApi<TState>,
+  api: WithReact<StoreApi<TState>>,
   selector: StateSelector<TState, StateSlice> = api.getState as any,
   equalityFn?: EqualityChecker<StateSlice>
 ) {
   const slice = useSyncExternalStoreWithSelector(
     api.subscribe,
     api.getState,
-    // TODO avoid `any` and add type only in react.ts
```
dai-shi, line 27: We simply allow specifying optional `getServerState` property. Note this has nothing to do with vanilla.ts.

devanshj (approved):
Looks good.

> I originally thought this requires StoreMutator, but it doesn't actually mutate. So, it's rather simple?

Yep, only middlewares require `StoreMutator`.

src/react.ts
Andarist: Out of curiosity - do you have any examples of how Zustand consumers should utilize `getServerState` nowadays? When I think about it (in the context of just any state management library) it seems quite cumbersome to provide this, especially if a given library supports multiple stores. I wonder if we could figure out a standard way for approaching this problem.

dai-shi: > Out of curiosity - do you have any examples of how Zustand consumers should utilize `getServerState` nowadays?

No. Here, the API is just exposed, hoping to find a way to utilize it somehow.

> I wonder if we could figure out a standard way for approaching this problem.

Definitely. In the zustand case, I was hoping if someone come up with some ideas or concrete use cases. There might be a standard way.

PR #944, opened 2022-05-01
Strict `replace` types

devanshj:
Opening a PR as requested here: https://github.com/pmndrs/zustand/pull/725#issuecomment-1099704025


dai-shi:
Thanks for submitting this.
This is important to have even if we don't merge it soon, as I can't do this.

After noticing #915, I think one of the issue is `Draft` type. We might need to inline it. I still think this is too complicated to realize strict `replace`.

Does anyone else follow this? have questions?

devanshj:
> After noticing https://github.com/pmndrs/zustand/pull/915, I think one of the issue is Draft type. We might need to inline it.

Inlining is not a good idea because we'd have to update it everytime immer updates. I think we'd have to declare immer as a peer dependency. There are other solutions but there are very hacky.

Merge or reject, I'm open to whatever you think is the right thing to do here.

dai-shi:
Closing this, but this is good to know we have such an option. Thanks.

devanshj:
No problems

PR #995, opened 2022-06-06
Update readme.md to reflect Maps and Sets usage

Gabriel-Alves-Cunha:
Hi, there. I was using this lib with React and I found that there is a specific way to use Maps and Sets, so I thought that a little update could help others. What do you guys think?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -529,6 +529,51 @@ const useStore = create<BearState>()(devtools(persist((set) => ({
 
 A more complete TypeScript guide is [here](https://github.com/pmndrs/zustand/blob/main/docs/typescript.md).
 
+## Map and Set Usage
+
+You need to wrap Maps and Sets inside an object, and when you want it's update to be reflected (e.g. in React),
+one way you can do it is calling the setState on it:
+
+```js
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -529,6 +529,51 @@ const useStore = create<BearState>()(devtools(persist((set) => ({
...
+  // use foo/bar...
+  // Here, the `.set()` function will update them, no need to call setState on them:
+  bar.set("bar");
+  foo.set("foo", "bar");
+  foo.delete("foo");
+  
+  // If you want to update some React component that uses `useFooBar`, you have to call setState
+  // to let React know that an update happened:
+  setFooBar({ foo, set });
+  // You never need to create a new Map or Set to update them!
```
dai-shi, line 560: This is not recommended as it's mutating Map/Set. You need to create a new Map/Set to update them to follow React rule.

dai-shi (changes requested):
Besides, we don't want to add more content in readme.md. Please make a doc in docs folder or wiki pages, and put a link in the best practices section.

PR #997, opened 2022-06-07
Create maps-and-sets-usage.md

Gabriel-Alves-Cunha:
Hi, there. I was using this lib with React and I found that there is a specific way to use Maps and Sets, so I thought that a little update could help others. What do you guys think?


docs/maps-and-sets-usage.md
```diff
@@ -0,0 +1,48 @@
...
+```js
+import create from "zustand";
+
+const useFooBar = create(() => ({ foo: new Map(), bar: new Set() }));
+
+const { getState: getFooBar, setState: setFooBar } = useFooBar;
+  
+function doSomething() {
+  // Following React's best practices, you should create a new Map/Set when updating them:
+  const foo = new Map(getFooBar().foo);
```
dai-shi, line 15: Let's do this as it's often written so.
```suggestion
  const foo = new Map(useFooBar.getState().foo);
```

docs/maps-and-sets-usage.md
```diff
@@ -0,0 +1,48 @@
...
+  const bar = new Set(getFooBar().bar);
+  
+  // use foo/bar...
+  
+  bar.set("bar");
+  foo.set("foo", "bar");
+  foo.delete("foo");
```
dai-shi, line 22: Deleting it right after setting??

docs/maps-and-sets-usage.md
```diff
@@ -0,0 +1,48 @@
...
+  
+  // If you want to update some React component that uses `useFooBar`, you have to call setState
+  // to let React know that an update happened:
+  setFooBar({ foo, set });
```
dai-shi, line 26:
```suggestion
  useFooBar.setState({ foo });
```
Should that ☝️ be fine?

I would actually like this pattern.
```js
useFooBar.setState((prev) => ({ foo: new Map(prev.foo).set("newKey", "newValue") }));
```

docs/maps-and-sets-usage.md
```diff
@@ -0,0 +1,48 @@
...
+
+This will not work if you want to tell React there was an update:
+
+```js
+import create from "zustand";
+
+const useFoo = create(() => new Map());
+
+function doSomething() {
+  // everything else corrected to follow the example above...
```
dai-shi, line 38: I think we can repeat code instead. what do you think?

dai-shi (changes requested):
Thanks for the suggestion. Here are some requests for clarifications and some styling preferences.

docs/maps-and-sets-usage.md
Gabriel-Alves-Cunha: Now that you said it, I tested here, and indeed it doesnt fail to update :) I should've used like that.

docs/maps-and-sets-usage.md
```diff
@@ -0,0 +1,22 @@
...
+  
+function doSomething() {
+  // doing something...
+
+  // If you want to update some React component that uses `useFooBar`, you have to call setState
+  // to let React know that an update happened.
+  // Following React's best practices, you should create a new Map/Set when updating them:
+  useFooBar.setState(prev => ({
+    foo: new Map(prev.foo).set("newKey", "newValue"),
+    bar: new Set(prev.bar).set("newKey"),
```
dai-shi, line 19: Set doesn't have `set` method. `add` may not return Set...

dai-shi (commented):
I think it's more valuable if you create a codesandbox example and add a link in "Examples" section in the doc.

Gabriel-Alves-Cunha:
> I think it's more valuable if you create a codesandbox example and add a link in "Examples" section in the doc.

Agreed, should I just as an example in an existing one?

dai-shi:
Please create a new codesandbox and add it in `docs/maps-and-sets-usage.md`.

Fix this too: https://github.com/pmndrs/zustand/pull/997#discussion_r892941784

dai-shi (approved):
Fixed the test. Should be good to go.

PR #1004, opened 2022-06-13
fix(types): use interface instead of type alias for `StoreApi` to fix errors in ts 4.8

devanshj:
See https://github.com/microsoft/TypeScript/issues/49490

I'm not sure exactly what the issue is but somehow TS 4.8 doesn't understand that `StoreApi` is covariant on `T` but with `interface` instead of type alias and with explicit `out` annotation TS now understands and accepts that `T` is indeed covariant which fixes the issue.

I think we should bump to stable 4.7 to use `out` annotation and then commit and publish.


dai-shi (commented):
Is it possible to add a test to catch this? (because otherwise, I might try to use type alias in the future. well, this also makes me think preferring type alias is a bad idea...)

dai-shi:
> You could also try to ship different types for different TS versions using `typesVersions`. Tools like [`downlevel-dts`](https://github.com/sandersn/downlevel-dts) could help you out to achieve this fairly easy.

So, `out` annotation is only available since 4.7, right?

https://github.com/sandersn/downlevel-dts/pull/74 okay, cool. I'll add `downlevel-dts` in zustand. It's already used in jotai and valtio.

devanshj:
> Is it possible to add a test to catch this? (because otherwise, I might try to use type alias in the future. well, this also makes me think preferring type alias is a bad idea...)

The existing tests are enough, in particular [this one](https://github.com/pmndrs/zustand/blob/2b698fb24916141dbb7f1488c3f7bebc6a5134eb/tests/types.test.tsx#L176-L190). But there's no non-hacky way to write a test to check whether or not an interface is being used, nor would I recommend writing such test. And yes, it's good to prefer interfaces instead of type aliases (for performance, error messages, etc).

dai-shi:
slightly off topic:

> it's good to prefer interfaces

A reason to prefer type aliases is that interfaces can be overloaded unexpectedly. (for the same reason, I'd use const arrow function when I don't need it to be overloaded like internal functions, instead of function declaration.) 
Are error messages still good with interfaces? Do you have convincing examples?

Thanks for sharing best practices anyway. I want to keep this PR minimal, and will tackle interfaces in the future (which affects all my too many projects.)

devanshj:
> Are error messages still good with interfaces? Do you have convincing examples?

I don't have any on top of my head, but generally because interfaces are guaranteed to not resolve to their definition (as I had said [here](https://github.com/pmndrs/zustand/pull/940#discussion_r861791454)) it makes debugging a little better. Some other refs: [this tweet by Daniel](https://twitter.com/drosenwasser/status/1319205169918144513), [this section of ts's performance wiki](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections).

Regarding being able to merge interface declarations, one could write a lint rule to warn against it. And of course if external users are wrongly using module augumentation to extend interfaces (ie the following) then it's really on them, just like it's on them when they make any assertions in their codebase.

```ts
declare module "zustand" {
  export interface SomeStuff {
    nonExistentProperty: string
  }
}
```

dai-shi:
> I don't have any on top of my head

I actually find a real issue in jotai: https://github.com/pmndrs/jotai/pull/1234

https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

uh huh.

> one could write a lint rule to warn against it.

That would be great. I wonder if there exists any plugin.

dai-shi:
> I don't have any on top of my head

I actually find a real issue in jotai: https://github.com/pmndrs/jotai/pull/1234

https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

uh huh.

> one could write a lint rule to warn against it.

That would be great. I wonder if there exists any plugin.

devanshj:
> I actually find a real issue in jotai: https://github.com/pmndrs/jotai/pull/1234

Ah yeah I saw that, as I said ts can inline type aliases in the quickinfo, so using interface fixes the issue.

> I wonder if there exists any plugin.

Probably not.

dai-shi:
> I think we should bump to stable 4.7 to use out annotation and then commit and publish.

Let's do this. Please fix this PR.

package.json
```diff
@@ -145,54 +153,55 @@
     ]
   },
   "dependencies": {
-    "use-sync-external-store": "1.1.0"
+    "date-fns": "2.28.0",
+    "use-sync-external-store": "1.2.0"
```
dai-shi, line 157: Can we not update dependencies in this PR? `date-fns` shouldn't be here.

package.json
```diff
@@ -13,6 +13,14 @@
       "*": [
         "ts3.4/*"
       ]
+    },
+    "<4.7": {
+      "esm/*": [
+        "ts4.7/*"
```
dai-shi, line 19: I think we can just have `ts3.4`, no? I mean change `"<4.0"` to `"<4.7"` at L9.

dai-shi (commented):
Thanks for working on this.
Can you only update `typescript` in dependencies?

package.json
devanshj: Some `devDependencies` need to be updated so that they recognise `out` annotation and the build is successful, I wasn't sure which ones in particular are needed so I updated all of them. If you know which in particular are required I can update only them.

And yeah we don't need to update `dependencies` and not sure how `date-fns` ended up there, I'll revert this bit.

package.json
devanshj: I'm actually not really familiar with downlevel-dts but I'm guessing it should look like this...

```json
"typesVersions": {
  "<4.0": {
    "esm/*": [
      "ts3.9*"
    ],
    "*": [
      "ts3.9/*"
    ]
  },
  "<4.7": {
    "esm/*": [
      "ts4.6/*"
    ],
    "*": [
      "ts4.6/*"
    ]
  }
}
```

If we make it the following...

```json
"typesVersions": {
  "<4.7": {
    "esm/*": [
      "ts3.9*"
    ],
    "*": [
      "ts3.9/*"
    ]
  }
}
```

Then users with ts version 4.6 will have to use types of ts version 3.9 when they could have used types with ts version 4.6

Am I getting this right?

Also why do we have <4.0 in the first place? which 4.0 feature we're using? and why it was 3.4 instead of 3.9? Am I missing something?

If we're not using any 4.0 feature per se then I think it'd look like the latter version.

dai-shi: > Then users with ts version 4.6 will have to use types of ts version 3.9 when they could have used types with ts version 4.6

Yes, would it be too much trouble?

> Also why do we have <4.0 in the first place? which 4.0 feature we're using?

I don't remember which, but I added it when I thought we used some 4.0 feature. The reason targeting 3.9 is it's the default minimum version that downlevel-dts supports.

So, we expect people use the latest typescript version, if it's not possible, we offer the lowest TS version as we can.

devanshj: > Yes, would it be too much trouble?

Yeah that would unnecessarily downlevel some features afaiu. For eg let's say we are using some feature from 4.5, the user is using 4.6, but if we downlevel <4.7 to 3.9 then it'll also downlevel the feature were using from 4.5 even if the user has support for it. So I'd say we should downlevel <4.7 to 4.6.

And by that same logic I'd suggest you to consider downleveling <4.0 to 3.9 instead of 3.4. And also consider not downleveling <4.0 at all, at the end of the day if users want they can downlevel themselves.

dai-shi: > Yeah that would unnecessarily downlevel some features

I understand it, but not sure the real problem. It's just non-ideal.
For maintainability, I would like to have a straightforward config for all zustand/jotai/valtio.

> downleveling <4.0 to 3.9 instead of 3.4.

Would that break if people use 3.4? It's rare lately. But, I remember we had issues with old CRAv4 (especially with codesandbox back then), or something.

> at the end of the day if users want they can downlevel themselves.

Not sure if I follow this. How can users do it by themselves?

At this point, I still want it to be <4.7 to 3.4 for long-term simplicity and compatibility.

devanshj: > I understand it, but not sure the real problem. It's just non-ideal.

I'm not exactly sure how downlevel-dts works, for instance at some places we're using `[...A, ...B]` which is not available for 3.4 afaik, so I'm not sure what will downlevel-dts do with it (probably leave it as it is?). The point I'm trying to make is downlevel-dts will mess with our types uncessarily even if users have version as latest as 4.6, that could be a real problem.

> Not sure if I follow this. How can users do it by themselves?

By writing a postinstall script that overwrites .d.ts with downlevel-dts or something like that.

> Would that break if people use 3.4? It's rare lately. But, I remember we had issues with old CRAv4 (especially with codesandbox back then), or something.

It could I guess. That's another problem with `typesVersion`, like not only we have to care about our source code for breaking changes but also the `typesVersion`. Like say if we publish a version with some `typesVersion`, then removing it in the next versions could also be a breaking change.

Frankly speaking I personally wouldn't bother with downleveling my types if it were my package, it's user's responsibility to use the required version or downlevel themselves.

> At this point, I still want it to be <4.7 to 3.4 for long-term simplicity and compatibility.

I'm fine with that, just I think downleveling for users that use very recent ts versions if not latest is a bit problematic.

But anyway now that we have #1029 we don't need `out` annotation xD

dai-shi: > By writing a postinstall script that overwrites .d.ts with downlevel-dts or something like that.

I see. As far as I recall, the reason we introduced downlevel-dts was codesandbox (or cra4) didn't support newer TS back then. So, `postinstall` wouldn't be a reasonable option.

> I think downleveling for users that use very recent ts versions if not latest is a bit problematic.

It makes sense, especially "very recent" part, and I get the point.

> But anyway now that we have #1029 we don't need `out` annotation and hence no downleveling xD

I'll check it. So, if we merge #1029, we don't need #1004?

devanshj: > I'll check it. So, if we merge #1029, we don't need #1004?

Yes in fact #1004 is now actually not the right fix, 4.8 changed between two version `-dev` versions as I mention in #1029. Also if ts fixes the possible bug I mention in #1029 then we won't even need #1029 xD

Issue #1008, opened 2022-06-14
Move the docs from Github's wiki to the ./docs directory


LubomirGeorgiev:
It will be way easier for people to search something in the docs if they are part of the repo itself instead of relying on GitHub to host the docs.

dai-shi:
Yeah, that's my plan too. We have `./docs` folder just recently.
Would you like to help the migration? I think it's good time for us to review the contents thoroughly. Maybe, we should try moving one by one. (one PR per one page).

chann44:
can i help in this

dai-shi:
sure!

chann44:
so i have read about wiki i think i'll need access to wiki am i right 
also do i have to manually  write these files from wiki to ./docs or we can just change the file path or something

dai-shi:
you have read access to wiki. at least you can edit it.
so, steps are:
1) pick one wiki page
2) copy content to ./docs/*.md
3) create a PR
4) improve the content through review
5) then merge
6) loop back to 1.

chann44:
can you assign this to me now

chann44:
i created a pr now see f it's right the format

dai-shi:
@Carnageous is in.

@chann44 Are you still around? If you guys need coordination, please use this issue for discussion.

tbor00:
Hi, here I leave my little contribution [ #1073](https://github.com/pmndrs/zustand/pull/1073)

PR #1014, opened 2022-06-16
Home page of wiki added to docs issue  #1008

chann44:
Added home page to the docs directory


dai-shi:
sorry for confusion. we don't need this for docs.

PR #1016, opened 2022-06-16
docs: 3d party libraries page added to docs

chann44:


dai-shi (changes requested):
`docs/3rd-party-libraries.md` would be better?

dai-shi (approved):

PR #1029, opened 2022-06-23
Mark more methods bivariant

RyanCavanaugh:
Fixes some upcoming build breaks in TypeScript 4.8. I'm not entirely sure the true root cause, but it looks like we're computing the correct variance of more of these types, leading to a (correct) error issued in covariant use sites like this:
```ts
export function useStore<TState extends State, StateSlice>(
  api: WithReact<StoreApi<TState>>, // <- correctly marked as constraint violation
```

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -19,7 +19,7 @@ export type SetState<T extends State> = {
     replace?: boolean | undefined
   ): void
 }['_']
-export type BivariantStateComputer<T> = {
+type BivariantStateComputer<T> = {
   _(state: T): T | Partial<T>
 }['_']
 export type GetState<T extends State> = () => T
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -15,12 +15,9 @@ export interface Subscribe<T extends State> {
 
 export type SetState<T extends State> = {
   _(
-    partial: T | Partial<T> | BivariantStateComputer<T>,
+    partial: T | Partial<T> | { _(state: T): T | Partial<T> }['_'],
     replace?: boolean | undefined
   ): void
-}['_']
-type BivariantStateComputer<T> = {
-  _(state: T): T | Partial<T>
 }['_']
 export type GetState<T extends State> = () => T
 export type Destroy = () => void
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -9,9 +9,9 @@ export type StateListener<T> = (state: T, previousState: T) => void
  * @deprecated Use `StateListener<T>` instead of `StateSliceListener<T>`.
  */
 export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
-export type Subscribe<T extends State> = {
-  _(listener: StateListener<T>): () => void
-}['_']
+export interface Subscribe<T extends State> {
+  (listener: StateListener<T>): () => void
+}
 
 export type SetState<T extends State> = {
   _(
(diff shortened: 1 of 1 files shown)


dai-shi:
@devanshj what's your call?

src/vanilla.ts
```diff
@@ -9,16 +9,19 @@ export type StateListener<T> = (state: T, previousState: T) => void
...
+}['_']
 
 export type SetState<T extends State> = {
   _(
-    partial: T | Partial<T> | ((state: T) => T | Partial<T>),
+    partial: T | Partial<T> | BivariantStateComputer<T>,
     replace?: boolean | undefined
   ): void
 }['_']
+export type BivariantStateComputer<T> = {
```
dai-shi, line 22:
```suggestion
type BivariantStateComputer<T> = {
```

Can we not export this?

RyanCavanaugh:
Here's a minimal version of what changed in 4.8:
```ts
type Action<T> = (state: T) => T;
export type MakeBivariantMethod<T> = {
  signature(arg: T | Partial<T> | Action<T>): void;
}['signature']
export interface StoreApi<T> {
  prop: MakeBivariantMethod<T>;
}
type CheckExtendsStoreApi<T extends StoreApi<unknown>> = T;
type Test<T> = CheckExtendsStoreApi<StoreApi<T>>; // <- New correct error
```

src/vanilla.ts
```diff
@@ -9,16 +9,19 @@ export type StateListener<T> = (state: T, previousState: T) => void
...
 export type SetState<T extends State> = {
   _(
-    partial: T | Partial<T> | ((state: T) => T | Partial<T>),
+    partial: T | Partial<T> | BivariantStateComputer<T>,
     replace?: boolean | undefined
   ): void
 }['_']
+type BivariantStateComputer<T> = {
+  _(state: T): T | Partial<T>
+}['_']
```
devanshj, line 24: Let's inline `BivariantStateComputer`.

```suggestion
export type SetState<T extends State> = {
  _(
    partial: T | Partial<T> | { _(state: T): T | Partial<T> }["_"],
    replace?: boolean | undefined
  ): void
}['_']
```

src/vanilla.ts
```diff
@@ -9,16 +9,19 @@ export type StateListener<T> = (state: T, previousState: T) => void
  * @deprecated Use `StateListener<T>` instead of `StateSliceListener<T>`.
  */
 export type StateSliceListener<T> = (slice: T, previousSlice: T) => void
-export interface Subscribe<T extends State> {
-  (listener: StateListener<T>): () => void
-}
+export type Subscribe<T extends State> = {
+  _(listener: StateListener<T>): () => void
+}['_']
```
devanshj, line 14: This can be left as it is as mentioned [here](https://github.com/pmndrs/zustand/pull/1029#issuecomment-1165717355).

```suggestion
export interface Subscribe<T extends State> {
  (listener: StateListener<T>): () => void
}
```

src/vanilla.ts
dai-shi: Yeah, inlining seems simple enough and more consistent.

dai-shi: Yeah, inlining seems simple enough and more consistent.

dai-shi (commented):
well, i'm not a ts wizard, but it looks good.

dai-shi (approved):
Thanks all for helping with types.

PR #1050, opened 2022-07-03
docs(typescript): Add mutators reference

### LINKED ISSUE #1046: Suggestion from my own confusion
The docs say to use  this `StateCreator<MySlice, [["zustand/devtools", never]], []>` for independent slices when using middleware but the docs don't show how to use it in an example. Couple that with trying to partialize certain things and I believe there should be clarity there. 

If I have two slices (independent) and want to persist certain things from each one of the two slices, it may be helpful to show the intended way to accomplish this. 

Thanks a lot for your time!

devanshj:


dai-shi:
I will wait for a response to https://github.com/pmndrs/zustand/issues/1046#issuecomment-1173177879.

dai-shi (approved):
Thanks for the improvement of docs!

PR #1051, opened 2022-07-03
fix(types): remove `Partial` typo from `persist`

devanshj:
In #725 I fixed `persist` types by making the default persisted state same as the whole state (which reflects the default behavior), but looks like I forgot to remove `Partial` from one place.

This is breaking change, it probably won't break for many people but a quickfix is to pass `partialize: s => s as Partial<typeof s>` in the options.

Remember it's a "quickfix" that is "I don't want more strict types, give me the old types". So the correct fix might be something else depending on the code. And again "might" implies the quickfix could be correct fix, eg the fix I apply for a test in this PR is not a quickfix but a correct fix as the partialized state is truly `Partial<{ count: number }>`.


devanshj:
Not sure why prettier in the CI has a non-zero exit.

dai-shi:
It says:
```
$ prettier '*.{js,json,md}' '{src,tests,docs}/**/*.{ts,tsx,md,mdx}' --list-different
tests/persistSync.test.tsx
error Command failed with exit code 1.
```

Can you run `yarn run prettier`?

devanshj:
Ah so it's expected to exit non-zero because it's not formatted. I didn't catch that. I think it'd be nicer if the CI shows what exactly is ill-formatted (previously it used to do that).

(Also `yarn run prettier` didn't do anything, I had to `yarn run prettier . --write` and discard extra changes, maybe there's a better way)

dai-shi:
I hope it to run this script.

https://github.com/pmndrs/zustand/blob/1f1a42264d95af482d6df568d9b67ecc0014c732/package.json#L71

Maybe it works differently with different yarn versions?

dai-shi (approved):
LGTM

devanshj:
Ah but the PR's package.json doesn't have that prettier script probably that's why it didn't work.

dai-shi:
Oh, then, please merge `main` into this PR branch to make sure if everything is okay.

PR #1055, opened 2022-07-05
docs: reordered quick start and why zustand

qiuandyq:
Moved sections "Why zustand over redux?" and "Why zustand over context?" above the "First create a store" header for users to see the why first rather than code.

Added "Quick start" above "First create a store" for users to know that the code beneath is a quick start on zustand


dai-shi:
@drcmda any thoughts?

dai-shi:
Let's leave it as is. I think it's intentional (those sections are secondary.)

Thanks for your suggestion anyway.

PR #968, opened 2022-05-22
feat(testing): add testing utils and docs

alekangelov:
- Revised docks for testing + added example for tests and best practices
- Added /examples/site and /examples/react-jest <- they symlink back to dist
- Added a .dev rollup config so people can work as they have an example open
- Added dev command to package.json for rollup watch mode and disabled the aliasing so that importing doesn't get fudged up
- Change the config to be injected on build script instead of all the individual builds
- Added setup-tests that automatically mocks the the create function <- uses @testing-library/react - this is up for discussion

Testing md preview:
https://github.com/alekangelov/zustand/blob/main/docs/testing.md

Testing example app (couner + bound counters):
https://github.com/alekangelov/zustand/tree/main/examples/react-jest/src

PR #1056, opened 2022-07-05
fix(deps)(example): incorrect peer dependency version of three

rnike:
Saw a lot of warnings in the console on the demo site ([https://zustand-demo.pmnd.rs/](https://zustand-demo.pmnd.rs/)),

lots of warning

the reason is because the incorrect dependency of `three`,

incorrect deps

downgrade `three` to version `0.120.1` the warnings no longer exist.

Related issue: pmndrs/postprocessing#246

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -18,7 +18,7 @@
     "react-spring": "^8.0.27",
     "react-three-fiber": "^4.2.20",
     "react-three-gui": "^0.1.5",
-    "three": "^0.125.0",
+    "three": "^0.120.1",
     "threejs-meshline": "^2.0.11",
     "web-vitals": "^0.2.2",
     "zustand": "^3.0.0"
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
I'm afraid someone will update it in the future without noticing, but it looks fine for now.

rnike:
Another solution is to upgrade `react-postprocessing`, I've tried but it needs to upgrade a lot of related dependencies(including react & react-dom), and the tree shaking of the `three` package written in `config-overrides.js` also needs a refactor or rewrite, the diff will be massive if I am going to do that, I chose the simplest solution in this PR.

@dai-shi I have a question here, what do you or the team think of upgrading dependencies for the example? Want to know in your opinion if doing this is valuable or not.

dai-shi:
`examples` folder is not actively maintained by me or the team. It's improved by a few contributors including you.
I'm fine with dropping react-app-rewired and customize-cra, which complicate things.
#968 is also related.

rnike:
got it, might take a look when I have time

PR #1066, opened 2022-07-08
build(deps)(example): Bump versions for react and react-dom

rnike:
Bump versions for `react` and `react-dom` in `examples/`, will make the follow-up package upgrade easier.

The site runs without problem after upgrade on both `yarn start` and `yarn build && serve build`, I assume all things are going well.

**NOTE**
Although `import React from 'react'` is no longer needed for the newest version of react, I'm still not able to remove it without errors, perhaps there are some issues in the eslint setting (not sure yet), I'm not fixing this in this PR but maybe will open another PR for it.


dai-shi (approved):
LGTM

PR #1067, opened 2022-07-08
test: Add additional test for the new dom render method on react 18

rnike:
`ReactDOM.render` and `ReactDOM.unmountComponentAtNode` has been deprecated from react 18, added an additional test case for react version 18 new render api.

Using node require and disabled eslint on `require('react-dom/client')` because if I use `import { createRoot } from 'reac-dom/client'` the ci will fail when the react version is lower than 18.

Reference
- `ReactDOM.render` is deprecated
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis

- `ReactDOM.unmountComponentAtNode` has been replaced with `root.unmount `
https://reactjs.org/docs/react-dom.html#unmountcomponentatnode

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -1,6 +1,7 @@
 import {
   Component as ClassComponent,
   ReactNode,
+  version as reactVersion,
   useEffect,
   useLayoutEffect,
   useState,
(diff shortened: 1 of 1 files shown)


tests/basic.test.tsx
```diff
@@ -1,6 +1,7 @@
 import {
   Component as ClassComponent,
   ReactNode,
+  version as reactVersion,
```
dai-shi, line 4: I didn't know this. I should use this in [jotai](https://github.com/pmndrs/jotai) tests.

dai-shi: just did so: https://github.com/pmndrs/jotai/pull/1287

rnike: I like how you use regexp, gonna use that here too.

dai-shi:
can you merge `main` just in case?

rnike:
rebased

dai-shi:
Hmm, if I understand correctly,
https://github.com/pmndrs/zustand/blob/ef46c686411e6a089f8316eebe029cbe675a6cfc/package.json#L165
`@testing-library/react` uses new root api.
https://github.com/testing-library/react-testing-library/releases/tag/v13.0.0

So, I don't think this test is necessary. I assume you had some reason you needed it. Could you explain it? Sorry, if I missed anything.

rnike:
@dai-shi I'm also wondering if is it necessary, but my point is on the other side. 

First, reply to your mention about "`@testing-library/react` uses new root api", in this test case `@testing-library/react` has not been used, it renders component using `react-dom` directly, and this turns out a question which is my point: Is this a proper test?

I've looked through it over again and found out these two unit test (the original one and the one I added) are like testing `useLayoutEffect` from react, not sure why this unit test exists, but testing another library in this library to me is pretty weird.

I'm going to close this pr according to above, by the way, the reason I started making this pr is I saw warrings in `yarn test`.

Screen Shot 2022-07-18 at 10 17 51 AM

dai-shi:
> in this test case `@testing-library/react` has not been used

Oh! I overlooked that. Thanks for pointing out.

This test has been added because there was an issue with uLE, and the test is to check the regression. Though, I'm not totally sure if the test is really meaningful now.

In a sense, your PR was valid to test both with new api and legacy api.
Or, we could just live with the warning and eventually drop the test in the future when say React 19 comes.

PR #1068, opened 2022-07-08
build(deps)(example): remove used dependencies

rnike:
1. Remove the dependencies not used in the example site.
2. Move `@testing-library` packages to devDependencies.


dai-shi (approved):
I didn't confirm, but if it works, the change seems reasonable.

Issue #1064, opened 2022-07-08
Bug in `shallow` when comparing dates


nvie:
Shallow treats dates incorrectly:

```ts
shallow(new Date('1970-01-01'), new Date('2022-07-08'))  // => true 🤯 
```

These dates are obviously not the same.

This is due to the fact that Dates have no enumerable keys, and as such `Object.keys(new Date())` returns `[]`. So it technically even compares true against `[]`:

```ts
shallow(new Date(), [])   // => true 🤯  
```

This is not only true for Dates, but would be a problem for any object that has no enumerable keys.

Even if this isn't considered a bug, it's inconsistent in this way:

```ts
shallow(new Date(), new Date())  // true
shallow([new Date()], [new Date()])  // false
```

In similar vein:

```ts
shallow({}, [])  // true 🤔 
```

dai-shi:
I think it's originally designed for comparing two plain objects (or two arrays).
If you are interested fixing it, you could try it.

However, this makes me feel that we could deprecate `shallow` entirely and educate people to use other libraries such as lodash.

What I would do instead is to restrict what it supports.
```ts
export default function shallow<T extends unknown[], U extends unknown[]>(objA: T, objB: U): boolean
export default function shallow<T extends Record<string, unknown>, U extends Record<string, unknown>>(objA: T, objB: U): boolean
```
TS Playground: https://tsplay.dev/w8E2MN

nvie:
Ah, I see — that makes a lot of sense if those are the only intended use cases. When I saw [these unit tests](https://github.com/pmndrs/zustand/blob/74de9eb6ba665333845986d305b97046b1fd3f91/tests/shallow.test.tsx#L4-L13) they seemed to imply this function was intended to handle more than just the array/object cases and also handle simple object identity. But if it's okay to explicitly declare the only intended use cases are arrays/objects, then I agree that the TypeScript solution is probably the way 👍

PR #1071, opened 2022-07-09
tests(shallow): Compare functions test was added

tbor00:
In this test compare function with shallow


dai-shi (approved):
Hm, I'm not sure if this is a typical usage https://github.com/pmndrs/zustand/issues/1064#issuecomment-1178889160, but it should work.

So, I will need to update #1070 too.

dai-shi:
Please run `yarn run prettier`.

tbor00:
> Hm, I'm not sure if this is a typical usage [#1064 (comment)](https://github.com/pmndrs/zustand/issues/1064#issuecomment-1178889160), but it should work.
> 
> So, I will need to update #1070 too.

Yes, I think it is not a typical use. but it should work. In this case you need to update #1070

Issue #714, opened 2021-12-16
add storage event listener for persist


souporserious:
Hello, I'm working on a project that uses local storage and wondering if it makes sense for the `persist` middleware to subscribe to the window `storage` event if local storage is being used? This would make sure any other open browser windows will get the update. Right now I'm subscribing and parsing through `useEffect` which is fine, but would be cool if it was baked in. I'm happy to work on a PR for this if it sounds good 😄 

```js
React.useEffect(() => {
  function handleStorage(event) {
    if (event.key === 'playground') {
      const { state } = JSON.parse(event.newValue)
      setPosition(state.position)
    }
  }
  window.addEventListener('storage', handleStorage)
  return () => {
    window.removeEventListener('storage', handleStorage)
  }
}, [])
```

dai-shi:
(Linking to another issue in jotai for my reference: https://github.com/pmndrs/jotai/issues/882 )

garganurag89:
Hi, Can we have this feature please jotai has already added this.

dai-shi:
Would you like to work on it??

tbor00:
Hi, i open [PR](https://github.com/pmndrs/zustand/pull/1075) for this

PR #1075, opened 2022-07-10
feature(persist): add method getOptions on persist api and update documentation

tbor00:
This pr can be relationated to:

- Expose getOptions method on persist api
- Update Documentation with examples of storage event listener
- Add test for the new method
- options were given so that this [issue](https://github.com/pmndrs/zustand/issues/714) can be solved

### DIFF
--- a/docs/persisting-store-data.md
+++ b/docs/persisting-store-data.md
@@ -516,14 +516,12 @@ export const useStore = create(
 )
 ```
 
-### How can add storage event listener?
-
-There's a fiew different ways to do this.
+### How can I rehydrate on storage event?
 
 You can use the `persist` api to create your own implementation, similar to what we see below
 
 ```ts
-export const detectStorageEventChange = (name: string, rehydrate: () => Promise<void>) => {
+export const withStorageDOMEvents = (name: string, rehydrate: () => Promise<void>) => {
   const storageEventCallback = (e: StorageEvent) => {
     if (e.key === name && e.newValue) {
       rehydrate()
@@ -538,28 +536,5 @@ export const detectStorageEventChange = (name: string, rehydrate: () => Promise<
 }
 
 const useStore = create(persist(...))
-detectStorageEventChange(useStore.persist.getOptions().name, useStore.persist.rehydrate())
-```
-
-You can also create a custom `useDetectStorageEventChange` hook:
-
-```ts
-export const useDetectStorageEventChange = (name: string, rehydrate: () => Promise<void>) => {
-  useEffect(() => {
-    const storageEventCallback = (e: StorageEvent) => {
-      if (e.key === name && e.newValue) {
-        rehydrate()
-      }
-    }
-
-    window.addEventListener('storage', storageEventCallback)
-
-    return () => {
-      window.removeEventListener('storage', storageEventCallback)
-    }
-  }, [])
-}
-
-const useStore = create(persist(...))
-useDetectStorageEventChange(useStore.persist.getOptions().name, useStore.persist.rehydrate())
+withStorageDOMEvents(useStore.persist.getOptions().name, useStore.persist.rehydrate())
 ```
--- a/docs/persisting-store-data.md
+++ b/docs/persisting-store-data.md
@@ -298,6 +298,18 @@ export const useStore = create(
 
 The persist api enables you to do numbers of interactions with the persist middleware from inside or outside a React component.
 
+### `getOptions`
+
+> Schema: `() => Partial<PersistOptions>`
+
+This method can you get the options of the middleware.
+
+For example, it can be used to obtain the storage name:
+
+```ts
+useStore.persist.getOptions().name
+```
+
 ### `setOptions`
 
 > Schema: `(newOptions: PersistOptions) => void`
--- a/docs/persisting-store-data.md
+++ b/docs/persisting-store-data.md
@@ -525,8 +525,8 @@ You can use the `persist` api to create your own implementation, similar to what
 ```ts
 export const detectStorageEventChange = (name: string, rehydrate: () => Promise<void>) => {
   const storageEventCallback = (e: StorageEvent) => {
-    if (e.key === store.persist.getOptions().name && e.newValue) {
-      store.persist.rehydrate()
+    if (e.key === name && e.newValue) {
+      rehydrate()
     }
   }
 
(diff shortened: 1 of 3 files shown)


AnatoleLucet:
Hey thanks for the PR
Unfortunately, i don't think this is a good solution to the original issue.

I think the event listener should be directly in the `persist` function/middleware and call `hydrate()` in the event handler.

tbor00:
> Hey thanks for the PR Unfortunately, i don't think this is a good solution to the original issue.
> 
> I think the event listener should be directly in the `persist` function/middleware and call `hydrate()` in the event handler.

Hi Thanks for your comment, I don't know if I understood you correctly, do you think something like this will work?

Captura de Pantalla 2022-07-10 a la(s) 20 10 07

AnatoleLucet:
Yes pretty much, although i'm not sure what `options.subscribe.keySubscribe` is. Maybe you meant `options.name`?
Also what's the removeEventListener used for? It feels like we're only reacting to the first storage change and omitting future ones.

tbor00:
> Yes pretty much, although i'm not sure what `options.subscribe.keySubscribe` is. Maybe you meant `options.name`? Also what's the removeEventListener used for? It feels like we're only reacting to the first storage change and omitting future ones.

You're right, the changes are already up. I'm waiting for more comments, Thanks

dai-shi:
`persist` middleware supports AsyncStorage in ReactNative, and possibly other non-DOM storages. So, you can't make it depend on `window`. Can you please make it an option somehow? (Note that a possible method to check `typeof window` is not recommended because some tools mock it unexpectedly.)

tbor00:
> `persist` middleware supports AsyncStorage in ReactNative, and possibly other non-DOM storages. So, you can't make it depend on `window`. Can you please make it an option somehow? (Note that a possible method to check `typeof window` is not recommended because some tools mock it unexpectedly.)

Hmm, You're right. So I'll work on finding the best way to do this.

tbor00:
> `persist` middleware supports AsyncStorage in ReactNative, and possibly other non-DOM storages. So, you can't make it depend on `window`. Can you please make it an option somehow? (Note that a possible method to check `typeof window` is not recommended because some tools mock it unexpectedly.)

@dai-shi Do you think adding such a function would be

Captura de Pantalla 2022-07-10 a la(s) 22 39 11
 worth it?

dai-shi:
No, I wouldn't recommend "detecting" anything in the library code. We should just provide a capability (option/method) so that developers can use as they want.

tbor00:
> No, no recomendaría "detectar" nada en el código de la biblioteca. Simplemente deberíamos proporcionar una capacidad (opción/método) para que los desarrolladores puedan usarla como quieran.


And a boolean in the options that if it is true execute this block? do you think it could work?

dai-shi:
I think I'd like to avoid any DOM related code in `persist` middleware which is just useless for RN users. A separate helper function can be an option.

```js
import { someHelperFunctionWithDOM } from 'zustand/middleware/persist'

const useFooStore = create(persist(
  ...,
  {
    someOption: someHelperFunctionWithDOM(...),
  }
))
```

Note that we could have other "subscribe" methods than DOM one. For example, one I can imagine is a storage based on firestore. The point is `persist` middleware is not only for `localStorage`/`sessionStorage`.

tbor00:
@dai-shi  Ok, so I think something like what's in the first commit might work

dai-shi:
Yeah, something like that. There are still many things to consider, but I think you first want to try creating a working demo?

AnatoleLucet:
@dai-shi what about:
```ts
// impl
const withStorageDOMEvent = (store) => {
  window.addEventListener("storage", () => {
    // TODO: add .getOptions method on persist's api
    if (e.key === store.persist.getOptions().name && e.newValue) {
      store.persist.rehydrate()
    }
  })

  return // TODO: removeEventListener
}

// usage
import { withStorageDOMEvent } from "zustand/middleware"

const useMyStore = create(persist(...))

withStorageDOMEvent(useMyStore)
```

tbor00:
> @dai-shi what about:
> 
> ```ts
> // impl
> const withStorageDOMEvent = (store) => {
>   window.addEventListener("storage", () => {
>     // TODO: add .getOptions method on persist's api
>     if (e.key === store.persist.getOptions().name && e.newValue) {
>       store.persist.rehydrate()
>     }
>   })
> 
>   return // TODO: removeEventListener
> }
> 
> // usage
> import { withStorageDOMEvent } from "zustand/middleware"
> 
> const useMyStore = create(persist(...))
> 
> withStorageDOMEvent(useMyStore)
> ```

I think what you propose sounds good too. Here I leave an [example](https://codesandbox.io/s/suscribeeventchangewithdom-example-forked-lql95t?file=/src/App.js:0-1133) of how it could be used with the changes I made

dai-shi:
@AnatoleLucet 's idea looks good. I wasn't aware this feature should be with persist api, instead of options. (To clarify, my point was only about that `persist` middleware shouldn't have storage specific code.)

tbor00:
@dai-shi  @AnatoleLucet  Hello, I don't know if you saw the example that I left you. If you want I can also make an example like the one you comment on.

[CODESANDBOX](https://yfxbv5.csb.app/)

tbor00:
@AnatoleLucet @dai-shi 

I already made the last changes. It works well in both ways, which of the two do we leave? 
Also, to be honest I'm not the best with typescript. Could you give me some reference on how to type it correctly?

Example:
[codesandbox](https://2qimn3.csb.app/)

tbor00:
Hi @AnatoleLucet , were you able to review the examples?

AnatoleLucet:
After thinking a bit about it, i think we shouldn't include this function for the same reasons we didn't include a [`useHydration`](https://github.com/pmndrs/zustand/wiki/Persisting-the-store's-data#how-can-i-check-if-my-store-has-been-hydrated) hook.
As Daishi said, we don't want to lean into a specific context (web, mobile, server, etc.), so i'm thinking about only exposing the `getOptions` getter and leave the implementation to the user (with an example in the doc). After all, that's why we added the persist api :slightly_smiling_face: 

Thoughts @tbor00 @dai-shi?

tbor00:
If it's OK, I think we can expose the getOptions and let developers make their own functions to detect changes. We could document an example of a hook for this in the documentation, what do you think? @AnatoleLucet @dai-shi

tbor00:
ready, you can check it again? @AnatoleLucet @dai-shi

dai-shi (approved):
LGTM. Good work. Thanks for your contribution.

Please update PR title&description.

AnatoleLucet (changes requested):

AnatoleLucet (approved):
Looks good, thanks for taking the time to make the changes!

PR #1057, opened 2022-07-05
Migrate docs from wiki: auto generating selectors

Carnageous:
Like you said in #1008 you want to migrate the docs from Github's wiki to the docs folder as markdown files and I'd like to help! This one is for [Auto generating selectors](https://github.com/pmndrs/zustand/wiki/Auto-Generating-Selectors).
I will work on all the pages in my fork and open PRs if that's fine?

A few things:
-  I removed some semicolons from the wiki page in the last commit (aabaaad1b92a7b33b0aa07803856b2fb42bbdfc3). These semicolons are added back in by prettier for some reason
- The prettier and eslint scripts do not run on windows for me, but they ran fine in a ubuntu wsl.
- I did not test a lot of the code provided in the examples. Should I do so or is it save to assume that it's still valid?

Looking forward to your suggestions

PR #1078, opened 2022-07-12
docs: update createSelectors example function to match V4 types

Thanaen:
UseBoundStore now accepts only one generic, whereas it used to accept two. I updated the createSelectors function in the doc to work with V4.


dai-shi:
https://github.com/pmndrs/zustand/pull/1057#issuecomment-1180483383
@Carnageous would also be working on this? In any case, we want to update the codesanbox demo too.

Carnageous:
Yep, I wanted to update both but did not yet came to it. Can do later today

Issue #964, opened 2022-05-18
`createSelectors` types do not work with new typescript interfaces


lukesmurray:
**Current Create Selectors** from the wiki

```ts
import create, { StateCreator, State, StoreApi, UseStore } from 'zustand'

interface Selectors<StoreType> {
  use: {
    [key in keyof StoreType]: () => StoreType[key]
  }
}

function createSelectors<StoreType extends State>(
  store: UseStore<StoreType>,
) {
  ;(store as any).use = {}

  Object.keys(store.getState()).forEach(key => {
    const selector = (state: StoreType) => state[key as keyof StoreType]
    ;(store as any).use[key] = () => store(selector)
  })

  return store as UseStore<StoreType> & Selectors<StoreType>
}
```

**My attempt to rewrite**

My attempt doesn't preserve the store type. Any help rewriting this correctly would be appreciated!

```ts
import { State, StoreApi, UseBoundStore } from "zustand";

interface Selectors<StoreType> {
  use: {
    [key in keyof StoreType]: () => StoreType[key];
  };
}

export function createSelectors<StoreType extends State>(
  store: UseBoundStore<StoreApi<StoreType>>
) {
  (store as any).use = {};

  Object.keys(store.getState()).forEach((key) => {
    const selector = (state: StoreType) => state[key as keyof StoreType];
    (store as any).use[key] = () => store(selector);
  });

  return store as UseBoundStore<StoreApi<StoreType>> & Selectors<StoreType>;
}

```

devanshj:
This should probably work... The idea is to "lift" the generic from state type to store type. (Feel free to use more `any`s in the implementation)

```ts
import { State, StoreApi, UseBoundStore } from "zustand";

type WithSelectors<S> = S extends { getState: () => infer T }
  ? S & { use: { [K in keyof T]: () => T[K] } }
  : never

type CreateSelectors =
  <S extends UseBoundStore<StoreApi<State>>>
  (s: S) =>
    WithSelectors<S>

const createSelectors: CreateSelectors = _store => {
  type S = typeof _store
  type T = S extends { getState: () => infer T } ? T : never

  let store = _store as WithSelectors<S>
  store.use = {}
  for (let k of Object.keys(store.getState())) {
    ;(store.use as any)[k] = () => store(s => s[k as keyof typeof s])
  }

  return store
}
```

dai-shi:
If you have further questions, please open a [new discussion](https://github.com/pmndrs/zustand/discussions/new).

PR #1079, opened 2022-07-12
fix(docs): Auto generating selectors docs updated to match v4 syntax

Carnageous:
I updated the docs and the [code sandbox](https://codesandbox.io/s/zustand-auto-generate-selectors-9i0ob3) to match the v4 syntax as suggested by @devanshj in https://github.com/pmndrs/zustand/issues/964#issuecomment-1132854020

(Feel free to compare to the other PR opened https://github.com/pmndrs/zustand/pull/1078#issuecomment-1181784002)

I also noticed that the readme seems to be out of date regarding the `create()(...)` syntax? I'd point out the syntax change in the readme (if it is a breaking change)


Carnageous:
I ran prettier but apparently the check still failed. Not sure what's going on

docs/auto-generating-selectors.md
```diff
@@ -11,27 +11,28 @@ However, writing these could be tedious, but you can auto-generate them
...
 
-  Object.keys(store.getState()).forEach((key) => {
-    const selector = (state: StoreType) => state[key as keyof StoreType]
-    ;(store as any).use[key] = () => store(selector)
-  })
+const createSelectors: CreateSelectors = _store => {
+  type S = typeof _store
+  type T = S extends { getState: () => infer T } ? T : never
+
+  let store = _store as WithSelectors<S>
```
devanshj, line 29: Let's make it more compact

```suggestion
let store = _store as WithSelector<typeof _store>
```

devanshj:
> (Feel free to compare to the other PR opened https://github.com/pmndrs/zustand/pull/1078#issuecomment-1181784002)

My version is more correct, so this PR should supersede #1078 (the PR is appreciated though)

docs/auto-generating-selectors.md
devanshj: You also might want to merge the type declaration and implementation in case that's the convention followed elsewhere...

```ts
const createSelector = <S extends UseBoundStore<StoreApi<State>>>(_store: S) =>
...
```

It's just that I was commenting on the issue and my style is to separate types and implementations.

Carnageous:
Sorry for pinging, but I cannot re-request review. Your comments are adressed @devanshj

docs/auto-generating-selectors.md
```diff
@@ -44,11 +39,11 @@ interface BearState {
   increment: () => void
 }
 
-const useStoreBase = create<BearState>((set) => ({
+const useStoreBase = create<BearState>()((set) => ({
   bears: 0,
   increase: (by) => set((state) => ({ bears: state.bears + by })),
-  increment: () => set((state) => state.increase(1)),
-}))
+  increment: () => set((state) => ({ bears: state.bears += 1 }))
```
devanshj, line 45: It doesn't matter much but let's do `+` instead of `+=` to avoid unnecessary mutation or confusions

devanshj:
> Sorry for pinging, but I cannot re-request review. Your comments are adressed @devanshj

Thanks, looks good.

> Also I still cannot get prettier/eslint to run properly for me.

Not sure about this, @dai-shi could help with that.

docs/auto-generating-selectors.md
Carnageous: I removed the use of 'increase' here just to make the example more simple, but for sure can add it back!

dai-shi: Should be this in that case.
```suggestion
  increment: () => set((state) => ({ bears: state.bears + 1 }))
```

dai-shi:
> Also I still cannot get prettier/eslint to run properly for me.

somebody says the script doesn't run on windows. we should fix it...

dai-shi (approved):

PR #1083, opened 2022-07-13
docs(readme): symbol supplement in code example

xlboy:
Located at _`Reading/writing state and reacting to changes outside of components`_

Supplement the ending symbol of the `Component` function 😄

---
image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -171,9 +171,9 @@ unsub1()
 useStore.destroy()
 
 // You can of course use the hook as you always would
-function Component() {
+const Component = () => {
   const paw = useStore((state) => state.paw)
-}
+  ...
 ```
 
 ### Using subscribe with selector
@@ -241,13 +241,14 @@ The subscribe function allows components to bind to a state-portion without forc
 ```jsx
 const useStore = create(set => ({ scratches: 0, ... }))
 
-function Component() {
+const Component = () => {
   // Fetch initial state
   const scratchRef = useRef(useStore.getState().scratches)
   // Connect to the store on mount, disconnect on unmount, catch state-changes in a reference
   useEffect(() => useStore.subscribe(
     state => (scratchRef.current = state.scratches)
   ), [])
+  ...
 ```
 
 ## Sick of reducers and changing nested state? Use Immer!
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -172,7 +172,8 @@ useStore.destroy()
 
 // You can of course use the hook as you always would
 function Component() {
-  const paw = useStore(state => state.paw)
+  const paw = useStore((state) => state.paw)
+}
```
dai-shi, line 176:
```suggestion
  // ...
```
Wouldn't it better if we were to add one line?

xlboy: > Wouldn't it better if we were to add one line?

I was thinking about something, Is it necessary to add corresponding hints? (`(Other codes in the Component)`)
After adding it, although it looks quite obvious. But it may bring about the situation of "too much content, not concise". 🤔
```ts
// ... (Other codes in the `Component`)
```
or
```ts
// ...
```

xlboy:
> This was intentional by the author to reduce lines, but since some people get confused, I'm fine to add it. Is this only this place? No other places with similar pattern?

This is one of the places I found when I read the document. I don't know if there will be any other places. Maybe you can wait for me to take a look.

xlboy:
I found three other similarities. ↓

image
---
image

---

They seem to be a little inconsistent. `Component` is created using `const` and `function` respectively. 

`const ` the function created contains the ending symbol, And `...`

I wonder, Is it necessary to unify a style? 🤔

dai-shi:
I think several people changed it by chance.
Let's make everything consistent. My proposition is
```jsx
const Component = () => {
  const { slice } = useStore()
  ...
```
style.

dai-shi (changes requested):
https://github.com/pmndrs/zustand/pull/1083#issuecomment-1183296705

xlboy:
See if there is anything else to add. :)

dai-shi (commented):
Please find other `function` components such as `App` and `BearCounter`. What do you think?

xlboy:
> Please find other `function` components such as `App` and `BearCounter`. What do you think?

I think it's ok, In this case, let all components in document be unified as `cosnt`? (readme.md, docs/*.md)

dai-shi (approved):
Hm, changing other files would be out of the scope of this PR. So, let's merge what we have now.

xlboy:
> Hm, changing other files would be out of the scope of this PR. So, let's merge what we have now.

OK, I will start another PR later. Thank you very much for your assistance this time. 😁

PR #1085, opened 2022-07-14
export the WithReact type

blendsdk:
Please export the `WithReact` type so we can use zustand store types in our own typed stores.
Extending from `StoreApi<State>` is not possible without `WithReact` being exported

For example:

    export type TTranslationStore<KeyIndex = any> = UseBoundStore<ITranslator<KeyIndex>>;

    export interface ITranslator<KeysIndex> extends WithReact<StoreApi<State>> {
        .....
    }

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -12,7 +12,7 @@ import createStore, {
 
 type ExtractState<S> = S extends { getState: () => infer T } ? T : never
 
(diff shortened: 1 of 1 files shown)


dai-shi:
> Extending from `StoreApi<State>` is not possible without `WithReact` being exported

Hm, there might be my misunderstanding.

PR #1086, opened 2022-07-14
feat(example): Upgrade threejs related packages to the latest version

rnike:
I've upgraded the `three` related packages to the latest version, put a lot of effort into reading the changelogs of each package because honestly I'm not so familiar with `three`. 

Reviewing this PR by commit will be easier, I've written down some notes in each of the commit messages, hope it will help speeding up the review.

I removed the tree-shaking stuff for 'three' in this PR, if we really need it I prefer doing this with another approach (maybe `pmndrs/react-three-babel`).

Found a weird thing during the upgrade, the color displays differently on some monitors after the upgrade (with the same code), it has been fixed in the commit witch changes the `gl_FragColor` in `layerMaterial.js`, I've tested it on all my devices (even on my Television), if it can be double confirmed by anyone else would be wonderful.

Although I've tried a lot still cannot make it looks 100% the same as before. The resolution of textures has improved after the upgrade, turning out the white of textures becomes brighter, and also I couldn't make the blur effect as same as before, or maybe I'm just missing something out.


dai-shi (approved):
Changes look good to me. It's cleaner.

@tbor00 would you like to review?

tbor00 (approved):
I like it, it looks much cleaner

PR #1090, opened 2022-07-16
doc(types): Remove independent slices pattern

### LINKED ISSUE #1046: Suggestion from my own confusion

devanshj:
Closes #1046

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -345,40 +345,7 @@ const useStore = create<
 })))
 ```
 
-### Independent slices pattern
-
-```ts
-import create, { StateCreator } from 'zustand'
-
-interface BearSlice {
-  bears: number
-  addBear: () => void
-}
-const createBearSlice: StateCreator<BearSlice, [], []> = (set) => ({
-  bears: 0,
-  addBear: () => set((state) => ({ bears: state.bears + 1 })),
-})
-
-interface FishSlice {
-  fishes: number
-  addFish: () => void
-}
-const createFishSlice: StateCreator<FishSlice, [], []> = (set) => ({
-  fishes: 0,
-  addFish: () => set((state) => ({ fishes: state.fishes + 1 })),
-})
-
-const useStore = create<BearSlice & FishSlice>()((...a) => ({
-  ...createBearSlice(...a),
-  ...createFishSlice(...a),
-}))
-```
-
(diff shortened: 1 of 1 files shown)


dai-shi:
Should we somewhere have interdependent slices pattern you showed in https://github.com/pmndrs/zustand/issues/1046#issuecomment-1185977777?

devanshj:
> Should we somewhere have interdependent slices pattern you showed in [#1046 (comment)](https://github.com/pmndrs/zustand/issues/1046#issuecomment-1185977777)?

Yes it's already there in the doc, I've only removed the independent version not interdependent. (And the one I showed in the issue is same as the one in the doc).

docs/typescript.md
```diff
@@ -345,40 +345,7 @@ const useStore = create<
...
-  ...createFishSlice(...a),
-}))
-```
-
-If you have some middlewares then replace `StateCreator<MySlice, [], []>` with `StateCreator<MySlice, Mutators, []>`. Eg if you're using `devtools` then it'll be `StateCreator<MySlice, [["zustand/devtools", never]], []>`. See the ["Middlewares and their mutators reference"](#middlewares-and-their-mutators-reference) section for a list of all mutators.
-
-Also you can even write `StateCreator<MySlice>` instead of `StateCreator<MySlice, [], []>` as the second and third parameter have `[]` as their default value.
-
-### Interdependent slices pattern
+### Slices pattern
```
dai-shi, line 348: Ah, I see. Got it now.

dai-shi (approved):

rothsandro:
The link to the #independent-slices-pattern chapter that is now broken:
https://github.com/pmndrs/zustand/wiki/Splitting-the-store-into-separate-slices

Should this link to #slices-pattern instead?

devanshj:
> Should these links refer to #slices-pattern instead?

Yes

dai-shi:
Feel free to update the wiki.

PR #1092, opened 2022-07-16
style(examples): run prettier on the examples

rnike:
1. Add directory `examples` and file types `js` and `jsx` to the prettier script
2. Remove deprecated config `jsxBracketSameLine`
3. Create and add `/examples/build` to `.prettierignore` because we don't want to prettier the production code
4. `yarn prettier`


.prettierignore
```diff
@@ -0,0 +1 @@
+/examples/build
```
dai-shi, line 1: Can you configure to run prettier only for `examples/src` instead?

examples/package.json
```diff
@@ -39,7 +39,6 @@
     "semi": false,
     "trailingComma": "es5",
     "singleQuote": true,
-    "jsxBracketSameLine": true,
```
dai-shi, line 42: We have `"bracketSameLine": true,` in other projects. Could you add it?

package.json
```diff
@@ -68,8 +68,8 @@
     "build:shallow": "rollup -c --config-shallow",
     "build:context": "rollup -c --config-context",
     "postbuild": "yarn copy",
-    "prettier": "prettier '*.{js,json,md}' '{src,tests,docs}/**/*.{ts,tsx,md,mdx}' --write",
-    "prettier:ci": "prettier '*.{js,json,md}' '{src,tests,docs}/**/*.{ts,tsx,md,mdx}' --list-different",
+    "prettier": "prettier '*.{js,json,md}' '{examples,src,tests,docs}/**/*.{js,jsx,ts,tsx,md,mdx}' --write",
+    "prettier:ci": "prettier '*.{js,json,md}' '{examples,src,tests,docs}/**/*.{js,jsx,ts,tsx,md,mdx}' --list-different",
     "eslint": "eslint --fix '*.{js,json}' '{src,tests}/**/*.{ts,tsx}'",
     "eslint:ci": "eslint '*.{js,json}' '{src,tests}/**/*.{ts,tsx}'",
```
dai-shi, line 74: I wonder if we can run eslint for `examples/src`. If it's not trivial, never mind.

dai-shi (changes requested):
Thanks so much for working on this. A few suggestions:

package.json
rnike: It does work by simply adding `examples/src` and `js,jsx` in the line, I'd like to look it through later and do this in another pull request.

examples/config-overrides.js
```diff
@@ -1,6 +1,5 @@
 const { override } = require('customize-cra')
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
 
-module.exports = override(
-  //addWebpackPlugin(new BundleAnalyzerPlugin())
-)
+module.exports = override()
+//addWebpackPlugin(new BundleAnalyzerPlugin())
```
dai-shi, line 5: let's revert this one as it looks weird.
```suggestion
module.exports = override(
  //addWebpackPlugin(new BundleAnalyzerPlugin())
)
```

dai-shi: btw, I'm fine to remove the plugin entirely. furthermore, remove customize-cra entirely.

package.json
dai-shi: sounds good.

dai-shi (approved):

examples/config-overrides.js
rnike: Done.

I'm currently working on adding `@react-three/babel` which is still using `customize-cra`, this file on my branch is now like

```js
// examples/config-overrides.js
const { addWebpackPlugin, override, addBabelPlugin } = require('customize-cra')
const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin

const overrides = [
  addBabelPlugin('module:@react-three/babel'),
  process.env.BUNDLE_ANALYZE === 'true' ? addWebpackPlugin(new BundleAnalyzerPlugin()) : undefined,
]

module.exports = override(...overrides.filter((config) => config))
```

remove `customize-cra` then I'll need to add it back or have another approach, I'm leaving it there if no other reason to remove it.

dai-shi: if you need it, no problem.

PR #1096, opened 2022-07-17
perf(example): Reduce bundle size

rnike:
Have done some work to reduce the bundle size of the example website

#### 1. `build:analyze` script

The script `build:analyze` is added to help visualize the build bundle, and also bumped the `webpack-bundle-analyzer` version to latest.

To use:

> yarn build:analyze

#### 2. @react-three/babel

Add `@react-three/babel` to reduce bundle size, this babel plugin is required using [createRoote](https://docs.pmnd.rs/react-three-fiber/api/canvas#custom-canvas) from `@react-three/fiber` to render the canvas or it will not work ([document](https://github.com/pmndrs/react-three-babel#limitations)).

#### 3. Unnecessary package`prismjs` removal
Remove unnecessary `prismjs` package because the using theme can be import from `prism-react-renderer` instead.

### The result

The bundle gzip size reduces 8.54 kB by this pr

result

by the way, the bundle gzip size from [my first contribution commit](https://github.com/pmndrs/zustand/tree/b06a8a017d92ed58ce5fb2f14c13b88b37d2b215) was over 380 kB

b06a8a017d92ed58ce5fb2f14c13b88b37d2b215

### DIFF
--- a/examples/src/components/CodePreview.js
+++ b/examples/src/components/CodePreview.js
@@ -1,12 +1,10 @@
 import Highlight, { defaultProps } from 'prism-react-renderer'
 import CopyButton from './CopyButton'
-import 'prismjs'
-import 'prismjs/components/prism-jsx.min'
-import 'prismjs/themes/prism-okaidia.css'
+import theme from 'prism-react-renderer/themes/okaidia'
 
 export default function CodePreview({ code, ...props }) {
   return (
-    <Highlight {...defaultProps} className="language-jsx" code={code} language="jsx" theme={undefined}>
+    <Highlight {...defaultProps} className="language-jsx" code={code} language="jsx" theme={theme}>
       {({ className, style, tokens, getLineProps, getTokenProps }) => (
         // define how each line is to be rendered in the code block,
         // position is set to relative so the copy button can align to bottom right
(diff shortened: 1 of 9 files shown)


dai-shi (commented):
Looks okay to me.
Can someone review?

dai-shi:
`yarn add @babel/runtime` fixed it.

https://zustand-demo-lq07ve8ws-pmndrs.vercel.app/

Hmm, the font doesn't look good.

## previous one

image

## new one

image

@rnike Is this intentional?

rnike:
no, it must be a UI bug, let me take a look

examples/src/components/CodePreview.js
```diff
@@ -1,12 +1,10 @@
 import Highlight, { defaultProps } from 'prism-react-renderer'
 import CopyButton from './CopyButton'
-import 'prismjs'
```
dai-shi, line 3: maybe we can't remove this?

rnike: You're right, I believe is the last commit which is causing this issue, going to revert it.

rnike: Looks like it needs additional configuration after the `prismjs` removal to keep the font the same, but I don't have time on it now, the pr #1110  is made to revert this. I'll come back later for it, sorry for the bug.

PR #1101, opened 2022-07-18
fix(types): persist middleware interfaces

### LINKED ISSUE #1100: Store factory typings
So I'm trying to create a store factory which will essentially adds just one helper method to the initial state object:

```
import create, { type State } from "zustand";
import { combine, persist } from "zustand/middleware";

export const createPreferencesStore = <TState extends State>(initialState: TState, storageKey = "preferences") =>
  create(
    persist(
      combine(initialState, (setState) => ({
        setPreference: <TKey extends keyof TState>(key: TKey, value: TState[TKey]) => {
          setState({ [key]: value } as unknown as Partial<TState>);
        },
      })),
      { name: storageKey }
    )
  );
```

Before the last update everything worked fine, but now I have an error: _"TS4023: Exported variable 'createPreferencesStore' has or is using name 'StorePersist' from external module ".../node_modules/zustand/middleware/persist" but cannot be named."_ I understand that you have removed the StorePersist type from the exports? Is there any way to type the factory? I also think that the "as unknown as Partial<TState>" is not the correct way to type the function.

dai-shi:
close #1100.

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -11,7 +11,7 @@ export interface StateStorage {
   removeItem: (name: string) => void | Promise<void>
 }
 
-interface StorageValue<S> {
+export interface StorageValue<S> {
   state: S
   version?: number
 }
@@ -77,7 +77,7 @@ export interface PersistOptions<S, PersistedState = S> {
 
 type PersistListener<S> = (state: S) => void
 
-interface StorePersist<S extends State, Ps> {
+export interface StorePersist<S extends State, Ps> {
   persist: {
     setOptions: (options: Partial<PersistOptions<S, Ps>>) => void
     clearStorage: () => void


tbor00 (approved):

devanshj:
@dai-shi I think it's better to revert #1012 and use type aliases for the internal types, we don't want to export types that user is not supposed to import directly. (Also I think there should be an option of "--inlineMissingInterfaces" which makes interfaces that are not exported act like type aliases while emitting declarations)

I missed the fact people use `--declarations` which needs all interfaces to be exported. (Although I don't follow why they use `--declarations` for apps/products that aren't going to imported, usually it makes sense to use `--declarations` only for libraries)

dai-shi:
> I think it's better to revert #1012 and use type aliases for the internal types,

Yeah, it sounds better. I thought we could always prefer interfaces...

dai-shi:
closing in favor of #1106.

PR #1110, opened 2022-07-19
fix(example): unexpected font changes

rnike:
This fix the unexpected font changes, the issue is described in #1096.


dai-shi:
Looks good now. 
image

PR #1117, opened 2022-07-20
fix(shallow): types with useStore again

### LINKED ISSUE #1107: types error after update to zustand rc1 => rc2
Hello

I used rc1 for a while without an issue. After updating to rc2 all my functions where I am using `shallow` are breaking:

```
const villages = useStore(state => state.villages, shallow)

TS2322: Type 'Village[]' is not assignable to type 'Record<string, unknown>'.   Index signature for type 'string' is missing in type 'Village[]'.
```
Are there additional steps necessary to make this work?

dai-shi:
close #1107 again

### DIFF
--- a/src/shallow.ts
+++ b/src/shallow.ts
@@ -1,12 +1,12 @@
-function shallow<T extends (...args: any[]) => any>(objA: T, objB: T): boolean
-
-function shallow<T extends string | number | boolean>(objA: T, objB: T): boolean
-
-function shallow<T extends any[]>(objA: T, objB: T): boolean
-
-function shallow<T extends Record<string, any>>(objA: T, objB: T): boolean
-
-function shallow<T>(objA: T, objB: T) {
+function shallow<
+  T extends
+    | string
+    | number
+    | boolean
+    | ((...args: any[]) => any)
+    | any[]
+    | Record<string, any>
+>(objA: T, objB: T) {
   if (Object.is(objA, objB)) {
     return true
   }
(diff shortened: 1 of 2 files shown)


dai-shi:
I just noticed `Record<string, any>` accepts `new Date()`.

image

So, #1108 reverted the change for #1064 anyway. Can't be helped. 🤷

src/shallow.ts
```diff
@@ -1,12 +1,7 @@
-function shallow<T extends (...args: any[]) => any>(objA: T, objB: T): boolean
-
-function shallow<T extends string | number | boolean>(objA: T, objB: T): boolean
-
-function shallow<T extends any[]>(objA: T, objB: T): boolean
-
-function shallow<T extends Record<string, any>>(objA: T, objB: T): boolean
-
-function shallow<T>(objA: T, objB: T) {
+function shallow<T extends string | number | boolean | Record<string, any>>(
```
devanshj, line 1: Why not also add `unknown[]` and even `undefined | null | symbol | bigint`? `Object.is` can compare them afaik.

dai-shi: I actually wanted to make it to accept only array and object, because it should cover 90% use cases.
Using `shallow` for string only is unnecessary. `string | number | boolean` is added to avoid breaking existing tests.

All that said, this doesn't work as I hoped in #1064, and I added a test a07a7d9, so I'm actually okay to revert it.

```suggestion
function shallow<T>(
```

PR #1129, opened 2022-07-25
perf(example): remove unnecessary imports of "prismjs"

rnike:
This reduces 8.54kb of the bundle size for the example site.

reduce
 
Cannot use the new theme from `'prism-react-renderer/themes/okaidia'` although they have the same name `okaidia`, the style still looks different. Followed [the introduction](https://github.com/FormidableLabs/prism-react-renderer#faq) and used the old Prism css theme, removed unnecessary imports of "prismjs".

The output is as same as before

compare


dai-shi (approved):
LGTM

PR #1144, opened 2022-07-27
feat(vanilla): non-object state

dai-shi:
and simplify types (eliminating `Cast`).

---

There might not be many use cases, but maybe we can support non-object state.

Counter example: https://codesandbox.io/s/react-typescript-forked-uvyys3?file=/src/App.tsx

```tsx
import create from "zustand";

const useCount = create(() => 0);
const inc = () => useCount.setState((c) => c + 1);

const Counter = () => {
  const count = useCount();
  return (
    <div>
      {count} <button onClick={inc}>+1</button>
    </div>
  );
};
```

### TODOs

- [x] vanilla types&impl
- [x] react types
- [x] middleware
  - combine: object only
  - [x] devtools: theoretically accepts non-object
  - [x] immer: theoretically accepts non-object
  - [x] persist: supports non-object
  - [x] redux: theoretically accepts non-object
  - [x] subscribeWithSelector: supports non-object
- [x] tests/basic.test.tsx


src/vanilla.ts
```diff
@@ -5,7 +5,7 @@ type SetStateInternal<T> = {
   ): void
 }['_']
 
-export interface StoreApi<T extends object = object> {
+export interface StoreApi<T = any> {
```
dai-shi, line 8: Can this be `= unknown`?

dai-shi:
@devanshj Can you help typing middlewares? This is not in hurry.

devanshj:
> @devanshj Can you help typing middlewares? This is not in hurry.

I think things should change much, this would mainly involve removing `extends State` from places. Give it a try and then if you're finding it difficult then let me know and I'll help.

src/vanilla.ts
devanshj: It should be just `StoreApi<T>`, the default type parameter is not needed.

dai-shi:
Yeah, I already tried and lost confident, especially around `Cast`.
I can try that again and ask for review if that's easier.

src/vanilla.ts
dai-shi: We use it for `extends StoreApi` in other places. Should `T` be `any` or `unknown` there?

devanshj:
> Yeah, I already tried and lost confident, especially around Cast.

Ah okay then I'll look into it hopefully soon.

> I can try that again and ask for review if that's easier.

Yeah if you can tell me specific places where you got stuck then I can solve it and explain it to you and it will also make you understand the types better.

src/middleware/persist.ts
```diff
@@ -312,14 +312,14 @@ declare module '../vanilla' {
   }
 }
 
-type Write<T extends object, U extends object> = Omit<T, keyof U> & U
+type Write<T, U> = Omit<T, keyof U> & U
 type Cast<T, U> = T extends U ? T : U
 
 type WithPersist<S, A> = S extends { getState: () => infer T }
-  ? Write<S, StorePersist<Cast<T, object>, A>>
+  ? Write<S, StorePersist<Cast<T, unknown>, A>>
```
dai-shi, line 319: @devanshj Does it make sense to keep `Cast`? Did it even make sense previously?

dai-shi: > Did it even make sense previously?

Ah, it didn't pass because of `extends object` constraint.

dai-shi: okay, I think it's fine to drop Cast entirely.

devanshj: Yes, we casted `T` to `object` so that we can pass it to `StorePersist` which required an `object`. So now that it doesn't require `object` we can simply drop the casting and just pass `T`.

src/vanilla.ts
devanshj: `unknown`. Also ideally one should write `StoreApi<unknown>` and maybe we can make the type parameter required in v5.

dai-shi: Thanks. It was my wrong change in #1122 then. I consider this as bug fix.

src/middleware/persist.ts
dai-shi: Good. I feel like it's more maintainable now.

src/vanilla.ts
devanshj: Ah I see so it was accidentally introduced it in #1122, and it's not that I missed removing it in #725 haha.

> I consider this as bug fix

Yes, but it'd still be a breaking change for v4.0.0 users (not v3.x.x), so would you wait for v5 or publish it in v4.0.1?

dai-shi: If people use it like `<StoreApi>`, it's breaking, but it's totally unexpected and "wrong" usage per se. So, I think it's good to "fix" this in v4.

devanshj: It still would technically violate semver, perhaps one could argue it doesn't violate on basis that users are not supposed to rely on undocumented behavior ie "wrong usage" as you said, nevertheless it's worth violating if at all it's a violation.

src/middleware/redux.ts
```diff
@@ -1,29 +1,20 @@
...
 
-type WithRedux<S, A> = Write<Cast<S, object>, StoreRedux<Cast<A, Action>>>
+type WithRedux<S, A> = Write<S, StoreRedux<A>>
 
-type Redux = <
-  T extends object,
-  A extends Action,
-  Cms extends [StoreMutatorIdentifier, unknown][] = []
->(
+type Redux = <T, A, Cms extends [StoreMutatorIdentifier, unknown][] = []>(
```
devanshj, line 17: @dai-shi Is removing the constraint in `A` intentional? This would allow non-redux actions (and non-redux reducers), which might not work well with `devtools`. I don't think it's a big deal to remove the constraint, but I'd prefer to keep it.

devanshj:
Let's also change the `State` type and the deprecation documentation...

```diff
 /**
- * @deprecated Use `object` instead of `State`
+ * @deprecated Use `unknown` instead of `State`
  */
- export type State = object
+ export type State = unknown
```

This should be probably be non-breaking as we have changed `object` to `unknown` everywhere. If you want to keep `State` as `object` then it's worth updating the deprecation documentation at least.

src/middleware/redux.ts
dai-shi: Yes, this is intentional. (to remove `Cast`, which is more important for me.)
For 99% use cases, there should be no issues. btw, we have a type guard and work around to force `{ type: unknown }`.

devanshj: > to remove `Cast`, which is more important for me.

Instead of `Cast<A, Action>` we can write `A & Action` too. We just have to satisfy the type-checker. Both `Cast<A, Action>` and `A & Action` are subtype of `Action` (which is what the type-checker wants), and when `A` is subtype of `Action` (which it'll always be) both resolve to just `A`.

> btw, we have a type guard and work around to force `{ type: unknown }`.

That makes it worse, right now we're sending `isObjectWithTypeProperty(x) ? x : { type: x }` that means if someone has an action `{ kind: "INCREMENT" }` (which is now allowed) what will get sent is `{ type: { kind: "INCREMENT" } }`.


> For 99% use cases, there should be no issues.

That's true. It's not a huge problem, it's just a matter of correctness. And I'm fine eitherway.

dai-shi: > if someone has an action `{ kind: "INCREMENT", by: 10 }` (which is now allowed)

I implicitly and intentionally make it loose. Redux Devtools allows `{ type: { kind: "INCREMENT", by: 10 } }`, right?
(So, I really wanted to make `State=unknown` and `Action=unknown`.)

devanshj: > Redux Devtools allows `{ type: { kind: "INCREMENT", by: 10 } }`, right?

Yes but the action is `{ kind: "INCREMENT", by: 10 }` and not `{ type: { kind: "INCREMENT", by: 10 } }`. So when the user jumps or dispatches an action from the redux devtools, what we'll receive from the extenstion is `{ type: { kind: "INCREMENT", by: 10 } }` which we'll dispatch and forward to the reducer but the reducer will simply won't recognise it.

The point I'm trying to make is all tooling, libraries, code around redux is written keeping in mind that the action will have a `type` property, so we can't allow violating the redux spec.

I personally see no point in violating the spec just for removing `& Action` from our code. But I'm fine if you're making that choice conciously. In that case maybe let's send `typeof x === "object" ? x : { type: x }` instead.

dai-shi: > which we'll dispatch and forward to the reducer but the reducer will simply won't recognise it.

Ah, I overlooked that.

Hm, I thought "our" redux middleware doesn't need to follow the `type` property spec, but this your comment makes me change the idea. Let me see how `& Action` works.

dai-shi: #1206

PR #1152, opened 2022-07-28
fix typo in typescript.md

garrett-hopper:
Not sure if `useBoundStore` had some special meaning in other places, but there were incorrect references to it in the `combine` comment that didn't match the example. (I just replaced it with `useBearStore` everywhere assuming it was a typo.)

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -12,7 +12,7 @@ interface BearState {
   increase: (by: number) => void
 }
 
-const useBearStore = create<BearState>()((set) => ({
+const useBoundStore = create<BearState>()((set) => ({
   bears: 0,
   increase: (by) => set((state) => ({ bears: state.bears + by })),
 }))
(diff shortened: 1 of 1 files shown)


dai-shi:
`useBoundStore` was intentional but I knew it's confusing. I'm fine with `useBearStore`, but do you mean the animal bear 🐻 , right?

garrett-hopper:
Ah, ok, it may just need to be changed to `useBoundStore` here: https://github.com/pmndrs/zustand/pull/1152/files#diff-ae03a9f2eb638a25c1d8d7ba853c78220f1ed44a19ad97a8d780cbc0c00a4260R124
The "But be a little careful" comment below it references `useBoundStore` rather than `useBearStore` as is in the `combine` example. (I'm guessing to match the `BearState` 🐻)
You could keep `useBoundStore` for the first and last change I'd made or even just change the `useBearStore` example to be `useBoundStore` matching its comment. 🤷
Feel free to close this and update it as you see fit.

garrett-hopper:
https://github.com/pmndrs/zustand/blob/main/docs/typescript.md?plain=1#L124

dai-shi:
> it may just need to be changed to useBoundStore here

Can you update the PR with your new suggestion?

garrett-hopper:
Done; now only the references that are actually incorrect have been updated. (Still feels weird to have both `useBoundStore` and `useBearStore` names as examples, but 🤷)

dai-shi (approved):
LGTM
The naming of `useBoundStore` is pretty new in #1125. I'm not confident either. Might be a better idea, or just an education problem.

PR #1163, opened 2022-08-02
Fix grammar

satvikpendem:
Change to "If you are a TypeScript user" from "If you are TypeScript users" on README.


dai-shi (approved):
Thanks! I don't know if the grammar was wrong, but can merge.

Feel free to fix other docs (readme.md and docs/*.md).

PR #1165, opened 2022-08-02
update persist merge and migrate types

charkour:
Hi @dai-shi, @devanshj, and @AnatoleLucet,

Thank you for the hard work on v4 of zustand. I have a question and proposal about the persist middleware option types.

Currently, the `merge` and `migrate` callback options are both access `persistedState` in the callback and are typed as `unknown`. However, we know that the `persistedState` is the type `PersistedState` in this context. Since the type is a contract and we cannot know for sure the state returned from the storage engine will perfectly match the structure of `PersistedState`, I've opted to wrap it in `Partial` so the user of the persist middleware will check to make sure the fields exist before access.

What do you think of this proposal? Is there a better way to leverage typescript when using the persist middleware `merge` and `migrate` options? Thank you for your time.

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -62,12 +62,15 @@ export interface PersistOptions<S, PersistedState = S> {
    * A function to perform persisted state migration.
    * This function will be called when persisted state versions mismatch with the one specified here.
    */
-  migrate?: (persistedState: unknown, version: number) => S | Promise<S>
+  migrate?: (
+    persistedState: Partial<PersistedState>,
+    version: number
+  ) => S | Promise<S>
   /**
    * A function to perform custom hydration merges when combining the stored state with the current one.
    * By default, this function does a shallow merge.
    */
-  merge?: (persistedState: unknown, currentState: S) => S
+  merge?: (persistedState: Partial<PersistedState>, currentState: S) => S
 }
 
 type PersistListener<S> = (state: S) => void


devanshj:
> Since the type is a contract and we cannot know for sure the state returned from the storage engine will perfectly match the structure of PersistedState, I've opted to wrap it in Partial so the user of the persist middleware will check to make sure the fields exist before access.

That's precisely why it's `unknown`, which is more correct/strict that `Partial<PersistedState>`.

For instance let's say the `PersistedState` is `{ count: number }`, now `Partial<{ count: number }>` becomes `{ count?: number }` but even that is a lie, we can't be sure if `count` exists it must be `number`, it's possible someone changed the storage to contain `{ version: 0, state: { count: "hello" } }` in that case the following code—which compiles—will give a runtime error making it unsound...

```ts
declare let persistedState: Partial<{ count: number }>

if (persistedState.count) {
  persistedState.count.toFixed()
  // Uncaught TypeError: persistedState.count.toFixed is not a function
}
```

The value that comes from storage can be literally anything, not only object but even string for example, because it's possible that the storage was compromised to `{ version: 1, state: "gotcha" }`. Hence the type of `persistedState` is `unknown` because it's actually "unknown".

> However, we know that the persistedState is the type `PersistedState` in this context.

So no it's not `PersistedState` but rather `unknown`.

And the correct way to go from `unknown` to your expected type is by parsing it with something like [zod](https://github.com/colinhacks/zod), see https://github.com/pmndrs/zustand/discussions/1142#discussioncomment-3260338

Also as I said it in the answer, you should not only parse the shape of your current persisted state, but rather the shape of all previous version too... So for example, your parsing code might look like the following if your current version is `2` (ie you're passing `{ version: 2 }` in the option)...

```ts
const v0PersistedStateSchema =
  z.object({ featureA: z.number() })

// featureB was added in v1 of our app
const v1PersistedStateSchema =
  z.object({ featureA: z.number(), featureB: z.number() })
  
// featureA was removed in v2 of our app
const v2persistedStateSchema =
  z.object({ featureB: z.number() })

const migrate = persistedStateUnchecked => {
  // When the user opens our app it's possible the last version they opened was either 0, 1 or 2
  // So the uncompromised persisted state is not just { featureB: number } but even 
  // { featureA: number } would be uncompromised, all versions are valid
  // hence we'd parse a union of all
  
  let persistedState:
    | z.infer<v0PersistedStateSchema>
    | z.infer<v1PersistedStateSchema>
    | z.infer<v2PersistedStateSchema>;
    
  try {
    persistedState = z.union([
      v0PersistedStateSchema,
      v1PersistedStateSchema,
      v2persistedStateSchema
    ]).parse(persistedState)
  } catch (e) {
    // if it doesn't matches any of the previous version it's compromised
    // handle what to do here and what to return
    return /* ? */;
  }
  
  if (persistedState.featureA !== undefined) delete persistedState.featureA
  if (persistedState.featureB === undefined) persistedState.featureB = initialValueOfFeatureB
 
  return persistedState as z.infer<v2PersistedStateSchema>
}
```

So even the parsed state is not `PersistedState`, it's `V0PersistedState | V1PersistedState | ... | VCurrentPersistedState`

Also this again is a simplistic version, you will have to parse things much more correctly. For instance you're persisting a user id which is a string, then `z.object({ userId: z.string() })` won't make the cut, because even `{ userId: "lolol" }` will parse, so you'll need this...

```ts
z.object({
  userId: z.string().refine(async id => {
    // check in the database if the id is legit,
    // if yes return true or else return false.
  })
})
```

And again I'm telling what is the most correct way to do things, being correct is not being righteous, for some situtation it could be righteous to not write the most correct code but rather correct enough and ship stuff.

Okay as usual I went too deep into shit xD—the main point being the type of `persistedState` is and should be `unknown`.

charkour:
Hi @devanshj, 

Thank you for the quick reply! I appreciate the thoroughness of your response and your explanation of soundness makes sense. I learned something new today!

The zod discussion is helpful, and I'll consider switching to a runtime check. 

I can tell you have a passion for this subject! Thank you for sharing your knowledge with me (and others)!

devanshj:
> I can tell you have a passion for this subject!

Haha that's true!

Thanks for your kind words, and I'm glad I could help and share the little I know.

PR #1177, opened 2022-08-08
[Documentation] Add detail example for "Slice Pattern"

detautama:
Add detailed examples for "Slice Pattern" and "Slice Pattern" with "Middlewares"

Check here to view changed documentation
https://github.com/pmndrs/zustand/blob/1609447555b224bf218ff220aae2069965939143/docs/typescript.md#slices-pattern

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -349,16 +349,21 @@ const useBearStore = create<
 ```
 
 ### Slices pattern
+You can divide your store into different files if your store getting bigger and harder to maintain.
 
+#### Example
+In this example, I want to separate "Bears" store and "Fish" store, and then bounding two of them into one store.
 ```ts
-import create, { StateCreator } from 'zustand'
+// store-bears.ts
+import { StateCreator } from "zustand";
+import { FishSlice } from "./store-fish";
 
-interface BearSlice {
-  bears: number
-  addBear: () => void
-  eatFish: () => void
+export interface BearSlice {
+  bears: number;
+  addBear: () => void;
+  eatFish: () => void;
 }
-const createBearSlice: StateCreator<
+export const createBearSlice: StateCreator<
   BearSlice & FishSlice,
   [],
   [],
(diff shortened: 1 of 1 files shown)


dai-shi:
Can you run `yarn run prettier` to fix CI?

detautama:
> Can you run `yarn run prettier` to fix CI?

Yup, nothing fixed
image

dai-shi:
Hmm, looks like it's not working well. Let me try.

dai-shi (approved):
LGTM
@devanshj Can you check if it's correct, and it's aligned with the intention of the doc?

devanshj:
I appreciate the PR but I think this is not the best way to go about it, I'll explain why.

I see you were inspired to make this PR based on the difficulty you encountered in https://github.com/pmndrs/zustand/discussions/1176. I understand the “I got stuck so I don't want someone else to get stuck” sentiment. But it's important to know whether it's “The document was missing something” or “I was missing something”.

Here the case is latter... Because in some sense the document was not targeted to you, the document is meant for people who already know what is slices pattern and how to use middlewares (ie they know already that you need to compose `devtools` with `create`)

This document only documents **additional typescript instructions** required to use zustand, it's not a document for teaching zustand or it's patterns.

I think the problem is that the readme (and some other docs too) link to the typescript document for slices pattern...

> You may wonder how to organize your code for better maintenance: [Splitting the store into separate slices](https://github.com/pmndrs/zustand/blob/main/docs/typescript.md#slices-pattern)

Instead there should be a seperate document for slices pattern (just like there used to be a wiki) or a section in readme. So the content of this PR will be more apt there.

It's important to keep this document succinct, minimal, and typescript-only because it'll be mostly read by people who have already read the readme ie the (javascript) zustand usage, so a tutorial-like prose is not apt here.

Just my two cents.

dai-shi:
> I think the problem is that the readme (and some other docs too) link to the typescript document for slices pattern...

Yeah, I think this is weird.

Let's do that.
Would @detautama or anyone be able to work on a new doc for slice pattern? I prefer it to explain it in JS first and then TS.

Issue #710, opened 2021-12-11
Proposal: Encoding store mutations type-level


devanshj:
Middlewares in zustand can and do mutate the store, which makes it impossible to type them with without some workaround, this proposal tries to encode mutations type-level to make zustand almost 100% type-safe and still extensible.

I'll walk you through how this came to be, because I think it makes one understand how it works and why certain elements of it are necessary.

PR #662 already implements this proposal so you can try it out if you want. And [here](https://tsplay.dev/NlxjeN)'s the final playground to try things out.

## Index

This is less of a "table of content" and more of an "index", meaning the whole thing is single-read and there are no independent "sections" per se, they are made as such to aid the reader to track their progress because this document is rather too long to be a single-read.

- Independent mutations
  - Forwarding mutations up the tree
  - Forwarding mutations down the tree
  - Ordered writes
  - Subtractive mutations
- Dependent mutations
  - Higher kinded mutations
  - Higher kinded mutators
- Zustand-adapted version
- To-dos
  - Caveats
  - Others
- Action point

## Independent mutations

Consider this scenario (I'll be making the api simpler only for the explanation, there are no api changes required)...

```js
const withA = (a, f) => store => {
  store.a = a,
  return f(store);
}

const withB = (b, f) => store => {
  store.b = b,
  return f(store);
}

let storeOut = create(
  withB("b",
    withA("a",
      storeIn =>
        ({ count: 0 })
    )
  )
)
```

Here both `storeOut` and `storeIn` should be of type `Store<unknown> & { a: string } & { b: string }` (we don't care about inferring the state `{ count: number }` for now)

So how would we approach this? First let's take a simpler version

```js
const withA = (a, f) => store => {
  store.a = a,
  return f(store);
}

let storeOut = create(
  withA("a",
    storeIn =>
      ({ count: 0 })
  ),
)
```

One could begin with something like...

```ts
// https://tsplay.dev/w6XJ6m

declare const create:
  <T, M>
    ( initializer:
        & ((store: Store<T>) => T)
        & { __mutation?: M }
    ) =>
    Store<T> & M

type Store<T> =
  { get: () => T
  , set: (t: T) => void
  }
  
declare const withA: 
  <A, T>
    ( a: A
    , initializer:
        (store: Store<T> & { a: A }) => T
    ) =>
      & ((store: Store<T>) => T)
      & { __mutation?: { a: A } }


let storeOut = create(
  withA("a",
    storeIn =>
      ({ count: 0 })
[...]

PR #1183, opened 2022-08-10
fix(types)(middleware/devtools): fix action type in devtools's setState

lucasrabiec:
- adjusted setState's `actionType` typing to contain `payload` (devtools middleware)
- added appropriate info to documentation
- added appropriate test (+ little tests refactor)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -412,6 +412,19 @@ const createBearSlice = (set, get) => ({
 })
 ```
 
+You can also log action's type along with its payload:
+
+```jsx
+const createBearSlice = (set, get) => ({
+  addFishes: (count) =>
+    set(
+      (prev) => ({ fishes: prev.fishes + count }),
+      false,
+      { type: 'bear/addFishes', payload: count }
+    ),
+})
+```
+
 If an action type is not provided, it is defaulted to "anonymous". You can customize this default value by providing an `anonymousActionType` parameter:
 
 ```jsx
--- a/readme.md
+++ b/readme.md
@@ -420,7 +420,7 @@ const createBearSlice = (set, get) => ({
     set(
       (prev) => ({ fishes: prev.fishes + count }),
       false,
-      { type: 'bear/addFishes', payload: count }
+      { type: 'bear/addFishes', count }
     ),
 })
 ```
--- a/readme.md
+++ b/readme.md
@@ -417,11 +417,10 @@ You can also log action's type along with its payload:
 ```jsx
 const createBearSlice = (set, get) => ({
   addFishes: (count) =>
-    set(
-      (prev) => ({ fishes: prev.fishes + count }),
-      false,
-      { type: 'bear/addFishes', count }
-    ),
+    set((prev) => ({ fishes: prev.fishes + count }), false, {
+      type: 'bear/addFishes',
+      count,
+    }),
 })
 ```
 
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -46,7 +46,7 @@ type StoreDevtools<S> = S extends {
       setState(
         ...a: [
           ...a: TakeTwo<A>,
-          actionType?: string | { type: unknown; payload?: unknown }
+          action?: string | { type: unknown; [key: string]: unknown }
         ]
       ): Sr
     }
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -40,11 +40,11 @@ type TakeTwo<T> = T extends []
 type WithDevtools<S> = Write<Cast<S, object>, StoreDevtools<S>>
 
 type StoreDevtools<S> = S extends {
-  setState: (...a: infer Sa) => infer Sr
+  setState: (...a: infer A) => infer Sr
 }
   ? {
-      setState<A extends string | { type: unknown }>(
-        ...a: [...a: TakeTwo<Sa>, action?: A]
+      setState<AT extends string | { type: unknown; [key: string]: unknown }>(
+        ...a: [...a: TakeTwo<A>, action?: AT]
       ): Sr
     }
   : never
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -40,11 +40,11 @@ type TakeTwo<T> = T extends []
 type WithDevtools<S> = Write<Cast<S, object>, StoreDevtools<S>>
 
 type StoreDevtools<S> = S extends {
-  setState: (...a: infer A) => infer Sr
+  setState: (...a: infer Sa) => infer Sr
 }
   ? {
-      setState<AT extends string | { type: unknown; [key: string]: unknown }>(
-        ...a: [...a: TakeTwo<A>, action?: AT]
+      setState<A extends string | { type: unknown }>(
+        ...a: [...a: TakeTwo<Sa>, action?: A]
       ): Sr
     }
   : never
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -43,11 +43,8 @@ type StoreDevtools<S> = S extends {
   setState: (...a: infer A) => infer Sr
 }
   ? {
-      setState(
-        ...a: [
-          ...a: TakeTwo<A>,
-          action?: string | { type: unknown; [key: string]: unknown }
-        ]
+      setState<AT extends string | { type: unknown; [key: string]: unknown }>(
+        ...a: [...a: TakeTwo<A>, action?: AT]
       ): Sr
     }
   : never
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -44,7 +44,10 @@ type StoreDevtools<S> = S extends {
 }
   ? {
       setState(
-        ...a: [...a: TakeTwo<A>, actionType?: string | { type: unknown }]
+        ...a: [
+          ...a: TakeTwo<A>,
+          actionType?: string | { type: unknown; payload?: unknown }
+        ]
       ): Sr
     }
   : never
--- a/tests/devtools.test.tsx
+++ b/tests/devtools.test.tsx
@@ -78,26 +78,45 @@ describe('If there is no extension installed...', () => {
     expect(console.warn).not.toBeCalled()
   })
 })
-
 describe('When state changes...', () => {
-  it("sends { type: setStateName || 'anonymous` } as the action with current state", () => {
-    const api = create(
-      devtools(() => ({ count: 0, foo: 'bar' }), {
-        name: 'testOptionsName',
-        enabled: true,
-      })
-    )
-    api.setState({ count: 10 }, false, 'testSetStateName')
+  it("sends { type: 'testSetStateName' } as the action with current state", () => {
+    const { setState } = createTestState()
+    setState({ count: 10 }, false, 'testSetStateName')
     expect(extension.send).toHaveBeenLastCalledWith(
       { type: 'testSetStateName' },
       { count: 10, foo: 'bar' }
     )
-    api.setState({ count: 5, foo: 'baz' }, true)
+  })
+
+  it("sends { type: 'testSetStateName', payload: 15 } as the action with current state", () => {
+    const { setState } = createTestState()
+    setState({ count: 15 }, false, {
+      type: 'testSetStateName',
+      payload: 15,
+    })
+    expect(extension.send).toHaveBeenLastCalledWith(
+      { type: 'testSetStateName', payload: 15 },
+      { count: 15, foo: 'bar' }
+    )
+  })
+
+  it("sends { type: 'anonymous' } as the action with current state", () => {
+    const { setState } = createTestState()
+    setState({ count: 5, foo: 'baz' }, true)
     expect(extension.send).toHaveBeenLastCalledWith(
       { type: 'anonymous' },
       { count: 5, foo: 'baz' }
     )
   })
+
+  function createTestState() {
+    return create(
+      devtools(() => ({ count: 0, foo: 'bar' }), {
+        name: 'testOptionsName',
+        enabled: true,
+      })
+    )
+  }
 })
 
 describe('when it receives an message of type...', () => {
--- a/tests/devtools.test.tsx
+++ b/tests/devtools.test.tsx
@@ -78,19 +78,23 @@ describe('If there is no extension installed...', () => {
     expect(console.warn).not.toBeCalled()
   })
 })
+
 describe('When state changes...', () => {
-  it("sends { type: 'testSetStateName' } as the action with current state", () => {
-    const { setState } = createTestState()
-    setState({ count: 10 }, false, 'testSetStateName')
+  it("sends { type: testSetStateName || 'anonymous`, ...rest } as the action with current state", () => {
+    const api = create(
+      devtools(() => ({ count: 0, foo: 'bar' }), {
+        name: 'testOptionsName',
+        enabled: true,
+      })
+    )
+
+    api.setState({ count: 10 }, false, 'testSetStateName')
     expect(extension.send).toHaveBeenLastCalledWith(
       { type: 'testSetStateName' },
       { count: 10, foo: 'bar' }
     )
-  })
 
-  it("sends { type: 'testSetStateName', payload: 15 } as the action with current state", () => {
-    const { setState } = createTestState()
-    setState({ count: 15 }, false, {
+    api.setState({ count: 15 }, false, {
       type: 'testSetStateName',
       payload: 15,
     })
@@ -98,25 +102,13 @@ describe('When state changes...', () => {
       { type: 'testSetStateName', payload: 15 },
       { count: 15, foo: 'bar' }
     )
-  })
 
-  it("sends { type: 'anonymous' } as the action with current state", () => {
-    const { setState } = createTestState()
-    setState({ count: 5, foo: 'baz' }, true)
+    api.setState({ count: 5, foo: 'baz' }, true)
     expect(extension.send).toHaveBeenLastCalledWith(
       { type: 'anonymous' },
       { count: 5, foo: 'baz' }
     )
   })
-
-  function createTestState() {
-    return create(
-      devtools(() => ({ count: 0, foo: 'bar' }), {
-        name: 'testOptionsName',
-        enabled: true,
-      })
-    )
-  }
 })
 
 describe('when it receives an message of type...', () => {
--- a/tests/devtools.test.tsx
+++ b/tests/devtools.test.tsx
@@ -80,7 +80,7 @@ describe('If there is no extension installed...', () => {
 })
 
 describe('When state changes...', () => {
-  it("sends { type: testSetStateName || 'anonymous`, ...rest } as the action with current state", () => {
+  it("sends { type: setStateName || 'anonymous`, ...rest } as the action with current state", () => {
     const api = create(
       devtools(() => ({ count: 0, foo: 'bar' }), {
         name: 'testOptionsName',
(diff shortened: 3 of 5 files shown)


src/middleware/devtools.ts
```diff
@@ -44,7 +44,10 @@ type StoreDevtools<S> = S extends {
 }
   ? {
       setState(
-        ...a: [...a: TakeTwo<A>, actionType?: string | { type: unknown }]
+        ...a: [
+          ...a: TakeTwo<A>,
+          actionType?: string | { type: unknown; payload?: unknown }
+        ]
       ): Sr
```
devanshj, line 51: I think the problem is that the following does not compile...

```ts
import create from "zustand"
import { devtools } from "zustand/middleware"

interface BearState {
  bears: number
  increase: (by: number) => void
}

const useBearStore = create<BearState>()(devtools((set) => ({
  bears: 0,
  increase: by => set(
    (state) => ({ bears: state.bears + by }),
    false,
    { type: "increase", by }
//                      ~~[1]
  ),
})))

// [1]:
// Argument of type '{ type: string; by: number; }' is not assignable to parameter of type 'string | { type: unknown; } | undefined'.
//   Object literal may only specify known properties, and 'by' does not exist in type '{ type: unknown; }'.
```

This is because of TypeScript's "excess property checks", the correct way to fix is to tell typescript that the type of `action` is not `{ type: unknown }` but rather `extends { type: unknown }` like so...

```suggestion
      setState<A extends string | { type: unknown }>(
        ...a: [...a: TakeTwo<A>, actionType?: string | { type: unknown }]
        ...a: [
          ...a: TakeTwo<A>,
          action?: A
        ]
      ): Sr
```

Another way to fix like this...

```ts
setState(
  ...a: [...a: TakeTwo<A>, actionType?: string | { type: unknown }]
  ...a: [
    ...a: TakeTwo<A>,
    action?: string | { type: unknown } & { [_ in string]: unknown }
  ]
): Sr
```

But this seems a little less correct and in general it's good to stay away from index signatures (ie { [_ in string]: unknown }).

And the current fix `{ type: unknown, payload?: unknown }` doesn't make the cut because it doesn't really solve the actual problem showed above, `{ type: "increase", by }` should also be assignable to `action` parameter.

tests/devtools.test.tsx
```diff
@@ -78,26 +78,45 @@ describe('If there is no extension installed...', () => {
...
   })
+
+  function createTestState() {
+    return create(
+      devtools(() => ({ count: 0, foo: 'bar' }), {
+        name: 'testOptionsName',
+        enabled: true,
+      })
+    )
+  }
```
devanshj, line 119: I think this refactor is a quite unncessary, the only change required should be this...

```diff
  describe('When state changes...', () => {
-   it("sends { type: setStateName || 'anonymous` } as the action with current state", () => {
+   it("sends { type: setStateName || 'anonymous`, ...rest } as the action with current state", () => {
      const api = create(
        devtools(() => ({ count: 0, foo: 'bar' }), {
          name: 'testOptionsName',
          enabled: true,
        })
      )
      api.setState({ count: 10 }, false, 'testSetStateName')
      expect(extension.send).toHaveBeenLastCalledWith(
        { type: 'testSetStateName' },
        { count: 10, foo: 'bar' }
      )
+
+     api.setState({ count: 10 }, false, { type: 'testSetStateName', bar: 'foo' })
+     expect(extension.send).toHaveBeenLastCalledWith(
+       { type: 'testSetStateName', bar: 'foo' },
+       { count: 10, foo: 'bar' }
+     )
+
      api.setState({ count: 5, foo: 'baz' }, true)
      expect(extension.send).toHaveBeenLastCalledWith(
        { type: 'anonymous' },
        { count: 5, foo: 'baz' }
      )
    })
  })
```

readme.md
```diff
@@ -412,6 +412,19 @@ const createBearSlice = (set, get) => ({
...
 
+You can also log action's type along with its payload:
+
+```jsx
+const createBearSlice = (set, get) => ({
+  addFishes: (count) =>
+    set(
+      (prev) => ({ fishes: prev.fishes + count }),
+      false,
+      { type: 'bear/addFishes', payload: count }
```
devanshj, line 423: I think it's good to have the payload key something other than "payload" to imply "payload" is not a special key, it can be anything...

```suggestion
      { type: 'bear/addFishes', count }
```

tests/devtools.test.tsx
lucasrabiec: Does acting after assertion isn't a bad idea? It can lead to unexpected results. I think we should start with a fresh state and test one thing per `it`.

readme.md
lucasrabiec: It's Redux-style logging and the payload is a special key for action in Redux Devtools. IMO it would be great to keep it like that while using Redux Devtools.
image
Here is a link to the conversation on the pmndrs DC: https://discord.com/channels/740090768164651008/740093228904218657/1006824679111135343

src/middleware/devtools.ts
lucasrabiec: Try `{ type: "increase", payload: by }`

devanshj: I don't follow, `{ type: "increase", by }` should compile, ofc users can write `{ type: "increase", payload: by }` if they wish.

readme.md
devanshj: In what way is it special? Like how does redux devtools treat `payload` key different from `count` key.

tests/devtools.test.tsx
devanshj: It doesn't matter much nor do current tests follow such rule. So I think it's better to keep it minimal, at least for this PR.

I usually don't write my tests with that rule, but if @dai-shi wants prefers to have such rule then you can open another PR that refactors all tests to follow the rule and not just this one.

readme.md
lucasrabiec: The main goal of this PR is logging the action's payload as the `payload` field, not any field name the user wants. To be consistent with the Redux convention (https://github.com/redux-utilities/flux-standard-action).
Ok, I am getting now that you want to abstract from it, right? The question is whether we want to have it consistent or to have whatever field's name the user wants but I think we should leave this decision for Zustand's owner.

tests/devtools.test.tsx
lucasrabiec: You are right

src/middleware/devtools.ts
lucasrabiec: I think this conversation is connected with the other one from this PR. Let's continue here: https://github.com/pmndrs/zustand/pull/1183#discussion_r943434166.

readme.md
devanshj: Okay so as I guessed redux devtools doesn't treat payload key in any special way.

Users allowed to choose the convention they want to follow, the one you've linked is just one of the many styles out there, it's not required. The only requirement and constraint of a redux action is that [it must have a `type` property](https://github.com/reduxjs/redux/blob/bb6233b726b2fda2b503364135899274ccc8973c/src/types/actions.ts#L1-L31).

I understand that you want to impose an convention for your logger, but the code being edited here affects everyone, and we can't impose a non-required convention for everyone.

> I think we should leave this decision for Zustand's owner

Sure, just giving my two cents.

tests/devtools.test.tsx
```diff
@@ -80,18 +80,29 @@ describe('If there is no extension installed...', () => {
 })
 
 describe('When state changes...', () => {
-  it("sends { type: setStateName || 'anonymous` } as the action with current state", () => {
+  it("sends { type: testSetStateName || 'anonymous`, ...rest } as the action with current state", () => {
```
devanshj, line 83: Nitpick: We can keep this too as it is (the test is that "it sends `setStateName`" and `setStateName` in turn is `"testSetStateName"` for the given tests, but it can be anything).
```suggestion
  it("sends { type: setStateName || 'anonymous`, ...rest } as the action with current state", () => {
```

dai-shi:
@lucasrabiec Thanks for opening a PR.
@devanshj Thanks for the review!

https://github.com/pmndrs/zustand/pull/1183#discussion_r943434166

We shouldn't follow FSA. `payload` is not special. It's up to users.
`{ type: 'add', by: number }` should be accepted as an action.

lucasrabiec:
Okay, thanks for the clarification, I will add amendments.

readme.md
lucasrabiec: I don't want to impose some of my conventions. I assumed that is a well-known standard and it would be nice to be consistent with it, but not. And it's okay.

I will add `[key: string]: unknown;` but you mention:
> in general it's good to stay away from index signatures

Could you elaborate? Is there the other way to achieve that?

dai-shi (approved):
LGTM

devanshj:
Fwiw my suggestion here https://github.com/pmndrs/zustand/pull/1183#discussion_r942852997 is ignored.

It's important to understand `extends { type: unknown }` and `{ type: unknown, [_: string]: unknown }` are not the same type.

We don't require an index signature on the action, and requiring it will create problems like it does in the following...

```ts
declare const dispatch1:
  <A extends { type: unknown }>(a: A) => void

declare const dispatch2:
  (a: { type: unknown, [_: string]: unknown }) => void

declare const fooAction1: FooAction1
interface FooAction1 {
  type: "FOO",
  foo: number
}

declare const fooAction2: FooAction2
type FooAction2 = {
  type: "FOO"
  foo: number
}

dispatch1(fooAction1)
dispatch1(fooAction2)
dispatch2(fooAction1)
//         ~~~~~~~~~~[1]
dispatch2(fooAction2)

// Argument of type 'FooAction1' is not assignable to parameter of type '{ [_: string]: unknown; type: unknown; }'.
//   Index signature for type 'string' is missing in type 'FooAction1'.
```

As you can see `fooAction1` doesn't compile with `dispatch2` implying `dispatch1` is the correct way to type `dispatch`. And of course I've just taken `dispatch` as a minimal example, same applies in our case too.

So we should use `extends { type: unknown }` as I suggested in my linked comment.

And in case linter complains that `A` is "unused" then the linter is wrong because it's clearly "used" and has a purpose in the way I showed, so it's correct to suppress the warning. I won't recommend renaming `A` to `_A` if that silents the linter, `A` is public here having an underscore will imply otherwise.

dai-shi (changes requested):
https://github.com/pmndrs/zustand/pull/1183#issuecomment-1213132016

Oh, I didn't follow that thread. Yeah, I wondered if `extends ...` should work. Let's do that.

lucasrabiec:
@devanshj I didn't know about this difference, thank you for an explanation. I added a change, please take a look if it fits with your comment.

src/middleware/devtools.ts
```diff
@@ -43,8 +43,8 @@ type StoreDevtools<S> = S extends {
   setState: (...a: infer A) => infer Sr
 }
   ? {
-      setState(
-        ...a: [...a: TakeTwo<A>, actionType?: string | { type: unknown }]
+      setState<AT extends string | { type: unknown; [key: string]: unknown }>(
```
devanshj, line 46: This is still incorrect, we don't need or want an index signature at all... Also let's rename `A` to `Sa` to avoid name clashing and let the public generic be `A`...

```diff
 type StoreDevtools<S> = S extends {
-  setState: (...a: infer A) => infer Sr
+  setState: (...a: infer Sa) => infer Sr
 }
   ? {
-      setState(
+      setState<A extends string | { type: unknown }>(
-        ...a: [...a: TakeTwo<A>, actionType?: string | { type: unknown }]
+        ...a: [...a: TakeTwo<Sa>, action?: A]
       ): Sr
     }
   : never
```

devanshj (changes requested):

readme.md
devanshj: Ah I missed this comment, it's mostly resolved now but I'll reply for sake of completeness...

> I don't want to impose some of my conventions. I assumed that is a well-known standard and it would be nice to be consistent with it, but not. And it's okay.

I never said it was "your" convention, I'm well aware about FSA, and not everyone uses it. My point was simply this: Redux does not require users to follow FSA so we should also not.

> Could you elaborate? Is there the other way to achieve that?

Now I did here https://github.com/pmndrs/zustand/pull/1183#issuecomment-1213132016

devanshj:
And maybe it's better to title the PR more appropriately, some suggestions...

- "fix(types): allow passing an action with additional properties other than `type` to `devtools`'s `setState`"
- "fix(types): circumvent excess-property-check for action in `devtools`'s `setState`"
- "fix(types): fix action type in `devtools`'s `setState`"
- etc

src/middleware/devtools.ts
lucasrabiec: Good point, TY.

lucasrabiec:
> And maybe it's better to title the PR more appropriately, some suggestions...
> 
> * "fix(types): allow passing an action with additional properties other than `type` to `devtools`'s `setState`"
> * "fix(types): circumvent excess-property-check for action in `devtools`'s `setState`"
> * "fix(types): fix action type in `devtools`'s `setState`"
> * etc

Good point, TY.

src/middleware/devtools.ts
lucasrabiec: When we remove index signature we still won't be able to add action's payload: `Argument of type '{ type: string; count: number; }' is not assignable to parameter of type 'string | { type: unknown; } | undefined'.`

devanshj: That's not true, I tried making the change locally on your branch works fine. You're probably not making the exact change I showed above. Make the change and push we'll see if the CI complains.

lucasrabiec: Ok, my bad. I didn't mention that the error occurs if devtools are used with immer:
```ts
const createBearSlice: StateCreator<
    { bears: { count: number } },
    [['zustand/devtools', never], ['zustand/immer', never]],
    [],
    { count: number }
  > = (set) => ({
    count: 0,
    setBears: (count: number) =>
      set(
        (state) => {
          state.bears.count = count
        },
        false,
        { type: 'bears/setBears', count } // <-- TS2345 
      ),
  })
```
I added some changes also to `immer.ts`. Could you take a look at them?

devanshj: It's recommended to use `immer` middleware before `devtools`, so that would fix your problem. You can change the test to use `immer` before `devtools`. I'll also open a PR to add this recommendation in the typescript doc.

As for the change you made in immer types, it's not correct, nor is it needed, so please revert it.

If we want to support using `devtools` before `immer` than the solution is much more complex. I'll see if I can open a PR for it later, but for now we don't need to support using `devtools` before `immer`

Also please don't make changes that you don't understand, it's okay if you don't fully understand the types here, it's better to ask for a solution than to guess a solution.

lucasrabiec: > It's recommended to use immer middleware before devtools, so that would fix the problem. You can change the test to use immer before devtools. I'll also open a PR to add this recommendation in the typescript doc.

Ok, didn't know.

> As for the change you made in immer types, it's not correct, nor is it needed, so please revert it.
>
>If we want to support using devtools before immer than the solution is much more complex. I'll see if I can open a PR for it later, but for now we don't need to support using devtools before immer
>
>Also please don't make changes that you don't understand, it's okay if you don't fully understand the types here, it's better to ask for a solution than to guess a solution.

No offense, but it is also good to explain what mistake a person made, not just to point out that they did something wrong. Cause for me now it infers the type well while omitting the first two if the third is given. (Maybe I understood something wrong but definitely not guessing the "right" solution). It looks like you know the whole code very well so please explain why we should use immer before devtools and what is wrong with my commit?

devanshj: Only mistakes can be corrected like I did here https://github.com/pmndrs/zustand/pull/1183#issuecomment-1213132016. But the change you made is entirely incorrect, there's nothing to correct there. So first you will have to tell why you think it's correct, and then I can tell what is wrong in your understanding. But unfortunately, I'm not available for such OSS interactions (hence I didn't ask you why you think it's correct), especially when I can see the person did a guesswork (ie make a change to the code they don't understand). And it's okay if you don't think it was guesswork.

For now I can give you just one false positive case... With your change, `immer`'s `setState` will accept an action even when `devtools` is not present... The following code should not compile, but with your change it'll.

```ts
import create from 'zustand'
import { immer } from 'zustand/middleware'

interface BearState {
  bears: number
  increase: (by: number) => void
}

const useBearStore = create<BearState>()(immer((set) => ({
      bears: 0,
      increase: (by) => set((state) => ({ bears: state.bears + by }), false, "increase"),
    }
))
```

If you want to understand the types then see #710. Using immer before devtools is recommended precisely so that we don't have to do anything extra in immer types to support cases like this.

No offense taken or meant.

lucasrabiec: Ok, I expressed myself wrong, it would be good to give an example when such a solution fails, which you did, so thank you. Now I see why it's a bad solution. Thanks for pointing out #710 I will dig into it. To be honest, these typings which you did is not the easiest to understand without fully digging into them (which can be time-consuming), but it's definitely a commendable effort of work (+1). It would be good to add that issue and PR in the documentation for development.

devanshj: > it would be good to give an example when such a solution fails

Yes I kinda agree, hence I made the edit...

> Maybe I should have pointed this out before itself, my apologies

> To be honest, these typings which you did is not the easiest to understand without fully digging into them (which can be time-consuming)

Yes, and hence users or contributors are not required to understand them. Only if one wishes to make a significant types contribution then it's required to understand them.

> it's definitely a commendable effort of work (+1)

Thanks!

> It would be good to add that issue and PR in the documentation for development.

It's actually already there in the typescript doc in the ["advanced usage" section](https://github.com/pmndrs/zustand/blob/main/docs/typescript.md#authoring-middlewares-and-advanced-usage).

lucasrabiec: > It's actually already there in the typescript doc in the "advanced usage" section.

Woah, my bad, didn't noticed that 😞

lucasrabiec:
@dai-shi @devanshj I think it is ready for merge, thanks for discussion and constructive critisism.

tests/middlewareTypes.test.tsx
```diff
@@ -262,15 +262,17 @@ describe('counter state spec (double middleware)', () => {
...
-              state.count = get().count + 1
-            }),
-        })),
-        { name: 'prefix' }
+      immer(
+        devtools(
+          (set, get) => ({
+            count: 0,
+            inc: () =>
+              set((state) => {
```
devanshj, line 270: Let's add an action with payload to this `set` call (like you did previously) as a test for types.

devanshj (changes requested):

src/middleware/devtools.ts
devanshj: No problems!

tests/middlewareTypes.test.tsx
```diff
@@ -259,18 +259,53 @@ describe('counter state spec (double middleware)', () => {
...
+      useBoundStore().inc()
+      useBoundStore.getState().count * 2
+      useBoundStore.getState().inc()
+      useBoundStore.setState({ count: 0 }, false, 'reset')
+      return <></>
+    }
+    TestComponent
+  })
+
+  it('immer & devtools, action with type and payload', () => {
```
devanshj, line 291: My suggestion was to add action in the above `set` call itself xD, another test case wasn't necessary, but I think it doesn't hurt much to have one, so it's okay I guess.

devanshj (approved):

tests/middlewareTypes.test.tsx
lucasrabiec: Haha, okay xD I will change it.

tests/middlewareTypes.test.tsx
```diff
@@ -1,4 +1,4 @@
-import create, { StoreApi } from 'zustand'
+import create, { StoreApi, UseBoundStore } from 'zustand'
```
dai-shi, line 1:
```suggestion
import create, { StoreApi } from 'zustand'
```

dai-shi (approved):
Thanks for working on this!

PR #1193, opened 2022-08-15
Fix Persist in TypeScript Usage on readme

spenceradolph:
Persist middleware requires the name option to keep storage unique.

See https://github.com/pmndrs/zustand/blob/main/docs/persisting-store-data.md#name

This change adds the 'name' option to the Typescript Usage example to update it.
Additionally, the persist example already has this required option.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -461,10 +461,15 @@ interface BearState {
 
 const useBearStore = create<BearState>()(
   devtools(
-    persist((set) => ({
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Nice catch. Can you run `yarn run prettier` to fix CI?

dai-shi (approved):
👍

PR #1206, opened 2022-08-19
fix(middleware): revert types of redux middleware

dai-shi:
In #1144, I made some types loosen, which includes not only `State` but also `Action` in the redux middleware. @devanshj pointed out that it can cause misusage with the combination of `redux` and `devtools` middleware.

So, I'm convinced to revert the `Action` part of the change.

### DIFF
--- a/src/middleware/redux.ts
+++ b/src/middleware/redux.ts
@@ -3,25 +3,31 @@ import { NamedSet } from './devtools'
 
 type Write<T, U> = Omit<T, keyof U> & U
 
-type ReduxState<A> = {
-  dispatch: StoreRedux<A>['dispatch']
-}
+type Action = { type: unknown }
 
-type StoreRedux<A> = {
+type StoreRedux<A extends Action> = {
   dispatch: (a: A) => A
   dispatchFromDevtools: true
 }
 
-type WithRedux<S, A> = Write<S, StoreRedux<A>>
+type ReduxState<A extends Action> = {
+  dispatch: StoreRedux<A>['dispatch']
+}
 
-type Redux = <T, A, Cms extends [StoreMutatorIdentifier, unknown][] = []>(
+type WithRedux<S, A extends Action> = Write<S, StoreRedux<A>>
+
+type Redux = <
+  T,
+  A extends Action,
+  Cms extends [StoreMutatorIdentifier, unknown][] = []
+>(
   reducer: (state: T, action: A) => T,
   initialState: T
 ) => StateCreator<Write<T, ReduxState<A>>, Cms, [['zustand/redux', A]]>
 
 declare module '../vanilla' {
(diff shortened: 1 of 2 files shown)


src/middleware/redux.ts
```diff
@@ -3,25 +3,31 @@ import { NamedSet } from './devtools'
...
+  Cms extends [StoreMutatorIdentifier, unknown][] = []
+>(
   reducer: (state: T, action: A) => T,
   initialState: T
 ) => StateCreator<Write<T, ReduxState<A>>, Cms, [['zustand/redux', A]]>
 
 declare module '../vanilla' {
   interface StoreMutators<S, A> {
-    'zustand/redux': WithRedux<S, A>
+    'zustand/redux': WithRedux<S, A & Action>
```
dai-shi, line 30: I wasn't so sure, but is this what you mean? @devanshj

devanshj: Yes.

Bonus if you use `Cast` instead. To see the difference write a store with redux middleware, however on the `dispatch` function, you'll see in the action parameter there's an extra `& Action`, but if you use `Cast` the action type will be as it is. If you want I can also explain how `Cast` works for the record.

dai-shi: I think I know what Cast does, but I don't get how it improves tooltip. Can this be explained with TS playground?

dai-shi:
@devanshj How's this change? b11f3ac

src/middleware/redux.ts
devanshj: There you go https://tsplay.dev/m3XyjW

devanshj:
> @devanshj How's this change? https://github.com/pmndrs/zustand/commit/b11f3acdb702883ddd6225edb9c671a8725c9ecb

It'd kinda work because all those types are not exported, so we need not worry much about being correct. However I personally would still be correct even for internal types and keep the constraint.

src/middleware/redux.ts
dai-shi: Thanks. I understand this one, and why `Cast` is preferable.

dai-shi:
> > @devanshj How's this change? [b11f3ac](https://github.com/pmndrs/zustand/commit/b11f3acdb702883ddd6225edb9c671a8725c9ecb)
> 
> It'd kinda work because all those types are not exported, so we need not worry much about being correct. However I personally would still be correct even for internal types and keep the constraint.

I kinda like what I did. In my mental model, the `redux` middleware internal types are based on `Action extends unknown` because there're no restrictions. The public API is based on `Action extends { type: unknown }`, because it's required by `devtools` middleware.

So, unless we have obvious limitations like quickinfo one, I'd go with this.

devanshj:
Yeah that mental model is not bad, so yeah it's okay to go with this.

PR #1239, opened 2022-08-29
fix(docs): comparison

### LINKED ISSUE #1008: Move the docs from Github's wiki to the ./docs directory

dbritto-dev:
## Related Issues

Fixes #1008 

## Summary

- Move difference between zustand and valtio to ./docs
- Updating zustand vs redux
- Adding other comparisons

## Check List

### DIFF
--- a/docs/getting-started/comparison.md
+++ b/docs/getting-started/comparison.md
@@ -222,8 +222,8 @@ const Component = () => {
 
 ### State Model
 
-There is a major difference between Zustand and Valtio. Zustand is based on the
-immutable state model, while Valtio is based on the mutable state model.
+There is a major differences between Zustand and Valtio. Zustand is based on
+the immutable state model, while Valtio is based on the mutable state model.
 
 ```ts
 import create from 'zustand'
--- a/docs/getting-started/comparison.md
+++ b/docs/getting-started/comparison.md
@@ -14,7 +14,7 @@ differences between these two libraries.
 There are no big difference between Zustand and Redux. Both are base on
 immutable state model.
 
-```tsx
+```ts
 import create from 'zustand'
 
 type State = {
@@ -40,7 +40,7 @@ const Component = () => {
 }
 ```
 
-```tsx
+```ts
 import create from 'zustand'
 
 type State = {
@@ -80,7 +80,7 @@ const Component = () => {
 }
 ```
 
-```tsx
+```ts
 import { createStore } from 'redux'
 import { useSelector, useDispatch } from 'react-redux'
 
@@ -146,7 +146,7 @@ The other difference between Valtio and Zustand is, Valtio made render
 optimizations through property access. But, with Zustand you need to do manual
 render optimizations through selectors.
 
-```tsx
+```ts
 import create from 'zustand'
 
 const useStore = create(() => ({
@@ -159,7 +159,7 @@ const Component = () => {
 }
 ```
 
-```tsx
+```ts
 import { proxy, useSnapshot } from 'valtio'
 
 const state = proxy({
--- a/docs/getting-started/comparison.md
+++ b/docs/getting-started/comparison.md
@@ -40,7 +40,7 @@ const Component = () => {
 }
 ```
 
-```ts
+```tsx
 import create from 'zustand'
 
 type State = {
@@ -80,7 +80,7 @@ const Component = () => {
 }
 ```
 
-```ts
+```tsx
 import { createStore } from 'redux'
 import { useSelector, useDispatch } from 'react-redux'
 
@@ -146,7 +146,7 @@ The other difference between Valtio and Zustand is, Valtio made render
 optimizations through property access. But, with Zustand you need to do manual
 render optimizations through selectors.
 
-```ts
+```tsx
 import create from 'zustand'
 
 const useStore = create(() => ({
@@ -159,7 +159,7 @@ const Component = () => {
 }
 ```
 
-```ts
+```tsx
 import { proxy, useSnapshot } from 'valtio'
 
 const state = proxy({
--- a/docs/getting-started/comparison.md
+++ b/docs/getting-started/comparison.md
@@ -12,7 +12,7 @@ differences between these two libraries.
 ### State Model
 
 There are no big difference between Zustand and Redux. Both are base on
-immutable state model.
+immutable state model. Also, Redux needs to wrap you app in context providers.
 
 ```ts
 import create from 'zustand'
@@ -291,8 +291,9 @@ differences between these two libraries.
 
 ### State Model
 
-The major difference is the same as Zustand and Jotai. Also, Recoil depends on
-atom string keys instead of atom object referential identities.
+The major difference is the same as Zustand and Jotai is: Recoil depends on
+atom string keys instead of atom object referential identities. Also, Recoil
+needs to wrap your app in a context provider.
 
 ```ts
 import create from 'zustand'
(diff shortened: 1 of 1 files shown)


Exalted100 (commented):
- For the difference between Zustand and Redux, I think some exist (the differences that were mentioned earlier and removed from the document). While Redux Toolkit removes some differences, it still involves wrapping a context provider around the app and other differences could still apply. The only difference I think Redux toolkit takes away besides that is that it makes Redux almost as simple to implement as Zustand.

Everything else looks fine to me.

dbritto-dev:
@Exalted100 about Redux since we're just comparing Redux (React Redux) with Zustand, anything related to Redux shouldn't be considered as part of this section.

Exalted100:
Since that's the case, we could include the differences from before since they still apply, right?

dbritto-dev:
@Exalted100 yeah with an exception about hooks, because now Redux has support for hooks.

dbritto-dev:
@Exalted100 ready to review :D

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,356 @@ description:
...
+import { proxy } from 'valtio'
+
+const state = proxy({ obj: { count: 0 } })
+
+state.obj.count += 1
+```
+
+### Render Optimization
+
+The other difference between Zustand and Valtio is: Valtio makes render
```
chrisk-7777, line 189: nit: I _think_ the ":" isn't required?

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,356 @@ description:
...
+  }
+}
+
+const countStore = createStore(countReducer)
+```
+
+### Render Optimization
+
+There are no difference between Zustand and Redux. In both you need to do manual
+render optimizations through selectors.
```
chrisk-7777, line 100: minor: As above "difference" -> "differences"

I would maybe reword slightly to make it clear we're talking about optimization. Perhaps something like:

> When it comes to render optimizations within your app, there are no major differences in approach between Zustand and Redux. In both libraries it is recommended you manually render optimizations by using selectors.

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,356 @@ description:
...
+const Component = () => {
+  const { count } = useSnapshot(state)
+  // ...
+}
+```
+
+## Jotai
+
+Zustand and Jotai are state management libraries for React. Here are some
+differences between these two libraries.
```
chrisk-7777, line 222: I feel like by this point we're repeating this line. 

How would you feel about an opening line, right at the top of the page, with something like:

> Zustand is one of many state management libraries for React. On this page we'll discuss Zustand in comparison to some of these libraries, including Redux, Valtio, Jotai, and Recoil.
>
> Each library has its own strengths and weaknesses, and we will compare key differences and similarities between each. 
>

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,356 @@ description:
...
+}
+
+const useCountStore = create<State & Actions>((set) => ({
+  count: 0,
+  dispatch: (action: Action) => set((state) => countReducer(state, action)),
+}))
+```
+
+```ts
+import { createStore } from 'redux'
```
dai-shi, line 71: This is traditional Redux. If you make a comparison note now, you want to compare with modern Redux.

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,356 @@ description:
...
+
+## Jotai
+
+Zustand and Jotai are state management libraries for React. Here are some
+differences between these two libraries.
+
+### State Model
+
+There is a major difference between Zustand and Jotai. Zustand is a single
+store, while Jotai consits of primitive atoms and allows composing them
```
chrisk-7777, line 227: "consits" -> "consists"

chrisk-7777 (commented):
This is looking really good, I just had some minor grammatical suggestions. 
Let me know your thoughts :) 

Also, do you know in Github, will it squash all the WIP commits?

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,356 @@ description:
...
+together.
```
dai-shi, line 228: Another more major difference is Zustand store is global in memory, but Jotai atoms are not. Atoms are definitions that don't hold values. That's why you can't read atom values outside React (as it can be multiple values.)

dai-shi (commented):
I'm not sure what kind of comparison we should provide and what readers expect.
I don't have big opinions on what it should be. What I feel is there are many (or mostly) code snippets.
I'd rather let the people who work on docs to discuss.
cc @chrisk-7777

docs/getting-started/comparison.md
dbritto-dev: Cool, good to know

docs/getting-started/comparison.md
dbritto-dev: What you mean with modern Redux? Actually, Redux toolkit use createStore behind scenes. I can update the code to use the Redux Toolkit version but I want to show the basic example because that's why people often use even these days.

dai-shi: I mean Redux Toolkit. The bare createStore should be marked as deprecated.

dbritto-dev: I added a piece of code for Redux Toolkit as well

dbritto-dev:
Hey folks, ready for review

@dai-shi @chrisk-7777 @Exalted100

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,409 @@ description:
...
+
+## Jotai
+
+### State Model
+
+There are two major difference between Zustand and Jotai. The first one is
+Zustand is a single store, while Jotai consists of primitive atoms and allows
+composing them together. The last one is Zustand store is global in memory, but
+Jotai atoms are not (are difinitions that do not hold values) and that's why
+you can use it outside React.
```
dai-shi, line 284: you can with zustand. you can't with jotai.

dbritto-dev:
@dai-shi done

dai-shi (commented):
Looks okay to me. I'll wait for others to review.

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,409 @@ description:
...
+  const { count } = useSnapshot(state)
+  // ...
+}
+```
+
+## Jotai
+
+### State Model
+
+There are two major difference between Zustand and Jotai. The first one is
```
Exalted100, line 280: It should be "differences", not "difference".

Exalted100 (commented):
It looks fine to me. Just one small typo which I left a comment for.

dbritto-dev:
@Exalted100 done

Exalted100 (approved):

dai-shi:
Let's wait a final call from @chrisk-7777.

docs/getting-started/comparison.md
```diff
@@ -4,11 +4,409 @@ description:
...
+```
+
+## Jotai
+
+### State Model
+
+There are two major differences between Zustand and Jotai. The first one is
+Zustand is a single store, while Jotai consists of primitive atoms and allows
+composing them together. The last one is Zustand store is global in memory, but
+Jotai atoms are not (are difinitions that do not hold values) and that's why
```
dai-shi, line 283:
```suggestion
Jotai atoms are not (are definitions that do not hold values) and that's why
```

chrisk-7777:
Thanks for actioning the feedback @dbritto-dev , I have reread the PR and have no further feedback

PR #1244, opened 2022-08-30
fix(docs): Issue 1120 docs guides

Exalted100:
## Related Issues

Fixes #.

## Summary

Reviewed the following documentation:
- maps and sets usage
- practice with no store actions
- typescript
- updating nested state object values

## Check List

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -28,7 +28,7 @@ const useBearStore = create<BearState>()((set) => ({
 
 **TLDR**: Because state generic `T` is invariant.
 
-Consider this minimal version `create...`
+Consider this minimal version `create`:
 
 ```ts
 declare const create: <T>(f: (get: () => T) => T) => T
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -5,7 +5,7 @@ nav: 9
 
 ## Basic usage
 
-The difference when using TypeScript is instead of writing `create(...)`, you have to write `create<T>()(...)` where `T` would be type of the state so as to annotate it. Example...
+The difference when using TypeScript is that instead of typing `create(...)`, you have to type `create<T>()(...)`. `T` would be type of the state, so as to annotate it. Foe example:
 
 ```ts
 import create from 'zustand'
@@ -22,13 +22,13 @@ const useBearStore = create<BearState>()((set) => ({
 ```
 
 <details>
-  <summary>Why can't we just simply infer the type from initial state?</summary>
+  <summary>Why can we not simply infer the type from initial state?</summary>
 
   <br/>
 
 **TLDR**: Because state generic `T` is invariant.
 
-Consider this minimal version `create`...
+Consider this minimal version `create...`
 
 ```ts
 declare const create: <T>(f: (get: () => T) => T) => T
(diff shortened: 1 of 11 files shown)


dbritto-dev (approved):

Exalted100:
@dai-shi I've made changes based on your comments.

dai-shi (approved):
Thanks!

docs/guides/typescript.md
```diff
@@ -190,11 +190,11 @@ const useBearStore = create<BearState>()(
 )
 ```
 
-Also it's recommended to use `devtools` middleware as last as possible, in particular after `immer` middleware, ie it should be `immer(devtools(...))` and not `devtools(immer(...))`. The reason being that `devtools` mutates the `setState` and adds a type parameter on it, which could get lost if other middlewares (like `immer`) mutate `setState` before `devtools`.
+Also, we recommend using `devtools` middleware as an argument, rather than passing arguments to it. For example, when you use it with `immer` as a middleware, it should be `immer(devtools(...))` and not `devtools(immer(...))`. This is because`devtools` mutates the `setState` and adds a type parameter on it, which could get lost if other middlewares (like `immer`) mutates `setState` before `devtools`.
```
devanshj, line 193: > we recommend using `devtools` middleware as an argument, rather than passing arguments to it

This sounds a bit wrong, even if you use `devtools` at the end you're still passing an argument to it namely your state creator.

I think the original wording is better, I'm fine if someone can come up with a better one but the one is PR isn't really making the cut imo.

Also "mutate" was correct because it's "middlewares".

```suggestion
Also, we recommend using `devtools` middleware as last as possible. For example, when you use it with `immer`, it should be `immer(devtools(...))` and not `devtools(immer(...))`. This is because `devtools` mutates `setState` and adds a type parameter on it, which could get lost if other middlewares (like `immer`) also mutate `setState` before `devtools`.
```

docs/guides/typescript.md
```diff
@@ -208,11 +208,11 @@ const useBearStore = create(foo(() => ({ bears: 0 }), 'hello'))
...
+Zustand middlewares can mutate the store. But how could we possibly encode the mutation on the type-level? That is to say how could do we type `foo` so that this code compiles?
 
-For an usual statically typed language this is impossible, but thanks to TypeScript, Zustand has something called an "higher kinded mutator" that makes this possible. If you're dealing with complex type problems like typing a middleware or using the `StateCreator` type, then you'll have to understand this implementation detail, for that check out [#710](https://github.com/pmndrs/zustand/issues/710).
+For a usual statically typed language, this is impossible. However, with TypeScript, Zustand has something called a "higher kind mutator" that makes this possible. If you are dealing with complex type problems, like typing a middleware or using the `StateCreator` type, you will have to understand this implementation detail. For this, you can [check out #710](https://github.com/pmndrs/zustand/issues/710).
```
devanshj, line 213: The original phrasing "thanks to" is better as it implies it's a TypeScript feature more significantly than a mere "with".

And it's [called "higher kinded"](https://www.google.com/search?q=higher+kinded+type) and not "higher kind". Putting a hyphen to make it more correct.

```suggestion
For a usual statically typed language, this is impossible. But thanks to TypeScript, Zustand has something called a "higher-kinded mutator" that makes this possible. If you are dealing with complex type problems, like typing a middleware or using the `StateCreator` type, you will have to understand this implementation detail. For this, you can [check out #710](https://github.com/pmndrs/zustand/issues/710).
```

devanshj (commented):
Thanks for the edit. I've left my two cents. Most of them are trying to preserve and improve the expression and intent as the original author. But of course I do understand I don't own the document haha so no pressure. Other than that...

- I like the punctuations, makes it more readable.
- I also like the more simplified phrasing at places.
- I personally don't like the use of non-contractions ie "do not" instead of "don't". Reading "do not" feels too passive to me. But I'm fine it others are okay with it.
- I also personally prefer "..." instead of ":" but again I'm fine as long as we're consistent.

cc @dai-shi

dai-shi:
@Exalted100 Can you please take the above comments into account?

PR #1247, opened 2022-09-01
fix(docs): Issue 1120 docs guides

Exalted100:
## Related Issues

Fixes #.

## Summary

* Made changes based on the comments left by @devanshj in #1244

## Check List

- [X] `yarn run prettier` for formatting code and docs

PR #1252, opened 2022-09-01
docs(typescript): Rewrite the inference explanation

devanshj:
The original had this bit...

> Now one can argue it's impossible to write an implementation for `createFoo`, and that's true. But then it's also impossible to write Zustand's `create`... Wait but Zustand exists? So what do I mean by that?

Which I think deserves a proper explanation. So I elaborated that bit and also rewrote the latter part. So now it goes even more deeper only in the "More about the inference" collapsible.

I'm also open to remove that whole "More about the inference" collapsible entirely (ie only keep till "Here again, `x` is `unknown` instead of `string`" and remove the details tag after it), as it's just for the people curious and interested in TypeScript. Let me know what you think.


docs/guides/typescript.md
```diff
@@ -44,20 +44,31 @@ const x = create((get) => ({
 // }
 ```
 
-Here, if you look at the type of `f` in `create`, i.e. `(get: () => T) => T`, it returns `T`. However, it also "takes" `T` via `get`. Typescript wonders where `T` comes from, like that chicken or egg problem. At the end TypeScript, gives up and infers `T` as `unknown`.
+Here, if you look at the type of `f` in `create`, i.e. `(get: () => T) => T`, it "gives" `T` via return (making it covariant), but it also "takes" `T` via `get` (making it contravariant). "So where does `T` come from?" TypeScript wonders. It's like that chicken or egg problem. At the end TypeScript, gives up and infers `T` as `unknown`.
 
-So, as long as the generic to be inferred is invariant, TypeScript will be unable to infer it. Another simple example would be this:
+So, as long as the generic to be inferred is invariant (i.e. both covariant and contravariant), TypeScript will be unable to infer it. Another simple example would be this:
```
devanshj, line 49: This change is already accomodated in #1247 btw, it's just not merged yet.

dai-shi (approved):
nested collapsible 😄 
yeah, let's keep it.
thanks for working on this.

Issue #1216, opened 2022-08-24
[documentation update] Missing images & dead links audit


chrisk-7777:
Core conversation here: https://github.com/pmndrs/zustand/discussions/1033

There are missing images (like on the introduction page) and dead links (like on the v4 migration page).
The person on this issue should (manually) scan through the docs pages looking for missing images or broken links.

A single PR should be able to apply all the fixes

Blocked by: https://github.com/pmndrs/zustand/issues/1215

chrisk-7777:
@lowellwashere2020 I can't assign it to you via Github, but I'll assume you're working on this one :)

lowellwashere2020:
@chrisk-7777 Indeed. I'm on it.

chrisk-7777:
Hi @drcmda , would you happen to know what the content of https://codesandbox.io/s/amazing-kepler-swxol was originally? It is referenced a few times in the Zustand docs but is no longer available. 

I haven't had much luck plucking it from waybackmachine.

lowellwashere2020:
> Hi @drcmda , would you happen to know what the content of https://codesandbox.io/s/amazing-kepler-swxol was originally? It is referenced a few times in the Zustand docs but is no longer available.
> 
> I haven't had much luck plucking it from waybackmachine.

I haven't been able to find it either. I asked in the discord as well.

PR #1257, opened 2022-09-03
docs: updates to fix remaining broken links across docs site

chrisk-7777:
## Related Issues

https://github.com/pmndrs/zustand/issues/1216

## Summary

Fixes the remaining broken links mentioned in 1216. After which, we can close issue 1216. 

Note: These is one missing codesandbox. I have asked drcmda, but if they don't have a copy I will recreate it as a seperate PR. For now, I have removed reference to that sandbox, I think its better to omit than have broken links. 

This PR also fixes a broken MDX file (typescript), where `<br>` was used instead of `<br/>`, which causes issues building locally. 

## Check List

### DIFF
--- a/docs/guides/flux-inspired-practice.md
+++ b/docs/guides/flux-inspired-practice.md
@@ -19,7 +19,7 @@ const useBoundStore = create((set) => ({
 }))
 ```
 
-See [Splitting the store into separate slices](/zustand/guides/typescript#slices-pattern) for how to define a store with separate slices.
+See [Splitting the store into separate slices](./typescript.md#slices-pattern) for how to define a store with separate slices.
 
 ## Flux like patterns / "dispatching" actions
 
--- a/docs/guides/flux-inspired-practice.md
+++ b/docs/guides/flux-inspired-practice.md
@@ -19,7 +19,7 @@ const useBoundStore = create((set) => ({
 }))
 ```
 
-See [Splitting the store into separate slices](./typescript.md#slices-pattern) for how to define a store with separate slices.
+See [Splitting the store into separate slices](/zustand/guides/typescript#slices-pattern) for how to define a store with separate slices.
 
 ## Flux like patterns / "dispatching" actions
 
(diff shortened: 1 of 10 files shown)


docs/guides/flux-inspired-practice.md
```diff
@@ -19,7 +19,7 @@ const useBoundStore = create((set) => ({
 }))
 ```
 
-See [Splitting the store into separate slices](./typescript.md#slices-pattern) for how to define a store with separate slices.
+See [Splitting the store into separate slices](./guides/typescript.md#slices-pattern) for how to define a store with separate slices.
```
dai-shi, line 22: This makes the url `guides/guides/`. I think the proper fix is to remove `.md`. Hope it works in GitHub.
```suggestion
See [Splitting the store into separate slices](./typescript#slices-pattern) for how to define a store with separate slices.
```

dai-shi (changes requested):

docs/guides/flux-inspired-practice.md
chrisk-7777: You're absolutely right. I'll get this fixed when I'm back at my desk

dai-shi: If it doesn't work in GitHub, we should keep it as is, and fix it in the doc site generation tool.

chrisk-7777: With an absolute path (in this case `/zustand/guides/typescript`) it appears to work correctly. I don't think any changes will be needed to the website repo

dai-shi: That's not acceptable. Breaks the link in https://github.com/chrisk-7777/zustand/blob/issue-1216-docs/docs/guides/flux-inspired-practice.md, which we would want to keep working.

dai-shi: Let's fix the website, it's how we did in jotai too.

dai-shi: It's in this branch: https://github.com/pmndrs/website/tree/docs

chrisk-7777: Right, I see what you mean by working in Github in the above comment. 

Ok, I'll take a look at fixing it at the website end

dai-shi: Sorry, I wasn't very clear about "working in GitHub"...

chrisk-7777: I've made a PR over on the site here: https://github.com/pmndrs/website/pull/251
I'm not 100% sold on the approach, so I'm open to suggestions. I believe this is what you meant by fixing it on the site side of things

dai-shi: the approach in general? or the approach in the PR?
yes, this is what i meant. the logic in PR can be improved as Cody commented.

chrisk-7777: With https://github.com/pmndrs/website/pull/251 now merged into the site repo, this PR has been updated to fix any outstanding broken relative links. These links now work on both GH and the docs site

dai-shi (approved):
LGTM

chrisk-7777:
Conflicts resolved

PR #1258, opened 2022-09-03
fix(context): store property missing from `useStoreApi`

exuanbo:
## Related Issues

Fixes #.

## Summary

When the store is created with persist middleware, property `storeApi.persist` is missing, which does not match its type.

The internal function `useBoundStore` should be named as `useContextStore`, because the store API is not assigned to it.

## Check List

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -44,7 +44,7 @@ function createContext<S extends StoreApi<unknown>>() {
     )
   }
 
-  const useBoundStore = (<StateSlice = ExtractState<S>>(
+  const useContextStore: UseContextStore<S> = <StateSlice = ExtractState<S>>(
     selector?: (state: ExtractState<S>) => StateSlice,
     equalityFn?: (a: StateSlice, b: StateSlice) => boolean
   ) => {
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Oh, nice catch! Both are valid points. Thanks for the fixes!

src/context.ts
```diff
@@ -68,21 +68,12 @@ function createContext<S extends StoreApi<unknown>>() {
...
-      () =>
-        ({
-          getState: store.getState,
-          setState: store.setState,
-          subscribe: store.subscribe,
-          destroy: store.destroy,
-        } as WithoutCallSignature<S>),
-      [store]
-    )
+    return useMemo<WithoutCallSignature<S>>(() => ({ ...store }), [store])
```
dbritto-dev, line 71: If you want to return all the props in the store, would be better just return the store directly

```suggestion
    return useMemo<WithoutCallSignature<S>>(() => store, [store])
```

dbritto-dev (changes requested):

Himself65:
when can this PR land?

dai-shi:
[v4.1.2](https://github.com/pmndrs/zustand/milestone/7) is currently planned at Sep 12.

Feel free to use the codesandbox build until then.
https://ci.codesandbox.io/status/pmndrs/zustand/pr/1258
Find "Local Install Instructions" ☝️

dai-shi:
> v4.1.2 is currently planned at Sep 12.

Sorry for the delay. We would like to merge another change, which is taking some more time.
Please use the codesandbox build meanwhile.

PR #1275, opened 2022-09-09
fix(types): simplify internal TakeTwo and SkipTwo types in middleware

### LINKED ISSUE #1013: TS issue with the set function when using devtools (with older TS versions)
Hey there zustand team

I'm having a problem when using zustand with devtools in my next.js project.

When running the store locally, everything works fine and I'm seeing devtools. Awesome.

But, when building my app for production with nx/next.js, it throws the following error:

```bash
Type error: Expected 3-4 arguments, but got 1.

  62 |       damageLocation: TDamageLocation
  63 |     ) => {
> 64 |       set({ damageType, damageLocation, view: 'station' });
     |       ^
  65 |     },
  66 |     updateStation: (station: IStationState) => {
  67 |       set({ station, view: 'timeSlot' });
```

### What have I do so far:
1. Tried to follow this https://github.com/pmndrs/zustand/discussions/976 and looked at the `tests/middlewareTypes.test.tsx` and `tests/devtools.test.tsx` examples but not successful.
2. Tried to type the store like in the example above: `create<InitialState, [["zustand/devtools", InitialState]]>`
3. Tried changing the set function called on line 64 (example above) but then I get the same error on a different function

Using the following core dependencies:

- next: 12.1.0,
- nx: 13.9.5,
- react: 17.0.2,
- zustand: ^4.0.0-rc.1

Let me know if you need more details or if this makes no sense.

dai-shi:
should fix #1013.

my guess is `TakeTwo` is too complex.

https://tsplay.dev/mx3VBN I think we can simplify it. @devanshj can confirm?

Also, `SkipTwo` is simplified, which might solve issues with old TS.

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -92,12 +92,6 @@ type TakeTwo<T> = T extends []
   ? [...a0: T, a1: undefined]
   : T extends [unknown?]
   ? [...a0: T, a1: undefined]
-  : T extends [unknown, unknown]
-  ? T
-  : T extends [unknown, unknown?]
-  ? T
-  : T extends [unknown?, unknown?]
-  ? T
   : T extends [infer A0, infer A1, ...unknown[]]
   ? [A0, A1]
   : T extends [infer A0, (infer A1)?, ...unknown[]]
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -18,18 +18,14 @@ declare module '../vanilla' {
 }
 
 type Write<T, U> = Omit<T, keyof U> & U
-type SkipTwo<T> = T extends []
-  ? []
-  : T extends [unknown]
-  ? []
-  : T extends [unknown?]
-  ? []
-  : T extends [unknown, unknown, ...infer A]
+type SkipTwo<T> = T extends [unknown, unknown, ...infer A]
   ? A
   : T extends [unknown, unknown?, ...infer A]
   ? A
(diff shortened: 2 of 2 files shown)


devanshj:
> https://tsplay.dev/mx3VBN I think we can simplify it. @devanshj can confirm?

This would remove the parameters names if we only have two parameters (and vanilla `setState` indeed has two parameters): https://tsplay.dev/wQxDVW (see `T41` vs `T42`). So the parameter names of `setState` would become something like `a0`, `a1` instead of `partial`, `replace`; which isn't great.

Let's first properly diagnose what exactly the issue is, ie let's find out in what ts version the issue occurs and why it occurs (eg maybe we're using a new feature). Once we know the concrete issue we can try different solutions instead of doing trial-and-error.

If you do the diagnosis and report your findings here I can give some suggestions, or else I'll do it when I find the time for it.

dai-shi:
Oh, I wasn't aware of the parameter name issue...

> I'll do it when I find the time for it.

Please, I'll count on you.

I did some rough play with TS playground, and it feels like this might fix for old TS versions (v4.4 and v4.0 for example)

devanshj:
> Please, I'll count on you.

Haha sure, I'll look into it in a couple of days.

devanshj:
There you go #1348, looks like it was more than a couple of days :P

dai-shi:
closing in favor of #1348.

PR #1283, opened 2022-09-10
fix(docs): move gotchas to docs/getting-started section

### LINKED ISSUE #1008: Move the docs from Github's wiki to the ./docs directory

dbritto-dev:
## Related Issues

Fixes #1008 

## Summary

Move gotchas to docs/getting-started section

## Check List

### DIFF
--- a/docs/getting-started/gotchas.md
+++ b/docs/getting-started/gotchas.md
@@ -1,0 +1,21 @@
+---
+title: Gotchas
+description: 
+nav: 18
+---
+
+On this page we can somethings that we need to keep in mind when you are using 
+Zustand with or without other libraries.
+
+## Immer
+
+### My subscriptions aren't being called
+
+If you are using ImmerJS, make sure you are actually following the rules of 
+ImmerJS.
+
+For example, you have to add [immerable] = true for class objects to work. If 
+you don't to this, ImmerJS will still mutate the object, but not as a proxy, so 
+it will also update the base state. Zustand checks if the state has actually 
+changed, so since both the base state as well as the next state are equal (if 
+you don't do it correctly), it will skip calling the subscriptions.
--- a/docs/getting-started/gotchas.md
+++ b/docs/getting-started/gotchas.md
@@ -1,21 +1,21 @@
 ---
 title: Gotchas
-description: 
+description:
 nav: 18
 ---
 
-On this page we can somethings that we need to keep in mind when you are using 
+On this page we can somethings that we need to keep in mind when you are using
 Zustand with or without other libraries.
 
 ## Immer
 
 ### My subscriptions aren't being called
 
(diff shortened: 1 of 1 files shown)


dai-shi (changes requested):
I don't think we should have "gotchas" in getting-started, because it shouldn't be something users have to read when they get started. Would there be a better place?

dbritto-dev:
@dai-shi could be integrations a better place for gotchas?

dai-shi:
I don't think so.
@chrisk-7777 has any ideas?

chrisk-7777:
Initially I would think integrations too, because that particular gotcha is talking about immer. But gotcha's as a whole (when others are added overtime) might not fit into the concept of integrations. 

How about the last link in "guides"? Plus a cross link from https://docs.pmnd.rs/zustand/guides/updating-nested-state-object-values#with-immer to this gotchas page?

Perhaps after the line "What a reduction!". For example "Please take note of the gotchas listed here"

dbritto-dev:
I like that idea. I will move gotchas to docs/guides section.

dbritto-dev:
@dai-shi done

docs/guides/gotchas.md
```diff
@@ -0,0 +1,21 @@
...
+### My subscriptions aren't being called
+
+If you are using ImmerJS, make sure you are actually following the rules of
+ImmerJS.
+
+For example, you have to add [immerable] = true for class objects to work. If
+you don't to this, ImmerJS will still mutate the object, but not as a proxy, so
+it will also update the base state. Zustand checks if the state has actually
+changed, so since both the base state as well as the next state are equal (if
+you don't do it correctly), it will skip calling the subscriptions.
```
dai-shi, line 21: Hmm, after reading this again, we should have a doc for `immer` middleware and add this note there?
I kind of feel like avoiding general gotchas at all.

dbritto-dev: We should, since we have `Immer` in our main page -> https://github.com/pmndrs/zustand#sick-of-reducers-and-changing-nested-state-use-immer so people often use `Zustand` with `Immer`.

dbritto-dev: Maybe an integrations page for common integrations with some libraries like `Immer` and put this gotchas in that page would be good idea. What you think?

dai-shi: Yes, as we have "persist middleware" there, let's add "immer middleware"

dbritto-dev:
@dai-shi done

dai-shi (approved):
LGTM
anyone else can review?

chrisk-7777 (approved):

chrisk-7777:
This looks good to me

Issue #15, opened 2019-04-19
Does not trigger rerender


dy:
I intended to use zustand as singleton hook for history, as so:

```js
import { createBrowserHistory } from 'history'
import delegate from 'delegate-it'
import create from 'zustand'

// global-ish history tracker
export const [ useHistory ] = create(set => {
  let history = createBrowserHistory()

  const unlisten = history.listen((location, action) => {
    // NOTE: passing history object as-is does not trigger rerendering
    set(history)
  })

  let delegation = delegate('a', 'click', e => {
    e.preventDefault()
    history.push(e.delegateTarget.getAttribute('href'))
  });
  return history
})
```
The problem is that passing unchanged state object to `set` does not trigger rerendering, as noted in the commentary. For that purpose we have to create a clone as
```js
set({...history})
```
Is that planned behavior? That isn't what expected.
Thanks

drcmda:
It’s using reference equality for change detection. Setting the same object won’t cause Rezensentin, similar to how Redux reduces state.

dy:
The problem persists is even if passing clone of the object
```js
// global-ish history tracker
export const [ useHistory ] = create(set => {
  let history = createBrowserHistory()

  const unlisten = history.listen((location, action) => {
    set({...history})
  })

  let delegation = delegate('a', 'click', e => {
    e.preventDefault()
    history.push(e.delegateTarget.getAttribute('href'))
  });
  return {...history}
})
```

JeremyRH:
It isn't updating because all the enumerable properties on `history` are the same. The check to update state is a "shallow equal" check. It will first check if the value is equal and if not, check all enumerable properties for equality. Even though you created a new history object, the enumerable properties were the same so it did not set the state.

It looks like you are trying to call `set` when the history event is dispatched. You should create your own state with `{ location }` and set the location value in the listener. If you need access to the history object outside of the store, create it outside of the store in a module and import it whenever you need it.

PR #1291, opened 2022-09-13
Add usage with context to docs

Cuppachino:
## Related Issues
#1180 

## Summary
This page covers:
- Vaguely when this approach should be used
- Making a store creator using `createStore` 
- Creating a React.context
- Passing the vanilla store to a context provider
- Consuming a provider's state with `useContext`

This page additionally covers:
- Making a provider wrapper component
- Initializing state with props from a component.
- Extracting context logic into a custom hook (a la `create`)
- Consuming a provider's state with the custom hook.

Notes:
- The examples are fully typed.
- I've expanded the custom hook from #1180 so that it may optionally accept an equality function. 
- The examples do not clearly represent _why_ context is being used for this store. I think people like Zustand because it works without providers, and I'm worried that this page might cause confusion. I'm ignoring this for now in favor of the examples being `typeof yourStore` agnostic (feedback appreciated).
- 
## Check List

### DIFF
--- a/docs/guides/initialize-state-with-props.md
+++ b/docs/guides/initialize-state-with-props.md
@@ -1,14 +1,11 @@
 ---
-title:
+title: initialize-state-with-props
 # nav: N
 ---
 
-<!-- In cases where React.context is required for dependency injection, or when a store should be initialized from component props, the recommended approach is to use a vanilla store. -->
-
-In cases that model [dependency injection](), such as when a store should be initialized with component props, the recommended approach is to use a vanilla store with React.context.
+In cases that model [dependency injection](), such as when a store should be initialized with props from a component, the recommended approach is to use a vanilla store with React.context.
 
 > **Note**: *Because the store returned by* `create` *is a hook, passing it to a context provider may violate the* [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html).
-
 
 ## Store creator with `createStore`
 ---
--- a/docs/guides/initialize-state-with-props.md
+++ b/docs/guides/initialize-state-with-props.md
@@ -86,7 +86,7 @@ function BearProvider({ children, ...props }: BearProviderProps) {
 }
 ```
 
-### Extracting context logic into custom hook
+### Extracting context logic into a custom hook
 
 ```tsx
 // Mimic the hook returned by `create`
--- a/docs/guides/initialize-state-with-props.md
+++ b/docs/guides/initialize-state-with-props.md
@@ -34,9 +34,9 @@ const createBearStore = (initProps?: Partial<BearProps>) => {
 ## Creating a context with `React.createContext`
 
 ```ts
-import { createContext } from 'react';
+import { createContext } from 'react'
 
-export const BearContext = createContext<BearStore | null>(null);
+export const BearContext = createContext<BearStore | null>(null)
 ```
 
 ## Basic component usage
--- a/docs/guides/initialize-state-with-props.md
+++ b/docs/guides/initialize-state-with-props.md
@@ -1,0 +1,129 @@
+---
+title:
+# nav: N
+---
+
+<!-- In cases where React.context is required for dependency injection, or when a store should be initialized from component props, the recommended approach is to use a vanilla store. -->
+
+In cases that model [dependency injection](), such as when a store should be initialized with component props, the recommended approach is to use a vanilla store with React.context.
+
+> **Note**: *Because the store returned by* `create` *is a hook, passing it to a context provider may violate the* [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html).
+
+
+## Store creator with `createStore`
+---
+```ts
+import { createStore } from 'zustand';
+
+interface BearProps {
+  bears: number;
+}
+
+interface BearState extends BearProps {
+  addBear: () => void;
+}
+
+type BearStore = ReturnType<typeof createBearStore>;
+
+const createBearStore = (initProps?: Partial<BearProps>) => {
+  const DEFAULT_PROPS: BearProps = {
+    bears: 0,
+  };
+  return createStore<BearState>()((set) => ({
+    ...DEFAULT_PROPS,
+    ...initProps,
+    addBear: () => set((state) => ({ bears: ++state.bears })),
+  }));
+};
+```
+
+## Basic component usage
+---
+```tsx
+// Provider implementation
+import { useRef } from 'react';
+
+function App() {
+  const store = useRef(createBearStore()).current;
+  return (
+    <BearContext.Provider value={store}>
+      <BasicConsumer />
+    </BearContext.Provider>
+  );
+}
+```
+```tsx
+// Consumer component
+import { useContext } from 'react';
+import { useStore } from 'zustand';
+
+function BasicConsumer() {
+  const store = useContext(BearContext);
+  if (!store) throw new Error('Missing BearContext.Provider in the tree');
+  const bears = useStore(store, (s) => s.bears);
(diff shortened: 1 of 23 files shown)


dai-shi:
cc @arvinxx @dbritto-dev @chrisk-7777

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,126 @@
+---
+title: initialize-state-with-props
```
chrisk-7777, line 2: Can we make this sentence case to make the rest? 
Can we also uncomment the line below and give it a number

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,126 @@
...
+# nav: N
+---
+
+In cases that model [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), such as when a store should be initialized with props from a component, the recommended approach is to use a vanilla store with React.context.
+
+> **Note**: *Because the store returned by* `create` *is a hook, passing it to a context provider may violate the* [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html).
```
chrisk-7777, line 8: Is it possible to remove the quote block here? I don't think its adding much and the website styles look to be struggling when combined with `>` + `****`

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,126 @@
...
+
+function BearProvider({ children, ...props }: BearProviderProps) {
+  const storeRef = useRef<BearStore>();
+  if (!storeRef.current) {
+    storeRef.current = createBearStore(props);
+  }
+  return <BearContext.Provider value={storeRef.current} children={children} />;
+}
+```
+### Extracting context logic into custom hook
```
chrisk-7777, line 88: PR: "into custom hook"
Suggestion: "into a custom hook"

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,126 @@
...
+  selector: (state: BearState) => T,
+  equalityFn?: (left: T, right: T) => boolean
+): T {
+  const store = useContext(BearContext);
+  if (!store) throw new Error('Missing BearContext.Provider in the tree');
+  return useStore(store, selector, equalityFn);
+}
+```
+```tsx
+// Consumer usage with hook
```
chrisk-7777, line 104: PR: "with hook"
Suggestion: "of the custom hook"

chrisk-7777 (changes requested):
Thanks @Cuppachino ! 
I just made a few minor suggestions, let me know if any need clarification

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,126 @@
+---
+title: initialize-state-with-props
+# nav: N
+---
+
+In cases that model [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), such as when a store should be initialized with props from a component, the recommended approach is to use a vanilla store with React.context.
```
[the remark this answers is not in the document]
Cuppachino: great change!

docs/guides/initialize-state-with-props.md
Cuppachino: no problem. `>` + `****` was from the v3 context doc

docs/guides/initialize-state-with-props.md
Cuppachino: > Can we make this sentence case to match the rest?

Is the title okay besides the format? Should it mention of context?

> Can we also uncomment the line below and give it a number

I wasn't sure what number this refers to. What should it be?

chrisk-7777: I think the rest of the title is good 👍 

The number is the overall position in the left hand nav. Unfortunately, it's a bit manual, but find the page you want this page to appear after, grab increment the nav number by one. You should then update any subsequent nav numbers (this is the unfortunate bit, I'm not sure if there is an alternative in the core website atm)

Cuppachino: I would like to put it at the bottom of the "GUIDES" section, under "How to reset state", but it's unclear how the nav number correlates to the entry's position on the site. I've put together an excalidraw sketch highlighting what I mean. 
The numbers on the left are the nav numbers in each doc entry. The numbers on the right are how the nav appears to be ordered. 

zustand-pr-1291-navs excalidraw

chrisk-7777: I'm not super familiar with the core website, but you can checkout the ordering logic here: 

https://github.com/pmndrs/website/blob/docs/src/utils/docs.ts#L144

I'm not sure why that #15 isn't bugging out, I wonder if it has something to do with the way pages are grouped into parents... 

[react three fiber](https://github.com/pmndrs/react-three-fiber/tree/master/docs) also use the same website for their docs, and they seem to follow an order nav count (similar to the above without the mistakes you highlighted). 

I will try to go through an audit the nav numbers for Zustand, unless you want to do it while you're in there and adding this page?

Cuppachino: > I will try to go through an audit the nav numbers for Zustand, unless you want to do it while you're in there and adding this page?

Yes I'd like to do that 😁. Just to clarify one more time, the correct order is on the right side of the image?

chrisk-7777: Yep, looks good to me, except, keep "createContext from zustand/context" for now. So, the final two numbers down the right should be 18 + 19

chrisk-7777 (approved):
Thanks for the back and forth 👍

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,131 @@
...
+## Basic component usage
+
+```tsx
+// Provider implementation
+import { useRef } from 'react'
+
+function App() {
+  const store = useRef(createBearStore()).current
+  return (
+    <BearContext.Provider value={store}>
```
dai-shi, line 44: Should we show how the context is exactly created?
```js
import { createContext } from 'react'

const BearContext = createContext()
```
(or in TS.)

docs/guides/initialize-state-with-props.md
```diff
@@ -0,0 +1,131 @@
...
+import { useRef } from 'react'
+
+type BearProviderProps = React.PropsWithChildren<BearProps>
+
+function BearProvider({ children, ...props }: BearProviderProps) {
+  const storeRef = useRef<BearStore>()
+  if (!storeRef.current) {
+    storeRef.current = createBearStore(props)
+  }
+  return <BearContext.Provider value={storeRef.current} children={children} />
```
dai-shi, line 85:
```suggestion
  return (
    <BearContext.Provider value={storeRef.current}>
      {children}
    </BearContext.Provider>
  )
```
may seem familiar.

dai-shi (commented):
Thanks for working on this.
Please also run `yarn run prettier`.

docs/guides/initialize-state-with-props.md
Cuppachino: Not sure how I left that out. It definitely should be here, and it is now.

docs/guides/initialize-state-with-props.md
```diff
@@ -31,6 +31,13 @@ const createBearStore = (initProps?: Partial<BearProps>) => {
   }))
 }
 ```
+## Creating a context with `React.createContext`
+
+```ts
+import React, { createContext } from 'react';
```
Cuppachino: 1. No reason to import default from React.
2. Run prettier

Issue #1059, opened 2022-07-05
Trying to migrate to v4 but hitting issue with mocking


httpete:
I was using the __mocks__/zustand.ts that contained the boilerplate from the doc, but now I get 
 TypeError: store.getState is not a function

```
import { act } from 'react-dom/test-utils';
import actualCreate from 'zustand';

// a variable to hold reset functions for all stores declared in the app
const storeResetFns = new Set();

// when creating a store, we get its initial state, create a reset function and add it in the set
const create = createState => {
  const store = actualCreate(createState);
  const initialState = store().getState();
  storeResetFns.add(() => store.setState(initialState, true));

  return store;
};

// Reset all stores after each test run
afterEach(() => {
  act(() => storeResetFns.forEach(resetFn => resetFn()));
});

export default create;

**Comment by dai-shi** [comment]
Hm, not sure, but does this help?
```js
const store = actualCreate(createState || () => ({}));
```

httpete:
OK Ill try soon thank you. I absolutely love the simplicity of Zustand 3, but I am finding that Zustand 4 just seems to be harder to use and I am not sure of the benefit. I really like to keep up to date, but my first shot at migrating us to 4rc1 didn't go too well. I will try again. When do you foresee a final 4?

dai-shi:
> When do you foresee a final 4?

I will merge #953, #1029, #1051, when they are ready, and release rc.2.
If we don't get any issues with it, we will finalize it.

httpete:
Thank you @dai-shi I will try right away and provide feedback.

devanshj:
> I am not sure of the benefit.

You can see the diff in `middlewareTypes.test.tsx` between v3.6.9 and v4. Here's an excerpt...

Before (v3.6.9):

```ts
const useStore = create<
  CounterState,
  SetState<CounterState>,
  GetState<CounterState>,
  StoreApiWithSubscribeWithSelector<CounterState> &
    StoreApiWithPersist<CounterState> &
    StoreApiWithDevtools<CounterState>
>(
  devtools(
    subscribeWithSelector(
      persist(
        (set, get) => ({
          count: 0,
          inc: () => set({ count: get().count + 1 }, false),
        }),
        { name: 'count' }
      )
    ),
    { name: 'prefix' }
  )
)
```

After (v4.0.0-rc1):
```ts
const useStore = create<CounterState>()(
  devtools(
    subscribeWithSelector(
      persist(
        (set, get) => ({
          count: 0,
[...]

Issue #271, opened 2021-01-06
[Proposal] 🚀 Testing utilities for zustand


3nvi:
## Background

`zustand` is an amazing package that doesn't depend on React Context providers at all. The stores live "outside" of React and can be accessed from everywhere, even outside of React components. 

Because of this, when testing React components via a testing library, the states of the stores don't reset across tests. That means that the code of one test may affect the outcome of another test. It also means that within a test suite (i.e. `describe` block), test order matters, since a test might modify the state of a store that another test may be depending upon.

There's currently no official way of resetting store state before/after each test. A custom solution would include manually extracting the `initialState` of a store and using it to force-set a store's state before each test run:

```js
const initialStoreState = useStore.getState();

describe('Suite', () => {
  
  beforeEach(() => {
    useStore.setState(initialStoreState, true);
  });

  ... // tests

});
```

This can quickly become tedious since:

1. You have to manually do that for each test suite/file
2. You have to do that for each and every store in your app or remember the exact stores your tests will be using

An alternative solution involves "mocking" zustand and automatically resetting all stores after each test by keeping track of all stores in the app. This involves creating a `__mocks__/zustand.js` file and putting the following code in there:

```jsx
import actualCreate from 'zustand';

// a variable to hold reset functions for all stores declared in the app
const storeResetFns = new Set();

// when creating a store, we get its initial state, create a reset function and add it in the set
const create = createState => {
  const store = actualCreate(createState);
  const initialState = store.getState();
  storeResetFns.add(() => store.setState(initialState, true));
  return store;
};

// Reset all stores after each test run
afterEach(() => {
  storeResetFns.forEach(resetFn => resetFn());
});
```

This has the benefits of:

1. Not having to modify your source code or tests in any shape or form
2. Not having to do anything for each new store you create

## Goal

Have an officially supported way of testing zustand, while limiting the developer's effort to achieve that

## Proposal

We create a separate package named `zustand/testing` which will contain a polished & safe version of the snippet above. From there there are 2 options:

### Option 1 - Automatic mocking

This option automatically mocks `zustand`'s default export, without the developer needing to create any mock files.

Whenever `process.env.NODE_ENV` has a value of `test` (which currently stands true for all major test runners), `zustand` will silently replace the default `create` function with a "mocked" version of `create` (available via `zustand/testing`) that will enable the reset functionality portrayed above.

[...]

PR #1303, opened 2022-09-21
docs: testing for using typescript

mugi-uno:
## Related Issues

(None)

## Summary

Add an explanation of how to mock in TypeScript to the Testing doc.
(If follow the TypeScript Guide, create is executed in a different way.)

## Check List


docs/guides/testing.mdx
```diff
@@ -38,6 +38,34 @@ The way you mock a dependency depends on your test runner/library.
 
 In [jest](https://jestjs.io/), you can create a `__mocks__/zustand.js` and place the code in that file. If your app is using `zustand/vanilla` instead of `zustand`, then you'll have to place the above code in `__mocks__/zustand/vanilla.js`.
 
+### TypeScript Usage
```
chrisk-7777, line 41: nit: can we change this to sentence case? "TypeScript usage"
(I'm slowly trying to get this consistent across the docs)

docs/guides/testing.mdx
```diff
@@ -38,6 +38,34 @@ The way you mock a dependency depends on your test runner/library.
...
+
+If you are using zustand according to [TypeScript Guide](./typescript.md), use the following code.
```
chrisk-7777, line 43: Would you be ok with a slight tweak here? I'm thinking something like:

```
If you are using zustand, as documented in [TypeScript Guide](./typescript.md), use the following code:

```

docs/guides/testing.mdx
```diff
@@ -38,6 +38,34 @@ The way you mock a dependency depends on your test runner/library.
...
+  return store
+}
+
+// Reset all stores after each test run
+beforeEach(() => {
+  act(() => storeResetFns.forEach((resetFn) => resetFn()))
+})
+
+export default create
+```
```
chrisk-7777, line 67: I need to test this locally, but can I confirm this is essentially identical to the above example, the only change is line 54?

chrisk-7777 (commented):
Thanks @mugi-uno !
I just had a few comments/questions 👍

chrisk-7777:
I'm just testing this locally, as well as reading through this thread (https://github.com/pmndrs/zustand/issues/1059), and I think the testing doc in general needs a pretty big uplift

dai-shi:
For reference: #271

docs/guides/testing.mdx
chrisk-7777, line 67: I get type errors locally for this snippet. 

I believe something like this would work:

```suggestion
import actualCreate, { StateCreator } from 'zustand'
// const actualCreate = jest.requireActual('zustand') // if using jest
import { act } from 'react-dom/test-utils'

// a variable to hold reset functions for all stores declared in the app
const storeResetFns = new Set<() => void>()


// when creating a store, we get its initial state, create a reset function and add it in the set
const create = <S>(createState: StateCreator<S>) => {
  const store = actualCreate<S>(createState)
  const initialState = store.getState()
  storeResetFns.add(() => store.setState(initialState, true))
  return store
}

// Reset all stores after each test run
beforeEach(() => {
  act(() => storeResetFns.forEach((resetFn) => resetFn()))
})
```

chrisk-7777:
Thanks dai-shi. I'll read through that today and try to come up with a consolidated testing page over the next few days or so

docs/guides/testing.mdx
mugi-uno: I was not able to take into account the type.
I tried it and it worked correctly, including type checking 👌

mugi-uno:
@chrisk-7777 
Thanks for review.
I fixed all.
f08f316

chrisk-7777 (approved):
Thank you for looking at the feedback

PR #1314, opened 2022-09-26
fix: should use Object.is to compare two value

hchlq:
## Related Issues

Fixes #.

## Check List


dai-shi (approved):
Good point. We used to assume state is an object, but since v4.1, we support any values.

bybruno (approved):

PR #1325, opened 2022-09-29
fix(docs): type replace Jotai with Recoil

DiemasMichiels:
## Related Issues

Fixes #.

## Summary

Typo fix: Jotai is used instead of Recoil in the Recoil section

## Check List


dai-shi (approved):
Nice catch!

PR #1326, opened 2022-09-29
docs: expose `enabled` devtools options in the documentation

FKSI:
## Summary

Document the `enabled` option for devtools middleware.

## Check List

### DIFF
--- a/docs/recipes/recipes.mdx
+++ b/docs/recipes/recipes.mdx
@@ -379,6 +379,8 @@ import { devtools } from 'zustand/middleware'
 const useStore = create(devtools(store))
 // Usage with a redux store, it will log full action types
 const useStore = create(devtools(redux(reducer, initialState)))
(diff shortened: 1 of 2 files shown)


chrisk-7777:
Looks really good @FKSI 

We're currently in the process of transitioning more and more of the readme/wiki over into the docs (https://docs.pmnd.rs/zustand/recipes/recipes#redux-devtools)

Could we capture the above in here too: https://github.com/pmndrs/zustand/blob/main/docs/recipes/recipes.mdx

FKSI:
@chrisk-7777  didn't know about the transition (which I think provide def. a better dev experience). I made the requested change :).

chrisk-7777 (approved):
Winner, thanks for making that addition. 

Also 100% agree about the docs

dai-shi:
Looks good.
Please note that we would prefer minimizing readme with moving some additional contents to docs out of readme.

PR #1330, opened 2022-09-30
docs: add instruction on how to install package using yarn in getting started section

Rec0iL99:
## Summary
Show users how to install zustand using yarn in the documentation website.

## Check List


dai-shi (commented):
I think this is a doc file still under construction.

chrisk-7777:
This one is still under construction. 

My only hesitation around adding the `yarn` version is its inconsistent with the other pmndrs doc sites, which only show `npm` installs. React Three Fiber, Jotai, etc only have `npm`. 

I'd probably lean towards consistency, what are your thoughts dai-shi?

dai-shi:
I don't mind having both npm and yarn _if_ that helps people. (unless it's in readme which we want to minimize.)

---
tbh, I assume yarn users know how to convert npm commands, but jotai and valtio readme files shows yarn commands, I guess.

chrisk-7777 (approved):

PR #1342, opened 2022-10-04
Typed redux action object

marckassay:
This PR introduces one exported TypeScript type (`ReduxAction`) that developers can use for using redux middleware. It requires one type parameter in the same shape as `FeatureEventActions`.

As mentioned in the modified doc of this PR:

"By leveraging typings as above eliminates the need to use any type assertions in reducers. It also enforces only valid values for properties of `action` objects that can increase productivity by eliminating mistakes by any developer."  

## Summary

It would be redundant to have a summary here since the modified doc and test files can be read clearly. At least I believe so.

## Check List

### DIFF
--- a/src/middleware/redux.ts
+++ b/src/middleware/redux.ts
@@ -3,44 +3,27 @@ import type { NamedSet } from './devtools'
 
 type Write<T, U> = Omit<T, keyof U> & U
 
-type LitObjAction = {
-  type: unknown
-} & Partial<{ [key: string]: unknown }>
-type TupleAction = [unknown, unknown?]
-type Action = LitObjAction | TupleAction
-type Tupleize<T> = T extends LitObjAction
-  ? [T]
-  : T extends TupleAction
-  ? T
-  : never
+type Action = { type: unknown }
 
-type StoreRedux<AIn, AOut> = {
-  dispatch: (...a: Tupleize<AIn>) => AOut
-
+type StoreRedux<A> = {
+  dispatch: (a: A) => A
   dispatchFromDevtools: true
 }
 
-type ReduxState<AIn, AOut> = {
-  dispatch: StoreRedux<AIn, AOut>['dispatch']
+type ReduxState<A> = {
+  dispatch: StoreRedux<A>['dispatch']
 }
 
-type WithRedux<S, A> = A extends [unknown, unknown]
-  ? Write<S, StoreRedux<A[0], A[1]>>
-  : never
+type WithRedux<S, A> = Write<S, StoreRedux<A>>
 
 type Redux = <
   T,
-  AIn extends Action,
-  AOut = AIn,
+  A extends Action,
   Cms extends [StoreMutatorIdentifier, unknown][] = []
 >(
-  reducer: (state: T, action: Required<AIn>) => T,
+  reducer: (state: T, action: A) => T,
   initialState: T
-) => StateCreator<
-  Write<T, ReduxState<AIn, AOut>>,
-  Cms,
-  [['zustand/redux', [AIn, AOut]]]
->
+) => StateCreator<Write<T, ReduxState<A>>, Cms, [['zustand/redux', A]]>
 
 declare module '../vanilla' {
   interface StoreMutators<S, A> {
@@ -54,36 +37,20 @@ type PopArgument<T extends (...a: never[]) => unknown> = T extends (
   ? (...a: A) => R
   : never
 
-type ReduxImpl = <T, AIn extends Action, AOut = AIn>(
-  reducer: (state: T, action: Required<AIn>) => T,
+type ReduxImpl = <T, A extends Action>(
+  reducer: (state: T, action: A) => T,
   initialState: T
-) => PopArgument<StateCreator<T & ReduxState<AIn, AOut>, [], []>>
+) => PopArgument<StateCreator<T & ReduxState<A>, [], []>>
 
 const reduxImpl: ReduxImpl = (reducer, initial) => (set, _get, api) => {
   type S = typeof initial
-  ;(api as any).dispatch = <A>(...action: Tupleize<A>) => {
-    let actionOut: any = action
-
-    const isLitObjAction = (arg: any): arg is [LitObjAction] =>
-      Array.isArray(arg) && arg.length === 1 && arg[0] instanceof Object
-
-    if (isLitObjAction(action)) {
-      actionOut = action.pop()
-    }
-
-    ;(set as NamedSet<S>)(
-      (state: S) => reducer(state, actionOut),
-      false,
-      actionOut
-    )
-
-    return actionOut
+  type A = Parameters<typeof reducer>[1]
+  ;(api as any).dispatch = (action: A) => {
+    ;(set as NamedSet<S>)((state: S) => reducer(state, action), false, action)
+    return action
   }
   ;(api as any).dispatchFromDevtools = true
 
-  return {
-    dispatch: <A>(...a: Tupleize<A>) => (api as any).dispatch(...a),
-    ...initial,
-  }
+  return { dispatch: (...a) => (api as any).dispatch(...a), ...initial }
 }
 export const redux = reduxImpl as Redux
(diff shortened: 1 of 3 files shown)


marckassay:
Hello @dai-shi! I just made another commit that reduces the footprint of this PR. I just realized TypeScript can infer the type from the `action` parameter of a reducer to the `dispatch` function of a store. So now the only change to the source code that would be modified if merged, is the type for the `a` parameter:

```typescript
dispatch: (a: PayloadOptionalIfUndefined<A>) => A
```

The type, `PayloadOptionalIfUndefined<A>` only enables when `A` is in the shape of `{type: any, payload: undefined}`, otherwise `A` is used. Although, `{type: any, payload: undefined}` can still be used if desired or if it's existing code (backward compatible).

I'm not sure exactly what you mean when you mentioned, "...`Ain` and `Aout` types". Other than "Action in" and "Action out" but exactly where is still unknown to me. At any rate, I don't believe there is a more minimal way than what I have now for `dispatch`.

The types from the previous commit are still in this PR. Again, the only exported type (`ReduxAction`) is the one the developer uses on the `action` parameter of a reducer if desired. I suppose those can be externalized as you suggested. Perhaps having a package that is simply a module augmentation of zustand's redux module that provides what this PR is proposing? If you prefer that then please close this PR and I'll do just that.

dai-shi:
Yeah, my suggestion wasn't clear, and it's because it's not yet clear to me either. Let me try something and if it works, I may add some commits.

dai-shi:
e9b0f4d is what I meant, but on second thought, it's not necessary and not correct typing anyway.

dai-shi:
> If you prefer that then please close this PR and I'll do just that.

So, you can either a) update docs accordingly, or b) close this and work on an external package.

marckassay:
Thanks dai-shi for [`e9b0f4d`](https://github.com/pmndrs/zustand/commit/e9b0f4dbb91c94e61ec37556239c087a0bc2ec68)! I have been experimenting with it by incorporating external TS types as you demonstrated in 'middlewareTypes.test.tsx'. Last week when attempting to implement a module augmentation of zustand, I noticed your comment and worked off your commit instead.

Having `AIn`and `AOut` is good! I noticed types can now be externalized especially for `AIn`. As for `AOut`, I’m not sure what benefit if any it can bring to zustand. I see that it’s used as the return type for `dispatch` and the `StateCreator` type. I also noticed that `AIn` is used for `dispatch`’s parameter and `reducer`’s action parameter. I’m not sure if that was intended. It might be best to keep `reducer`’s action parameter inferred and have your `AIn` be used for `dispatch`’s parameter. Therefore TS will infer `the reducer`’s action parameter and this opens up the possibility to override the `dispatch`’s parameter. This can allow having the “payload” parameter optional when `type` is set from a discriminated union. 

I’ll push here shortly, but in addition, I’ve added something else. I changed `dispatch` to be a variadic function. In addition to `action`/`a` parameter accepting a literal object with type property, it allows tuples (lengths of 1 or 2) to be used. Doing so required some changes to the implementation code. See the updated tests for more info. Let me know if this is something you’re interested in or not.

dai-shi:
I don't think `AIn`/`AOut` is good. It's kind of a lie.
I'm actually fine with making `A` type extends `unknown` if there is a use case. ( @devanshj suggested to follow redux convention, because it will not work well with `devtools`.)
I would hesitate to change the implementation code. If you have a big idea for new benefits/patterns, I would suggest to create a 3rd-party library.

src/middleware/redux.ts
```diff
@@ -3,27 +3,44 @@ import type { NamedSet } from './devtools'
...
+type TupleAction = [unknown, unknown?]
+type Action = LitObjAction | TupleAction
+type Tupleize<T> = T extends LitObjAction
+  ? [T]
+  : T extends TupleAction
+  ? T
+  : never
+
+type StoreRedux<AIn, AOut> = {
+  dispatch: (...a: Tupleize<AIn>) => AOut
```
dai-shi, line 18: I wonder it can be something like this:
```suggestion
type StoreRedux<A, Rest> = {
  dispatch: (a: A, ...rest: Rest) => A
```

src/middleware/redux.ts
```diff
@@ -37,20 +54,36 @@ type PopArgument<T extends (...a: never[]) => unknown> = T extends (
...
-    ;(set as NamedSet<S>)((state: S) => reducer(state, action), false, action)
-    return action
+  ;(api as any).dispatch = <A>(...action: Tupleize<A>) => {
+    let actionOut: any = action
+
+    const isLitObjAction = (arg: any): arg is [LitObjAction] =>
+      Array.isArray(arg) && arg.length === 1 && arg[0] instanceof Object
+
+    if (isLitObjAction(action)) {
+      actionOut = action.pop()
```
dai-shi, line 71: shouldn't mutate `action`. (but it's just a nit.)

dai-shi (changes requested):
I'm not sure if I understand the precise goal here,
but changing the implementation and types too much is not acceptable.
If it's really what you want and it values, I'd suggest to create a third-party library.

What I wonder is if unlimiting the type is enough or not:
```diff
- type Action = { type: unknown }
+ type Action = unknown
```

marckassay:
> ...but changing the implementation and types too much is not acceptable.

I concur. 

For this suggestion:

```typescript
type StoreRedux<A, Rest> = {
  dispatch: (a: A, ...rest: Rest) => A 
  ...
```

Would require changing the implementation code too. TS will see we're trying to lie to it again. :D And for `type Action = unknown`, there is something in zustand I believe would need to be changed. A discriminated union of `string | {type: unknown} | undefined` if I recall correctly.

But I just realized, the second type parameter of `redux` can have a different type than what the action parameter of the reducer is. So there *may* not be a need to change any code to have typed literal object actions available. Although tuple actions as we know would require changes in implementation code. I say *may* because the commit I just made, is a revert of redux.ts to what it was prior to this PR. But the test errors by TS when `devtools(...)` is uncommented. I'm not sure why that is exactly at this point but just wanted to get back to you before the end of the day for me here. Perhaps the error has something to do with the 3rd type parameter of `redux`?:

```
Cms extends [StoreMutatorIdentifier, unknown][] = []
```

dai-shi:
> for `type Action = unknown`, there is something in zustand I believe would need to be changed. A discriminated union of `string | {type: unknown} | undefined` if I recall correctly.

Yes, that breaks `devtools`, so we need to somehow deal with it.

Can you explain what was the issue with 225fec7?

marckassay:
I believe you are asking about [`e9b0f4d`](https://github.com/pmndrs/zustand/commit/e9b0f4dbb91c94e61ec37556239c087a0bc2ec68)? If so, there is no issue with that commit. However, I don't know what additional benefits it can provide. Perhaps I misunderstand its intent.

So this test works:
```typescript
    const useBoundStore = create(
      devtools(
        redux(
          (
            state: { count: number },
            { type, payload }: AppAction<FeatureEventActions>
          ) => {
            switch (type) {
              case 'grumpiness/increase':
                return { count: state.count + payload }
              case 'grumpiness/decrease':
                return { count: state.count - payload }
              case 'grumpiness/reset':
                return { count: 0 }
              default:
                return state
            }
          },
          { count: 0 }
        )
      )
    )
```

So apparently having no generic type parameters for `redux` prevents the TS error from occurring. 

Conclusion for this PR:

Since type constraints have been decided to be externalized and since there is no need to change the source code to accomplish this for either `dispatch` or a `reducer`'s action parameter, and changes to implementation code would be required to have `dispatch` as a variadic function, I think this PR can be closed. And having any documents demonstrating this would require all of the types to be shown which may be distracting to the reader, in my opinion. But creating an npm package that can be just those types, will be sufficient.

Thanks for your time @dai-shi! Hopefully, not too much time was wasted on your end. I definitely learned more about zustand's internals!

dai-shi:
> And having any documents demonstrating this would require all of the types to be shown which may be distracting to the reader, in my opinion. But creating an npm package that can be just those types, will be sufficient.

Okay, I get that.

I meant https://github.com/marckassay/zustand/tree/225fec71f77cc5aa2db8cbfccc9dd65467bfa2a6, which doesn't change the zustand code, but just makes your test pass with distracting types.

marckassay:
@dai-shi Ah, I understand. I just published [zustand-redux-extends-type](https://github.com/marckassay/zustand-redux-extends-type) to the npm registry. Thanks again for all your input!

PR #1343, opened 2022-10-05
Update outdated LICENSE year

Serjobas:
Updated the outdated copyright year to the present.  
So there will not be a need for any further license year updates


dai-shi (commented):
AFAIK, copyright year is only meaningful for the creation year, so we don't need to update it.

PR #1348, opened 2022-10-07
Rewrite `TakeTwo`, `SkipTwo` and `Mutate` to make them work for older ts versions too

### LINKED ISSUE #1013: TS issue with the set function when using devtools (with older TS versions)

devanshj:
Fixes #1013. Superceeds #1275.

## What is the issue?

The core of the issue is this: TypeScript changed subtyping rules of tuples with optional elements (probably) in v4.4.4. The type `Test` in the following snippet resolves to `"yes"` in v4.3.5 and before but it resolves to `"no"` in 4.4.4 and after.

```ts
type Test = [string, number?] extends [unknown] ? "yes" : "no"
```
This should most probably be intentional as the change is so old and 4.4.4 also happens to be the version that came with `--exactOptionalProperties` so it's possible it touched upon tuples with optional element. It's also possible that this is a "fix" or that it increases the "strictness" of the subtyping rules. (Also turning `--exactOptionalProperties` off or on does not have any effect on the subtyping.)

Now how does it affect us? We have a type `TakeTwo<T>` that does a type-level equivalent of...

```js
const takeTwo = _x => {
  let y = [], x = [..._x];
  while (y.length < 2) y.push(x.shift())
  return y
}

// This is similar to `.slice(0, 2)` except we fill the remaining spaces with `undefined`s
// so that the final result is of length 2.
```

Now the problem is `TakeTwo` doesn't work for pre-4.4.4 version...

```ts
type TakeTwo<T> =
  T extends [] ? [undefined, undefined] :
  T extends [unknown] ? [...a0: T, a1: undefined] :
  T extends [unknown?] ? [...a0: T, a1: undefined] :
  T extends [unknown, unknown] ? T :
  T extends [unknown, unknown?] ? T :
  T extends [unknown?, unknown?] ? T :
  T extends [infer A0, infer A1, ...unknown[]] ? [A0, A1] :
  T extends [infer A0, (infer A1)?, ...unknown[]] ? [A0, A1?] :
  T extends [(infer A0)?, (infer A1)?, ...unknown[]] ? [A0?, A1?] :
  never

type Test = TakeTwo<[state: number, replace?: boolean]>
// Test is [state: number, replace: boolean | undefined, a1: undefined] <4.4.4
// and [state: number, replace: boolean | undefined] >= 4.4.4
```

So now even though our code is "correct" it doesn't work with pre-4.4.4. Same is the case for `SkipTwo` that does a type-level `.slice(2)`.

## How do we solve it?

Now in some sense we can't write a code that works for both versions because typescript changed the math itself.

Imagine you have a function `f` as `const f = (a, b) => a <> b`. Now in wild-land-4.3 `<>` did subtraction, but now in wild-lang-4.4 `<>` does addition. Now you can't write a function `f` that does addition in both wild-lang-4.3 and wild-land-4.4.

So if we want to write `TakeTwo` that works for both pre-change and post-change versions, we'll have to use only those rules that are same in both, some subtyping rules for tuples with optional element are not same so we can't use them.

What is consistent in both is the length property, eg `[string, number?]["length"]` is `1 | 2` for both versions. So we can compare the lengths instead of comparing the tuples themselves...

```ts
type Test = [string, number?] extends { length: 1 | 2 }  ? "yes" : "no"
// Test is "yes" in <4.4.4 and >=4.4.4
```

So the new `TakeTwo` looks like this...

```ts
type TakeTwo<T> =
  T extends { length: 0 } ? [undefined, undefined] :
  T extends { length: 1 } ? [...a0: Cast<T, unknown[]>, a1: undefined] :
  T extends { length: 0 | 1 } ? [...a0: Cast<T, unknown[]>, a1: undefined] :
  T extends { length: 2 } ? T :
  T extends { length: 1 | 2 } ? T :
  T extends { length: 0 | 1 | 2 } ? T :
  T extends [infer A0, infer A1, ...unknown[]] ? [A0, A1] :
  T extends [infer A0, (infer A1)?, ...unknown[]] ? [A0, A1?] :
  T extends [(infer A0)?, (infer A1)?, ...unknown[]] ? [A0?, A1?] :
  never

type Cast<T, U> =
  T extends U ? T : U

type Test = TakeTwo<[state: number, replace?: boolean]>
// Test is [state: number, replace?: boolean] in both <4.4.4 and >= 4.4.4
```

(The last three cases remain the same as the subtyping rule is consistent for them, and nor there's another way of writing them.)

And we make a similar change for `SkipTwo` also.

## How do we test it

Now usually libraries only target the latest typescript version, or whatever version the repo is running on, but here we have to target an old version.

This usualy would have been simple and we could test it with a script like so...

```sh
tsc --noEmit # type-check with current ts version
./node_modules/typescript-4.3.5/bin/tsc --noEmit # type-check with old ts version
```

But now the problem is the version 4.3.5 is so old that there would be some current tests that would be failing because we're using new features, eg `--exactOptionalProperties`. So now those type-check errors are expected, so we can't just rely on the exit code. What we'll have to do see if the tests type-check except some errors that we expect.

To do this I'm using inline snapshots. That is the `oldTsc --noEmit` output should only contain type errors we expect.

To confirm that this change in `TakeTwo` and `SkipTwo` fixes the problem, see the snapshot in first commit, it has 38 errors, but now after we make this change, the new snapshot in the second commit only has 6 expected errors that we can't fix because it's an older tsc that doesn't support the new features and changes.


devanshj:
Also having two `typescript` packages somehow makes the `tsc` binary point to `4.3.5` instead of `4.7.4`. I'm not sure if there's a way to tell yarn to not install the binary for `4.3.5` version. As a fix I've added a `tsc` script that points to the correct binary...

```diff
+ "tsc": "./node_modules/typescript/bin/tsc",
- "pretest": "tsc --noEmit",
+ "pretest": "yarn tsc --noEmit",
```

Let me know if there's a better way.

Also if you want to remove the this test and testing strategy then I'm fine with that too, but I think it's nice to have it.

dai-shi (commented):
Wow, nice hack. I have a different testing strategy which only works on workflows. Let me try.

devanshj:
> Wow, nice hack.

Haha thanks.

> I have a different testing strategy which only works on workflows. Let me try.

Sure.

If you want that test to run only on ci, then I think there are several ways to do, probably the simplest would be make a new test script in package.json for ci that uses `--testPathIgnorePatterns ci-only` and rename `issue1013.test.tsx` to `issue1013.ci-only.test.tsx`.

devanshj:
> [2a28f8e](https://github.com/pmndrs/zustand/pull/1348/commits/2a28f8ea1765672282bfcbec29cb0b40c11eb0db) add a workflow to test with old tsc

Ha interesting. But unfortunately this won't work because ther's [an error in the source too](https://github.com/devanshj/zustand/blob/6d1af698cb027a32eb3224443c42078f86d538f3/tests/issue1013.test.tsx#L15-L18)...

```
"src/vanilla.ts:104:17 - error TS2590: Expression produces a union type that is too complex to represent.
104   createState ? createStoreImpl(createState) : createStoreImpl) as CreateStore
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dai-shi:
> Ha interesting. But unfortunately this won't work because ther's [an error in the source too](https://github.com/devanshj/zustand/blob/6d1af698cb027a32eb3224443c42078f86d538f3/tests/issue1013.test.tsx#L15-L18)...
> 
> ```
> "src/vanilla.ts:104:17 - error TS2590: Expression produces a union type that is too complex to represent.
> 104   createState ? createStoreImpl(createState) : createStoreImpl) as CreateStore
>                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

Yeah, I notice that too. But, then it means that this PR doesn't fix for older ts versions, doesn't it?

devanshj:
> Yeah, I notice that too. But, then it means that this PR doesn't fix for older ts versions, doesn't it?

Technically yes, but that type error won't happen for the user as the user only has `zustand.d.ts` which has only the type not the implementation. So this PR does fix for older ts versions.

dai-shi:
> we'll have to keep in mind old versions too while authoring.

That's my intention actually.

devanshj:
> That's my intention actually.

Ha, well if you want to take supporting older versions seriously then it's not a bad idea :P

devanshj:
> [e5075f1](https://github.com/pmndrs/zustand/pull/1348/commits/e5075f14e719882a1b86b8609aa1f2fc6a783e7d) make Mutate more complete wrt to the domain `StoreMutatorIdentifier[]` to help the type-checker

Let me explain this a bit, imagine we're writing this function...

```ts
const join =
  <L extends string[], D extends string>
  (l: L, d: D) =>
    l.join(d) as Join<L, D>

type Join<L extends string[], D extends string> = 
  L extends [infer H extends string]
    ? H :
  L extends [infer H extends string, ...infer T extends string[]]
    ? `${H}${D}${Join<T, D>}` :
  never

let x = join(["a", "b"] as ["a", "b"], " | ")
// typeof x is "a | b"
```

Often when we write types like `Join` we forget that the domain is `string[]`, which also includes the non-tuple arrays...

```ts
type Test0 = Join<["a", "b"], " | ">
// "a | b"

type Test1 = Join<"a"[], " | ">
// never, should be string

type Test2 = Join<string[], " | ">
// never, should be string
```

The last two arguments are indeed in the domain `string[]` but they have no been handled in the implementation which results in bad results like `never`. This also affects the function `join` of course...

```ts
let x: "a"[] = ["a", "a"]
let test1 = join(x, " | ")
// typeof test1 is never, should be string

let y: string[] = ["a", "b"]
let test2 = join(y, " | ")
// typeof test2 is never, should be string
```

So a more complete version of `Join` would be this...

```ts
type Join<L extends string[], D extends string> =
  number extends L["length"]
    ? string :
  L extends [infer H extends string]
    ? H :
  L extends [infer H extends string, ...infer T extends string[]]
    ? `${H}${D}${Join<T, D>}` :
  never
```

Now the result will always be `string` even for arrays.

Now one might ask the question... But what if we never pass arrays to `Join`? And that indeed is the case with `Mutate`, we always pass tuples, never arrays, so why this change matters and fixes the issue?

The thing is even if we don't pass tuples, typescript does... How? Let's look at the following code...

```ts
const join =
  <L extends string[], D extends string>
  (l: L, d: D) =>
    l.join(d) as Join<L, D>

const identitiyString =
  <T extends string>(t: T) => t

const join2 =
  <L extends string[], D extends string>
  (l: L, d: D) =>
    identitiyString(join(l, d))

type Join<L extends string[], D extends string> = 
  L extends [infer H extends string]
    ? H :
  L extends [infer H extends string, ...infer T extends string[]]
    ? `${H}${D}${Join<T, D>}` :
  number

let x = join2(["a", "b"] as ["a", "b"], " | "))
// typeof x is "a | b"
```

Here one would assume that the only thing that ever gets passed to `Join` is `["a", "b"]`, but that's incorrect. The easiest way to check it is by replacing `never` with `number`. Surely if only `["a", "b"]` is the only thing passed to `Join` we'd never reach that last case and it shouldn't matter what we write there... Let's see what happens...

```ts
const join =
  <L extends string[], D extends string>
  (l: L, d: D) =>
    l.join(d) as Join<L, D>

const identitiyString =
  <T extends string>(t: T) => t

const join2 =
  <L extends string[], D extends string>
  (l: L, d: D) =>
    identitiyString(join(l, d))
//                  ~~~~~~~~~~[1]

type Join<L extends string[], D extends string> = 
  L extends [infer H extends string]
    ? H :
  L extends [infer H extends string, ...infer T extends string[]]
    ? `${H}${D}${Join<T, D>}` :
  number

let x = identitiyString(join(["a", "b"] as ["a", "b"], " | "))
// typeof x is "a | b"

/* 
[1]:
Argument of type 'Join<L, D>' is not assignable to parameter of type 'string'.
  Type 'string | (L extends [infer H extends string, ...infer T extends string[]] ? `${H}${D}${Join<T, D>}` : number)' is not assignable to type 'string'.
    Type 'L extends [infer H extends string, ...infer T extends string[]] ? `${H}${D}${Join<T, D>}` : number' is not assignable to type 'string'.
      Type 'number | `${string}${D}${number}`' is not assignable to type 'string'.
        Type 'number' is not assignable to type 'string'.
          Type 'number' is not assignable to type 'string'.
            Type 'number' is not assignable to type 'string'.
*/
```

Ah ha, the fact that it reached to `number` implies that our `Join` was indeed passed a non-tuple... The thing is even if we ourselves never pass non-tuple, typescript has to type-check our code... while it was type-checking `join2`, it saw that `join(l, d)` must be a string because it's passed to `identityString` that only accepts `string`, so now it calculates what would be the type of `join(l, d)`, the type of `l` is `extends string[]` and the type of `d` is `extends string`, so the type of `join(l, d)` would be `extends Join<extends string[], extends string>` and here `string[]` gets passed to `Join`, which we never expected.

So same happens with our `Mutate`, and hence we were getting this error...

> "src/vanilla.ts:104:17 - error TS2590: Expression produces a union type that is too complex to represent.
> 104   createState ? createStoreImpl(createState) : createStoreImpl) as CreateStore
>                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

(This is the same error in the previous CI in context.test.tsx, typescript seems to report this error only once if the origin is same, because we tested on the build there was no implementation where typescript could encounter it, so the first time it encountered it as in context.test.tsx, but it's the same error.)

Here while type-checking our code, it must have found the need to find out what `Mutate<S, Mos>` resolves to, so it passed the constraint of `Mos` ie `[StoreMutatorIdentifier, unknown][]` to `Mutate` and there because we didn't handle non-tuples it's result was `never` and this probably creates problems and our code didn't type-check.

So we make this change to `Mutate`...

```diff
  export type Mutate<S, Ms> =
+   number extends Ms['length' & keyof Ms] ? S :
    // don't mutate if we get an array, it's just for the type-checker
    // and except that it'll never happen
    Ms extends [] ? S :
    Ms extends [[infer Mi, infer Ma], ...infer Mrs]
      ? Mutate<StoreMutators<S, Ma>[Mi & StoreMutatorIdentifier], Mrs> :
    never
```

Now `Mutate` is more complete as it is now implemented for the whole domain `[StoreMutatorIdentifier, unknown][]` (tuples and non-tuples) instead of just a part of the domain (non-tuples).

dai-shi:
Wow, what "a bit". Thanks for the fix. This is great!

dai-shi (approved):
Let's ship it.

devanshj:
> Wow, what "a bit".

Hahahah, little did I know when I started writing the explanation, how much of "a bit" it'll be xD


> Thanks for the fix. This is great!

Nps and thanks!

PR #1372, opened 2022-10-19
types: add `UnknownBecauseOfGenericMutators`

### LINKED ISSUE #1368: zustand 4.1.2 create() initializer type problems with store mutators
I have a function that extracts out the initializer for a store to a function so that I can create many variants of a particular store, in this case various cache stores.

I was previously working with zustand 4.1.1 and the only type errors that I had were the getter and setter possibly being undefined, so I could suppress them using assertions.

After updating to zustand 4.1.2, there are now new type errors that prevent building, and I cannot for the life of me figure out what's going on with them.

I made a [minimal CodeSandbox demo of the problem available here](https://codesandbox.io/s/zustand-create-initializer-with-storemutator-problem-zer3y6?file=/src/App.tsx).

A possibly notable observation is that without the `M` parameter, the type of `set` is reported to be `(partial: Cache<K, V> | Partial<Cache<K, V>> | ((state: Cache<K, V>) => Cache<K, V> | Partial<Cache<K, V>>), replace?: boolean | undefined) => void`.

With the `M` parameter, it instead becomes `Get<Mutate<StoreApi<Cache<K, V>>, M>, "setState", undefined>`.

The closest I've gotten to figuring out a possible cause is that the `Get` type might not be inferring that the `"setState"` and `"getState"` fields should (I think) be guaranteed to be present, but that's really just grasping at straws.

devanshj:
Closes #1368. Implements the option 1 mentioned in the issue.

### DIFF
--- a/src/middleware/redux.ts
+++ b/src/middleware/redux.ts
@@ -53,4 +53,4 @@ const reduxImpl: ReduxImpl = (reducer, initial) => (set, _get, api) => {
 
   return { dispatch: (...a) => (api as any).dispatch(...a), ...initial }
 }
-export const redux = reduxImpl as Redux
+export const redux = reduxImpl as unknown as Redux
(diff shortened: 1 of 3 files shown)


devanshj:
Closing because we're going with #1371

PR #1373, opened 2022-10-19
fix(types): fix `StateCreator` subtyping

devanshj:
I originally had the phantom type because I assumed it would be required. I'm still a little skeptic about this change but just want to expermiment and see the CI results hence opening the PR.

This PR's branch is based on #1372's branch. Note that this does not require the change in #1372, but has it as base because if we're going with #1371 this PR would probably be much less useful.

### DIFF
--- a/tests/types.test.tsx
+++ b/tests/types.test.tsx
@@ -1,4 +1,5 @@
 import create, { StateCreator, StoreApi, UseBoundStore } from 'zustand'
+import { persist } from 'zustand/middleware'
 
 it('can use exposed types', () => {
   type ExampleState = {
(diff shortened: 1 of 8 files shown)


tests/types.test.tsx
```diff
@@ -188,3 +194,42 @@ it('state is covariant', () => {
...
+  }
+
+  const foo: () => StateCreator<State, []> = () => (set, get) => ({
+    count: 0,
+    increment: () => {
+      set({ count: get().count + 1 })
+    },
+  })
+
+  create<State>()(persist(foo()))
```
devanshj, line 231: So fixing the `StateCreator` subtyping no longer errors on this line as it does today... https://tsplay.dev/mq3pjw

devanshj:
Now that we're going with #1371 this is not much useful, we can revisit later if we want, closing for now.

dai-shi:
> if we're going with https://github.com/pmndrs/zustand/pull/1371 this PR would probably be much less useful.

I get it, but removing `PopArgument` is something looking nice.

devanshj:
I was initially a little skeptic because I misunderstood some nuance, now I'm less skeptic about this change, so we can reopen this if you want. (Ofc not the `UnknownBecauseOfGenericMutators` part but just removing the phantom parameter).

Not only it removes the `PopArgument` stuff but also improves the `StateCreator` in terms of subtyping so it's actually a good change.

Reopening, let me know what you think.

dai-shi (approved):
LGTM

PR #1387, opened 2022-10-26
fix(build): use *.d.mts for esm ('import' condition only)

### LINKED ISSUE #1381: TSC throw errors when compiling with "moduleResolution": "NodeNext"
When I want to import the default `create` function from zustand/vanilla while setting `NodeNext` for both `module` and `moduleResolution` in `tsconfig.json`, TSC will throw an error: `error TS2349: This expression is not callable.`

I mentioned this issue in https://github.com/microsoft/TypeScript/issues/50058#issuecomment-1288155712. However, according to the reply, this is a zustand problem:

> `@zustand/vanilla` has incorrect types, again, it only declares types for a cjs entrypoint for everything, and that is the result of that mismatch.

Kindly request a fix on this problem, thanks!

In the meantime, I use [`default-import`](https://www.npmjs.com/package/default-import) as a temporary patch.

### this doesn't work
```ts
import create from "zustand/vanilla";
export const store = create(() => ({ foo: "bar" }));
```

### this works
```ts
import zustand from "zustand/vanilla";
import { defaultImport } from "default-import";
const create = defaultImport(zustand);
export const store = create(() => ({ foo: "bar" }));
```

A minimal repo to reproduce this problem: https://stackblitz.com/edit/node-vj368k?file=src/index.ts

dai-shi:
close #1381

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "test:dev": "jest --watch --no-coverage",
     "test:coverage:watch": "jest --watch",
     "copy": "shx cp -r dist/src/* dist/esm && shx cp -r dist/src/* dist && shx rm -rf dist/src && shx rm -rf dist/{src,tests} && downlevel-dts dist dist/ts3.4 && shx cp package.json readme.md LICENSE dist && json -I -f dist/package.json -e \"this.private=false; this.devDependencies=undefined; this.optionalDependencies=undefined; this.scripts=undefined; this.prettier=undefined; this.jest=undefined;\"",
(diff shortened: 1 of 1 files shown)


dai-shi:
I wish if `tsc` has an option to produce esm compatible .d.mts files. 🤔

dai-shi:
Thanks to @Sec-ant, this should be good to go. (waiting for a confirmation with middleware.)

Sec-ant:
> Thanks to @Sec-ant, this should be good to go. (waiting for a confirmation with middleware.)

I tried `React`, `subscribeWithSelector`, `persist` and `devtools` and find no problems. Thank you for such quick fix. I'm impressed by your work! @dai-shi

dai-shi:
@Sec-ant Thanks for your help too! We'll merge this and release it a bit later.
Meanwhile, feel free to use codesandbox build if you can.

PR #1388, opened 2022-10-26
Update recipes.mdx: Using subscribe with selector

guenyoo:
Examples used from README and are now copy/pasteable without throwing errors in sandbox

## Related Issues
#1378 @chrisk-7777 @dai-shi 
Fixes #.

## Check List


chrisk-7777 (approved):

dai-shi (approved):
we may need to revisit recipes.mds entirely, but for now it looks good.

PR #1398, opened 2022-10-29
Persisting store doc href hash fixes and tweaks

wmcb91:
## Related Issues

Issue not created, but the problem can be replicated by visiting https://docs.pmnd.rs/zustand/integrations/persisting-store-data, then clicking on the "Hydration and asynchronous storages" link in the second paragraph.

#### Main fix
Make section hash hrefs all lowercase so in-page links work properly on Persisting Store page. The problem likely stemmed from porting Github's implementation of markdown, which is capitalization agnostic about URL hashes, to the new docs framework.

#### Secondary enhancements
- Remove a "see below for more details" link that is linked to the next section.
- Add additional context to a different "see below" statement.

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -409,7 +409,7 @@ Why does it matter?
 Asynchronous hydration can cause some unexpected behaviors.
 For instance, if you use Zustand in a React app, the store will _not_ be hydrated at the initial render. In cases where you app depends on the persisted value at page load, you might want to wait until the store has been hydrated before showing anything (e.g. your app might think the user is not logged in because it's the default, while in reality the store has not been hydrated yet).
 
-If your app does depends on the persisted state at page load, see [How can I check if my store has been hydrated?](#how-can-i-check-if-my-store-has-been-hydrated?) in the Q/A.
+If your app does depends on the persisted state at page load, see **_How can I check if my store has been hydrated?_** in the Q/A section.
 
 ## Q/A
 
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -28,8 +28,6 @@ export const useFishStore = create(
 )
 ```
 
-See [Options](#options) for more details.
-
 ## Options
 
 ### `name`
@@ -206,8 +204,7 @@ export const useBoundStore = create(
 
 If you want to introduce a breaking change in your storage (e.g. renaming a field), you can specify a new version number.
 By default, if the version in the storage does not match the version in the code, the stored value won't be used.
-
-See the `migrate` option below for more details about handling breaking changes.
+You can use the `migrate` option to handle breaking changes in order to persist previously stored data.
 
 ### `migrate`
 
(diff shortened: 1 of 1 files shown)


docs/integrations/persisting-store-data.md
```diff
@@ -412,7 +409,7 @@ Why does it matter?
 Asynchronous hydration can cause some unexpected behaviors.
 For instance, if you use Zustand in a React app, the store will _not_ be hydrated at the initial render. In cases where you app depends on the persisted value at page load, you might want to wait until the store has been hydrated before showing anything (e.g. your app might think the user is not logged in because it's the default, while in reality the store has not been hydrated yet).
 
-If your app does depends on the persisted state at page load, see [How can I check if my store has been hydrated?](#how-can-I-check-if-my-store-has-been-hydrated) in the Q/A.
+If your app does depends on the persisted state at page load, see [How can I check if my store has been hydrated?](#how-can-i-check-if-my-store-has-been-hydrated?) in the Q/A.
```
chrisk-7777, line 412: The inclusion of the "?" is an interesting one. For a quick bit of background, we want the docs to work on the docs site, but also, we want them to continue to work within Github. 

At the moment, the fix here makes sense in the context of the docs (which is currently broken for that link), but it wouldn't work in the context of Github, because it looks like Github strips the "?". 

I wonder if we can introduce a fix on https://github.com/pmndrs/website/ so it is consistent with github? Technically a question mark is allowed in the fragment identifier but it would be good to be consistent across both GH/docs. 

Sorry, I know that is a bit of a rabbit hole for a case fix! Is it something you want to raise/fix with the website repo? If not, I can look at doing it

chrisk-7777:
Thanks! Good eye on the case in those links. I just had a quick comment about the "?"

docs/integrations/persisting-store-data.md
chrisk-7777: A quick update @wmcb91 , I've raised a query over on the website repo: https://github.com/pmndrs/website/issues/259
I want to get Cody's 2c on the approach. Sorry about the hold up

wmcb91:
Hi @chrisk-7777, that was my fault for not checking if the behavior still worked in Github. It's probably worthwhile continuing to explore how to make the website hashes and Github hashes compatible.

If we want to allow that work to be done asynchronously while making this PR mergeable, a solution could be to remove the link and leave the text since the section is directly below and it would be difficult to miss. It could be bold, italic or both to indicate a title section.

Screen Shot 2022-11-01 at 5 49 51 PM

chrisk-7777:
> Hi @chrisk-7777, that was my fault for not checking if the behavior still worked in Github. It's probably worthwhile continuing to explore how to make the website hashes and Github hashes compatible.
> 
> If we want to allow that work to be done asynchronously while making this PR mergeable, a solution could be to remove the link and leave the text since the section is directly below and it would be difficult to miss. It could be bold, italic or both to indicate a title section.
> 
> Screen Shot 2022-11-01 at 5 49 51 PM

Hey @wmcb91 I think this is a great way forward while the site is updated in parallel 👍

chrisk-7777 (approved):

PR #1404, opened 2022-10-31
Docs: added section in slices-pattern explaining how to affect multiple stores in one function

Rec0iL99:
## Related Issues

fixes #1365 

## Check List

### DIFF
--- a/docs/guides/slices-pattern.md
+++ b/docs/guides/slices-pattern.md
@@ -62,9 +62,9 @@ function App() {
 export default App
 ```
 
-### Changing multiple stores together
+### Updating multiple stores
 
-You can affect multiple stores together in a single function itself.
+You can update multiple stores, at the same time, in a single function.
 
 ```js
 import { createBearSlice } from './bearSlice'
--- a/docs/guides/slices-pattern.md
+++ b/docs/guides/slices-pattern.md
@@ -62,6 +62,37 @@ function App() {
 export default App
 ```
 
+### Changing multiple stores together
+
+You can affect multiple stores together in a single function itself.
+
+```js
+import { createBearSlice } from './bearSlice'
+import { createFishSlice } from './fishSlice'
+
+export const createBearFishSlice = (set) => ({
+  addBearAndFish: () => {
+    createBearSlice(set).addBear()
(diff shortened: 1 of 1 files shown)


Aliemeka (commented):
Can you include in this doc an implementation for returning all the stores back to default. In other words like  a `clear` or `reset` functionality.

Rec0iL99:
> Can you include in this doc an implementation for returning all the stores back to default. In other words like a `clear` or `reset` functionality.

The `reset` functionality can be found in the `How to reset state` doc. I can maybe provide a link to the doc from here. Will this be required @dai-shi @chrisk-7777?

docs/guides/slices-pattern.md
```diff
@@ -62,6 +62,37 @@ function App() {
 export default App
 ```
 
+### Changing multiple stores together
+
+You can affect multiple stores together in a single function itself.
```
chrisk-7777, line 67: What do you think about:

"You can update multiple stores, at the same time, in a single function."

 - "affect" may hold a slightly different meaning (for example, like that in useEffect or a side effect)
 - "itself" feels redundant at the end of the sentence. 

In terms of the heading, do you think something succinct like:

"Updating multiple stores"

Thoughts @Rec0iL99 ?

Rec0iL99: After thinking about it, yes I agree.

chrisk-7777 (approved):
Thanks for actioning that feedback. The rest looks good 👍

docs/guides/slices-pattern.md
```diff
@@ -62,6 +62,37 @@ function App() {
...
+You can update multiple stores, at the same time, in a single function.
+
+```js
+import { createBearSlice } from './bearSlice'
+import { createFishSlice } from './fishSlice'
+
+export const createBearFishSlice = (set) => ({
+  addBearAndFish: () => {
+    createBearSlice(set).addBear()
+    createFishSlice(set).addFish()
```
dai-shi, line 76: Interesting. I haven't thought about this pattern.

dai-shi (approved):
Thanks!

PR #1405, opened 2022-11-03
feat: :sparkles: implement createLocal store-hook creator

luciancaetano:
## Summary

This is an implementation proposal for the creation of the createLocal function which has the purpose of creating a store hook to maintain the component's local state for use in design patterns such as mvvm

## Check List

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -1,4 +1,4 @@
-import { useDebugValue } from 'react'
+import { useDebugValue, useState } from 'react'
 // import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector'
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Thanks for the suggestion.
I think this is out of the scope of zustand library.
You can do that on your end without problems. (I would prefer using useRef instead of useState.)

However, it's worth noting in docs as one practice.

Closing this PR, but feel free to open a PR to add docs.

dai-shi:
On second thought, I'm not sure how this is better than using useState. Using zustand instead of useState is technically slow. Unless we have a reasonable benefit over the slowness, it shouldn't be recommended (shouldn't be in docs.)

PR #1425, opened 2022-11-11
docs: fix error usage in typescript at persist

moonrailgun:
## Summary

in my real usage, we must define it with `zustand/persist` and storage state type

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -544,7 +544,7 @@ withStorageDOMEvents(useBoundStore)
 
 ### How do I use with TypeScript?
 
-Basic typescript usage doesn't require anything special except for writing `create<State>()(...)` instead of `create(...)`.
+Basic typescript usage require define middleware and state like below.
 
 ```tsx
 import create from 'zustand'
@@ -555,7 +555,10 @@ interface MyState {
   addAFish: () => void
 }
 
-export const useFishStore = create<MyState>()(
+export const useFishStore = create<
+  MyState,
+  [['zustand/persist', Pick<MyState, 'fishes'>]]
+>()(
   persist(
     (set, get) => ({
       fishes: 0,
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -544,7 +544,7 @@ withStorageDOMEvents(useBoundStore)
 
 ### How do I use with TypeScript?
 
-Basic typescript usage require define middleware and state like below.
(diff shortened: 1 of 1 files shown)


docs/integrations/persisting-store-data.md
```diff
@@ -544,7 +544,7 @@ withStorageDOMEvents(useBoundStore)
 
 ### How do I use with TypeScript?
 
-Basic typescript usage doesn't require anything special except for writing `create<State>()(...)` instead of `create(...)`.
+Basic typescript usage require define middleware and state like below.
```
chrisk-7777, line 547: Thanks @moonrailgun .  I think there is a word or two missing in this updated sentence. 

How about something along this lines: 

"To use the persist middleware with TypeScript, you should describe both the shape of your store as well as each middleware being introduced. `create` accepts two generic parameters. Refer to the example below.

chrisk-7777 (changes requested):
Thanks @moonrailgun ! I have requested one change

docs/integrations/persisting-store-data.md
moonrailgun: well. your word is more detail and nice, look good to me and i follow it

moonrailgun: and a detail is `generic` is not same with `generics`. i think `generics` is more fit in this case

dai-shi:
Thanks @devanshj , let us close this.

moonrailgun:
Oh you are right, @devanshj. Thanks for your reply and i get it.

I just check the first `create(...) ` and follow type checker to change it. and didn't realize with those tiny diff between them.

chrisk-7777:
Thanks @devanshj 

I think this highlights that we need to extend the contribution checklist. 
Is it too extreme to ask for a working Codesandbox snippet for changes that include a code sample?

moonrailgun:
@chrisk-7777 
Sorry but in this case, use manual generics like `[['zustand/persist', Pick<MyState, 'fishes'>]]` is also available, and i think its  typescipt-friendly... because after i change my style from `create<MyState>({...})` to `create<MyState>()({...})`, sometime my typescript language server will broken and need restart.

For example:

my code:
```typescript
interface State {
  list: string[]
}

const useMyStore = create<State>()(
  persist(
    immer((set, get) => ({
      list: []
    }))
  )
)
```

sometimes language server will think my list type is `never[]` instead of `string[]` , and tell me its not match. If i restart language server, everything will be work fine, and after some time pass, still will throw this error. so i have to declare it with `list: [] as string[]`.

i think its maybe have too deep type inference and when if typescript check low memory, will throw away. I am not sure but its a problem.

PR #1428, opened 2022-11-14
docs/recipes: persist example should reference 'fishes', not 'fish'

boatilus:
## Summary

In the 'Persist middleware' example, there is no `fish` property; it should be `fishes`.

## Check List


dai-shi (approved):
LGTM.

Note: This doc is still under construction.

PR #1431, opened 2022-11-15
fix(middleware/devtools): do not use optional catch binding syntax

### LINKED ISSUE #1427: Zustand produces SyntaxError in Chrome<66 (optional catch binding)
## SyntaxError in older browsers due to optional catch binding not transpiled

Zustand codebase utilizes "optional catch binding" JS feature (eg. here: https://github.com/pmndrs/zustand/blob/main/src/middleware/devtools.ts#L163) which is [ECMAScript Stage 4 proposal](https://github.com/tc39/proposal-optional-catch-binding) and is not supported in Chrome versions less than 66: https://caniuse.com/?search=optional%20catch%20binding

Probably due to wrong babel configuration, `try {} catch {}` syntax is not correctly transpiled to `try {} catch (e) {}`. This should be done automatically by [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env#modules) which contains [@babel/plugin-proposal-optional-catch-binding](https://babeljs.io/docs/en/babel-plugin-proposal-optional-catch-binding), but is not: https://unpkg.com/browse/zustand@4.1.4/esm/middleware.js

Wrong syntax produces "SyntaxError" in browsers like Chrome 65:

> Uncaught SyntaxError: Unexpected token {

syntax

image

mskrzypek:
## Related Issues

Fixes #1427.

## Summary

Remove optional catch binding syntax usage in devtools.ts, since it's the only one place it has been used and it is not transpilled correctly to backward compatible syntax.

## Check List

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -160,7 +160,7 @@ const devtoolsImpl: DevtoolsImpl =
     try {
       extensionConnector =
         (enabled ?? __DEV__) && window.__REDUX_DEVTOOLS_EXTENSION__
-    } catch {
+    } catch (e) {
       // ignored
     }
 


dai-shi (approved):
Thanks for the PR!

I will merge it some time later and release it.

PR #1435, opened 2022-11-17
feat(middleware/devtools): Better redux devtools. One connection for all zustand stores

pavlobu:
## Summary
* one redux devtools connection for all zustand stores
* add `store` devtools option to name different stores in devtoolsOptions
* backward compatibility with old api usage
* add devOnlyDevtools export that helps to exclude devtools from prod builds without changing code
* add demo project in documentations
* temporary npmjs package with this PR features: https://www.npmjs.com/package/@pavlobu/zustand

## Check List

- [v] `yarn run prettier` for formatting code and docs

Thank you for this awesome tool - zustand. I really enjoyed to use it and I hope my update helps!


dai-shi:
This looks like a nice improvement.
What do you guys think? @Aslemammad @devanshj 

---

I'm not a big fan of the complexity of the implementation.
- any possibility to make it better?
- tests would be required anyway.
- `process.env` is not always accessible and we can't depend on it. (and we decided to let developers to control disabling the middleware)

pavlobu:
@dai-shi, thank you for your feedback.

I could not find any better solution as of now, any suggestions?
I can try to fix those eslint failures and write some tests.
Regarding process.env do I have to use something else instead? I thought it might be useful for users to disable zustand devtools for react builds this way, but I can remove that feature if it's cumbersome.

pavlobu:
> @pavlobu Can we have a Codesandbox example?

I can create and add codesandbox example here if you want:
https://github.com/pmndrs/zustand/blob/main/.codesandbox/ci.json

If you want to check it now, you can refer to example that I've included in my fork branch of zustand:
https://github.com/pavlobu/zustand/tree/%40pavlobu-zustand-devtools-middleware/examples/dev-env-devtools-many-stores
I've also added example demo in this PR

Aslemammad:
Thank you so much, an example in codesandbox would be so much better.

pavlobu:
@dai-shi the eslint build failure occurs on latest master
it's not due to my latest changes. Here is the command result:

```
zustand   main  npx yarn eslint:ci
yarn run v1.22.19
$ eslint '*.{js,json}' '{src,tests}/**/*.{ts,tsx}'

Oops! Something went wrong! :(

ESLint: 8.26.0

ESLint couldn't find the config "react-app" to extend from. Please check that the name of the config is correct.

The config "react-app" was referenced from the config file in "/zustand/examples/demo/package.json".

If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
 zustand   main  
```

pavlobu:
Hmm, but in the pipeline I see different results
https://github.com/pmndrs/zustand/actions/runs/3484199626/jobs/5836345592

Screenshot 2022-11-17 at 17 31 56

pavlobu:
> Thank you so much, an example in codesandbox would be so much better.

I tried code sandbox, but unfortunately it does not connect to redux devtools extension :(
https://codesandbox.io/s/pavlobu-zustand-demo-frutec

local build works fantastic though

devanshj:
> This looks like a nice improvement.
What do you guys think? @Aslemammad @devanshj

I've never been a user of the redux devtools or zustand, so I can't tell if like or dislike this version or even the existing version :P — an user would be able to gauge better

Aslemammad:
This PR looks good to me, the only concern I have is the complexity of the DX, would it confuse users who only want to watch one store, instead all of the other stores?

pavlobu:
> This PR looks good to me, the only concern I have is the complexity of the DX, would it confuse users who only want to watch one store, instead all of the other stores?

Thanks!
Users who only want to see one store can always type a different `connection name` and don't assign `store` in devtoolsOptions. This implementation is backwards compatible.
We can mention this use case in docs

Aslemammad (approved):
With proper testing reviewed by @dai-shi, This PR would look fully good to me.

dai-shi:
@devanshj

> I've never been a user of the redux devtools or zustand

Oh, never mind then. I thought you have some familiarity as you filed #716, at least more then me.

---

@pavlobu You might be interested in looking into #716 after finishing this PR. Not sure what's left to do there.

dai-shi:
@pavlobu 

> I could not find any better solution as of now, any suggestions?

I'd like to propose something to make the devtools middleware more flexible. It may also make you re-think to organize the code better, less complex, and more readable.

It's to support multiple connections for groups of stores.

image

pavlobu:
@dai-shi I got your point, I will find a way to do such improvement.
For the code, I can do some refactoring. Np

pavlobu:
> @pavlobu
> 
> > I could not find any better solution as of now, any suggestions?
> 
> I'd like to propose something to make the devtools middleware more flexible. It may also make you re-think to organize the code better, less complex, and more readable.
> 
> It's to support multiple connections for groups of stores.
> 
> image

Ok, @dai-shi we have this feature now.
I'm going to write tests before doing refactoring.
Regarding unit tests, what do you think should be tested and how many tests should we have for this feature ?

dai-shi:
The goal of tests here is, say, if I were to refactor the code (no intention to break, but can break with mistakes), I would confirm if I did anything wrong with tests. If tests pass, we will let the release go.

readme.md
```diff
@@ -389,6 +389,19 @@ const usePlainStore = create(devtools(store))
 const useReduxStore = create(devtools(redux(reducer, initialState)))
 ```
 
+One redux devtools connection for multiple stores
+```jsx
+import { devtools } from 'zustand/middleware'
+
+// Usage with a plain action store, it will log actions as "setState"
+const usePlainStore1 = create(devtools(store), { name, store: storeName1 })
```
dai-shi, line 397: Do we have any other naming idea for `store` property?

src/middleware/devtools.ts
```diff
@@ -147,13 +148,37 @@ type DevtoolsImpl = <T>(
 
 export type NamedSet<T> = WithDevtools<StoreApi<T>>['setState']
 
+type ConnectResponse = ReturnType<
+  NonNullable<Window['__REDUX_DEVTOOLS_EXTENSION__']>['connect']
+>
+const connections: Record<string, ConnectResponse> = {}
```
dai-shi, line 154: My suggestion for refactor is to use `Map` and deal `undefined` as a normal case.

pavlobu: @dai-shi 
Ok I will use thé Map

Regarding namings, I thaught of 
storeName
slice
prefix

dai-shi: `name` is to distinguish devtools instances.
`store` is to distinguish states and actions in a single instance.

I want to make that clear with naming. `prefix` might be better, but it's not a prefix. Another idea is `subName`, but not super comfortable. In a sense, it's a key. `storeKey`. Hm, then `store` would be better... ok, let's leave it for now.

readme.md
devanshj: Afaiu, how about instead of having `{ name, store }` we have `{ name, instanceName }`? So the implementation would have bits like `extension.connect({ name: options.instanceName ?? options.name })` and `if (options.instanceName !== undefined) currentStates[options.name] = ....`

pavlobu: I just thought, we already have `type` to name actions, then we could rename `store` to `typePrefix`. Because indeed this property is added as prefix to `type`

pavlobu:
@dai-shi, I've added some tests, please check if this pr is ok to go.
I agree with you that the implementation is complex and adds too many edge cases that need to be handled. I did what I could for now.

src/middleware/devtools.ts
```diff
@@ -147,13 +148,50 @@ type DevtoolsImpl = <T>(
...
+const connections: Map<string | undefined, ConnectResponse> = new Map()
+
+type ConnectionStoreApis = (StoreApi<any> & { store: string })[]
+const connectionStoreApis: Map<string | undefined, ConnectionStoreApis> =
+  new Map()
+type ConnectionInitialStoreStates = (any & { store: string })[]
+const connectionInitialStoreStates: Map<
+  string | undefined,
+  ConnectionInitialStoreStates
+> = new Map()
```
dai-shi, line 163: Can this be just a single map?
```ts
type Name = string | undefined // `name` option
type Store = string | undefined // `store` option
type Connections = Map<Name, [ConnectResponse, Record<Store, StoreApi<unknown>>]>
```
Only having one store api where `store=undefined` is a special case.

---
Oh, we need to allow having multiple `name=undefined` instances? That seems tricky.
Then this? 🤔 
```ts
type Connections = Set<[Name, ConnectResponse, Record<Store, StoreApi<unknown>>]>
```

dai-shi (commented):
I feel like we should refactor a lot and remove edge cases, before merging this PR. Otherwise, nobody will be able to maintain it.

pavlobu:
> I feel like we should refactor a lot and remove edge cases, before merging this PR. Otherwise, nobody will be able to maintain it.

Hi, agree.
Unfortunately I don't know when I will have time for this. 

Here are things i thought about how to do it.
Before doing any major refactoring we need to have a clear goal of what should be implemented. Maybe some diagram would help.

The edge cases are:
1. Name undefined for multiple stores devtools() calls (default case not introduced by this pr). Connections should be separated from each other
2. Name defined - connect stores for different names, reuse same connection for same name for different stores (this case was introduced by suggestion of improvement by @dai-shi ). Connections should be separated from each other
3. Name undefined but store is defined in devtools call, different actions should be called when using devtools controls. Connections should be separated from each other
4. store is defined and anonimousActionType is defined in state action. This action type should be added with store name as prefix for action type

src/middleware/devtools.ts
pavlobu: Yeah, multiple name undefined instances is a default behavior which is not introduced by this pr. And it is not fully covered by tests

dai-shi:
I'm struggling with how to proceed this.
If this is really useful for developers (I personally like the idea, but don't use it. Not sure how many people use it), we might consider introducing breaking changes.
If nobody has time to work on this, it might worth considering to provide it as a third party package for now and get feedback.

cwstra:
@pavlobu: I thought I might take a stab at the refactor, so I made my own fork and [a PR to merge it into yours](https://github.com/pavlobu/zustand/pull/3). Let me know what you think.

Did not add extra test cases yet, since I didn't want to invest too much before knowing if I was stepping on toes.

pavlobu:
@cwstra 
Thanks. No extra tests were added by me so you are ok to try your approach 👍

pavlobu:
Hi @dai-shi,
We have an update.
Thanks, @cwstra for your refactoring work!

dai-shi:
@pavlobu @cwstra 
Thanks for working on this!
Let's try to merge this. I'll review for it.

I'd appreciate if you would be willing to work on adding tests and more refactors even after merging this.

src/middleware/devtools.ts
```diff
@@ -173,22 +202,54 @@ const devtoolsImpl: DevtoolsImpl =
...
+      if (existingConnection) {
+        return { type: 'tracked' as const, store, ...existingConnection }
+      }
+      const newConnection: ConnectionInformation = {
+        connection: extensionConnector.connect(options),
+        stores: {},
+      }
+      trackedConnections.set(options.name, newConnection)
+      return { type: 'tracked' as const, store, ...newConnection }
+    })()
```
dai-shi, line 224: Are you sure we need IIFE here?

src/middleware/devtools.ts
```diff
@@ -147,13 +148,41 @@ type DevtoolsImpl = <T>(
...
+
 const devtoolsImpl: DevtoolsImpl =
   (fn, devtoolsOptions = {}) =>
   (set, get, api) => {
-    type S = ReturnType<typeof fn>
+    const { enabled, anonymousActionType, store, ...options } = devtoolsOptions
+
+    type S = ReturnType<typeof fn> & {
+      [store: string]: S
+    }
```
dai-shi, line 183: Sure about this type? Seems very weird.

src/middleware/devtools.ts
```diff
@@ -241,7 +318,34 @@ const devtoolsImpl: DevtoolsImpl =
...
+                  )
+                }
+                if ((action.state as S)[store] === undefined) {
+                  return
+                }
+                const stateFromDevtools = (action.state as S)[store]
+                if (stateFromDevtools) {
+                  if (
+                    JSON.stringify(api.getState()) !==
+                    JSON.stringify(stateFromDevtools)
```
dai-shi, line 344: This seem pretty unfortunate. Wonder if we have any other solution than deep equal. Should be fine for this PR.

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "target": "esnext",
+    "module": "esnext",
```
dai-shi, line 4: What would happen if we don't add this?

src/middleware/devtools.ts
```diff
@@ -147,13 +148,41 @@ type DevtoolsImpl = <T>(
...
 export type NamedSet<T> = WithDevtools<StoreApi<T>>['setState']
 
+type Connection = ReturnType<
+  NonNullable<Window['__REDUX_DEVTOOLS_EXTENSION__']>['connect']
+>
+type ConnectionName = string | undefined
+type StoreName = string
+type StoreInformation = {
+  api: StoreApi<unknown>
+  initialState: unknown
```
dai-shi, line 158: I'm not confident if we need this `initialState`. Can it be done with "current state" i.e. `api.getState()`?

dai-shi (commented):
There should be more ways to do things better with refactors, but for this PR, we want small changes, and it looks good.

Please see the comments below.

dai-shi (changes requested):
@pavlobu @cwstra
Can you reply to my previous comments? We want to merge this after resolving those.

src/middleware/devtools.ts
cwstra: Extracted this into a helper function in another PR.

src/middleware/devtools.ts
cwstra: I agree, though it's tricky to get rid of entirely. I _did_ manage to get it working without recursively referring to S, at least. @pavlobu might be able to explain it more; I was basically just going with it when doing the initial changes.

src/middleware/devtools.ts
cwstra: Think I managed to hammer this away in the aforementioned PR to the fork.

src/middleware/devtools.ts
cwstra: Agreed. If there's a better way, it's probably in the docs for the redux devtools somewhere.

tsconfig.json
cwstra: From what I can gather by commenting it out, I _think_ it was added so that we could do dynamic imports for the devtools tests. We want to dynamically import the tools to avoid needing to clean up connections from previous tests. @pavlobu would know for sure.

cwstra:
Sorry for the delay, hit a wall for a bit there. Addressed some concerns in [this PR on pavlobu's fork](https://github.com/pavlobu/zustand/pull/4), and replied to them all.

tsconfig.json
pavlobu: Correct @cwstra , that was done for that. To use dynamic imports inside tests.
Thanks!

pavlobu:
@cwstra's changes are merged in my fork PR.
Thank you all and Happy New year!

tsconfig.json
dai-shi: Thanks for clarification!

@pavlobu Can you elaborate why we need dynamic imports in tests?

dai-shi (approved):
Thanks for your work!
Let's merge it and work on other improvements afterwards.

- @pavlobu @cwstra Adding tests and then refactoring would be very nice.
- @chrisk-7777 Adding docs for full usage and simplifying readme would be nice.

chrisk-7777:
>     * @chrisk-7777 Adding docs for full usage and simplifying readme would be nice.

Not a problem, I'll get that organized

PR #1438, opened 2022-11-18
fix(docs): Fix 'State' declaration position

hicucu:
## Related Issues

Fixes #.

## Check List

### DIFF
--- a/docs/guides/auto-generating-selectors.md
+++ b/docs/guides/auto-generating-selectors.md
@@ -14,7 +14,9 @@ However, writing these could be tedious. If that is the case for you, you can au
 ## create the following function: `createSelectors`
 
 ```typescript
-import { State, StoreApi, UseBoundStore } from 'zustand'
+import { StoreApi, UseBoundStore } from 'zustand'
+
+type State = object;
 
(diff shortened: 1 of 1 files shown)


docs/guides/auto-generating-selectors.md
```diff
@@ -14,7 +14,9 @@ However, writing these could be tedious. If that is the case for you, you can au
 ## create the following function: `createSelectors`
 
 ```typescript
-import { State, StoreApi, UseBoundStore } from 'zustand'
+import { StoreApi, UseBoundStore } from 'zustand'
+
+type State = object;
```
dai-shi, line 19: How about making L25 like this?
```ts
const createSelectors = <S extends UseBoundStore<StoreApi<object>>>(
```

dai-shi (commented):
Thanks for the suggestion. It generally looks good. Please see the comment below.

docs/guides/auto-generating-selectors.md
dai-shi: @hicucu Can you update the PR to follow it?

dai-shi (approved):
LGTM
Thanks for your contribution!

PR #1451, opened 2022-11-29
feat(shallow): Add support of Maps and Sets to shallow comparator

anton-lunev:
## Summary
Zustand allows storing Maps and Sets, however, the `shallow` comparator returns incorrect comparison results for these type.
This PR adds support for shallow comparison of ES6 `Map` and `Set`

## Check List

### DIFF
--- a/src/shallow.ts
+++ b/src/shallow.ts
@@ -15,7 +15,7 @@ function shallow<T>(objA: T, objB: T) {
     if (objA.size !== objB.size) return false
 
     for (const [key, value] of objA) {
-      if (value !== objB.get(key)) {
+      if (!Object.is(value, objB.get(key))) {
         return false
       }
     }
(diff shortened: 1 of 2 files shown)


dai-shi:
See some discussions in #1064.
I'm not sure if we should keep modifying `shallow` or simply drop it and suggest alternatives.
But, indeed we have maps and sets usage in docs, so this should be good to go.

Do anyone have any insights?

src/shallow.ts
```diff
@@ -10,6 +10,29 @@ function shallow<T>(objA: T, objB: T) {
   ) {
     return false
   }
+
+  if (objA instanceof Map && objB instanceof Map) {
+    if (objA.size !== objB.size) return false
+
+    for (const [key, value] of objA) {
+      if (value !== objB.get(key)) {
```
dai-shi, line 18: Let's make it consistent:
```suggestion
      if (!Object.is(value, objB.get(key))) {
```

dai-shi (approved):
should be good.

PR #1459, opened 2022-12-03
refactor(vanilla/types): replace hardcoded types with StoreApi type

Arsikod:
Add binding of createStoreImpl methods types to StoreApi without hardcoding.

## Related Issues
In case of changing something in StoreApi type we dont have to change it again inside createStoreImpl

Fixes #.

## Summary
Removing arrow function types reduces number of arrows, and makes it easier to read.
 Makes it easier to read and reason about where the type comes from and belongs to. 

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -60,7 +60,7 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
   let state: TState
   const listeners: Set<Listener> = new Set()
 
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Looks good to me.
Mentioning @devanshj just in case.

devanshj:
Looks fine to me too.

PR #1460, opened 2022-12-04
Fix: Programmatic `module.exports` addition for named and default exports in CJS

### LINKED ISSUE #559: "TypeError: create is not a function" when using with jest configured for ESM
Consider the following setup:

`package.json`:
```js
{
  "name": "repro",
  "version": "1.0.0",
  "type": "module",
  "main": "index.js",
  "dependencies": {
    "jest": "27.1.1",
    "react": "17.0.2",
    "zustand": "3.5.10"
  }
}
```

`index.js`:
```js
import create from 'zustand'

export const useStore = create(set => ({
    foo: 0,
}))
```

`repro.test.js`:
```js
import { useStore } from "."

test("repro", () => {
    console.log(useStore.getState())
});
```

Then, when running `node --experimental-vm-modules node_modules/.bin/jest` I get:

```
    TypeError: create is not a function

      3 | console.log(create)
      4 |
    > 5 | export const useStore = create(set => ({
        |                         ^
      6 |     foo: 0,
      7 | }))
      8 |

      at index.js:5:25
      at TestScheduler.scheduleTests (node_modules/@jest/core/build/TestScheduler.js:333:13)
      at runJest (node_modules/@jest/core/build/runJest.js:387:19)
      at _run10000 (node_modules/@jest/core/build/cli/index.js:408:7)

  console.log
    { default: [Function: create] }
```

It seems that the import is not picking the esm bundle?

barelyhuman:
Patches the base index file to export the default as classic node cjs module

## Related Issues

Fixes #559 

## Check List

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -75,12 +75,15 @@ function createESMConfig(input, output) {
   }
 }
 
-function createCommonJSConfig(input, output, options = {}) {
-  const config = {
+function createCommonJSConfig(input, output, options) {
+  return {
     input,
     output: {
       file: `${output}.js`,
       format: 'cjs',
+      outro: options.addModuleExport
+        ? 'module.exports = exports.default; Object.assign(exports.default, exports)'
+        : '',
     },
     external,
     plugins: [
@@ -97,12 +100,6 @@ function createCommonJSConfig(input, output, options = {}) {
       babelPlugin(getBabelOptions({ ie: 11 })),
     ],
   }
-  if (options.addModuleExport) {
-    config.output.outro =
-      'module.exports = exports.default; Object.assign(exports.default,exports)'
-  }
-
-  return config
 }
 
 function createUMDConfig(input, output, env) {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -82,7 +82,7 @@ function createCommonJSConfig(input, output, options) {
       file: `${output}.js`,
       format: 'cjs',
       outro: options.addModuleExport
-        ? 'module.exports = exports.default; Object.assign(exports.default, exports)'
+        ? 'module.exports = exports.default; Object.assign(exports.default, exports);'
         : '',
     },
     external,
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -75,10 +75,13 @@ function createESMConfig(input, output) {
   }
 }
 
-function createCommonJSConfig(input, output) {
-  return {
+function createCommonJSConfig(input, output, options = {}) {
+  const config = {
     input,
-    output: { file: `${output}.js`, format: 'cjs' },
+    output: {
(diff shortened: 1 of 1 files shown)


barelyhuman:
The base sandboxes work, now we need to see if the node based resolvers work. 

@Pagebakers , do you wish to try the same scenario that you had an issue with ?

Process, 
Just install the package from the codesandbox build results

https://ci.codesandbox.io/status/pmndrs/zustand/pr/1460/builds/322747

There's `Local install instructions` that can be used to test this PR's version

rollup.config.js
```diff
@@ -166,7 +175,9 @@ module.exports = function (args) {
   }
   return [
     ...(c === 'index' ? [createDeclarationConfig(`src/${c}.ts`, 'dist')] : []),
-    createCommonJSConfig(`src/${c}.ts`, `dist/${c}`),
+    createCommonJSConfig(`src/${c}.ts`, `dist/${c}`, {
+      addModuleExport: c === 'index',
```
dai-shi, line 179: We want this for `zustand`, `zustand/vanilla`, `zustand/shallow` and `zustand/shallow`.

barelyhuman: for some reason the `zustand/{vanilla,shallow}` already exports itself as `module.exports` in the build so the patch wasn't needed, also why I decided to go with a selective way to add this line. 

I do realise I have to add this to middleware but I wanted to check if it's working before I add that

dai-shi: > for some reason the `zustand/{vanilla,shallow}` already exports itself as `module.exports` in the build

https://unpkg.com/browse/zustand@4.1.4/vanilla.js
Oh, I didn't know that...
Maybe because there are no named exports.

> I do realise I have to add this to middleware

We don't have default exports in `zustand/middleware`, so it's not necessary.

barelyhuman:
UnderstoodOn 04-Dec-2022, at 5:18 PM, Daishi Kato ***@***.***> wrote:﻿
@dai-shi commented on this pull request.



In rollup.config.js:
> @@ -166,7 +175,9 @@ module.exports = function (args) {
   }
   return [
     ...(c === 'index' ? [createDeclarationConfig(`src/${c}.ts`, 'dist')] : []),
-    createCommonJSConfig(`src/${c}.ts`, `dist/${c}`),
+    createCommonJSConfig(`src/${c}.ts`, `dist/${c}`, {
+      addModuleExport: c === 'index',


for some reason the zustand/{vanilla,shallow} already exports itself as module.exports in the build

***@***.***/vanilla.js
Oh, I didn't know that...
Maybe because there are no named exports.

I do realise I have to add this to middleware

We don't have default exports in zustand/middleware, so it's not necessary.

PR #1463, opened 2022-12-05
feat(middleware/persist): new storage option

### LINKED ISSUE #800: Multiple persist slices under one store
Is there anyway to have multiple slices use different `persist` storage types? In my use case, I'm trying to have one subset use `localStorage` and another one `sessionStorage`.

Ideal usage:

```
const useStore = create((set, get) => {
  firstSlice: ...persist(firstSlice, { getStorage: () => localStorage }),
  secondSlice: ...persist(secondSlice, { getStorage: () => sessionStorage }),
})
```

What I've tried so far:
- Did something like example above
  - Resulted in reading/writing to both storages. Seems this is due to `hydrate` being bound to the `create` function
- Created two middlewares that run `useStore.persist.setOptions({ getStorage })` before every `get`/`set` with persist being at the top-level and using `partialize` on notable keys
  - Resulted in runtime errors and main state corruption

Would love to know if there's a way to do this. My interim approach will be to just have completely different stores for these two storage types.

### LINKED ISSUE #1210: (middlware/persist): Limitations are not documented
Based on #618 and #652 there are limitations to what can be stored. Aside from the actual reasons for this limitation, users should be made aware of this. If this can be done using the typesystem for TypeScript users, that'd be great (I'm used to Java). If not, please add a big fat warning so that users don't run into issues. Without proper documentation, there's no way to know if the code you're going to write is going to work. Trial-and-error isn't a viably alternative for me, YMMV.

From #618:

> By default, persist serializes the state using JSON.stringify.
Unfortunately JSON doesn't fit very well with Sets and Maps.

To me that sounds like a bug, not a documentation issue. If JSON doesn't do what you need it to do, don't use JSON (or fix it).

dai-shi:
## Related Issues

Closes #800 
Closes #1210 

## Summary

When we designed the persist middleware, we have (de)serialize options to customize.
Later, we realized it's tricky implementation-wise, and less flexible capability-wise.
We use `createJSONStorage` in [jotai](https://github.com/pmndrs/jotai), and this PR is a backport of it.

## Check List


src/middleware/persist.ts
```diff
@@ -11,10 +11,50 @@ type StorageValue<S> = {
   version?: number
 }
 
+export interface PersistStorage<S> {
+  getItem: (
+    name: string
+  ) => StorageValue<S> | null | Promise<StorageValue<S> | null>
```
TiagoCavalcante, line 17: There are 2 nulls here

TiagoCavalcante (approved):
This is a great improvement

src/middleware/persist.ts
dai-shi: No, there aren't. 😄

TiagoCavalcante: My mistake 😅

mooalot:
@dai-shi, I made [this](https://github.com/mooalot/zustand-multi-persist) in hopes to make it a bit easier to write to multiple storages from one store. It's built on top of zustand/persist.

dai-shi:
Thanks. Please send a PR to add it in the list: https://github.com/pmndrs/zustand/blob/main/docs/integrations/third-party-libraries.md

PR #1466, opened 2022-12-05
fix(types): Add typing for subscribe function with selector.

connorgmeehan:
## Related Issues

Fixes bad typing for the subscribe function of the store.  Now correctly types subscriptions with selectors.

## Summary

Created an interface for the new subscribe function type and updated the `StoreApi` type to use it.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -5,10 +5,18 @@ type SetStateInternal<T> = {
   ): void
 }['_']
 
+type SubscribeInternal<T> = {
+  (listener: (state: T, prevState: T) => void): () => void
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks for the suggestion.
However, vanilla.ts has nothing to do with selectors, and the suggestion is incorrect.
Please open [a new discussion](https://github.com/pmndrs/zustand/discussions/new) for questions about typing.

connorgmeehan:
Ahhh my bad, only now did I see that it requires the `subscribeWithSelector` middleware.  Sorry about that.

PR #1474, opened 2022-12-09
docs: fix testing Typescript code example

konclave:
## Summary

Fixed testing with Typescript code example

## Check List

### DIFF
--- a/docs/guides/testing.mdx
+++ b/docs/guides/testing.mdx
@@ -52,10 +52,10 @@ const storeResetFns = new Set<() => void>()
 
 // when creating a store, we get its initial state, create a reset function and add it in the set
 const create = <S,>(createState: StateCreator<S>) => {
-    const store = actualCreate<S>(createState)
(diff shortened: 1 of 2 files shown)


dai-shi (commented):
Hmm, maybe it is intentional.
Can you check #1303?

dai-shi:
I think this is about TypeScript usage, and the TypeScript usage in docs is `create()(...)`.

dai-shi (changes requested):
This is intentional for TypeScript usage `create()(...)`. The current docs is correct in that sense.
If you want to support `create(...)` usage, can you try this?

```ts
const create = <S,>(createState?: StateCreator<S>) => {
  if (!createState) return create
  const store = actualCreate<S>(createState)
  storeResetFns.add(() => store.setState(initialState, true))
  return store
}
```

konclave:
My bad. Seems like I just messed up the option of how it could be used and imported not a default import. That was the reason of my typing issues.

PR #1486, opened 2022-12-15
fix(build): disable esModule and re-export for nodejs

### LINKED ISSUE #1475: Parcel production builds can't bundle the latest version
## Summary
The latest version of Zustand is not compatible with [Parcel's](https://github.com/parcel-bundler/parcel) builds. The following error is thrown when the bundled code is run:

```
index.js:41 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.assign (<anonymous>)
    at index.js:41:42
````

## Link to reproduction
https://github.com/nicobohne/zustand-parcel-reproduction

## Check List

- [X] I understand this is not a question.

Please fill this template if you're filling an issue regarding TypeScript.

- [ ] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction. Or some other way of reproduction if these two don't suffice.
- [ ] I've read the [typescript guide](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md), in particular that `create` is to be used as `create<T>()(...)` and not `create<T>(...)`.

### PR DESCRIPTION
**Pull request by barelyhuman**
## Related Issues

Fixes #1475

## Summary

Patches the issue where exports.default might not get translated because of the `_esModule` property

## Check List

## Additional Notes 

Tested with 
- Parcel v2 
- Webpack (Current CRA) 
- All CSB Additions from the CI

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -81,8 +81,9 @@ function createCommonJSConfig(input, output, options) {
     output: {
       file: `${output}.js`,
       format: 'cjs',
+      esModule: false,
       outro: options.addModuleExport
-        ? 'module.exports = exports.default; Object.assign(exports.default, exports);'
+        ? ';(module.exports = (exports && exports.default) || {}),\n  Object.assign(module.exports, exports)'
         : '',
     },
     external,

### CODE REVIEW

File: rollup.config.js [hunk 1]
```diff
@@ -81,8 +81,9 @@ function createCommonJSConfig(input, output, options) {
     output: {
       file: `${output}.js`,
       format: 'cjs',
+      esModule: false,
       outro: options.addModuleExport
-        ? 'module.exports = exports.default; Object.assign(exports.default, exports);'
+        ? ';(module.exports = (exports && exports.default) || {}),\n  Object.assign(module.exports, exports)'
```
dai-shi, line 86: why `,\n` instead of `;`? Mostly out of curiosity.

barelyhuman: I wrote it in a minified bundle locally and copied it from there so that’s the minified version

dai-shi: I'm not convinced that `,\n` is more minified than `;`, but don't care much..

dai-shi (approved):
Thanks for the fix!
Will merge when other PRs for v4.2.0 are ready.

rollup.config.js
barelyhuman: Yeah, I didn't do it 😂, it was from the minifier of the I was trying all this with and most of them do that 
Screenshot 2022-12-17 at 5 52 29 PM

dai-shi: that's actually minified. but adding `;(...)` increased 3 bytes.

What I would do is something like:
```suggestion
        ? ';module.exports=exports&&exports.default||{};Object.assign(module.exports,exports)'
```

Hmm, we can actually combine it?
```suggestion
        ? ';Object.assign((module.exports=exports&&exports.default||{}),exports)'
```

(I'm just enjoying the puzzle. Whatever works is okay to merge.)

barelyhuman: That'll need another addition of `,` cause the return of the above would be either exports or an empty object and not actually assign stuff to `exports` so 
`Object.assign(((module.exports = (exports && exports.default) || {}), module.exports),exports);`

Would be the working one and compared to that the original commit code is smaller 😂

PR #1492, opened 2022-12-20
fix(docs): fix typo and redundant periods

scoobytux:
## Related Issues

Fixes #.

## Summary

- Correct boilerplatey -> boilerplate
- Remove redundant dots at the end line

## Check List


dai-shi:
I think both are intentional. Let's leave them unless they are super confusing.

PR #1495, opened 2022-12-23
Update how to reset state doc

### LINKED ISSUE #1494: Typescript Error when using Immer with custom "Reset Multiple Stores" function
## Summary

I'm using the `create` function defined in [this guide for resetting all stores](https://github.com/pmndrs/zustand/blob/main/docs/guides/how-to-reset-state.md). When I wrap a store with immer, I get a typescript error.

## Link to reproduction

The following sandbox is forked from the advanced example in the 'reset-state' guide.

https://codesandbox.io/s/zustand-with-immer-reset-all-stores-tme1b4

The typescript error is:

```
Argument of type 'StateCreator<StateSalmon & ActionsSalmon, [], [["zustand/immer", never]], StateSalmon & ActionsSalmon>' is not assignable to parameter of type 'StateCreator<StateSalmon & ActionsSalmon, [], [], StateSalmon & ActionsSalmon> | StoreApi<StateSalmon & ActionsSalmon>'.
  Type 'StateCreator<StateSalmon & ActionsSalmon, [], [["zustand/immer", never]], StateSalmon & ActionsSalmon>' is not assignable to type 'StateCreator<StateSalmon & ActionsSalmon, [], [], StateSalmon & ActionsSalmon>'.
    Type 'StateCreator<StateSalmon & ActionsSalmon, [], [["zustand/immer", never]], StateSalmon & ActionsSalmon>' is not assignable to type '{ $$storeMutators?: [] | undefined; }'.
      Types of property '$$storeMutators' are incompatible.
        Type '[["zustand/immer", never]] | undefined' is not assignable to type '[] | undefined'.
          Type '[["zustand/immer", never]]' is not assignable to type '[]'.
            Source has 1 element(s) but target allows only 0.ts(2345)
```

## Check List

- [x] I understand this is not a question.

Please fill this template if you're filling an issue regarding TypeScript.

- [x] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction. Or some other way of reproduction if these two don't suffice.
- [x] I've read the [typescript guide](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md), in particular that `create` is to be used as `create<T>()(...)` and not `create<T>(...)`.

dbritto-dev:
## Related Issues

Fixes #1494

## Summary

- Update how to reset state doc
- Update codesandbox advance demo
- Add codesandbox immer demo

## Check List


dbritto-dev:
@dai-shi @KevinMusgrave here's the PR :)

dbritto-dev:
@chrisk-7777 would you mind reviewing this PR?

dai-shi:
Let's have @devanshj eyes on this, in case there are better solutions.

Cuppachino:
Should the curry workaround have a mention? It allows multiple middleware.
```tsx
// Use a trailing comma after `T` in tsx files.
const create = <T,>() => <
  Mos extends [StoreMutatorIdentifier, unknown][] = []
>(
  initializer: StateCreator<T, [], Mos>
) => {
  const slice = _create(initializer);
  const initialSlice = slice.getState();
  resetters.push(() => {
    slice.setState(initialSlice, true);
  });

  return slice;
};
```
```tsx
const useSalmon = create<StateSalmon & ActionsSalmon>()(
  immer(
    devtools((set, get) => ({
      // ...
    }))
  )
);
```

dbritto-dev:
> Should the curry workaround have a mention? It allows multiple middleware.
> ```tsx
> // Use a trailing comma after `T` in tsx files.
> const create = <T,>() => <
>   Mos extends [StoreMutatorIdentifier, unknown][] = []
> >(
>   initializer: StateCreator<T, [], Mos>
> ) => {
>   const slice = _create(initializer);
>   const initialSlice = slice.getState();
>   resetters.push(() => {
>     slice.setState(initialSlice, true);
>   });
> 
>   return slice;
> };
> ```
> ```tsx
> const useSalmon = create<StateSalmon & ActionsSalmon>()(
>   immer(
>     devtools((set, get) => ({
>       // ...
>     }))
>   )
> );
> ```

Do you know if there's a way to avoid that formatting tools, like prettier, remove that trailing comma?

Cuppachino:
@dbritto-dev here's a couple ideas:

### Option 1. 
Use `.ts` extension and encourage users to separate stores from components.
### Option 2.
Tell prettier to skip formatting `create`
```diff
+ /* prettier-ignore */
const create = <T,>() => <
```

### Option 3.
Type `T` as `unknown` (I feel like this hasn't worked for me in the past, but it seems to now).
```diff
- const create = <T,>() => <
+ const create = <T as unknown> => <
```

### Option 4.
Tell eslint we *know* `_` does nothing.
```diff
+ /* eslint-disable @typescript-eslint/no-unused-vars */
- const create = <T,>() => <
+ const create = <T,_>() => <
```

### Option 5. 
Use a function expression (maybe not so consistent with other doc pages). This would mimic the declaration file for `Create`, which looks like an overloaded function given a variable name with const (without overloads, thoughts below).
```diff
- const create = <T,>() => <
-   Mos extends [StoreMutatorIdentifier, unknown][] = []
- >(
+  const create = function <T>() {
+   return <Mos extends [StoreMutatorIdentifier, unknown][] = []>(
    initializer: StateCreator<T, [], Mos>
  ) => {
    const slice = _create(initializer);
    const initialSlice = slice.getState();
    resetters.push(() => {
      slice.setState(initialSlice, true);
    });

    return slice;
+   };
};
```
Imo we shouldn't try to replicate the default export `_create` behavior because its really a TS issue. If you need type inference and you're using typescript, use the curried version until TS gets smarter, otherwise, the issue isn't even an issue, right? The curried form should always be a function that returns a function, and it should be a separate example, if included at all/better solution isn't found.

### TLDR: use option D, don't overload.

dai-shi:
To make prettier happy without changing configs, I prefer option 3. It means `T extends unknown`, right?

Cuppachino:
> To make prettier happy without changing configs, I prefer option 3. It means `T extends unknown`, right?

Can you double check? There was a typo where I said option "D"... and all the options are numbered. I also added option 6, but the type of `_Mos` isn't actually enforced.

dai-shi:
I mean this.

```diff
- const create = <T,>() => <
+ const create = <T extends unknown> => <
```

dai-shi:
> Do you know if there's a way to avoid that formatting tools, like prettier, remove that trailing comma?

Does changing "ts" to "tsx" after three backquotes help?

ref: https://github.com/prettier/prettier/pull/6115

Cuppachino:
> > Do you know if there's a way to avoid that formatting tools, like prettier, remove that trailing comma?
> 
> Does changing "ts" to "tsx" after three backquotes help?
> 
> ref: [prettier/prettier#6115](https://github.com/prettier/prettier/pull/6115)

Yes, but anyone who implements it in a tsx file will have the problem if they're also using prettier. The codesandboxes linked in #1494 autoformat on save, breaking the example.

dbritto-dev:
Changing the `ts` to `tsx` actually does nothing because tsx is just an alias of ts on blockquotes

devanshj (changes requested):
The type of `create` here has the same type as the original `create`, so there's no need of retyping it (the retyping is also incorrect as it misses the `Mis` type parameter), so one should just use `typeof originalCreate`. And it should also implement the curried usage.

So the final code would look something like this... (feel free to change formating, variable names, etc)

```ts
import _create, { StateCreator } from 'zustand'

const resetters: (() => void)[] =
  []

export const create = (<T extends unknown>(f: StateCreator<T> | undefined) => {
  if (f === undefined) return create
  const store = _create(f)
  resetters.push(() => {
    store.setState(initialState, true)
  })
  return store
}) as typeof _create

export const resetAllSlices = () => {
  for (const resetter of resetters) {
    resetter()
  }
}
```

Also the code snippet above this one should use the curried version ie `create<State & Actions>()(...)` instead of `create<State & Actions>(...)`. The latter is not anywhere in the TypeScript docs and is not recommended.

dbritto-dev:
> The type of `create` here has the same type as the original `create`, so there's no need of retyping it (the retyping is also incorrect as it misses the `Mis` type parameter), so one should just use `typeof originalCreate`. And it should also implement the curried usage.
> 
> So the final code would look something like this... (feel free to change formating, variable names, etc)
> 
> ```ts
> import _create, { StateCreator } from 'zustand'
> 
> const resetters: (() => void)[] =
>   []
> 
> export const create = (<T extends unknown>(f: StateCreator<T> | undefined) => {
>   if (f === undefined) return create
>   const store = _create(f)
>   const initialState = store.getState()
>   resetters.push(() => {
>     store.setState(initialState, true)
>   })
>   return store
> }) as typeof _create
> 
> export const resetAllSlices = () => {
>   for (const resetter of resetters) {
>     resetter()
>   }
> }
> ```
> 
> Also the code snippet above this one should use the curried version ie `create<State & Actions>()(...)` instead of `create<State & Actions>(...)`. The latter is not anywhere in the TypeScript docs (except for "create without curried workaround" section) and is not recommended.

Thanks I'll apply the changes

dbritto-dev:
@devanshj the changes are done :D

docs/guides/how-to-reset-state.md
```diff
@@ -78,3 +72,4 @@ export const resetAllSlices = () => {
 
 - Basic: https://codesandbox.io/s/zustand-how-to-reset-state-basic-demo-rrqyon
 - Advanced: https://codesandbox.io/s/zustand-how-to-reset-state-advanced-demo-gtu0qe
+- Immer: https://codesandbox.io/s/how-to-reset-state-advance-immer-demo-nyet3f
```
devanshj, line 75: Here `useTuna` should be `create<StateTuna & ActionsTuna>()(immer(...` and not `create(immer<StateTuna & ActionsTuna>(...`

devanshj (changes requested):
Left an addtional suggestion as a comment.

And let's also change `create<State & Actions>(...` into `create<State & Action>()(...` in the first snippet of the document (can't comment on that line as it's not part of the change) as I suggested earlier.

Other than these changes it looks good.

docs/guides/how-to-reset-state.md
dbritto-dev: Sure

dbritto-dev:
> Left an addtional suggestion as a comment.
> 
> And let's also change `create<State & Actions>(...` into `create<State & Action>()(...` in the first snippet of the document as I suggested earlier (can't comment on that line as it's not part of the change).
> 
> Other than these changes it looks good.

docs/guides/how-to-reset-state.md
dbritto-dev: @devanshj now we are good

devanshj: Yep looks fine. Now this one change remains...

> And let's also change `create<State & Actions>(...` into `create<State & Action>()(...` in the first snippet of the document as I suggested earlier (can't comment on that line as it's not part of the change).

devanshj:
This one is still missing:

> And let's also change `create<State & Actions>(...` into `create<State & Action>()(...` in the first snippet of the document as I suggested earlier (can't comment on that line as it's not part of the change).

dbritto-dev:
> This one is still missing:
> 
> > And let's also change `create<State & Actions>(...` into `create<State & Action>()(...` in the first snippet of the document as I suggested earlier (can't comment on that line as it's not part of the change).

yeah, I updated the first codesandbox demo as well. Now we are good.

devanshj:
No I'm talking about line 30 in how-to-reset-state.md, that still needs a change, let me know if it's still not clear.

dbritto-dev:
> No I'm talking about line 30 in how-to-reset-state.md, that still needs a change, let me know if it's still not clear.

I see. I thought you were talking about the codesandbox demo haha, my bad

dbritto-dev:
@devanshj all yours

devanshj (approved):

dbritto-dev:
@dai-shi all yours :D

PR #1498, opened 2022-12-29
Update readme.md

andresin87:
## Related Issues

Fixes #.

## Summary

The readme for using zustand store with vanillaJS needs a fix. To create the store you cannot use the create method cuz it will directly return the hook wich needs to be fired in react fn component ONLY. The doc should inform about the createStore method

## Check List


dai-shi:
> The readme for using zustand store with vanillaJS needs a fix. To create the store you cannot use the create method cuz it will directly return the hook wich needs to be fired in react fn component ONLY. The doc should inform about the createStore method

You can use the hook version to read state outside react. The doc is correct. Maybe there are other ways to improve it though.

PR #1512, opened 2023-01-06
doc: update documentation of persist

wcastand:
## Related Issues

Fixes #1510 

## Summary

Update doc to include how to handle Map and Set with the new persist API.

issue since i have yarn v3 installed, it creates `.yarn` and `.yarnrc.yml` and need me to set pacakge to v1 with `yarn set version classic` so not sure i can run prettier easily but i will run it and delete the files we don't want commited.

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -574,7 +574,7 @@ export const useBearStore = create<MyState>()(
 
 ### How do I use deal with Map/Set with the persist api
 
-With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into Array so they could be parsed into proper JSON.
+With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into an Array so they could be parsed into proper JSON.
 The new persist API has deprecated `serialize`/`deserialize`.
 
 Now, you will need to use the `storage` prop, let's say your state use a `Map` to handle a list of `transactions`, then you can convert the Map into an Array in the storage prop:
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -574,7 +574,10 @@ export const useBearStore = create<MyState>()(
 
 ### How do I use it with Map and Set?
 
-With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into an Array so they could be parsed into proper JSON.
+With the previous persist API, you would use `serialize`/`deserialize`
+to deal with `Map` and `Set` and convert them into
+an Array so they could be parsed into proper JSON.
+
 The new persist API has deprecated `serialize`/`deserialize`.
 
 Now, you will need to use the `storage` prop.
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -577,9 +577,11 @@ export const useBearStore = create<MyState>()(
 With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into an Array so they could be parsed into proper JSON.
 The new persist API has deprecated `serialize`/`deserialize`.
 
-Now, you will need to use the `storage` prop. Let's say your state uses `Map` to handle a list of `transactions`, then you can convert the Map into an Array in the storage prop:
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Thanks for the PR. LGTM

docs/integrations/persisting-store-data.md
```diff
@@ -571,3 +571,34 @@ export const useBearStore = create<MyState>()(
   )
 )
 ```
+
+### How do I use deal with Map/Set with the persist api
+
+With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into Array so they could be parsed into proper JSON.
```
chrisk-7777, line 577:
```suggestion
With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into an array so they could be parsed into proper JSON.
```

chrisk-7777, line 577:
```suggestion
With the previous persist API, you would use `serialize`/`deserialize` to deal with Map and Set and convert them into an Array so they could be parsed into proper JSON.
```

docs/integrations/persisting-store-data.md
```diff
@@ -571,3 +571,34 @@ export const useBearStore = create<MyState>()(
...
+The new persist API has deprecated `serialize`/`deserialize`.
+
+Now, you will need to use the `storage` prop, let's say your state use a `Map` to handle a list of `transactions`, then you can convert the Map into an Array in the storage prop:
```
chrisk-7777, line 580:
```suggestion
Now, you will need to use the `storage` prop. Let's say your state uses `Map` to handle a list of `transactions`, then you can convert the Map into an Array in the storage prop:
```

chrisk-7777 (changes requested):
Looks good, I just have two really nit-picky change requests

docs/integrations/persisting-store-data.md
```diff
@@ -571,3 +571,34 @@ export const useBearStore = create<MyState>()(
...
+
+```typecript
```
sewera, line 582: There is a typo in `typecript`. Please fix it either to `typescript`, or better yet, simply use `ts`.

docs/integrations/persisting-store-data.md
```diff
@@ -571,3 +571,34 @@ export const useBearStore = create<MyState>()(
   )
 )
 ```
+
+### How do I use deal with Map/Set with the persist api
```
sewera, line 575: I'm not completely sure how to understand it, but I think it will be more readable with a shorter question:

```suggestion
### How do I use it with Map and Set?
```

We implicitly use the `persist` API on this doc page, so I omitted it.

docs/integrations/persisting-store-data.md
sewera, line 577: If I understood it correctly, this is how I'd rephrase it.

```suggestion
With the previous persist API, you would use `serialize`/`deserialize`
to deal with mapping and setting the data and converting it
into an Array, so it could be parsed into a proper JSON.
```

docs/integrations/persisting-store-data.md
sewera, line 580:
```suggestion
Now, you will need to use the `storage` prop.
For example, let's say your state uses `Map` to handle
a list of transactions, then you can convert that Map
into an Array in the `storage` prop:
```

sewera (changes requested):

docs/integrations/persisting-store-data.md
wcastand: Not really, i'm talking about the JS Object `Map` and `Set` not mapping and setting.

sewera: Alright, you can wrap it in backticks, so `Map` and `Set`, so that it's clear that they're JS objects.

sewera: And you can split this long line to shorter ones, to match the rest of this file.

sewera:
The PR looks good to me, nice work :)

@chrisk-7777, I think you can re-review and approve those changes now.

dai-shi:
I can merge an approval from either of you.

chrisk-7777 (approved):
thank you for actioning those changes - look good now 👍

PR #1514, opened 2023-01-07
feat: deprecate default export

dai-shi:
close #1506


judicaelandria (approved):

meypod:
hmm, shouldn't `zustand/vanilla` need this as well ?

dai-shi:
Do you mean #1531? Yeah, good point.

meypod:
I didn't mean that, It seems I was mistaken (`createStore` was actually exported and I didn't see it), but I'm actually glad if my comment could help with sth😅

dai-shi:
😄

PR #1521, opened 2023-01-09
Update persist documentation for `storage`

charkour:
## Summary

This replaces `getStorage` with `storage`.

This PR raises two questions:
- Should we write documentation on how to migrate from the `getStorage` API to the `storage` API?
- Should we version the middleware separately from the zustand core?

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -300,7 +300,7 @@ Or even to change the storage engine:
 
 ```ts
 useBoundStore.persist.setOptions({
-  getStorage: () => sessionStorage,
+  storage: createJSONStorage(() => sessionStorage),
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks for the fix! LGTM

dai-shi:
> * Should we write documentation on how to migrate from the `getStorage` API to the `storage` API?

That would be a good idea. Can you try? It might be tough to cover broader use cases.

> * Should we version the middleware separately from the zustand core?

We try to avoid breaking changes in JS in v4. Maybe the deprecation message seems very big to users.

charkour:
> That would be a good idea. Can you try? It might be tough to cover broader use cases.

I'll try my best.

PR #1524, opened 2023-01-10
docs: add migration example in v3 create context

dai-shi:
https://github.com/pmndrs/zustand/pull/1403#discussion_r1065627397

### DIFF
--- a/docs/previous-versions/zustand-v3-create-context.md
+++ b/docs/previous-versions/zustand-v3-create-context.md
@@ -105,7 +105,7 @@ export default function App({ initialBears }) {
 
 ## Migration
 
-https://github.com/pmndrs/zustand/discussions/1276
+Discussion: https://github.com/pmndrs/zustand/discussions/1276
 
 Here's the diff showing how to migrate from v3 createContext to v4 API.
 
--- a/docs/previous-versions/zustand-v3-create-context.md
+++ b/docs/previous-versions/zustand-v3-create-context.md
@@ -6,7 +6,7 @@ nav: 18
 A special `createContext` is provided since v3.5,
 which avoids misusing the store hook.
 
-> **Note**: This function will be deprecated in v4 and removed in v5.
+> **Note**: This function is deprecated in v4 and will be removed in v5. See [#migration](Migration).
 
 ```jsx
 import create from 'zustand'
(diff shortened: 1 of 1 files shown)


chrisk-7777 (approved):
👍

TkDodo:
@dai-shi what would be your recommendation to initialize a zustand store from props? We've used `createContext` for this up until now. Could we just wrap `createStore` with `useMemo` inside the App component?

```
const StoreContext = createContext()

const App = ({ initialState }) => {
  const store = useMemo(() => createStore(initialState), [initialState])

  return (
    <StoreContext.Provider value={store}>
      ...
    </StoreContext.Provider>
  )
}
```
?

dai-shi:
Yeah, that's the way to go. Probably, useRef would be better.

charkour:
@TkDodo, here's the [zustand guide](https://github.com/pmndrs/zustand/blob/main/docs/guides/initialize-state-with-props.md) for initializing a zustand store with props using useRef as Daishi mentioned. In preparation for the removal of `createContext` in zustand v5, my team released our [own package](https://github.com/charkour/zustand-di) for it. Hopefully, that is helpful for your team.

@dai-shi, should the React/context migration example include a link to the [guide "Initialize State With Props"](https://github.com/pmndrs/zustand/blob/main/docs/guides/initialize-state-with-props.md)? It has a nice walkthrough of how to use React Context to init stores. When I was trying to migrate our codebase, my team and I had trouble because the zustand + context documentation was separated in many different spots. What do you think?

TkDodo:
Thanks for the link. Just a small preference: I would choose `useState` over `useRef` in the docs:

```diff
-  const storeRef = useRef<BearStore>()
-  if (!storeRef.current) {
-    storeRef.current = createBearStore(props)
-  }
+  const [store] = useState() => createBearStore(props))
```

This has only upsides :)

charkour:
Would this also be equivalent in behavior? This is stretching my understanding of React's state behavior.

 ```diff
 -  const storeRef = useRef<BearStore>()
 -  if (!storeRef.current) {
-    storeRef.current = createBearStore(props)
 -  }
 +  const store = useRef(createBearStore(props)).current
 ```

TkDodo:
no it's not strictly equivalent, because `useRef` has no "lazy initializer". So in this case, `createBearStore(props)` is executed every time, then it is passed to `useRef`, where only the result of the first render is taken into account. It's minor, but it's still different.

with `useState`, you can have the initializer function, which is only evaluated once.

charkour:
Right, I always forget about lazy init. Thank you!

dai-shi:
@TkDodo https://github.com/pmndrs/zustand/pull/1524#issuecomment-1377815907

> This has only upsides :)

I don't prefer using useState for lazy init.
It should be less efficient technically. Not a right tool, as it's not "state".
My recommendation is [use-constant](https://www.npmjs.com/package/use-constant).

axelboc:
Is there an easy way to type the line below? I'd like `useMyStore` to have the same type as `useStore` (with the same generics and overrides). I've tried a few things, including `const useMyStore: typeof useStore = ...` but it doesn't seem to work:

```ts
export const useMyStore = (selector) => useStore(useContext(MyContext), selector);
```

dai-shi:
You don't need to replicate the type, instead:

```ts
const useMyStore = <T>(selector: (state: MyState) => T) => useStore(useContext(MyContext), selector);
```

Issue #1430, opened 2022-11-15
Suggestion: drop uncurried support of `create`


devanshj:
I've lost count of how many people have missed the extra parenthesis in `create()(...)` (and the wrong usage ie `create(...)` doesn't even get caught because it's allowed). I don't see how supporting the `create(...)` usage adds any value, it's just creating trouble for many people (including me). I know I've suggested this before but I want to make the suggestion official hence opening an issue.

dai-shi:
Thanks for opening a suggestion. Let's wait others to comment. One note is as such a change is breaking, it will be only possible in v5.

That said, my personal comment would be a) `create(...)` is still very valid and natural syntax for JS users, and b) `create()(...)` simply doesn't look comfortable, so I prefer to use `create(...)` without middleware, and I'm not a big fan of middleware. They are mostly preference, and my mind could be changed on this.

(I understand there are many troubles, and wonder if there's any other way to mitigate it, like warning for the usage of `create(...)` only in TS. Otherwise, `eslint-plugin-zustand` would be one solution. Or, hm, we could attach `api.middlwareIsUsed` property and warn it.)

devanshj:
There's also an option to support it in runtime but not include it in types (ie .d.ts won't include the uncurried overload)

Gustav-Eikaas:
Maybe i'm misunderstanding but is this the two types of syntaxes we're discussing?

```TS
type Test = {
  name: string;
};

const curried = create<Test>()((set) => ({ name: "Yes" }));

const uncurried = create<Test>((set) => ({
  name: "string",
}));
```

Personally I prefer the uncurried one and find the curried one weird..
Callback initializers are pretty common in the world of JS

dai-shi:
> Maybe i'm misunderstanding but is this the two types of syntaxes we're discussing?

Yeah, in summary:
- In TypeScript, the curried one is _required_ if you use middleware (not all of them, but with store mutation, but hard to know which).
- The curried one works for all cases, both in JavaScript and TypeScript
- In JavaScript, the uncurried one works for all cases.

ollwenjones:
We use TS, and have used the uncurried version almost exclusively, and have not had any problems, I guess because we aren't using middleware that requires it. We tend to use type inference on factory function return types for different slices, with good results.

charkour:
The idea of `eslint-plugin-zustand` would be great. I've been wanting to make one so we can add optional rules like `always-use-selector` when calling `useStore` in a React component.

[...]

PR #1529, opened 2023-01-10
Extra brackets in typescript example

Faithfinder:
This looks more correct

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -490,7 +490,7 @@ interface BearState {
   increase: (by: number) => void
 }
 
-const useBearStore = create<BearState>()(
+const useBearStore = create<BearState>(
   devtools(
     persist(
       (set) => ({
--- a/readme.md
+++ b/readme.md
@@ -481,6 +481,8 @@ const Component = () => {
 
 Basic typescript usage doesn't require anything special except for writing `create<State>()(...)` instead of `create(...)`...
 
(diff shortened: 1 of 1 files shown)


charkour:
Hey @Faithfinder,

Thanks for the PR, but the original is actually correct. The `create` function is curried to provide accurate types for the store when using typescript. Here is more [documentation](https://github.com/pmndrs/zustand#typescript-usage) to explain the reason in detail.

Faithfinder:
Ah! Thank you. How about this then? If not, feel free to close :)

dai-shi:
459320c
I don't think it's necessary, as it's too repetitive. But, thanks for the feedback. Better way to educate the syntax would be important. 🤔

Faithfinder:
I missed the extra brackets in the text blurb, because I thought you're trying to draw attention to the generic, and that's why I thought the extra brackets in the example are, well, extra >.>

dai-shi:
Yeah, maybe it should more redirect to "TypeScript guide".

What do you guys think? @chrisk-7777 @sewera @dbritto-dev @charkour

dbritto-dev:
I do not know what you think guys but having two syntax for the same could be confusing for some people. I would go for using the curried version for both, javascript and typescript. Then people does not need to rewrite the code to support typescript.

dai-shi:
We are debating in #1430.

PR #1541, opened 2023-01-13
docs: update shallow to defaultExport instead of named

Just-Moh-it:
## Summary

The docs had an incorrect example about using shallow, whereby, a named export of shallow was being imported from `zustand/shallow`. I fixed that to a default export :)

## Check List


sewera:
I think this named export is on purpose, because all the default exports were deprecated in #1514 (look for `src/shallow.ts` file).

Therefore, I suggest declining this PR.

dai-shi:
Yes, default export is deprecated and will be removed. (with a bit of regret, because it's not solely our fault.)
Closing.

PR #1550, opened 2023-01-18
docs: update testing

nus3:
## Related Issues

(None)

## Summary

Update testing doc because default exports is deprecated in v.4.3.0.

Mocking zustand(using v4.3.2) according to the TypeScript guide results in the following error.

image

So I used aliases and named export.
The test is working correctly in my environment.

## Check List


dai-shi (approved):
that's my mistake. thanks for the fix!

PR #1565, opened 2023-01-23
fix(docs/typescript): add typing usage with custom store hooks

holgergp:
## Related Issues

Discussion: https://github.com/pmndrs/zustand/discussions/1564

Fixes #.

## Summary

Providing documentation for accessing and using vanilla stores and bound stores using typescript

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -399,6 +399,59 @@ A detailed explanation on the slices pattern can be found [here](./slices-patter
 
 If you have some middlewares then replace `StateCreator<MyState, [], [], MySlice>` with `StateCreator<MyState, Mutators, [], MySlice>`. For example, if you are using `devtools` then it will be `StateCreator<MyState, [["zustand/devtools", never]], [], MySlice>`. See the ["Middlewares and their mutators reference"](#middlewares-and-their-mutators-reference) section for a list of all mutators.
 
+### Using a vanilla store as a bound store
+Create your vanilla store:
+```ts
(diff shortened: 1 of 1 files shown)


dai-shi:
Should we wait for the follow-up discussion? https://github.com/pmndrs/zustand/discussions/1564#discussioncomment-4753546

chrisk-7777 (approved):
This is a really nice outcome coming out of the big discussion

holgergp:
Thank you! 😀

docs/guides/typescript.md
```diff
@@ -399,6 +399,74 @@ A detailed explanation on the slices pattern can be found [here](./slices-patter
...
+}
+```
+
+> **_NOTE:_** We prefer function overloading here, as this closely follows the definition of `useStore` itself.  
+> If you are not familiar with this pattern, just have a look here: [Typescript Docs](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads)
+
+Now you can access your vanilla store (e.g. in your tests) like:
+
+```ts
+import { vanillaBearStore, initialBearState } from './BearStore'
```
sewera, line 442: Sorry for being so late, overall a great PR and discussion,
however here you import `initialBearState`, but you use `initialState`.

It's not a critical change, but it can be updated :)

sewera:
Great PR, great discussion. I've got only one comment (above).
@holgergp, @chrisk-7777, so that you've got it in mind.

holgergp:
@sewera Ah! Thanks! I made #1577 for this.

PR #1578, opened 2023-01-26
fix(docs/typescript): provide example how to use middlewares with #1565

holgergp:
## Related Issues
#1565 

Fixes #.

## Summary
Providing an example of using middlewares with vanilla stores using typescript

## Check List

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -444,7 +444,7 @@ import { vanillaBearStore, initialBearState } from './BearStore'
 describe('MyComponent should', () => {
   // remember to reset the store
   beforeEach(() => {
-    vanillaBearStore.setState(initialState)
+    vanillaBearStore.setState(initialBearState)
   })
 
   it('set the value', () => {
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -467,6 +467,22 @@ export const BearComponent = () => {
...
+
+export const vanillaBearStore = createStore<BearState>()(
+  devtools((set, getState) => ({
+    ...initialBearState,
+    increase: (by) => set((state) => ({ bears: state.bears + by })),
+  }))
+)
+```
+
+For the extra parantheses please see [this](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md#basic-usage)
```
sewera, line 484:
```suggestion
For more information about why there are extra parentheses,
please see the [Basic usage section](#basic-usage).
```

sewera (changes requested):
I believe linking to the section, so that it's available both on Github, and in the rendered <https://docs.pmnd.rs/zustand/guides/typescript#basic-usage> is a better idea.

sewera (approved):
Thanks for commit mention :)

holgergp:
@sewera NP :)
This was the inline GH "apply suggestion" workflow. Apparently it generates a co-author. 🤷  Which is nice! 😀

PR #1600, opened 2023-02-05
fix(build): reorder the types field in package.json

HiChen404:
## Summary
According to the typescript documentation: `Entry-point` for TypeScript resolution must occur first!

>Reference:  
> https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing

## Check List

- [ √ ] `yarn run prettier` for formatting code and docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,56 +21,56 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./index.d.ts",
       "import": {
         "types": "./esm/index.d.mts",
         "default": "./esm/index.mjs"
       },
-      "types": "./index.d.ts",
       "module": "./esm/index.js",
       "default": "./index.js"
     },
     "./vanilla": {
+      "types": "./vanilla.d.ts",
       "import": {
         "types": "./esm/vanilla.d.mts",
         "default": "./esm/vanilla.mjs"
       },
-      "types": "./vanilla.d.ts",
       "module": "./esm/vanilla.js",
       "default": "./vanilla.js"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -21,56 +21,56 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./index.d.ts",
       "import": {
         "types": "./esm/index.d.mts",
```
dai-shi, line 26: Can you make sure if "import" condition picks this types correctly?

dai-shi (commented):
thanks for the suggestion.
we should make jotai and valtio to follow this, if it works fine.
@barelyhuman what do you think?

barelyhuman:
That seems okay, since most editors today use [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server) to handle the typedef and crossfile info, we could in theory get rid of the whole type definition it self, though this helps in case someone does implement something new based on the mentioned specs! 👍🏼👍🏼

Looks good, @dai-shi  want me to replicate this in the other 2 (jotai, valtio) ?

dai-shi:
https://github.com/pmndrs/valtio/pull/655 is already open.

@HiChen404 Would you like to open the one in jotai repo?

HiChen404:
> [pmndrs/valtio#655](https://github.com/pmndrs/valtio/pull/655) is already open.
> 
> @HiChen404 Would you like to open the one in jotai repo?

Okay, thanks.

PR #1617, opened 2023-02-10
chore: unused import type {} from '@redux-devtools/extension'

zhangrenyang:
## Summary

Unused import type {} from '@redux-devtools/extension'

## Check List


dai-shi:
It's not unused...

(but, we may have issues with this approach. see: https://github.com/pmndrs/jotai/discussions/1721 )

PR #1632, opened 2023-02-14
fix(build): UMD names

dai-shi:
following https://github.com/pmndrs/valtio/pull/667

it fixes immer middleware. everything else is unchanged.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -112,7 +112,7 @@ function createUMDConfig(input, output, env) {
   const fileName = output.slice('dist/umd/'.length)
   const capitalize = (s) => s.slice(0, 1).toUpperCase() + s.slice(1)
   if (fileName !== 'index') {
-    name += fileName.replaceAll(/(\w+)\W*/g, (_, p) => capitalize(p))
+    name += fileName.replace(/(\w+)\W*/g, (_, p) => capitalize(p))
   }
   return {
     input,
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -127,6 +127,7 @@ function createUMDConfig(input, output, env) {
         'use-sync-external-store/shim/with-selector':
           'useSyncExternalStoreShimWithSelector',
         'zustand/vanilla': 'zustandVanilla',
+        immer: 'immer',
       },
     },
     external,
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -123,11 +123,11 @@ function createUMDConfig(input, output, env) {
       name,
       globals: {
         react: 'React',
+        immer: 'immer',
         // FIXME not yet supported
         'use-sync-external-store/shim/with-selector':
           'useSyncExternalStoreShimWithSelector',
         'zustand/vanilla': 'zustandVanilla',
-        immer: 'immer',
       },
     },
     external,
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -108,16 +108,19 @@ function createCommonJSConfig(input, output, options) {
 }
 
 function createUMDConfig(input, output, env) {
-  const c = output.split('/').pop()
+  const c = output.replace(/^dist\/umd\//, '').split(/\W/)
+  const name = c.reduce((acc, itm, idx) => {
+    if (idx === c.length - 1 && itm === 'index') {
+      return acc
+    }
+    return acc + `${itm.slice(0, 1).toUpperCase()}${itm.slice(1)}`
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -108,18 +108,22 @@ function createCommonJSConfig(input, output, options) {
 }
 
 function createUMDConfig(input, output, env) {
-  const c = output.split('/').pop()
+  const c = output.replace(/^dist\/umd\//, '').split(/\W/)
+  const name = c.reduce((acc, itm, idx) => {
```
sewera, line 112: I think you can extract the function inside `reduce`,
so that it's a named function. It can be extracted to a `const` inside this scope.

Something like:
```js
const toCamelCase = (acc, item, i) => { /* ... */ }
const name = c.reduce(toCamelCase, 'zustand')
```

rollup.config.js
```diff
@@ -108,18 +108,22 @@ function createCommonJSConfig(input, output, options) {
 }
 
 function createUMDConfig(input, output, env) {
-  const c = output.split('/').pop()
+  const c = output.replace(/^dist\/umd\//, '').split(/\W/)
```
sewera, line 111: I think using a longer name than `c` will be beneficial, because right now it's hard to understand.

rollup.config.js
dai-shi: I don't think it helps much, but how would you name it? `toCamelCase` doesn't sound correct for `(acc, item, i) => ...`.

rollup.config.js
dai-shi: Yeah, I agree. How would you name instead of `c` if you have an idea?

sewera: I don't know, because I don't know the _intent_. `outputConfigPath`? I'm only guessing from the usage, but I don't know what did you mean.

You know, there are two things hard in programming: cache invalidation and naming things ;)

rollup.config.js
sewera: I don't know the intent of the function, but from what I've read below in the implementation, it makes this: `['happy', 'hippo']` to this: `'zustandHappyHippo'`, right?

```js
['happy', 'hippo'].reduce((acc, itm, idx) => {
  if (idx === 2 - 1 && itm === 'index') {
    return acc
  }
  return acc + `${itm.slice(0, 1).toUpperCase()}${itm.slice(1)}`
}, 'zustand')

// => 'zustandHappyHippo'
```

That's why I named it `toCamelCase`. You can name it `toCamelCaseWithPrefix` as well.

dai-shi: I know what you mean, but it doesn't sound like a reducer function. Maybe, `toCamelCaseReducer`.

rollup.config.js
dai-shi: Yeah, I don't know what to name it. The intent is, output file name parts divided by non words.
Maybe there's a better solution with regex, which could be harder to read.

rollup.config.js
sewera: I'll be a bit philosophical now, so excuse me for that :P 
For me, `toCamelCaseReducer` would make sense if such function
was exported as a public interface.
Here, in the local block scope, `list.reduce(toCamelCaseWithPrefix, 'zustand')`
reads like a sentence, that's why I chose it.

With that said, `toCamelCaseReducer` is a better name than no name at all,
so you can go with it :)

dai-shi: To continue a bit philosophical discussion, what I don't prefer is
```js
const toCamelCase = (acc, itm, idx) => ...
```
which doesn't make sense about what are acc, itm, and even idx. I'd expect
```js
const toCameCase = (str, maybeSomethingRelatedWithString) => ...
```

sewera: I see your point. I'm used to TDD-style design, so if I have to,
I will trade the readability of the usage, even if the implementation
or declaration is less readable. In this context, my idea is more
readable _for me and my habits, in this context alone_,
but I perfectly understand where are you going from, so I think
your idea is as good, so go for it :)

rollup.config.js
sewera: My best guess then is `outputFilenameWords`, but it won't force it

sewera (approved):
My comments are quite minor discussions, so I'll approve this PR.

sewera (approved):
Much cleaner in my opinion now, good job :)

PR #1640, opened 2023-02-16
fix(docs): resolve different section linking behavior with special ch…

sewera:
## Summary

I've found it by accident.
The issue is that there is different behavior on Github
and Next.js when linking sections with special characters like `?`
in a single Markdown file.
Incidentally, in VSCode, both links with question marks and without them work.

That's why I propose dropping the question marks in headings
to avoid broken links [like this one in Hydration and async storages section](https://docs.pmnd.rs/zustand/integrations/persisting-store-data#hydration-and-asynchronous-storages),
which points to [this section](https://docs.pmnd.rs/zustand/integrations/persisting-store-data#how-can-i-check-if-my-store-has-been-hydrated?).

Another issue is that the `?` sign is special in URIs.
I don't know if it could break anything, but my inner future self
wants to avoid potential hard-to-debug issues like that.

There are also question marks in headings in `recipes.mdx` and `readme.md`,
but they are crucial for the meaning and the tone of those headings,
and are not linked anywhere I looked.

An example:

```md
# Example

[link to why which works on Github](#why)

[link to why which works on Next.js](#why?)

## Why?

Markdown implementations, why?
```

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -405,12 +405,12 @@ is not logged in because it's the default,
 but in reality the store has not been hydrated yet.
 
 If your app does depends on the persisted state at page load,
-see [_How can I check if my store has been hydrated?_](#how-can-i-check-if-my-store-has-been-hydrated)
+see [_How can I check if my store has been hydrated_](#how-can-i-check-if-my-store-has-been-hydrated)
 in the [FAQ](#faq) section below.
 
 ## FAQ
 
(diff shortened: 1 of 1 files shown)


chrisk-7777 (approved):
Good idea removing them. 

There was an issue raised against the site here: https://github.com/pmndrs/website/issues/259 , but I haven't come up with a sensible solution. 

I need to check in on that issue..., but this PR makes sense to me 👍

PR #1641, opened 2023-02-16
Fix Jest example's usage of default imports

adimit:

Inspired by #559

## Summary
Default exports and imports are deprecated. The default import of `zustand` generates a deprecation warning on the console. The default export that was suggested previously would actually fail to compile. See the comments at the very bottom of #559 for further discussion.

This commit fixes the example in the documentation by introducing named exports & imports instead of default exports & imports.

## Check List

### DIFF
--- a/docs/guides/testing.mdx
+++ b/docs/guides/testing.mdx
@@ -50,11 +50,11 @@ const storeResetFns = new Set<() => void>()
 
 // when creating a store, we get its initial state, create a reset function and add it in the set
 export const create = <S>(createState: StateCreator<S>) => {
-    const store = actualCreate(createState)
-    const initialState = store.getState()
-    storeResetFns.add(() => store.setState(initialState, true))
-    return store
-  }
+  const store = actualCreate(createState)
+  const initialState = store.getState()
+  storeResetFns.add(() => store.setState(initialState, true))
+  return store
+}
 
 // Reset all stores after each test run
 beforeEach(() => {
--- a/docs/guides/testing.mdx
+++ b/docs/guides/testing.mdx
@@ -12,7 +12,7 @@ Thus, there can be cases where the state of one test can affect another. To make
 
 ```jsx
 import { create as actualCreate } from 'zustand'
-// import { create as actualCreate } from 'zustand' // if using jest
+// const { create: actualCreate } = jest.requireActual('zustand') // if using jest
 import { act } from 'react-dom/test-utils'
 
 // a variable to hold reset functions for all stores declared in the app
@@ -42,7 +42,9 @@ If you are using zustand, as documented in [TypeScript Guide](./typescript.md), 
 
 ```typescript
 import { create as actualCreate, StateCreator } from 'zustand'
-// import { create as actualCreate } from 'zustand' // if using jest
+// if using Jest:
+// import { StateCreator } from 'zustand';
+// const { create: actualCreate } = jest.requireActual<typeof import('zustand')>('zustand');
 import { act } from 'react-dom/test-utils'
 
 // a variable to hold reset functions for all stores declared in the app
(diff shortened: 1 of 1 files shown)


docs/guides/testing.mdx
```diff
@@ -12,7 +12,7 @@ Thus, there can be cases where the state of one test can affect another. To make
 
 ```jsx
 import { create as actualCreate } from 'zustand'
-// const actualCreate = jest.requireActual('zustand') // if using jest
+// const { create: actualCreate } = jest.requireActual('zustand') // if using jest
```
sewera, line 15: Why `jest.requireActual`? Won't the line above
(`import { create as actualCreate } from 'zustand'`)
be sufficient even with `jest`?

From what I understood from [the Jest docs](https://jestjs.io/docs/jest-object#jestrequireactualmodulename),
`jest.requireActual` is for bypassing the `jest.mock`,
when we want to get to the actual implementation
from inside the mock.
Here, there is no `jest.mock('zustand')`,
so no way for Jest to confuse the implementations.

I see that it wasn't your code before,
but I wanted to drop in when we are at it :)

sewera (approved):
Nice catch! The changes LGTM, but I've got a small discussion started in a comment in code above (or below, depending on how Github views it).

docs/guides/testing.mdx
adimit: @sewera I've fixed the docs to use plain `import` statements. I also fixed some mistakes in the Typescript docmentation. The `create` function was curried there for no reason, and the file type was TSX, which it didn't need to be. 

(The deprecation warnings I was getting were due to me using the wrong type before and are no longer a problem.)

sewera (changes requested):
I think I found what was the idea behind `jest.requireActual`:
it is the advice to [mock the 'zustand' dependency](https://github.com/pmndrs/zustand/blob/main/docs/guides/testing.mdx?plain=1#L103).

So I think it should be brought back because of that.

dai-shi:
> * [ ]  ~`yarn run prettier` for formatting code and docs~ N/A, it's just docs changes.

For formatting docs. 😄

docs/guides/testing.mdx
giulianovarriale: Hello,
First, thanks for picking these fixes up. I was going to open a PR but I realised you were already working on it.

I believe the need to use the `jest.requireActual` is to prevent circular dependencies.
If you write a test for a store you've created, the mock is going to run and it will import `zustand` which is the mock which will import the zustand mock again and it goes like this forever. See the screenshot:
Screenshot 2023-02-23 at 17 00 41.

So, I would suggest to keep the commented line, but I would make it type safe by doing the following:
```ts
const actualCreate = jest.requireActual<typeof import("zustand")>("zustand").create; // if using jest
```

adimit: That's a great suggestion, thanks!

It's curious, I'm not getting that problem in our project. But [madge](https://www.npmjs.com/package/madge) flags it as a circular dependency.

I guess we'll keep it with `jest.useActual`—your suggestion removes the only drawback. I'll update the PR.

adimit:
> > * [ ]   ~`yarn run prettier` for formatting code and docs~ N/A, it's just docs changes.
> 
> For formatting docs. smile

:see_no_evil: Oh dear. There's a German proverb: those who can read are at a clear advantage. Cleary, I wasn't.

sewera (approved):

Scimonster:
This change, removing currying from the typescript docs, broke the mock setup for me. I'm using zustand 4.1.5. Is it related to a change since then?

adimit:
> This change, removing currying from the typescript docs, broke the mock setup for me. I'm using zustand 4.1.5. Is it related to a change since then?

Interesting. Could you specify how it "broke"? How are you calling the mocked `create`? I'm assuming you're using TS. I'm using TS here with 4.3.6, and it works well.

PR #1647, opened 2023-02-20
feat(middleware/persist): add skip hydration option #405

### LINKED ISSUE #405: Add ability to delete, or choose not to rehydrate the store in onRehydrateStorage

gmanninglive:

Fixes #405

## Summary

Add `skipHydration` boolean to persist options. This allows the developer to control when the persist middleware will hydate manually.

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -272,7 +272,7 @@ By default the store with be hydrated on initialization.
 In some applications you may need to control when the first hydration occurs.
 For example, in server-rendered apps.
 
-If you set `skipHydration` the initial call for hydration isn't called,
+If you set `skipHydration`, the initial call for hydration isn't called,
 and it is left up to you to manually call `reHydrate()`.
 
 ```ts
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -269,7 +269,8 @@ export const useBoundStore = create(
 
 By default the store with be hydrated on initialization.
 
-In some applications you may need to control when the first hydration occurs. For example in server rendered apps.
+In some applications you may need to control when the first hydration occurs.
+For example, in server-rendered apps.
 
 If you set `skipHydration` the initial call for hydration isn't called,
 and it is left up to you to manually call `reHydrate()`.
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -261,6 +261,49 @@ export const useBoundStore = create(
 )
 ```
 
+### `skipHydration`
+
+> Type: `boolean | undefined`
+
+> Default: `undefined`
+
+By default the store with be hydrated on initialization.
+
+In some applications you may need to control when the first hydration occurs. For example in server rendered apps.
+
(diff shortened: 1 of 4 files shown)


dai-shi:
> #1537

that's `delayHydration` and different from `skipHydration`.

docs/integrations/persisting-store-data.md
```diff
@@ -261,6 +261,49 @@ export const useBoundStore = create(
...
 
+### `skipHydration`
+
+> Type: `boolean | undefined`
+
+> Default: `undefined`
+
+By default the store with be hydrated on initialization.
+
+In some applications you may need to control when the first hydration occurs. For example in server rendered apps.
```
sewera, line 272:
```suggestion
In some applications you may need to control when the first hydration occurs.
For example, in server-rendered apps.
```

docs/integrations/persisting-store-data.md
```diff
@@ -261,6 +261,49 @@ export const useBoundStore = create(
...
+
+If you set `skipHydration` the initial call for hydration isn't called,
```
sewera, line 274:
```suggestion
If you set `skipHydration`, the initial call for hydration isn't called,
```

src/middleware/persist.ts
```diff
@@ -119,6 +119,16 @@ export interface PersistOptions<S, PersistedState = S> {
...
    */
   merge?: (persistedState: unknown, currentState: S) => S
+
+  /**
+   * An optional boolean that will prevent the persist middleware from triggering hydration on initialization,
+   * This allows you to call `rehydrate()` at a specific point in your apps rendering life-cycle.
+   *
+   * This is useful in SSR application.
+   *
+   * @default undefined
```
sewera, line 129: I know that _technically_ it is undefined, but I think it would be more understandable with `false`.

Please don't change it if you disagree, though.

sewera (commented):
I've only got some minor punctuation suggestions

dai-shi (changes requested):
@gmanninglive Can you follow @sewera 's suggestions?

dai-shi:
> @gmanninglive Can you follow @sewera 's suggestions?

Maybe I can simply apply @sewera 's suggestions.

src/middleware/persist.ts
dai-shi:
```suggestion
   * @default false
```

dai-shi (approved):

PR #1656, opened 2023-02-25
feat: add `useHydration` hook

KrzysztofMoch:
## Summary
While checking if my store has been hydrateed [here](https://github.com/pmndrs/zustand/blob/main/docs/integrations/persisting-store-data.md#how-can-i-check-if-my-store-has-been-hydrated) I though that it'll cool idea to have `useHydration` to all stores, to don't create for every store `useHydration` hook inside codebase.

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -414,7 +414,34 @@ in the [FAQ](#faq) section below.
 
 There are a few different ways to do this.
 
-You can use the [`onRehydrateStorage`](#onrehydratestorage)
+You can use the `useHydration` hook:
+
+```ts
+const useBoundStore = create(
+  persist(
+    (set, get) => ({
+      // ...
+    }),
+    {
+      // ...
+    }
+  )
+);
+
+export default function App() {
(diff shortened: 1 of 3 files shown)


dai-shi:
Hi, thanks for your suggestion.
To keep the repo minimal, we don't add middleware specific hooks to this repo.
Please consider adding it as a recipe in docs, or create a third-party library.

It might be acceptable to extend the middleware a little bit to help creating such hooks. Please open a discussion for your ideas if any.

KrzysztofMoch:
Hi,
so if we want to give the possibility to create hooks like this one, we need types `StorePersist` and `Write` from `middleware/persist` to create type for persistedStore (unless there is another way to do it)

dai-shi:
> Hi, so if we want to give the possibility to create hooks like this one, we need types `StorePersist` and `Write` from `middleware/persist` to create type for persistedStore (unless there is another way to do it)

Does this work? https://tsplay.dev/mxjEzw

KrzysztofMoch:
Yeah, it works perfectly

KrzysztofMoch:
So, we can close this PR

PR #1659, opened 2023-02-26
fix: add explicity export paths for cjs

barelyhuman:
## Related Issues or Discussions

Fixes #1563 

## Summary

Manually adds a set of `module.exports` for each file that needs a the CJS patch for exports. 
This is temporary and will be removed once zustand's API are all named instead of mixed.

## Check List

- [x] `yarn run prettier` for formatting code and docs

PR #1661, opened 2023-02-27
fix: re-export createStore from index

hiendv:

Fixes #1659

## Summary
@barelyhuman forgot to export `createStore` from the index and it caused a breaking change with code like this (after transpiled to CJS).
```js
import { createStore, useStore } from 'zustand'
```

## Check List

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -193,7 +193,12 @@ module.exports = function (args) {
     ...(c === 'index' ? [createDeclarationConfig(`src/${c}.ts`, 'dist')] : []),
     createCommonJSConfig(`src/${c}.ts`, `dist/${c}`, {
       addModuleExport: {
-        index: { default: 'react', create: 'create', useStore: 'useStore' },
+        index: {
+          default: 'react',
+          create: 'create',
+          useStore: 'useStore',
+          createStore: 'exports.createStore',
+        },
         vanilla: { default: 'vanilla', createStore: 'createStore' },
         shallow: { default: 'shallow$1', shallow: 'shallow' },
       }[c],


hiendv:
I'm using `exports.createStore` because of existing vanilla exports.
https://github.com/pmndrs/zustand/blob/0b55a3d74b598aba2e41b09854e9237a4ce0f990/src/index.ts#L1

barelyhuman:
That should've been resolved by the programmatic export but I think that failed for the same reason that the explicit exports were added for. 

Thanks for the quick PR! 

LGTM

dai-shi:
> That should've been resolved by the programmatic export but I think that failed for the same reason that the explicit exports were added for.

Well, I checked the diff against your build, though.

hiendaovinh:
> That should've been resolved by the programmatic export but I think that failed for the same reason that the explicit exports were added for.
> 
> Thanks for the quick PR!
> 
> LGTM

Seems like a problem with module.exports and exports though.
I think we should add an e2e tests for this?

hiendv:
Seems like a problem with module.exports and exports though.
I think we should add an E2E tests for this?

rollup.config.js
```diff
@@ -193,7 +193,12 @@ module.exports = function (args) {
     ...(c === 'index' ? [createDeclarationConfig(`src/${c}.ts`, 'dist')] : []),
     createCommonJSConfig(`src/${c}.ts`, `dist/${c}`, {
       addModuleExport: {
-        index: { default: 'react', create: 'create', useStore: 'useStore' },
+        index: {
+          default: 'react',
+          create: 'create',
+          useStore: 'useStore',
+          createStore: 'exports.createStore',
```
dai-shi, line 200: Let's do this:
```suggestion
          createStore: 'vanilla.createStore',
```

dai-shi (approved):
Nice catch!

barelyhuman:
> > That should've been resolved by the programmatic export but I think that failed for the same reason that the explicit exports were added for.
> > Thanks for the quick PR!
> > LGTM
> 
> Seems like a problem with module.exports and exports though. I think we should add an e2e tests for this?

We do have tests for this... 
There's an multi environment and multi node version test in the actions. We'll need  bundler specific tests as well which is a very deep hole

barelyhuman:
> > That should've been resolved by the programmatic export but I think that failed for the same reason that the explicit exports were added for.
> 
> Well, I checked the diff against your build, though.
> 
> @barelyhuman I don't think the programmatic solution can resolve this.

Was talking about this at the end of the exports. 

image

My bad though, I forgot this is how the generic module.exports is initialized so...
```js
var module = { exports: {} };
var exports = module.exports;
```
we replace the entire `module.exports` so the original `exports.` one are nullified... 

Thanks again to @hiendv for the quick response

hiendv:
No problem! I was doing a dependency update and noticed nextjs crying so hard lol.

barelyhuman:
> No problem! I was doing a dependency update and noticed nextjs crying so hard lol.

I normally go through the sandboxes to check these, I didn't yesterday and that, backfired. 

So again, apologies for anyone who does end up to this issue.

PR #1674, opened 2023-03-04
fix(docs): Update imports

kevin51jiang:

Fixes #

## Summary

Example needs createJSONStorage to run, but it isn't imported

## Check List


dai-shi (approved):
👍

PR #1687, opened 2023-03-12
fix(doc): update state-with-url-hash demo

hellohy:

The old demo in connect-to-state-with-url-hash.md is not available.

## Summary

I have written a new demo.

## Check List

### DIFF
--- a/docs/guides/connect-to-state-with-url-hash.md
+++ b/docs/guides/connect-to-state-with-url-hash.md
@@ -14,8 +14,8 @@ import { persist, StateStorage, createJSONStorage } from 'zustand/middleware'
 const hashStorage: StateStorage = {
   getItem: (key): string => {
     const searchParams = new URLSearchParams(location.hash.slice(1))
-    const storedValue = searchParams.get(key)
-    return storedValue ? JSON.parse(storedValue) : ''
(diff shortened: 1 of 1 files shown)


docs/guides/connect-to-state-with-url-hash.md
```diff
@@ -9,13 +9,13 @@ If you want to connect state of a store to URL hash, you can create your own has
...
 import { create } from 'zustand'
-import { persist, StateStorage } from 'zustand/middleware'
+import { persist, StateStorage, createJSONStorage } from 'zustand/middleware'
 
 const hashStorage: StateStorage = {
   getItem: (key): string => {
     const searchParams = new URLSearchParams(location.hash.slice(1))
     const storedValue = searchParams.get(key)
-    return JSON.parse(storedValue)
+    return storedValue ? JSON.parse(storedValue) : ''
```
sewera, line 18: I think a more concise and single-responsibility-per-line-esque solution will be:

```ts
const storedValue = searchParams.get(key) ?? ''
return JSON.parse(storedValue)
```

hellohy: Sure, thanks for the suggestion!

dai-shi (approved):
Thanks!

PR #1689, opened 2023-03-14
fix(middleware/persist): ensure `persist` does not drop updates in `onRehydrateStorage` when using a synchronous storage API

### LINKED ISSUE #1688: State updates within the 'onRehydrateStorage' callback using a synchronous storage API get dropped
## Summary

Updating a store's state within the `onRehydrateStorage` callback when using a synchronous storage API results in the state update getting overwritten by the persisted state. The overwritten state update will still get persisted, however, resulting in a scenario where (1) the update failed to be applied to the current in-memory store and (2) the in-memory state and the persisted state do not match.

This issue is caused by [the return value of `createState`](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/vanilla.ts#L101) (which in the synchronous case is [`stateFromStorage`](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/middleware/persist.ts#L347)) overwriting [state updates within the 'onRehydrateStorage' hook](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/middleware/persist.ts#L303). The in-callback updates do _initially_ update the current state (and get persisted) via [`set` and `setItem`](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/middleware/persist.ts#L398-L399), but they then get lost because [`stateFromStorage` is never updated to reflect those changes](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/middleware/persist.ts#L294-L306).

This doesn't occur in the asynchronous case because the `createState` function immediately returns, setting the initial state to `undefined`. Then, when the hydration promises resolve during subsequent event loop ticks, they never get overwritten because the `createState` method already returned.

I believe this is also the root cause of #1527 - I tested that issue (using the sandbox provided) without `partialize` defined but it still had the same problem.

## Link to reproduction

I forked the sandbox from #1527 and isolated the issue described above. Feel free to toggle between (1) synchronously and asynchronously updating the state within `onRehydrateStorage` and (2) `localStorage` and `asyncStorage` to see how synchronous/asynchronous APIs affect the issue.

https://codesandbox.io/s/zustand-4-3-x-bug-forked-45m5jq?file=/src/App.js

## Check List

- [X] I've already opened a [discussion](https://github.com/pmndrs/zustand/discussions) before opening this issue, or already discussed in other media.

Please include a minimal reproduction.

> N/A - I've got a reproducible example and opened a PR to fix.

- [X] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction.

Please check this if you're filing an issue regarding TypeScript.

- [X] I've read the [typescript guide](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md), in particular that `create` is to be used as `create<T>()(...)` and not `create<T>(...)`.

coffeebeats:

Fixes #1688

## Summary

See #1688 for a detailed description of the problem. This PR resolves the issue by ensuring `stateFromStorage` gets updated to the most recent value. I also added tests verifying this update pattern works for both synchronous and asynchronous storage APIs.

One note for reviewers: I added a comment describing how, in the asynchronous case, it's possible that the state passed to `postRehydrationCallback` is _not_ the most up-to-date state. This would occur if there was a state update between callbacks in the hydration promise chain. Please let me know if I should leave the comment as-is, make the change, or simply remove the comment.

## Check List

### DIFF
--- a/tests/persistAsync.test.tsx
+++ b/tests/persistAsync.test.tsx
@@ -611,6 +611,7 @@ describe('persist middleware with async configuration', () => {
       { count: 42, name: 'test-storage' },
       undefined
     )
+  })
 
   it('handles state updates during onRehydrateStorage', async () => {
     const storage = {
--- a/tests/persistAsync.test.tsx
+++ b/tests/persistAsync.test.tsx
@@ -561,4 +561,40 @@ describe('persist middleware with async configuration', () => {
     await useBoundStore.persist.rehydrate()
     expect(useBoundStore.persist.hasHydrated()).toBe(true)
   })
+
+  it('handles state updates during onRehydrateStorage', async () => {
+    const storage = {
+      getItem: async () => JSON.stringify({ state: { count: 1 } }),
+      setItem: () => {},
+      removeItem: () => {},
+    }
+
+    const useBoundStore = create<{ count: number; inc: () => void }>()(
+      persist(
+        (set) => ({
+          count: 0,
+          inc: () => set((s) => ({ count: s.count + 1 })),
+        }),
+        {
+          name: 'test-storage',
+          storage: createJSONStorage(() => storage),
+          onRehydrateStorage: () => (s) => s?.inc(),
+        }
+      )
+    )
+
+    function Counter() {
+      const { count } = useBoundStore()
+      return <div>count: {count}</div>
+    }
+
+    const { findByText } = render(
+      <StrictMode>
+        <Counter />
+      </StrictMode>
+    )
+
+    await findByText('count: 2')
+    expect(useBoundStore.getState().count).toEqual(2)
+  })
 })
--- a/tests/persistSync.test.tsx
+++ b/tests/persistSync.test.tsx
@@ -598,6 +598,7 @@ describe('persist middleware with sync configuration', () => {
       { count: 42, name: 'test-storage' },
       undefined
     )
+  })
 
   it('handles state updates during onRehydrateStorage', () => {
     const storage = {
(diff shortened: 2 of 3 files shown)


dai-shi (commented):
looks correct.
can anyone else review too?

sewera (approved):
Looks correct also to me.

sewera:
It's a tricky topic that will not have to be rediscovered with this comment, so I'd leave it as-is.

dai-shi:
> Does this also need to be fixed for the [old implementation](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/middleware/persist.ts#L303-L304)?

My plan is to leave the old implementation unchanged. People shouldn't use the deprecated one moving forward.

dai-shi (approved):

tests/persistSync.test.tsx
```diff
@@ -549,6 +549,7 @@ describe('persist middleware with sync configuration', () => {
     expect(onFinishHydrationSpy2).toBeCalledWith({ count: 2 })
   })
 
+
```
dai-shi, line 552:
```suggestion
```

tests/persistSync.test.tsx
```diff
@@ -598,5 +598,28 @@ describe('persist middleware with sync configuration', () => {
       { count: 42, name: 'test-storage' },
       undefined
     )
+
```
dai-shi, line 601:
```suggestion
  })

```

tests/persistAsync.test.tsx
```diff
@@ -611,5 +611,40 @@ describe('persist middleware with async configuration', () => {
       { count: 42, name: 'test-storage' },
       undefined
     )
+
```
dai-shi, line 614:
```suggestion
  })

```

PR #1692, opened 2023-03-14
fix(middleware/persist): ensure argument for `onRehydrateStorage` and `onHydrate` is defined on first hydration

### LINKED ISSUE #1691: The argument passed to `onRehydrateStorage` and `onHydrate` is always `undefined` during first hydration
## Summary

The argument passed to `onRehydrateStorage` and `onHydrate`, which is [supposed to be `state: S`](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/middleware/persist.ts#L105) (where `S` is the state type), is always `undefined` **during the first hydration**. This occurs for both the old and new persist implementations using either synchronous or asynchronous storage APIs.

The reason this happens is that calling `get()` within the first few lines of `hydrate` always returns the initial value of `state`, declared in the `createStoreImpl` function, which is `undefined`. It will always be `undefined` in this case because the store's `state` value doesn't get updated until [`createState` returns](https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/vanilla.ts#L101), which is _after_ the `persist` middleware (and `hydrate` function) returns.

There are two potential fixes for this:
1. Update the argument passed to `onRehydrateStorage` and `onHydrate` to be `get() ?? configResult`, so these callbacks have the latest _available_ state passed to them.
2. Change the type signatures of `onRehydrateStorage` and `onHydrate` to reflect that the `state` argument can be `undefined`.

I've opened a PR which implements the first approach, but please let me know if approach 2 (or some other resolution) is better.

## Link to reproduction

https://codesandbox.io/s/zustand-4-3-x-bug-forked-j8hqh5?file=/src/App.js

## Check List

> N/A - The issue is reproducible and I've opened a PR to fix it.

- [X] I've already opened a [discussion](https://github.com/pmndrs/zustand/discussions) before opening this issue, or already discussed in other media.

Please include a minimal reproduction.

- [X] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction.

Please check this if you're filing an issue regarding TypeScript.

- [X] I've read the [typescript guide](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md), in particular that `create` is to be used as `create<T>()(...)` and not `create<T>(...)`.

coffeebeats:

Fixes #1691

## Summary

This PR ensures that the argument passed to `onRehydrateStorage` and `onHydrate` is defined during the first hydration of a store. Instead of passing the result of `get()`, which will always be `undefined` before the first hydration, `get() ?? configResult` is passed. I also added tests validating that `onRehydrateStorage` is always passed a defined value for `state`.

One caveat is that it's actually _not possible_ to add an `onHydrate` listener which will be invoked prior to first hydration. As such, I can't test this case, but I've documented it in the test I added.

### Alternative Solution
An alternative solution would be to update the argument type signature of `onRehydrateStorage` and `onHydrate` and to `state?: S` instead of `state: S`, which reflects the potential for state to be `undefined`. This is probably not the right solution, however, as it both changes a public API and makes the `persist` API less useful.

## Check List

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -259,15 +259,14 @@ const oldImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
 
     // On the first invocation of 'hydrate', state will not yet be defined (this is
     // true for both the 'asynchronous' and 'synchronous' case). Pass 'configResult'
-    // as a backup so listeners and 'onRehydrateStorage' are called with the latest
-    // available state.
-    const currentState = get() ?? configResult
+    // as a backup  to 'get()' so listeners and 'onRehydrateStorage' are called with
+    // the latest available state.
 
     hasHydrated = false
-    hydrationListeners.forEach((cb) => cb(currentState))
+    hydrationListeners.forEach((cb) => cb(get() ?? configResult))
 
     const postRehydrationCallback =
-      options.onRehydrateStorage?.(currentState) || undefined
+      options.onRehydrateStorage?.(get() ?? configResult) || undefined
 
     // bind is used to avoid `TypeError: Illegal invocation` error
     return toThenable(storage.getItem.bind(storage))(options.name)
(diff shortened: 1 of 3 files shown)


dai-shi (changes requested):
I think it's the right fix as `merge()` does the same.

Can you make changes only for `newImpl`?

dai-shi (approved):
LGTM

PR #1707, opened 2023-03-22
Don't recommend memoizing selectors

pastelmind:

- https://github.com/pmndrs/zustand/discussions/971 (This PR was requested [here](https://github.com/pmndrs/zustand/discussions/971#discussioncomment-5388443))
- https://github.com/pmndrs/zustand/discussions/658

## Summary

As per the discussions above, memoizing selectors does not improve performance significantly as of v4. Thus we should not recommend memoizing selectors.

Note that the same paragraph was removed from the README in https://github.com/pmndrs/zustand/pull/550 and this section may have been left behind as an oversight.

P. S. I wish the docs stated "selectors do not benefit significantly from memoization". I feel this doesn't belong in the Recipes page, though; perhaps a FAQ section is needed?

## Check List

### DIFF
--- a/docs/recipes/recipes.mdx
+++ b/docs/recipes/recipes.mdx
@@ -65,29 +65,6 @@ const currentBear = useCredentialsStore((state) => state.currentBear)
 const bear = useBearStore((state) => state.bears[currentBear])
 ```
 
-## Memoizing selectors
-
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks!

PR #1737, opened 2023-04-06
fix(docs): fixes ExtractState in TypeScript guide

JacobWeisenburger:
fixes #1735


dai-shi (approved):
LGTM
Nice catch! I wonder why this has been left wrong.

PR #1738, opened 2023-04-07
fix(docs): update testing doc

### LINKED ISSUE #905: Zustand testing throws an error "TypeError: store.getState is not a function"
In the documentation it states that adding this in JEST would enable testing for ZUSTAND state.

```
const create = createState => {
  const store = actualCreate(createState);
  const initialState = store.getState();
  storeResetFns.add(() => store.setState(initialState, true));
  return store;
};
```

But it seems to throw an error "TypeError: store.getState is not a function"

I am using typescript not sure if this causes some issues. Also, it is not well documented on how to use this test in jest. Just want to have values on my states so that they can be used on my componentes during rendering of test components.

dbritto-dev:

Fixes #905 

## Summary

Update testing doc to support w/o middleware testing Zustand using Jest

## Check List

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -8,9 +8,11 @@ nav: 9
 
 When running tests, the stores are not automatically reset before each test run.
 
-Thus, there can be cases where the state of one test can affect another. To make sure all tests run with a pristine store state, you can mock `zustand` during testing and use the following code to create your store:
+Thus, there can be cases where the state of one test can affect another. To make sure all tests run
+with a pristine store state, you can mock `zustand` during testing and use the following code to
+create your store:
 
-```jsx
+```js
 import { create as actualCreate } from 'zustand'
 // const { create: actualCreate } = jest.requireActual('zustand') // if using jest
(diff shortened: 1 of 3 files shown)


docs/guides/testing.md
```diff
@@ -34,13 +36,16 @@ beforeEach(() => {
 
 The way you mock a dependency depends on your test runner/library.
 
-In [jest](https://jestjs.io/), you can create a `__mocks__/zustand.js` and place the code in that file. If your app is using `zustand/vanilla` instead of `zustand`, then you'll have to place the above code in `__mocks__/zustand/vanilla.js`.
+In [jest](https://jestjs.io/), you can create a `__mocks__/zustand.js` and place the code in that
+file. If your app is using `zustand/vanilla` instead of `zustand`, then you'll have to place the
+above code in `__mocks__/zustand/vanilla.js`.
```
sewera, line 41: Nitpick: I find it easier to read and edit if the newlines are inserted after logical pauses in sentences, like commas or full stops.

See <https://sembr.org/>, it's a great read :)

PR #1741, opened 2023-04-07
Simplify TS slices example

Runeii:
Maximise type inference, minimise manual type annotation

## Summary

I recently migrated a large legacy project over to TS, and was struggling to find a way to add TS support to a Zustand store based on slices. The goal was to add TS support in such a way that there was minimal need for type annotation, maximum inference, and no need to maintain manual unions of slice types in the create function.

Proposed change is the result. I wanted to suggest it as a more user friendly approach for the documentation of adding TS support to a slices based Zustand store.

Would welcome feedback/understand if this doesn't match the preferred style of Zustand/there are shortcomings this overlooks.

Thanks

## Check List

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -357,42 +357,25 @@ const useBearStore = create<
 ### Slices pattern
 
 ```ts
-import { create, StateCreator } from 'zustand'
-
-interface BearSlice {
-  bears: number
-  addBear: () => void
-  eatFish: () => void
-}
-const createBearSlice: StateCreator<
-  BearSlice & FishSlice,
-  [],
-  [],
-  BearSlice
-> = (set) => ({
+const createBearSlice = set => ({
   bears: 0,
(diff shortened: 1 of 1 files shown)


dai-shi:
Check out this related discussion: https://github.com/pmndrs/zustand/discussions/1714

Runeii:
Ah, looks like someone else had the same idea since I first tried to tackle this a few months ago – wish I'd re-researched when returning to it this last week.

As I understand it, the main opposition to this sort of approach relates to how middleware can change setState?

I'd concur with the other PR that this would be a huge help for users first integrating Zustand with TS. But I also understand there's more to consider when introducing a change to a library that's already widely used in production.

If the types library is being used to gauge interest in this approach, might it be a sensible option to include a link/reference to this in the main Zustand docs?

dai-shi:
Yeah, middleware can change anything. What's your proposition to docs now?

dai-shi:
Please open a new discussion or a PR for docs.

PR #1754, opened 2023-04-14
chore: prevent merging eslint configs

arjunvegda:
## Summary
[Our `eslint` config for Zustand](https://github.com/pmndrs/zustand/blob/main/.eslintrc.json) gets merged with [one of the example configs](https://github.com/pmndrs/zustand/blob/main/examples/demo/package.json#L24-L29). We fix it by explicitly passing the config to use.

## Check List


dai-shi (approved):
Thanks!!

PR #1763, opened 2023-04-21
feat(middleware/persist): improve createJSONStorage for Maps

### LINKED ISSUE #618: Persist middleware incompatible with immer enableMapSet.
Is this a KP? Wondering what the best workaround is? Even just importing persist seems to cause this.

lauhon:

Fixes #618 

## Summary
The related Issue is closed because a workaround exists. But I would love to see that possible without a workaround needed. Thats why I did this fix.

## Check List

- [x] Add tests to check persistance and rehydration in sync and async mode (I just extended existing tests instead of writing new ones, I hope thats ok)
- [x] implement reviver and replacer for JSON.stringify & JSON.parse

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -20,8 +20,8 @@ export interface PersistStorage<S> {
 }
 
 type JsonStorageOptions = {
-  reviver: (key: string, value: unknown) => unknown
-  replacer: (key: string, value: unknown) => unknown
+  reviver?: (key: string, value: unknown) => unknown
+  replacer?: (key: string, value: unknown) => unknown
 }
 
 export function createJSONStorage<S>(
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -35,7 +35,7 @@ export function createJSONStorage<S>(
         if (str === null) {
           return null
         }
-        return JSON.parse(str) as StorageValue<S>
+        return JSON.parse(str, reviver) as StorageValue<S>
       }
       const str = (storage as StateStorage).getItem(name) ?? null
       if (str instanceof Promise) {
@@ -44,7 +44,10 @@ export function createJSONStorage<S>(
       return parse(str)
     },
     setItem: (name, newValue) =>
-      (storage as StateStorage).setItem(name, JSON.stringify(newValue)),
+      (storage as StateStorage).setItem(
+        name,
+        JSON.stringify(newValue, replacer)
+      ),
     removeItem: (name) => (storage as StateStorage).removeItem(name),
   }
   return persistStorage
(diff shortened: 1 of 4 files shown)


dai-shi:
Hi, thanks for suggestion.
Our design decision with #1463 is to let users specify any storage, so that we don't need to support everything.

So, the general suggestion is to provide an alternative function to `createJOSNStorage` in third-party packages. We basically don't accept changes to the default `createJSONStorage`.

That said, what would be acceptable is to let `createJSONStorage` have a second argument `options` and allow specifying `reviver` and `replacer`. #1720 kind of suggests it.

Would you like to work on it? You would still like to create a third-party package that supports Maps out of the box.

lauhon:
Hey, thanks for the answer!

Alright, makes sense to me.

Just to double check if I understand correctly:

What I would do now is to 

1. extend `createJsonStorage` by an optional Object `Options` which for now can contain a `replacer` and a `reviver` function with the signatures of the code I already have in this PR.

2. The replacing/reviving code itself should not live in zustand, so what I would do is publish a new Lib that contains replacer & reviver (I found no existing npm package that contains that for Map/Set)

3. (OPTIONAL) pull the new lib as dev-dependency into zustand and use it with the new options parameter in the tests here. Then I can basically keep the tests I already did, maybe then it makes even more sense to create a new test just for the new options object..


Please let me know if those steps make sense for you, and if you want me to even do the third step. If not there would then be the question if and how I should test the options parameter 🤔

dai-shi:
@lauhon 
1: Yes. We'll see details during the review process.
2: You have two options for your package. One is to provide replacer/reviver and the other is to provide `createJSONStorageWithMaps` to be used instead of `createJSONStorage`.
3: No, please don't pull a new dev dependency. Just define replacer/reviver in the test code, which is the code you already have in the lib code.

dai-shi:
@lauhon Hi, just wondering how you are doing. No rush, but hope to have the update soon.

lauhon:
Hey @dai-shi, thanks for the push. Was a couple of busy days for me :)

Not sure if its in your interest what i did in the tests. Looking forward to your review!

src/middleware/persist.ts
```diff
@@ -19,8 +19,14 @@ export interface PersistStorage<S> {
   removeItem: (name: string) => void | Promise<void>
 }
 
+type JsonStorageOptions = {
+  reviver: (key: string, value: unknown) => unknown
+  replacer: (key: string, value: unknown) => unknown
```
dai-shi, line 24: I think we can make these optional:
```suggestion
  reviver?: (key: string, value: unknown) => unknown
  replacer?: (key: string, value: unknown) => unknown
```

tests/persistAsync.test.tsx
```diff
@@ -44,7 +45,11 @@ describe('persist middleware with async configuration', () => {
     const storage = {
       getItem: async (name: string) =>
         JSON.stringify({
-          state: { count: 42, name },
+          state: {
+            count: 42,
+            name,
+            map: { type: 'Map', value: [['foo', 'bar']] },
+          },
```
dai-shi, line 52: Could you create a new test case for Map and keep existing tests unmodified?

tests/persistAsync.test.tsx
```diff
@@ -1,8 +1,9 @@
-import { StrictMode, useEffect } from 'react'
 import { afterEach, describe, expect, it, jest } from '@jest/globals'
 import { act, render, waitFor } from '@testing-library/react'
+import { StrictMode, useEffect } from 'react'
```
dai-shi, line 3: Please check lint errors.

dai-shi (commented):
Overall looks good to me. Added some comments.

tests/persistAsync.test.tsx
lauhon: Done, i pretty much copied the old persist- and rehyadrate tests and changed it to persist and rehydrate a map, I hope thats fine

tests/persistAsync.test.tsx
lauhon: Done

dai-shi (approved):
Thanks for your contribution!

PR #1779, opened 2023-04-27
docs(typescript): fix bounded use store example

dannobytes:

Fixes https://github.com/pmndrs/zustand/discussions/1564#discussioncomment-5735802

## Summary

The overloaded return function from `createBoundedUseStore` had marked the first selector as optional. This was causing memoized selectors via `useCallback` to not be correctly typed. Removing the `?` solved the lost type inference.

## Check List

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -401,8 +401,6 @@ If you have some middlewares then replace `StateCreator<MyState, [], [], MySlice
 
 ### Bounded `useStore` hook for vanilla stores
 
-laksdfjlksadjfkl;asdjflksadjlkasf
-
 ```ts
 import { useStore } from 'zustand'
 import { createStore } from 'zustand/vanilla'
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -428,7 +428,7 @@ function useBearStore<T>(
 }
 ```
 
-You can also make an abstract `createBoundedUseStore` if you create bounded `useStore`s often and want to DRY things up...
+You can also make an abstract `createBoundedUseStore` generator function if you need to create bounded `useStore` hooks often and want to DRY things up...
```
sewera, line 431: A [generator function (`function*`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) is a specific concept in JS, so I don't think it's a good name for it.

sewera (changes requested):

docs/guides/typescript.md
dannobytes: Removed `generator`

sewera (approved):

PR #1792, opened 2023-05-05
docs(recipes): fix description of devtools store naming

Debbl:

Fixes #

## Summary

The doc's description that you can name the store by `devtools(store, "MySore")`, but the second parameter in the source is a `devtoolsOptions` object rather than a string.

https://github.com/pmndrs/zustand/blob/0784028ed7591daf890c22ad4c065205cb35d2be/src/middleware/devtools.ts#L203

## Check List

### DIFF
--- a/docs/recipes/recipes.mdx
+++ b/docs/recipes/recipes.mdx
@@ -410,8 +410,8 @@ const useStore = create(devtools(store, { enabled: false }))
 ```
 
 The `devtools` middleware takes the store function as its first argument.
-Optionally, you can name the store with a second argument:
-`devtools(store, "MyStore")`, which will be prefixed to your actions.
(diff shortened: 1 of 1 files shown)


docs/recipes/recipes.mdx
```diff
@@ -410,8 +410,8 @@ const useStore = create(devtools(store, { enabled: false }))
 ```
 
 The `devtools` middleware takes the store function as its first argument.
-Optionally, you can name the store with a second argument:
-`devtools(store, "MyStore")`, which will be prefixed to your actions.
+Optionally, you can name the store with a second argument `devtoolsOptions`:
+`devtools(store, { store: "MyStore" })`, which will be prefixed to your actions.
```
sewera, line 414: Good catch!

sewera (approved):
Looks good!

PR #1793, opened 2023-05-05
feat: add helper createSelector

moonrailgun:
## Summary

add helper fn **createSelector** to quick create typesafe selector fn.

```js
const { bears, fishes } = useBearStore(createSelector('bears', 'fishes'))
```

same as 

```js
const { bears, fishes } = useBearStore((state) => ({ bears: state.bears, fishes: state.fishes }))
```

The reason of create mapping selector is named export has better maintainability

## Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,15 @@
       },
       "module": "./esm/context.js",
       "default": "./context.js"
+    },
+    "./createSelector": {
+      "types": "./createSelector.d.ts",
+      "import": {
+        "types": "./esm/createSelector.d.mts",
+        "default": "./esm/createSelector.mjs"
+      },
+      "module": "./esm/createSelector.js",
+      "default": "./createSelector.js"
     }
   },
   "sideEffects": false,
--- a/readme.md
+++ b/readme.md
@@ -94,6 +94,8 @@ const { nuts, honey } = useBearStore(
   (state) => ({ nuts: state.nuts, honey: state.honey }),
   shallow
 )
+// or use helper with createSelector
+const { nuts, honey } = useBearStore(createSelector('nuts', 'honey'), shallow)
 
 // Array pick, re-renders the component when either state.nuts or state.honey change
 const [nuts, honey] = useBearStore(
--- a/src/createSelector.ts
+++ b/src/createSelector.ts
@@ -1,0 +1,18 @@
+/**
+ * @usage
+ * const { bears } = useBearStore(createSelector('bears'))
+ * const { bears, fishes } = useBearStore(createSelector('bears', 'fishes'))
+ *
+ * @param keys state keys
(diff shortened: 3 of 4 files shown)


dbritto-dev:
I don't see the point of having `createSelector`, since you are using typescript you can type safe the store

dai-shi:
Hi, thanks for the suggestion.
While it may be useful for some developers, we basically don't add features in this repo.
Please consider publishing a new package.
Once done, please feel free to list it in https://github.com/pmndrs/zustand/blob/main/docs/integrations/third-party-libraries.md

moonrailgun:
> I don't see the point of having `createSelector`, since you are using typescript you can type safe the store

my point is if we have much fields, we should write much same code with state like:

```js
create(state => ({
  a: state.a,
  b: state.b,
  c: state.c,
  d: state.d,
  e: state.e,
  ...
}))
```

`createSelector ` is a quick helper which like `shallow` can guidance user how to create selector

dai-shi:
fwiw, you might be interested in https://react-tracked.js.org.

```js
const useTrackedBearStore = createTrackedSelector(useBearStore)

// in a component/hook
  const { bears, fishes } = useTrackedBearStore()
```

dbritto-dev:
> > I don't see the point of having `createSelector`, since you are using typescript you can type safe the store
> 
> my point is if we have much fields, we should write much same code with state like:
> 
> ```js
> create(state => ({
>   a: state.a,
>   b: state.b,
>   c: state.c,
>   d: state.d,
>   e: state.e,
>   ...
> }))
> ```
> 
> `createSelector ` is a quick helper which like `shallow` can guidance user how to create selector

but that's your implementation detail shouldn't be part of zustand

moonrailgun:
> > > I don't see the point of having `createSelector`, since you are using typescript you can type safe the store
> > 
> > 
> > my point is if we have much fields, we should write much same code with state like:
> > ```js
> > create(state => ({
> >   a: state.a,
> >   b: state.b,
> >   c: state.c,
> >   d: state.d,
> >   e: state.e,
> >   ...
> > }))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `createSelector ` is a quick helper which like `shallow` can guidance user how to create selector
> 
> but that's your implementation detail shouldn't be part of zustand

well, i just look as a optimization utils for zustand like `shallow`.

moonrailgun:
> fwiw, you might be interested in https://react-tracked.js.org.
> 
> ```js
> const useTrackedBearStore = createTrackedSelector(useBearStore)
> 
> // in a component/hook
>   const { bears, fishes } = useTrackedBearStore()
> ```

A interesting library. i will be take a look and try to use, thanks for your recommendation

PR #1815, opened 2023-05-18
docs(guides): fix slices pattern types

sldk-yuri:
## Summary
There is no reason to use additional `BearSlice` interface for `createFishSlice`. We haven't got any `BearSlice` related methods or fields, so it could be confusing to use this interface.

## Check List

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -379,12 +379,7 @@ interface FishSlice {
   fishes: number
   addFish: () => void
 }
-const createFishSlice: StateCreator<
-  BearSlice & FishSlice,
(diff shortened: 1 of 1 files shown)


dai-shi:
While it may work, the intention t is to show precise typing with slice patterns. Maybe we could add an example that requires BearSlice type.

sldk-yuri:
> While it may work, the intention t is to show precise typing with slice patterns. Maybe we could add an example that requires BearSlice type.

I think that's good idea, I'll going to provide some example on this weekend.

dai-shi:
Please open a new PR.

PR #1825, opened 2023-05-25
feat(shallow): compare dates even if not referentially equal

nicksrandall:

Fixes #

## Summary
This adds a condition in `shallow` compare function to handle dates

## Check List

### DIFF
--- a/src/shallow.ts
+++ b/src/shallow.ts
@@ -32,6 +32,10 @@ export function shallow<T>(objA: T, objB: T) {
     }
     return true
   }
+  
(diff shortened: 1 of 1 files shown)


dai-shi:
It's requested once before #1064, and we decided to support only plain objects.
Please consider using custom equality function.

We have a test to be explicit:
https://github.com/pmndrs/zustand/blob/8cb150ae8aa4176645d7d33b54964256388ccc9f/tests/shallow.test.tsx#L124-L133

PR #1908, opened 2023-06-30
Update zustand-v3-create-context.md

shaaandi:
Working example: https://codesandbox.io/s/vibrant-taussig-cm5fcd?file=/src/working-store/context.js

Fixes #

## Summary

I see that the example given in the docs is not working correctly in the sense of using the context. I updated it so that you can use the stores using Providers so there state is separate. 

## Check List

### DIFF
--- a/docs/previous-versions/zustand-v3-create-context.md
+++ b/docs/previous-versions/zustand-v3-create-context.md
@@ -111,21 +111,28 @@ Here's the diff showing how to migrate from v3 createContext to v4 API.
 
 ```diff
 // store.tsx
-+ import { createContext, useContext } from "react";
++ import { createContext, useContext, useRef } from "react";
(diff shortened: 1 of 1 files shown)


dai-shi (changes requested):
I know what you mean, but the document is about v3, and the migration guide is a simple diff to illustrate the usage from v3. (the original code is one store anyway.)

I think what you want is to improve the guide in v4. https://github.com/pmndrs/zustand/blob/cf0bef6d5efc14606fd7780ac916b3a701715e33/docs/guides/initialize-state-with-props.md or there can be a new guide.

shaaandi:
@dai-shi  Understood. I actually wasn't able to find that link and was looking at old docs then. I am closing this PR as it's not required 👍 

Thanks.

PR #1913, opened 2023-07-01
📝 Fixed some grammar and punctuation

thedipankarroy:
Fixed some grammar and punctuation

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -8,7 +8,7 @@
 [![Downloads](https://img.shields.io/npm/dt/zustand.svg?style=flat&colorA=000000&colorB=000000)](https://www.npmjs.com/package/zustand)
 [![Discord Shield](https://img.shields.io/discord/740090768164651008?style=flat&colorA=000000&colorB=000000&label=discord&logo=discord&logoColor=ffffff)](https://discord.gg/poimandres)
 
-A small, fast and scalable bearbones state-management solution using simplified flux principles. Has a comfy API based on hooks that isn't boilerplatey or opinionated.
+A small, fast and scalable bearbones state-management solution using simplified flux principles. Has a comfy API based on hooks, isn't boilerplatey or opinionated.
 
 Don't disregard it because it's cute. It has quite the claws, lots of time was spent dealing with common pitfalls, like the dreaded [zombie child problem](https://react-redux.js.org/api/hooks#stale-props-and-zombie-children), [react concurrency](https://github.com/bvaughn/rfcs/blob/useMutableSource/text/0000-use-mutable-source.md), and [context loss](https://github.com/facebook/react/issues/13332) between mixed renderers. It may be the one state-manager in the React space that gets all of these right.
 
--- a/readme.md
+++ b/readme.md
@@ -8,9 +8,9 @@
 [![Downloads](https://img.shields.io/npm/dt/zustand.svg?style=flat&colorA=000000&colorB=000000)](https://www.npmjs.com/package/zustand)
 [![Discord Shield](https://img.shields.io/discord/740090768164651008?style=flat&colorA=000000&colorB=000000&label=discord&logo=discord&logoColor=ffffff)](https://discord.gg/poimandres)
 
-A small, fast and scalable bearbones state-management solution using simplified flux principles. Has a comfy API based on hooks, isn't boilerplatey or opinionated.
+A small, fast and scalable bearbones state-management solution using simplified flux principles. Has a comfy API based on hooks that isn't boilerplatey or opinionated.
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -8,9 +8,9 @@
 [![Downloads](https://img.shields.io/npm/dt/zustand.svg?style=flat&colorA=000000&colorB=000000)](https://www.npmjs.com/package/zustand)
 [![Discord Shield](https://img.shields.io/discord/740090768164651008?style=flat&colorA=000000&colorB=000000&label=discord&logo=discord&logoColor=ffffff)](https://discord.gg/poimandres)
 
-A small, fast and scalable bearbones state-management solution using simplified flux principles. Has a comfy API based on hooks, isn't boilerplatey or opinionated.
+A small, fast and scalable bearbones state-management solution using simplified flux principles. Has a comfy API based on hooks that isn't boilerplatey or opinionated.
```
sewera, line 11: It's a little bit different semantically now. Previously, not being boilerplatey referred to Zustand. Now it's about API.

But I think this version is also alright.

readme.md
```diff
@@ -8,9 +8,9 @@
...
 
-Don't disregard it because it's cute. It has quite the claws, lots of time was spent dealing with common pitfalls, like the dreaded [zombie child problem](https://react-redux.js.org/api/hooks#stale-props-and-zombie-children), [react concurrency](https://github.com/bvaughn/rfcs/blob/useMutableSource/text/0000-use-mutable-source.md), and [context loss](https://github.com/facebook/react/issues/13332) between mixed renderers. It may be the one state-manager in the React space that gets all of these right.
+Don't disregard it because it's cute. It has quite the claws, lots of time was spent dealing with common pitfalls, like the dreaded [zombie child problem](https://react-redux.js.org/api/hooks#stale-props-and-zombie-children), [react concurrency](https://github.com/bvaughn/rfcs/blob/useMutableSource/text/0000-use-mutable-source.md), and [context loss](https://github.com/facebook/react/issues/13332) between mixed renderers. It may be the one state-manager in the React space that gets all of these rights.
```
sewera, line 13:
```suggestion
Don't disregard it because it's cute. It has quite the claws, lots of time was spent dealing with common pitfalls, like the dreaded [zombie child problem](https://react-redux.js.org/api/hooks#stale-props-and-zombie-children), [react concurrency](https://github.com/bvaughn/rfcs/blob/useMutableSource/text/0000-use-mutable-source.md), and [context loss](https://github.com/facebook/react/issues/13332) between mixed renderers. It may be the one state-manager in the React space that gets all of these right.
```

Gets all of these (as in solutions to common pitfalls) **right**. "Right" as an adjective, not a noun.

Please revert this change.

sewera (changes requested):
The title of this PR is not correct. You fixed grammar and punctuation, not typos.

thedipankarroy:
Thanks for Reviewing the PR.
Apologies for not naming the issue correctly and changed the file as you suggested.

readme.md
thedipankarroy: Sure then!

thedipankarroy:
Thanks for Reviewing the PR.
And apologies for not naming the PR correctly and also changed the file as you suggested.

sewera:
No worries, that's what a review is for

sewera (approved):

thedipankarroy:
Is there anything else for me to do?

readme.md
dai-shi: I don't know but the original one is more intentional?

sewera:
No, that's why I gave you an approval, should appear like this:

image

sewera (approved):

readme.md
dai-shi: > I don't know but the original one shows the intention more?

Let me repost it. Is there a grammar fix to preserve the original meaning?

dai-shi: > I don't know but the original one shows the intention more?

Let me repost it. Is there a grammar fix to preserve the original meaning?

dai-shi (commented):
Can you check please?

readme.md
thedipankarroy: Here are some examples, let's see which works best in this scenario.

- It has a comfortable API based on hooks that isn't boilerplatey or opinionated.
- Has a nice way of working that uses hooks and isn't too standard or opinionated.
-  Has a slick API built on hooks that's not stale or snarky.
- Has a nice API that's not too formal or opinionated.

dai-shi: I think what's unopinionated is Zustand itself, not the API. @sewera Please help on this.

sewera: Technically speaking, the only way of interacting with Zustand is through its API. That's why I approved that version as well. With that said, if the intention was specifically to mark Zustand as not opinionated or boilerplatey, let's go with the previous version.

sewera: Now when I think of it, the previous version is probably better.

dai-shi: Yeah, I think Zustand is an unopinionated library and its API allows less boilerplate.
Let's revert the change.

dai-shi (changes requested):
Let's revert the first change.

sewera (approved):

dai-shi (approved):
Thanks for your contribution!
Please feel free to look into `docs` and fix some grammar.

PR #1957, opened 2023-07-25
Docs Example (Persist API): Avoid 'any', avoid "might be null" error and make example clearer

itayperry:
https://github.com/pmndrs/zustand/discussions/1955

This is a small fix for one of the examples in the docs (persist API).
The example:
image
Link: https://docs.pmnd.rs/zustand/integrations/persisting-store-data#how-do-i-use-it-with-map-and-set
## Fixes  

Fixes TypeScript Error in this example and clarifies a certain type that isn't really any.

## Summary
The type of `newValue` isn't exactly `StorageValue<any>` but `StorageValue<YourStateType>`.
This PR helps to clarify that and helps users have better types.
In addition, `str` in `getItem` could be `null` and that's why **this code won't compile**.

`zustand` is incredible and I do not wish to waste your time, I just thought this might help 💛

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -698,17 +698,28 @@ Let's say your state uses `Map` to handle a list of `transactions`,
 then you can convert the Map into an Array in the storage prop:
 
 ```ts
+
+interface BearState {
+  .
+  .
+  .
+  transactions: Map<any>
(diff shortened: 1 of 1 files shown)


chrisk-7777 (approved):
Looks good to me. Much clearer example overall - thanks!

PR #1963, opened 2023-07-28
fix: declare @types/react as peerDep to avoid phantom dep

leviosa-e:

Fixes [#2047 of jotai](https://github.com/pmndrs/jotai/issues/2047)

## Check List


dai-shi (approved):
Thanks!

sewera (approved):
good idea :)

PR #1985, opened 2023-08-09
fix(readme): description with createWithEqualityFn

dai-shi:

https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-6675672

## Summary

Clarify to use `createWithEqualityFn` for `shallow`.

## Check List

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -86,7 +86,7 @@ const honey = useBearStore((state) => state.honey)
 
 If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing the `shallow` equality function.
 
-To use a custom equality function, you need `createWithEqualityFn` from `zustand/traditional`. It takes the second argument for the default equality function.
+To use a custom equality function, you need `createWithEqualityFn` instead of `create`. Usually you want to specify `Object.is` as the second argument for the default equality function, but it's configurable.
 
 ```jsx
 import { createWithEqualityFn } from 'zustand/traditional'
--- a/readme.md
+++ b/readme.md
@@ -86,8 +86,21 @@ const honey = useBearStore((state) => state.honey)
 
 If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing the `shallow` equality function.
 
+To use a custom equality function, you need `createWithEqualityFn` from `zustand/traditional`. It takes the second argument for the default equality function.
+
 ```jsx
+import { createWithEqualityFn } from 'zustand/traditional'
 import { shallow } from 'zustand/shallow'
+
+// Use createWithEqualityFn instead of create
+const useBearStore = createWithEqualityFn(
+  (set) => ({
+    bears: 0,
+    increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
+    removeAllBears: () => set({ bears: 0 }),
+  }),
+  Object.is // Specify the default equality function, which can be shallow
+)
 
 // Object pick, re-renders the component when either state.nuts or state.honey change
 const { nuts, honey } = useBearStore(


readme.md
```diff
@@ -86,9 +86,22 @@ const honey = useBearStore((state) => state.honey)
 
 If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing the `shallow` equality function.
 
+To use a custom equality function, you need `createWithEqualityFn` from `zustand/traditional`. It takes the second argument for the default equality function.
```
sewera, line 89: I feel it's a bit confusing — "to use a custom equality function" and then "for the default equality function."

Is the default eq fn a default for that store, but you can still override it later? If so, I suggest:

```suggestion
To use a custom equality function that will become a new default for the store, you need `createWithEqualityFn` from `zustand/traditional`.
Pass an equality function as a second argument.
```

readme.md
```diff
@@ -86,9 +86,22 @@ const honey = useBearStore((state) => state.honey)
...
 import { shallow } from 'zustand/shallow'
 
+// Use createWithEqualityFn instead of create
+const useBearStore = createWithEqualityFn(
+  (set) => ({
+    bears: 0,
+    increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
+    removeAllBears: () => set({ bears: 0 }),
+  }),
+  Object.is // Specify the default equality function, which can be shallow
```
sewera, line 102: If the eq fn is shallow, what is the effect of using 'zustand/shallow'? Isn't the eq fn always shallow then, regardless of it?

sewera (changes requested):

readme.md
dai-shi: Thanks for the suggestion. Let me clarify and maybe you have some other suggestions.
- `create` (in v5) doesn't allow changing equality function. (It's always `Object.is` equivalent.)
- To use custom equality function, `createWithEqualityFn` is necessary.
  - The signature is `const useMyStore = createEqualityFn(..., defaultEqualityFn)`.
  - `useMyStore(selector, equalityFn)` can override the default equality function.
  - If you specify `Object.is` for the second argument of createEqualityFn, it behave the same as `create` < v4.4.0.

readme.md
dai-shi: If you do this:
```js
import { createWithEqualityFn } from 'zustand/traditional'
import { shallow } from 'zustand/shallow'

const useMyStore = createWithEqualityFn(..., shallow)
```
You don't need to specify shallow for useMyStore:
```js
  useMyStore(...)
  // that 👆 is the same as this 👇 
  useMyStore(..., shallow)
```

readme.md
sewera: Alright, so my proposal:

```suggestion
To use a custom equality function that will become a new default for the store, you need `createWithEqualityFn` from `zustand/traditional`.
Pass a desired equality function as a second argument, like so:
```

readme.md
sewera:
```suggestion
  customEqualityFunction // This will become the default equality function, unless you pass another one to `useBearStore` hook.
```

And then I'd add a `customEqualityFunction`.

sewera (changes requested):
I'd explicitly write some example of an actually different equality function than the `Object.is` to highlight the use case.

One use case I can think about that better highlights what a real-world scenario could be would be something like:

```js
const useNotificationStore = createWithEqualityFn(set => ({
    notifications: []
    addNotification: notification => set(state => ({ notifications: [...state.notifications, notification] }))
  }),
  hashEquality
)

function hashEquality(a, b) {
  return JSON.stringify(a.notifications.map(it => it.hash)) === JSON.stringify(b.notifications.map(it => it.hash))
}
```

provided that notification hash is computed in backend.

dai-shi:
> I'd explicitly write some example

That would be good, but our strong desire is to keep readme.md as minimal as possible.
So, what I'd like to do is to show something very essential in readme.md and link to docs for details.
See: https://github.com/pmndrs/zustand#typescript-usage

sewera:
You're right. I still suggest this rephrasal: https://github.com/pmndrs/zustand/pull/1985#discussion_r1288383694

readme.md
dai-shi: What I was concerning is the backward compatibility with previous version. But, I think it's not super important.

dai-shi: > for the store

It's not for the store (which is still Objet.is), but it's for the hook.

dai-shi: "To use a custom equality function, you need `createWithEqualityFn` instead of `create`. Usually, specify `Object.is` at the second argument for the default equality function, but it's configurable." How about that?

sewera: "To use a custom equality function, you need `createWithEqualityFn` instead of `create`. Usually you want to specify `Object.is` as the second argument for the default equality function, but it's configurable."

dai-shi: Thanks. I'll update it later.

sewera (approved):

PR #1997, opened 2023-08-16
fix(docs): Do not recommend deprecated useStore + equalityFn on initialize with props docs

michelts:
## Summary

Passing an equalityFn to the useStore hook is deprecated. The documentation to initialize the store from props is recommending the approach. It was updated to recommend using useStoreWithEqualityFn instead.

## Check List

BTW notice running prettier messed with a lot of files but didn't change the file I changed. I'm not pushing those other files formatting changes though.

### DIFF
--- a/docs/guides/initialize-state-with-props.md
+++ b/docs/guides/initialize-state-with-props.md
@@ -106,7 +106,7 @@ import { useContext } from 'react'
 import { useStore } from 'zustand'
 
 function useBearContext<T>(
-  selector: (state: BearState) => T,
+  selector: (state: BearState) => T
 ): T {
   const store = useContext(BearContext)
   if (!store) throw new Error('Missing BearContext.Provider in the tree')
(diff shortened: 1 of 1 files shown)


sewera:
> Or, maybe we should show both? because some people don't use `equalityFn`.

I agree, and I'd say that _most_ people don't use `equalityFn`, so showing both would probably make sense.

michelts:
Sure! I updated the content, how does it sound?

docs/guides/initialize-state-with-props.md
```diff
@@ -107,11 +107,10 @@ import { useStore } from 'zustand'
 
 function useBearContext<T>(
   selector: (state: BearState) => T,
```
sewera, line 109:
```suggestion
  selector: (state: BearState) => T
```

sewera (approved):

dai-shi (approved):
Thanks!!
Please run prettier.

michelts:
Done! Sorry not noticing it earlier.

PR #2029, opened 2023-09-06
fix(docs): Fix Immer import

fdb:
The `produce` function is a named export, not a default export:

https://immerjs.github.io/immer/produce

### DIFF
--- a/docs/recipes/recipes.mdx
+++ b/docs/recipes/recipes.mdx
@@ -221,7 +221,7 @@ Reducing nested structures is tiresome.
 Have you tried [Immer](https://github.com/immerjs/immer)?
 
 ```jsx
-import produce from 'immer'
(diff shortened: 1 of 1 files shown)


docs/recipes/recipes.mdx
```diff
@@ -221,7 +221,7 @@ Reducing nested structures is tiresome.
 Have you tried [Immer](https://github.com/immerjs/immer)?
 
 ```jsx
-import produce from 'immer'
+import { produce } from 'immer'
```
dai-shi, line 224: Looks good to me. It this the only place that needs to be changed?

fdb:
There are two other references to `produce` on the page as well. I also updated those.

dai-shi (approved):
Thanks for your contribution!

PR #2047, opened 2023-09-11
fix(docs): Update usage without React on recipes.mdx

ivanquirino:
I've noticed that the usage without React in the recipes section of the documentation shows a deprecated use, which is different from the usage from the readme. This PR makes the recipes usage inline with the readme one.

### DIFF
--- a/docs/recipes/recipes.mdx
+++ b/docs/recipes/recipes.mdx
@@ -183,7 +183,7 @@ const store = createStore(() => ({ ... }))
 const { getState, setState, subscribe, destroy } = store
 ```
 
-You can use a vanilla store with `useStore` hook available since v4.
+You can use a vanilla store in React with a `useStore` hook.
 
 ```jsx
 import { useStore } from 'zustand'
--- a/docs/recipes/recipes.mdx
+++ b/docs/recipes/recipes.mdx
@@ -183,14 +183,15 @@ const store = createStore(() => ({ ... }))
 const { getState, setState, subscribe, destroy } = store
 ```
 
-You can even consume an existing vanilla store with React:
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks!

I'm a bit worried about the duplication between readme.md and redupes.mdx.
There's a historical reason, but if someone has an idea to improve it, it's welcome.
cc: @dbritto-dev @sewera @chrisk-7777

docs/recipes/recipes.mdx
```diff
@@ -183,15 +183,16 @@ const store = createStore(() => ({ ... }))
 const { getState, setState, subscribe, destroy } = store
 ```
 
-You can even consume an existing vanilla store with React:
+You can use a vanilla store with `useStore` hook available since v4.
```
sewera, line 186: I think the docs should implicitly be for the current version of Zustand.
Also, I'll mention that we are using a vanilla store _in React_, because this section is all about the usage _without React_.

```suggestion
You can use a vanilla store in React with a `useStore` hook.
```

sewera (changes requested):

dai-shi:
> I think the docs should implicitly be for the current version of Zustand.

@sewera It's how described in readme.md. Maybe, it's time to revisit.

ivanquirino:
I don't know why it is failing. I've read the job log but no idea.

dbritto-dev:
> I don't know why it is failing. I've read the job log but no idea.

just run `yarn prettier`

ivanquirino:
> > I don't know why it is failing. I've read the job log but no idea.
> 
> just run `yarn prettier`

I did this PR directly on github and missed that. Thank you.

sewera (approved):

PR #2059, opened 2023-09-17
chore(context) shave bytes

charkour:
## Summary

This shaves bytes within the react context function by using logical OR assignment ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment)).

Thank you

## Check List

- [x] Verified tests passed

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -44,7 +44,7 @@ function createContext<S extends StoreApi<unknown>>() {
 
     return createElement(
       ZustandContext.Provider,
-      { value: storeRef.current ||= createStore() },
+      { value: (storeRef.current ||= createStore()) },
       children
     )
   }
--- a/src/context.ts
+++ b/src/context.ts
@@ -42,13 +42,9 @@ function createContext<S extends StoreApi<unknown>>() {
   }) => {
     const storeRef = useRef<S>()
 
-    if (!storeRef.current) {
-      storeRef.current = createStore()
-    }
-
     return createElement(
       ZustandContext.Provider,
-      { value: storeRef.current },
+      { value: storeRef.current ||= createStore() },
       children
     )
   }


charkour:
From GitHub Actions:

Error: compressed-size-action was unable to comment on your PR.
This can happen for PR's originating from a fork without write permissions.
You can copy the size table directly into a comment using the markdown below:


**Size Change:** -4 B (0%) 

**Total Size:** 43.5 kB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/context.js` | 785 B | +3 B (0%) |
| `dist/esm/context.js` | 591 B | +1 B (0%) |
| `dist/system/context.development.js` | 699 B | +2 B (0%) |
| `dist/system/context.production.js` | 383 B | -3 B (-1%) |
| `dist/umd/context.development.js` | 885 B | -2 B (0%) |
| `dist/umd/context.production.js` | 551 B | -5 B (-1%) |

src/context.ts
```diff
@@ -42,13 +42,9 @@ function createContext<S extends StoreApi<unknown>>() {
...
     const storeRef = useRef<S>()
 
-    if (!storeRef.current) {
-      storeRef.current = createStore()
-    }
-
     return createElement(
       ZustandContext.Provider,
-      { value: storeRef.current },
+      { value: (storeRef.current ||= createStore()) },
```
sewera, line 47: I find such shorthand less readable than an explicit `if`, which purposefully takes some space to highlight a certain condition.

Also, nullish coalescing assignment would probably make more sense here:

```suggestion
      { value: (storeRef.current ??= createStore()) },
```

But again, I disagree with this change. I don't want to do bike shedding here though, so I won't block it and leave the decision-making to @dai-shi :)

dai-shi: I actually like `||=` and I would use it when it makes sense.
For this PR, it's not important as we will be dropping it in the future version.

dai-shi:
image

If it increases the size for cjs and esm, there's no reason to merge this pr.

charkour:
Thanks for the review. I appreciate the explanation :)

PR #2076, opened 2023-09-24
fix doc: fix bearbones to barebones

c0sc0s:
There might be a spelling error in the document, it should be 'barebones', not 'bearbones'.


sewera:
It's an intentional pun, because a bear is the mascot of Zustand.

c0sc0s:
> It's an intentional pun, because a bear is the mascot of Zustand.

ohh! HAHA, it's so funny.

Issue #903, opened 2022-04-10
[test] add spec with ssr


dai-shi:
#886 was reported and fixed, but we don't have a test for it.

https://github.com/reduxjs/react-redux/pull/1888 can be a good reference.

amareshsm:
I would like to give it a try. I might need some time to complete it. Can I pick this?

arnaugomez:
Hi, because more than a year has passed since last activity, I will give it a try and make a PR

qqqzhch:
This is to write a demo, such as using zustand in next13？

dai-shi:
This is about adding a new test in https://github.com/pmndrs/zustand/tree/main/tests.

NaamuKim:
@dai-shi I would like to adding a test for this issue.
Can you assign me?
I will upload the PR before October 3rd.

PR #2088, opened 2023-09-28
test(ssr): Validate state synchronization between server and client in React 18 using Zustand (#903)

NaamuKim:

- #903

## Summary

This change tests state synchronization between server and client using React 18 and Zustand. It verifies that the server-rendered markup reflects the initial state correctly, and this state is accurately updated on the client side. This test is executed only when the version of React is 18.

Please suggest me if you have any other tests that might be helpful!

## Check List

### DIFF
--- a/.github/workflows/test-old-typescript.yml
+++ b/.github/workflows/test-old-typescript.yml
@@ -13,6 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         typescript:
+          - 5.1.6
+          - 5.0.4
           - 4.9.5
           - 4.8.4
           - 4.7.4
--- a/readme.md
+++ b/readme.md
@@ -495,6 +495,7 @@ Basic typescript usage doesn't require anything special except for writing `crea
 ```ts
 import { create } from 'zustand'
 import { devtools, persist } from 'zustand/middleware'
+import type {} from '@redux-devtools/extension' // required for devtools typing
 
 interface BearState {
   bears: number
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -5,75 +5,11 @@ import type {
   StoreMutatorIdentifier,
 } from '../vanilla.ts'
 
-// Copy types to avoid import type { Config } from '@redux-devtools/extension'
-// https://github.com/pmndrs/zustand/issues/1205
-type Action<T = any> = {
-  type: T
-}
-type ActionCreator<A, P extends any[] = any[]> = {
-  (...args: P): A
-}
-type EnhancerOptions = {
-  name?: string
-  actionCreators?:
-    | ActionCreator<any>[]
-    | {
-        [key: string]: ActionCreator<any>
-      }
-  latency?: number
-  maxAge?: number
-  serialize?:
-    | boolean
-    | {
-        options?:
-          | undefined
-          | boolean
-          | {
-              date?: true
-              regex?: true
-              undefined?: true
-              error?: true
-              symbol?: true
-              map?: true
-              set?: true
-              function?: true | ((fn: (...args: any[]) => any) => string)
-            }
-        replacer?: (key: string, value: unknown) => any
-        reviver?: (key: string, value: unknown) => any
-        immutable?: any
-        refs?: any
-      }
-  actionSanitizer?: <A extends Action>(action: A, id: number) => A
-  stateSanitizer?: <S>(state: S, index: number) => S
-  actionsBlacklist?: string | string[]
-  actionsWhitelist?: string | string[]
-  actionsDenylist?: string | string[]
-  actionsAllowlist?: string | string[]
-  predicate?: <S, A extends Action>(state: S, action: A) => boolean
-  shouldRecordChanges?: boolean
-  pauseActionType?: string
-  autoPause?: boolean
-  shouldStartLocked?: boolean
-  shouldHotReload?: boolean
-  shouldCatchErrors?: boolean
-  features?: {
-    pause?: boolean
-    lock?: boolean
-    persist?: boolean
-    export?: boolean | 'custom'
-    import?: boolean | 'custom'
-    jump?: boolean
-    skip?: boolean
-    reorder?: boolean
-    dispatch?: boolean
-    test?: boolean
-  }
-  trace?: boolean | (<A extends Action>(action: A) => string)
-  traceLimit?: number
-}
-type Config = EnhancerOptions & {
-  type?: string
-}
+type Config = Parameters<
+  (Window extends { __REDUX_DEVTOOLS_EXTENSION__?: infer T }
+    ? T
+    : any)['connect']
+>[0]
 
 declare module '../vanilla' {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
@@ -124,6 +60,7 @@ type StoreDevtools<S> = S extends {
   : never
 
 export interface DevtoolsOptions extends Config {
+  name?: string
   enabled?: boolean
   anonymousActionType?: string
   store?: string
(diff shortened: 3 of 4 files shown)


dai-shi:
Can anyone review this PR please?

tests/ssr.test.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useEffect } from 'react'
+import { act, screen } from '@testing-library/react'
+import { hydrateRoot } from 'react-dom/client'
+import { renderToString } from 'react-dom/server'
+import { describe, expect, it } from 'vitest'
+import { create } from 'zustand'
+
+const IS_REACT_18 = React.version.startsWith('18')
```
dbritto-dev, line 8: We can skip any react version under 18

```suggestion
import React, { useEffect } from 'react'

if (!React.version.startsWith('18')) {
  describe.skip('skipping test on react version < 18', () => {})
  return
}

import { act, screen } from '@testing-library/react'
import { hydrateRoot } from 'react-dom/client'
import { renderToString } from 'react-dom/server'
import { describe, expect, it } from 'vitest'
import { create } from 'zustand'
```

tests/ssr.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      increasePopulation()
+    }, [increasePopulation])
+
+    return <div>bears: {bears}</div>
+  }
+
+  if (!IS_REACT_18) {
+    it('Dummy test for React 17, ignore', () => {})
+    return
+  }
```
dbritto-dev, line 43: We can omit this statement

```suggestion
```

dbritto-dev (changes requested):
Some ideas to make it work on react versions under 18

tests/ssr.test.tsx
NaamuKim: Oh, thank you for your review.
I understand your intention.
I resolved it by using the only method from vitest and dynamic import.
What do you think about it?

dbritto-dev: @NaamuKim that approach doesn't work you should completely avoid import react-dom/client. Let's try with my suggestion.

NaamuKim: There's an issue with top-level returns not being allowed in ESM, causing errors. 
I think your suggestion of using describe.skip is excellent, and I've configured it to conditionally branch describe based on the version. 
Could you please confirm this?

dbritto-dev:
@dai-shi can we trigger the tests again?

tests/ssr.test.tsx
dbritto-dev: Looks good to me

tests/ssr.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const container = document.createElement('div')
+    document.body.appendChild(container)
+    container.innerHTML = markup
+
+    expect(container.textContent).toContain('bears: 0')
+
+    await act(async () => {
+      // Using dynamic import for 'react-dom/client' as it’s not accessible in React versions below 18.
+      // eslint-disable-next-line import/no-unresolved
+      const { hydrateRoot } = await require('react-dom/client')
```
dbritto-dev, line 55: Move to the top and use `vi.importActual` instead of require

```suggestion
```

tests/ssr.test.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { useEffect } from 'react'
+import { act, screen } from '@testing-library/react'
+import { renderToString } from 'react-dom/server'
+import { describe, expect, it } from 'vitest'
+import { create } from 'zustand'
+
+const describeTest = React.version.startsWith('18') ? describe : describe.skip
+
+describeTest('ssr behavior with react 18', () => {
+  interface BearStoreState {
```
dbritto-dev, line 10:
```suggestion
  const { hydrateRoot } = await vi.importActual('react-dom/client')

  interface BearStoreState {
```

tests/ssr.test.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { useEffect } from 'react'
+import { act, screen } from '@testing-library/react'
+import { renderToString } from 'react-dom/server'
+import { describe, expect, it } from 'vitest'
```
dbritto-dev, line 4: Add vi on imports

```suggestion
import { describe, expect, it, vi } from 'vitest'
```

dbritto-dev (changes requested):
Some minor tweaks

tests/ssr.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { create } from 'zustand'
+
+const describeTest = React.version.startsWith('18') ? describe : describe.skip
+
+describeTest('ssr behavior with react 18', () => {
```
dbritto-dev, line 9: Let's use `describe.skipIf`

```suggestion
describe.skipIf(!React.version.startsWith('18'))('ssr behavior with react 18', () => {
```

dbritto-dev (changes requested):

tests/ssr.test.tsx
NaamuKim: I want to use it like this, but if I change it to "skipIf" and it runs, it throws an error because it doesn't actually skip the context.
```
describe.skipIf(!React.version.startsWith('18'))(
  'ssr behavior with react 18',
  async () => {
    const {
      hydrateRoot,
    }: {
      hydrateRoot: (
        container: Element | Document,
        initialChildren: React.ReactNode
      ) => void
    } = await vi.importActual('react-dom/client')
```
throw error with a message `Error: Failed to load url react-dom/client (resolved id: react-dom/client). Does the file exist?`

If I move the other context inside the it context, it resolves the issue.

I've applied the changes with it. 
Please take a look and feel free to share any further suggestions or improvements

dbritto-dev: Yeah I mean to the top of the context

tests/ssr.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+  async () => {
+    it('should handle different states between server and client correctly', async () => {
+      const {
+        hydrateRoot,
+      }: {
+        hydrateRoot: (
+          container: Element | Document,
+          initialChildren: React.ReactNode
+        ) => void
+      } = await vi.importActual('react-dom/client')
```
dbritto-dev, line 47:
```suggestion
      const { hydrateRoot } = await vi.importActual<typeof import('react-dom/client')>('react-dom/client')
```

dbritto-dev (changes requested):

tests/ssr.test.tsx
NaamuKim: Thank you for the suggestion. I've made the requested changes as you suggested.

dbritto-dev (approved):

tests/ssr.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  )
+
+  useEffect(() => {
+    increasePopulation()
+  }, [increasePopulation])
+
+  return <div>bears: {bears}</div>
+}
+
+describe.skipIf(!React.version.startsWith('18'))(
```
dai-shi, line 36: Oh, I didn't know about `skipIf`!

We should use it in `jotai` too:
https://github.com/pmndrs/jotai/blob/d6afb777705eb3a90385817fd6db30db101f6caa/tests/react/transition.test.tsx#L12C1-L16

Free free to open a PR there.

dai-shi (approved):
Thanks for your contribution!

tests/ssr.test.tsx
NaamuKim: I am also glad to have learned about skipIf, and I will apply refactoring accordingly in Jotai as well

dbritto-dev: I guess we are good to merge the changes :D. BTW, good job.

NaamuKim: Thank you so much for your review. 
I'm delighted to learn about the useful functions of vitest. I'll continue to check if there are any opportunities for further contributions.

PR #2090, opened 2023-09-30
feat: add useShallow

FaberVitale:
## Summary

Adds a `useShallow` hook to prevent unnecessary re-renders when a selector outputs the same value according to shallow equal. 

### Basic usage

```tsx
import { useShallow } from 'zustand/shallow'

const useMyStore = create(() => ({ a: 1, b: 2, c: 3 }))

export const TestShallow = () => {
  const output = useMyStore(useShallow(Object.keys))

  return <div>{output.join(',')}</div>
}
```

See tests in `shallow.test.tsx` for more details.

---

- https://github.com/pmndrs/zustand/discussions/1937
- https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-7118242
- https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-6974554

## Check List

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -86,36 +86,31 @@ const honey = useBearStore((state) => state.honey)
 
 If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing the `shallow` equality function.
 
-To use a custom equality function, you need `createWithEqualityFn` instead of `create`. Usually you want to specify `Object.is` as the second argument for the default equality function, but it's configurable.
+To prevent unnecessary rerenders when the selector output does not change according to shallow equal, you can use [useShallow](./docs/guides/prevent-rerenders-with-use-shallow.md).
 
 ```jsx
-import { createWithEqualityFn } from 'zustand/traditional'
-import { shallow } from 'zustand/shallow'
+import { create } from 'zustand'
+import { useShallow } from 'zustand/shallow'
 
 // Use createWithEqualityFn instead of create
-const useBearStore = createWithEqualityFn(
-  (set) => ({
-    bears: 0,
-    increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
-    removeAllBears: () => set({ bears: 0 }),
-  }),
-  Object.is // Specify the default equality function, which can be shallow
-)
+const useBearStore = create((set) => ({
+  bears: 0,
+  increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
+  removeAllBears: () => set({ bears: 0 }),
+}))
 
 // Object pick, re-renders the component when either state.nuts or state.honey change
 const { nuts, honey } = useBearStore(
-  (state) => ({ nuts: state.nuts, honey: state.honey }),
-  shallow
+  useShallow((state) => ({ nuts: state.nuts, honey: state.honey }))
 )
 
 // Array pick, re-renders the component when either state.nuts or state.honey change
 const [nuts, honey] = useBearStore(
-  (state) => [state.nuts, state.honey],
-  shallow
+  useShallow((state) => [state.nuts, state.honey])
 )
 
 // Mapped picks, re-renders the component when state.treats changes in order, count or keys
-const treats = useBearStore((state) => Object.keys(state.treats), shallow)
+const treats = useBearStore(useShallow((state) => Object.keys(state.treats)))
 ```
 
 For more control over re-rendering, you may provide any custom equality function.
--- a/readme.md
+++ b/readme.md
@@ -84,9 +84,7 @@ const nuts = useBearStore((state) => state.nuts)
 const honey = useBearStore((state) => state.honey)
 ```
 
-If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing the `shallow` equality function.
-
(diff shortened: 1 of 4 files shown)


src/shallow.ts
```diff
@@ -59,3 +61,17 @@ export default ((objA, objB) => {
...
+  const prev = useRef<U>()
+
+  return useMemo(
+    () => (state) => {
+      const next = selector(state)
+      return shallow(prev.current, next)
+        ? (prev.current as U)
+        : (prev.current = next)
+    },
+    [selector]
```
dai-shi, line 75: I don't think we need useMemo nor useCallback for this, because Zustand doesn't require stable selector.

Do you prefer `useRef`? I think we should use useMemo like https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-6974554 , because React may prohibit updating ref during render in the future (L73 is usually not allowed).

dai-shi:
> Basic usage

`create` returns a hook. So, let's do this.
```jsx
import { useShallow } from 'zustand/shallow'

const useMyStore = create(() => ({ a: 1, b: 2, c: 3 }))

export const TestShallow = () => {
  const output = useMyStore(useShallow(Object.keys))

  return <div>{output.join(',')}</div>
}
```

Maybe, `useShallow((state) => Ojbect.keys(state))` might be easier to read even though it can be less efficient.

src/shallow.ts
FaberVitale: Version 1 of https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-6974554 requires useCallback to use inline selectors.

Version 2 of https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-6974554 works but we open the door to potential stale closure issues.


If there are no better alternatives version 2 is better.

Are there less cumbersome alternatives.

---

I wish we could use [useEvent or something similar](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md) during the render phase :/

dai-shi: Oops, that's my bad. I forgot about it. Let's use `useRef` then.

I think you can remove useMemo from your implementation.

dai-shi: > It's not necessary because the caller of useShallow can always wrap the selector with useCallabck.

We never need useCallback, it doesn't improve anything.

src/shallow.ts
```diff
@@ -59,3 +61,14 @@ export default ((objA, objB) => {
...
+export function useShallow<S, U>(selector: (state: S) => U): (state: S) => U {
+  const prev = useRef<U>()
+
+  return (state) => {
+    const next = selector(state)
+    return shallow(prev.current, next)
+      ? (prev.current as U)
+      : (prev.current = next)
+  }
+}
```
dai-shi, line 74: The code looks good to me.

FaberVitale:
@dai-shi  do you know why `renderHook` is not defined [here](https://github.com/pmndrs/zustand/actions/runs/6363218830/job/17278913539)?

It's rather odd.

I can implement those tests without `renderHook` if needed.

dai-shi:
> Why useShallow does not have a dependency array?

I don't think we need that section.

We need to update docs too. Would you like to work on this PR, or should it be a separate PR?

dai-shi:
> do you know why `renderHook` is not defined

I think it's because in the production mode.

> I can implement those tests without `renderHook` if needed.

Yeah, that should be the way to go.

FaberVitale:
> > Why useShallow does not have a dependency array?
> 
> I don't think we need that section.
> 
> We need to update docs too. Would you like to work on this PR, or should it be a separate PR?

I'll open another PR tomorrow.

Thanks

FaberVitale:
Added guide in [9526d87](https://github.com/pmndrs/zustand/pull/2090/commits/9526d87ab80b0ebb4a1260c5ab66def886f72ac2).

Let me know if it's ok.

Thanks

dai-shi:
> Added guide in [9526d87](https://github.com/pmndrs/zustand/pull/2090/commits/9526d87ab80b0ebb4a1260c5ab66def886f72ac2).
> 
> Let me know if it's ok.
> 
> Thanks

The content looks good. But, what's required in #1937 discussion, I think, is that
we should show the use of `useShallow` everywhere in the docs instead of using `createWithEqualityFn` and `shallow`. What do you think?

readme.md
```diff
@@ -86,36 +86,31 @@ const honey = useBearStore((state) => state.honey)
 
 If you want to construct a single object with multiple state-picks inside, similar to redux's mapStateToProps, you can tell zustand that you want the object to be diffed shallowly by passing the `shallow` equality function.
```
dai-shi, line 87: Needs to revise this line too?

readme.md
```diff
@@ -86,36 +86,31 @@ const honey = useBearStore((state) => state.honey)
...
-To use a custom equality function, you need `createWithEqualityFn` instead of `create`. Usually you want to specify `Object.is` as the second argument for the default equality function, but it's configurable.
+To prevent unnecessary rerenders when the selector output does not change according to shallow equal, you can use [useShallow](./docs/guides/prevent-rerenders-with-use-shallow.md).
 
 ```jsx
-import { createWithEqualityFn } from 'zustand/traditional'
-import { shallow } from 'zustand/shallow'
+import { create } from 'zustand'
+import { useShallow } from 'zustand/shallow'
 
 // Use createWithEqualityFn instead of create
```
dai-shi, line 95:
```suggestion
```

dai-shi (commented):
Thanks for working on docs too!
Below are some comments.

dai-shi (approved):
Thanks so much for your contribution!

PR #2098, opened 2023-10-04
Update testing docs

dbritto-dev:

Testing docs improvements

## Summary

Add some notes/warnings so people can be aware of that

## Check List

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -58,7 +58,7 @@ this means your application logic does not need to be changed or mocked when wri
 The mock provided below will enable the relevant test runner to reset the zustand stores after each test.
 
 > **Warning:** In the following examples we are only supporting the curried version of create `(create()(...))`,
-> due to that is supported on both JavaScript and Typescript. If you are using the uncurried version of
+> since it is supported on both JavaScript and Typescript. If you are using the uncurried version of
 > `create(...)` you will need to update your hooks to use the curried version, or update the mocks in order
 > to support the uncurried version.
 
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -141,7 +141,7 @@ In the next steps we are going to setup our Vitest environment in order to mock 
 > that, you need make sure that you are creating your `__mocks__` directory in the right place.
 > Let's say that you change the **root** to `./src`, that means you need to create a `__mocks__`
 > directory under `./src`. The end result would be `./src/__mocks__`, rather than `./__mocks__`.
-> Creating `__mocks__` directory in the wrong place could lead to issues while using Vitest.
+> Creating `__mocks__` directory in the wrong place can lead to issues when using Vitest.
 
 ```ts
 // __mocks__/zustand.ts
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -58,7 +58,7 @@ this means your application logic does not need to be changed or mocked when wri
 The mock provided below will enable the relevant test runner to reset the zustand stores after each test.
 
 > **Warning:** In the following examples we are only supporting the curried version of create `(create()(...))`,
-> due to that is compatible with both JavaScript and Typescript, If you are using the uncurried version of
+> due to that is supported on both JavaScript and Typescript. If you are using the uncurried version of
 > `create(...)` you will need to update your hooks to use the curried version, or update the mocks in order
 > to support the uncurried version.
 
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -137,8 +137,8 @@ export default config
 
 In the next steps we are going to setup our Vitest environment in order to mock Zustand.
 
-> **Warning:** In Vitest you can change the [root](https://vitest.dev/config/#root), due to
-> that, you need make sure that you are creating your `__mocks__` directory in the right place.
+> **Warning:** In Vitest you can change the [root](https://vitest.dev/config/#root).
+> Due to that, you need make sure that you are creating your `__mocks__` directory in the right place.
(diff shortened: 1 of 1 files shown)


docs/guides/testing.md
```diff
@@ -57,6 +57,11 @@ this means your application logic does not need to be changed or mocked when wri
 
 The mock provided below will enable the relevant test runner to reset the zustand stores after each test.
 
+> **Warning:** on these examples we are supporting only the curried version `create()(...)` since that
+> works for JavaScript and Typescript, if you are using the uncurried version `create(...)` you need
+> to update your hooks to use the curried version or update the mocks in order to support the uncurried
+> version.
+
```
chrisk-7777, line 64: This is looking good. A few tweaks:

> **Warning:** In the following examples we are only demostrating the curried version of create (`create()(...)`), because that is compatible with both 
> JavaScript and Typescript. If you are using the uncurried version of `create(...)` you will need
> to update your hooks to use the curried version, or update the mocks in order to support the uncurried
> version.

docs/guides/testing.md
```diff
@@ -130,7 +135,13 @@ export default config
...
 ### Vitest
 
-In the next steps we are going to setup our Vitest environment in order to mock Zustand
+In the next steps we are going to setup our Vitest environment in order to mock Zustand.
+
+> **Warning:** due to in Vitest you can change the [root](https://vitest.dev/config/#root) you
+> need make sure that you are creating your `__mocks__` folder in the right place. Let's say
+> that you change the **root** to `./src`, that means you need to create `__mocks__` folder
+> under `./src` ending up with `./src/__mocks__` rather than `./__mocks__`. Creating `__mocks__`
+> folder in the wrong place could lead you to some issues while using Vitest.
```
chrisk-7777, line 144: Just a few small tweaks here:

> **Warning:** In Vitest you can change the [root](https://vitest.dev/config/#root), because of this, you
> need make sure that you are creating your `__mocks__` directory in the right location. Let's say
> that you change the **root** to `./src`, that means you need to create a `__mocks__` directory
> under `./src`. The end result would be `./src/__mocks__`, rather than `./__mocks__`. Creating a `__mocks__`
> directory in the wrong location could lead to issues while using Vitest.

chrisk-7777 (commented):
Looking good, nice improvement. I just had some minor grammatical tweaks.

docs/guides/testing.md
chrisk-7777: Thanks for looking through the feedback. 

I think this section "due to that is compatible" is still not quite right? 

Also "both JavaScript and Typescript, If you" is not quite right either. If you want to keep the comma, "If" will need to be "if". However, I think that sentence is very long, and it would be worth breaking it up - similar to the suggestion above. 

What are your thoughts?

dbritto-dev: @chrisk-7777 what you think about "due to that is supported on both" instead of "due to that is compatible"

chrisk-7777: I think the main issue is "due to that is" doesn't really read well in my opinion. 
I'm happy to a second opinion though :)

docs/guides/testing.md
```diff
@@ -57,6 +57,11 @@ this means your application logic does not need to be changed or mocked when wri
 
 The mock provided below will enable the relevant test runner to reset the zustand stores after each test.
 
+> **Warning:** In the following examples we are only supporting the curried version of create `(create()(...))`,
+> due to that is supported on both JavaScript and Typescript. If you are using the uncurried version of
```
sewera, line 61:
```suggestion
> due to it being supported by both JavaScript and TypeScript. If you are using the uncurried version of
```

docs/guides/testing.md
```diff
@@ -130,7 +135,13 @@ export default config
 
 ### Vitest
 
-In the next steps we are going to setup our Vitest environment in order to mock Zustand
+In the next steps we are going to setup our Vitest environment in order to mock Zustand.
+
+> **Warning:** In Vitest you can change the [root](https://vitest.dev/config/#root), due to
+> that, you need make sure that you are creating your `__mocks__` directory in the right place.
```
sewera, line 141:
```suggestion
> **Warning:** In Vitest you can change the [root](https://vitest.dev/config/#root).
> Due to that, you need make sure that you are creating your `__mocks__` directory in the right place.
```

docs/guides/testing.md
```diff
@@ -130,7 +135,13 @@ export default config
...
+> Let's say that you change the **root** to `./src`, that means you need to create a `__mocks__`
+> directory under `./src`. The end result would be `./src/__mocks__`, rather than `./__mocks__`.
+> Creating `__mocks__` directory in the wrong place could lead to issues while using Vitest.
```
sewera, line 144:
```suggestion
> Creating `__mocks__` directory in the wrong place can lead to issues when using Vitest.
```

sewera (changes requested):

docs/guides/testing.md
dbritto-dev: @chrisk-7777 because of that doesn't have the same meaning of due to that, does it?

dbritto-dev: Let me reword it

docs/guides/testing.md
dbritto-dev: @sewera I changed to `since it is support on` instead of `due to that is`

dbritto-dev:
@sewera @chrisk-7777 all yours :D

sewera (approved):

dbritto-dev:
@dai-shi we need your help here :D

PR #2119, opened 2023-10-13
Update reset state docs

### LINKED ISSUE #2116: Resetting Multiple Stores Example Doesn't Work with the Currying Required Elsewhere for Typescript Support
Something is wrong with the documentation on how to reset all stores with respect to the currying needed for Typescript support:

https://docs.pmnd.rs/zustand/guides/how-to-reset-state

As is my existing code that has the extra currying () doesn't create the store properly using the example create() function. I was able to get around it by rewriting my code to not curry in my stores and to replicate the currying inside the custom create, which is actually probably better in my case because now I'm enforcing that my stores use persist and that they implement a reset function, but probably loses the extensibility desired by the example:

```
export const createPersistingStore = <T extends ResettableState>(
  f: StateCreator<T, [], [['zustand/persist', unknown]]>
) => {
  const store = _create<T>()(f);
  resetters.push(() => {
    store.getState().reset();
  });
  return store;
};
```

dbritto-dev:

Fixes #2116

## Summary

Update reset state docs

## Check List

### DIFF
--- a/docs/guides/how-to-reset-state.md
+++ b/docs/guides/how-to-reset-state.md
@@ -29,15 +29,12 @@ const initialState: State = {
 // create store
 const useSlice = create<State & Actions>()((set, get) => ({
   ...initialState,
-
   addSalmon: (qty: number) => {
     set({ salmon: get().salmon + qty })
   },
-
   addTuna: (qty: number) => {
     set({ tuna: get().tuna + qty })
   },
-
   reset: () => {
     set(initialState)
   },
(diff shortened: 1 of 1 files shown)


docs/guides/how-to-reset-state.md
```diff
@@ -47,33 +44,40 @@ const useSlice = create<State & Actions>()((set, get) => ({
 Resetting multiple stores at once
 
 ```ts
-import { create as _create, StateCreator } from 'zustand'
+import { create as _create, type StateCreator } from 'zustand'
```
sewera, line 47: This [verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) is not in Zustand's [tsconfig.json](https://github.com/pmndrs/zustand/blob/main/tsconfig.json).

Because in the documentation we follow Zustand's code style configuration, maybe splitting this import will do?

In Zustand split imports (import + import type) [are used](https://github.com/search?q=repo%3Apmndrs%2Fzustand+%2Fimport+%28%5BA-z%5D%2B%2C%29%3F+%5C%7B+.*type+%7Cimport+type%2F&type=code), e.g., in `src/middleware/immer.ts` or in `src/context.ts`.

```suggestion
import { create as _create } from 'zustand'
import type { StateCreator } from 'zustand'
```

docs/guides/how-to-reset-state.md
```diff
@@ -47,33 +44,40 @@ const useSlice = create<State & Actions>()((set, get) => ({
...
   }
-}
+}) as typeof _create
 ```
 
 Resetting bound store using Slices pattern
 
 ```ts
-import create, { StateCreator } from 'zustand'
+import create, { type StateCreator } from 'zustand'
```
sewera, line 72: Similar here:

```suggestion
import create from 'zustand'
import type { StateCreator } from 'zustand'
```

sewera (approved):
This is very nitpicky, but maybe you'll agree with me.

sewera (approved):

PR #2121, opened 2023-10-15
fix(middleware) remove warning if there is no window

bluelakee02:

Fixes # devtools middleware  running on next.js development (next dev command) 

## Summary
if you develop next.js app, you are getting constant flow of repeating message to install extensions in terminal
so I am proposing an early return, since devtools are useless if there is no window
image

## Check List

- [x ] `yarn run prettier` for formatting code and docs

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -141,6 +141,8 @@ const devtoolsImpl: DevtoolsImpl =
   (set, get, api) => {
     const { enabled, anonymousActionType, store, ...options } = devtoolsOptions
 
+    if (typeof window === 'undefined') return fn(set, get, api)
+
     type S = ReturnType<typeof fn> & {
       [store: string]: ReturnType<typeof fn>
     }


dai-shi:
Is it just about showing a warning in DEV mode? Do you think showing the warning is required if `window` is defined?

We have same warnings across zustand/jotai/valtio and we need three PRs at the same time. But, I would like to confirm the requirement and expectation before suggesting changes.

dai-shi:
On second thought,
```js
      if (import.meta.env?.MODE !== 'production' && enabled) {
        console.warn(
          '[zustand devtools middleware] Please install/enable Redux devtools extension'
        )
      }
```
The warning only shows when `enabled` is true.
Do you specify `enabled: true`? I think you don't want to specify it if you don't want to enable it explicitly if there can be no window. Or you can use `enabled: typeof window !== 'undefined'`

Closing this as it's an intended behavior.
We can discuss about the requirement further and re-open it.
(I actually wonder if we should remove the warning entirely.)

bluelakee02:
@dai-shi OK, sorry my bad.

PR #2137, opened 2023-10-21
fix(docs): support currying for testing mock

plrs9816:

[#905](https://github.com/pmndrs/zustand/issues/905#issuecomment-1499264087)

## Summary

Fix example jest/vitest mock on docs to handle currying

## Check List

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -91,7 +91,7 @@ export const create = (<T>() => {
   console.log('zustand create mock')
 
   // to support curried version of create
-  return typeof stateCreator === "function"
+  return typeof stateCreator === 'function'
     ? createUncurried(stateCreator)
     : createUncurried
 }) as typeof zustand.create
@@ -110,7 +110,7 @@ export const createStore = (<T>(stateCreator: zustand.StateCreator<T>) => {
   console.log('zustand createStore mock')
 
   // to support curried version of createStore
-  return typeof stateCreator === "function"
+  return typeof stateCreator === 'function'
     ? createStoreUncurried(stateCreator)
     : createStoreUncurried
 }) as typeof zustand.createStore
@@ -180,7 +180,7 @@ export const create = (<T>() => {
   console.log('zustand create mock')
 
   // to support curried version of create
-  return typeof stateCreator === "function"
+  return typeof stateCreator === 'function'
     ? createUncurried(stateCreator)
     : createUncurried
 }) as typeof zustand.create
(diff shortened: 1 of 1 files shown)


docs/guides/testing.md
```diff
@@ -77,30 +77,42 @@ const { create: actualCreate, createStore: actualCreateStore } =
 // a variable to hold reset functions for all stores declared in the app
 export const storeResetFns = new Set<() => void>()
 
+const createInternalFn = <T>(stateCreator: zustand.StateCreator<T>) => {
```
sewera, line 80: Consider renaming `createInternalFn` and `createStoreInternalFn`
to `createTestStore`, `createStoreUncurried`, or similar.

sewera (changes requested):

sewera (approved):

dai-shi:
@plrs9816 Please check the CI error.

dbritto-dev:
@plrs9816 try not to forget that these changes requires to update codesandbox demos too

dbritto-dev:
@dai-shi btw, I wouldn't merge these changes until v5 is ready but up to you

dai-shi:
> @dai-shi btw, I wouldn't merge these changes until v5 is ready but up to you

Why so? Doesn't it work on v4? v5 won't be ready any time soon. (Expect more than several months.)

dbritto-dev:
> > @dai-shi btw, I wouldn't merge these changes until v5 is ready but up to you
> 
> Why so? Doesn't it work on v4? v5 won't be ready any time soon. (Expect more than several months.)

@dai-shi do we want to support uncurried and curried versions of create/createStore on v5? IHMO, I prefer to support curried version because works well on both js and ts and we don't need to explain why curried version if the preferred way on ts but on js the preferred way is the uncurried version if you are not using middlewares. Furthermore, we can make the curried version the preferred way for both ts and js so people can easly migrate to ts from js; sometimes people get confuse by using the curried version on ts when they are migrating from js. In any case whatever you think is good is OK

BTW, I don't know if curried or uncurried version would be the only way to use create/createStore but would be good to have just one way to do the things and that way works well on both js and ts

dai-shi:
At this moment, we plan to support both curried and uncurried versions in v5.
I don't think we have a conclusion in #1865 yet.
On the other hand, I see people get confused with two ways.

> BTW, I don't know if curried or uncurried version would be the only way

Maybe, a new util like `applyMiddleware` in redux might help. The beauty of Zustand code is that it doesn't have any code for middleware, but maybe considerable. I guess it's too soon to do it in v5, no? I'm not even sure if that can avoid currying.

PR #2139, opened 2023-10-22
Improve types in Immer middleware docs

Brammm:

Fixes #2136

## Summary

Moves the generic type from the `immer` middleware call to `create`. If you use the code as mentioned previously in the docs, you lose IDE introspection on where Actions or State are being used. 

I also added an extra note on the extra parenthesis with a link to the Typescript guide, but if people feel that's superfluous, I can revert that commit.

## Check List

### DIFF
--- a/docs/integrations/immer-middleware.md
+++ b/docs/integrations/immer-middleware.md
@@ -34,8 +34,8 @@ type Actions = {
   decrement: (qty: number) => void
 }
 
-export const useCountStore = create(
-  immer<State & Actions>((set) => ({
+export const useCountStore = create<State & Actions>()(
+  immer((set) => ({
     count: 0,
     increment: (qty: number) =>
       set((state) => {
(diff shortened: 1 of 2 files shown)


docs/integrations/immer-middleware.md
```diff
@@ -19,6 +19,8 @@ npm install immer
 
 ## Usage
 
+(Note the extra parenthesis after the type parameter as mentioned in the [Typescript Guide](.../guides/typescript.md)).
```
sewera, line 22: This relative URL: `.../guides/typescript.md` is incorrect.
Two dots (`..`) move to the parent directory.

```suggestion
(Note the extra parenthesis after the type parameter as mentioned in the [Typescript Guide](../guides/typescript.md)).
```

sewera, line 22: The plural form of _parenthesis_ is _parentheses_, [as per Merriam-Webster Dictionary](https://www.merriam-webster.com/dictionary/parenthesis).

```suggestion
(Note the extra parentheses after the type parameter as mentioned in the [Typescript Guide](.../guides/typescript.md)).
```

sewera (changes requested):

docs/integrations/immer-middleware.md
Brammm: Woops, that was a typo, fixed.

Brammm: Ah, good point, fixed here and in typescript.md.

dbritto-dev (approved):

Brammm:
Looks like the builds failed during the github clone issue earlier today. What's the best way to rerun these?

dai-shi:
We could add an empty commit to trigger codesandbox ci, but as this is doc-only change. Let me force-merge it.

PR #2147, opened 2023-10-25
docs: update documentation of persist

cheatkey:
## Summary
Added example of using third-party libraries to the "How do I use it with Map and Set" section of the persist documentation

There's no need to write code to serialise or deserialise data that is not supported by JSON.stringify

## Check List

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -731,3 +731,32 @@ interface BearState {
     removeItem: (name) => localStorage.removeItem(name),
   },
 ```
+
+If writing serialisation and deserialisation code is tedious, you can use third-party libraries to serialise and deserialise different types of data.
+
+SuperJSON serialises data along with its type, allowing the data to be parsed back to its original type upon deserialisation
+
+```ts
+import superjson from "superjson";
+import { StorageValue } from "zustand/middleware";
+
+
+interface BearState {
+// ...
+}
+//...
+
+  storage: {
+    getItem: (name) => {
+      const str = localStorage.getItem(name);
+      if (!str) return null;
+      return {
+        state: superjson.parse<StorageValue<BearState>>(str).state,
+      };
+    },
+    setItem: (name, value) => {
+      localStorage.setItem(name, superjson.stringify(value));
+    },
+    removeItem: (name) => localStorage.removeItem(name),
+  }
+```


docs/integrations/persisting-store-data.md
```diff
@@ -731,3 +731,32 @@ interface BearState {
     removeItem: (name) => localStorage.removeItem(name),
   },
 ```
+
+If writing serialisation and deserialisation code is tedious, you can use third-party libraries to serialise and deserialise different types of data.
+
+SuperJSON serialises data along with its type, allowing the data to be parsed back to its original type upon deserialisation
```
sewera, line 737: It depends whether @dai-shi and other contributors want to encourage using another library for this use case in the official Zustand docs.

docs/integrations/persisting-store-data.md
```diff
@@ -731,3 +731,32 @@ interface BearState {
     removeItem: (name) => localStorage.removeItem(name),
   },
 ```
+
+If writing serialisation and deserialisation code is tedious, you can use third-party libraries to serialise and deserialise different types of data.
```
sewera, line 735: Please use the American English spelling — _serialization_ for this and other occurrences.

docs/integrations/persisting-store-data.md
```diff
@@ -731,3 +731,32 @@ interface BearState {
...
+
+SuperJSON serialises data along with its type, allowing the data to be parsed back to its original type upon deserialisation
+
+```ts
+import superjson from "superjson";
```
sewera, line 740: Please run prettier.

sewera (changes requested):
@dai-shi — please comment if you're okay with encouraging the usage of `superjson` external library.

docs/integrations/persisting-store-data.md
cheatkey: Should we change it to how to use persist with a custom serialization library? Like "How can I use a custom storage engine"?
스크린샷 2023-10-26 01-13-24

If you use a lot of variables like Map, Set, Date, etc. that JSON.stringify() doesn't support, it would be nice to have a way in the documentation to avoid writing your own serialization/deserialization code

No problem using other libraries (ex) [serialize-javascript](https://github.com/yahoo/serialize-javascript), [devalue](https://github.com/Rich-Harris/devalue) , ...etc)

dbritto-dev: @cheatkey I'm down with the idea of having this under "How can I use a custom storage engine?".

dai-shi:
> @dai-shi — please comment if you're okay with encouraging the usage of `superjson` external library.

I'm not familiar with superjson, and we shouldn't endorse a specific library, but a usage example may help people.

docs/integrations/persisting-store-data.md
```diff
@@ -731,3 +731,31 @@ interface BearState {
     removeItem: (name) => localStorage.removeItem(name),
   },
 ```
+
+If writing serialization and deserialization code is tedious, you can use third-party libraries to serialize and deserialize different types of data
+
+SuperJSON serialize data along with its type, allowing the data to be parsed back to its original type upon deserialization
```
dai-shi, line 737: So, how about adding this?
```suggestion
For example, [Superjson](https://github.com/blitz-js/superjson) can serialize data along with its type, allowing the data to be parsed back to its original type upon deserialization
```

dai-shi:
> @cheatkey I'm down with the idea of having this under "How can I use a custom storage engine?".

That sounds good too.

cheatkey:
I have committed the following changes
- Moved content below "How can I use a custom storage engine?"
- Specified that the code is an example
- Added the following types to the example code: Map, Set, Date, and RegExp

docs/integrations/persisting-store-data.md
```diff
@@ -630,6 +630,37 @@ export const useBoundStore = create(
...
+If you're using a type that JSON.stringify() doesn't support, you'll need to write your own serialization/deserialization code. However, if this is tedious, you can use third-party libraries to serialize and deserialize different types of data.
+
+For example, [Superjson](https://github.com/blitz-js/superjson) can serialize data along with its type, allowing the data to be parsed back to its original type upon deserialization
+
+```ts
+import superjson from "superjson"; //  can use anything: serialize-javascript, devalue, etc.
+import { StorageValue } from "zustand/middleware";
+
+interface BearState {
+  ...: Map<string, string>;
```
sewera, line 642: Instead of an ellipsis, use some example names, like "bears", "fish", "time", "query", or something like that.

docs/integrations/persisting-store-data.md
```diff
@@ -630,6 +630,37 @@ export const useBoundStore = create(
...
+  ...: Set<string>;
+  ...: Date,
+  ...: RegExp;
+}
+//...
+
+  storage: {
```
sewera, line 649: Wrap it with `const storage: StateStorage = {}` and add necessary imports.

docs/integrations/persisting-store-data.md
```diff
@@ -630,6 +630,37 @@ export const useBoundStore = create(
...
+    getItem: (name) => {
+      const str = localStorage.getItem(name);
+      if (!str) return null;
+      return {
+        state: superjson.parse<StorageValue<BearState>>(str).state,
```
sewera, line 654: I'm wondering if you need to return `{ state: superjson.parse().state }`, or if you could just return `superjson.parse().state`, without wrapping it in an object?

sewera (changes requested):

docs/integrations/persisting-store-data.md
cheatkey: You don't need to wrap it in an object. The result of `superjson.parse` returns the original json, so you can return the result of the `superjson.parse` function directly. 
I have updated the example code accordingly

sewera (approved):

Issue #1395, opened 2022-10-27
Zustand breaks with NextJS 13 when enabling appDir (Cannot read properties of null (reading 'useDebugValue'))


lucasmerlin:
I created a minimal reproduction here: 
https://stackblitz.com/edit/nextjs-egfcaj?file=package.json

It throws the following error: 
```bash
Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
TypeError: Cannot read properties of null (reading 'useDebugValue')
    at Module.useDebugValue (/home/projects/nextjs-egfcaj/node_modules/react/cjs/react.development.js:1659:23)
    at useStore (file:///home/projects/nextjs-egfcaj/node_modules/zustand/esm/index.mjs:37:10)
    at useBoundStore (file:///home/projects/nextjs-egfcaj/node_modules/zustand/esm/index.mjs:42:51)
    at __WEBPACK_DEFAULT_EXPORT__ (webpack-internal:///./pages/index.js:20:19)
```

If you remove the "appDir: true" line from next.config.js, the counter works as expected. The weirdest thing is that the project doesn't even use the app dir, just enabling the flag is enough for the error to happen.
After a bit of debugging in my project, comparing the useDebugValue that is imported by zustand and the one I imported in my component, they are not equal. Seems like the multiple react copies error, although there is only one react version installed.

I also tried zustand version 3, it throws a similar error, but with the useReducer hook instead of useDebugValue.
I also tried the version from here: https://github.com/pmndrs/zustand/issues/1392 but it didn't help.

Relevant NextJS discussion: https://github.com/vercel/next.js/discussions/41236

Sec-ant:
Does anyone like me think this might have something to do with the ["Dual Package Hazard"](https://nodejs.org/api/packages.html#:~:text=The%20reason%20that,to%2Dtroubleshoot%20bugs.) caused by different instances of a same dependency?

[...]

PR #2154, opened 2023-10-28
fix: importing CJS React in ESM

dai-shi:

Fixes https://github.com/pmndrs/zustand/issues/1395#issuecomment-1783619389

## Summary

To avoid errors when Zustand is used on server as ESM.

## Check List

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,14 +1,26 @@
-import {
-  createElement,
-  createContext as reactCreateContext,
-  useContext,
-  useMemo,
-  useRef,
-} from 'react'
+// import {
+//   createElement,
+//   createContext as reactCreateContext,
+//   useContext,
+//   useMemo,
+//   useRef,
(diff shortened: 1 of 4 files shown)


OlegLustenko:
@dai-shi it was incredibly fast!

So far, the fix looks good

tested:
- prod build Docker
- dev build local
- prod build local

However, I've got a weird issue in the dev build locally, 
I'm not sure how this could be related to the changes made

I've got this error
image

When I only have `page.tsx` without `layout.tsx` on the same level 

It wasn't the case for `4.4.4` version

Maybe this is a floating bug in Next.js  🤷‍♂️

OlegLustenko:
@dai-shi I would go ahead and merge it. Are there any concerns about pushing it forward?

dai-shi:
I'm not very confident if this causes some other unexpected issues, but let's proceed to know it.

PR #2157, opened 2023-10-28
Apply publint recommendations

dbritto-dev:
## Related Issues or Discussions

Publint warnings -> https://publint.dev/zustand@4.4.4

## Summary

Apply publint recommendations

## Check List

- [x] `yarn run prettier` for formatting code and docs

PR #2170, opened 2023-11-01
Update export types

dbritto-dev:

Fixes #2163

## Summary

The types for should be under each type of import. For instance "import", "module", "require" and "default"

## Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,32 +21,85 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./index.d.ts",
       "import": {
         "types": "./esm/index.d.mts",
         "default": "./esm/index.mjs"
       },
-      "module": {
-        "types": "./esm/index.d.ts",
-        "default": "./esm/index.js"
-      },
-      "default": {
-        "types": "./index.d.ts",
-        "default": "./index.js"
-      }
+      "module": "./esm/index.js",
+      "default": "./index.js"
     },
-    "./*": {
+    "./vanilla": {
+      "types": "./vanilla.d.ts",
       "import": {
-        "types": "./esm/*.d.mts",
-        "default": "./esm/*.mjs"
+        "types": "./esm/vanilla.d.mts",
+        "default": "./esm/vanilla.mjs"
       },
-      "module": {
-        "types": "./esm/*.d.ts",
-        "default": "./esm/*.js"
+      "module": "./esm/vanilla.js",
+      "default": "./vanilla.js"
+    },
+    "./middleware": {
+      "types": "./middleware.d.ts",
+      "import": {
+        "types": "./esm/middleware.d.mts",
+        "default": "./esm/middleware.mjs"
       },
-      "default": {
-        "types": "./*.d.ts",
-        "default": "./*.js"
-      }
+      "module": "./esm/middleware.js",
+      "default": "./middleware.js"
+    },
+    "./middleware/immer": {
+      "types": "./middleware/immer.d.ts",
+      "import": {
+        "types": "./esm/middleware/immer.d.mts",
+        "default": "./esm/middleware/immer.mjs"
+      },
+      "module": "./esm/middleware/immer.js",
+      "default": "./middleware/immer.js"
+    },
+    "./shallow": {
+      "types": "./shallow.d.ts",
+      "import": {
+        "types": "./esm/shallow.d.mts",
+        "default": "./esm/shallow.mjs"
+      },
+      "module": "./esm/shallow.js",
+      "default": "./shallow.js"
+    },
+    "./vanilla/shallow": {
+      "types": "./vanilla/shallow.d.ts",
+      "import": {
+        "types": "./esm/vanilla/shallow.d.mts",
+        "default": "./esm/vanilla/shallow.mjs"
+      },
+      "module": "./esm/vanilla/shallow.js",
+      "default": "./vanilla/shallow.js"
+    },
+    "./react/shallow": {
+      "types": "./react/shallow.d.ts",
+      "import": {
+        "types": "./esm/react/shallow.d.mts",
+        "default": "./esm/react/shallow.mjs"
+      },
+      "module": "./esm/react/shallow.js",
+      "default": "./react/shallow.js"
+    },
+    "./traditional": {
+      "types": "./traditional.d.ts",
+      "import": {
+        "types": "./esm/traditional.d.mts",
+        "default": "./esm/traditional.mjs"
+      },
+      "module": "./esm/traditional.js",
+      "default": "./traditional.js"
+    },
+    "./context": {
+      "types": "./context.d.ts",
+      "import": {
+        "types": "./esm/context.d.mts",
+        "default": "./esm/context.mjs"
+      },
+      "module": "./esm/context.js",
+      "default": "./context.js"
     }
   },
   "sideEffects": false,
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
-      "types": "./esm/index.d.ts",
+      "types": "./index.d.ts",
       "import": {
         "types": "./esm/index.d.mts",
         "default": "./esm/index.mjs"
--- a/package.json
+++ b/package.json
@@ -21,85 +21,130 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
-      "types": "./index.d.ts",
       "import": {
         "types": "./esm/index.d.mts",
         "default": "./esm/index.mjs"
       },
-      "module": "./esm/index.js",
-      "default": "./index.js"
+      "module": {
+        "types": "./esm/index.d.ts",
+        "default": "./esm/index.js"
+      },
+      "default": {
+        "types": "./index.d.ts",
+        "default": "./index.js"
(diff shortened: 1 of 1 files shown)


dbritto-dev:
@dai-shi here you go

dai-shi:
> ## Related Issues or Discussions
> Fixes #2164

Do you mean #2163 ?

package.json
```diff
@@ -21,12 +21,13 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./esm/index.d.ts",
```
dai-shi, line 24: Do you need this? Also, I just noticed that we didn't change other than ".".

dbritto-dev: For some reason esm types throw some errors, that set up the `index.d.ts` as default for everything but internally we use the ones that we had under types on each config

dbritto-dev:
> > ## Related Issues or Discussions
> 
> > Fixes #2164
> 
> 
> 
> Do you mean #2163 ?

Yeah, that one

dbritto-dev:
@dai-shi all yours

package.json
```diff
@@ -26,85 +26,27 @@
...
-      "import": {
-        "types": "./esm/react/shallow.d.mts",
-        "default": "./esm/react/shallow.mjs"
-      },
-      "module": "./esm/react/shallow.js",
-      "default": "./react/shallow.js"
-    },
-    "./traditional": {
-      "types": "./traditional.d.ts",
+    "./*": {
```
dai-shi, line 37: This requires newer version of Node, and we will be doing this change in v5.

package.json
```diff
@@ -26,85 +26,27 @@
         "default": "./esm/index.mjs"
       },
       "module": {
-        "types": "./index.d.ts",
+        "types": "./esm/index.d.ts",
```
dai-shi, line 29: Is this change the key to fix the issue? I mean the file content must be the same, so the file path was the root cause, right? Why wasn't it an issue before #2157?

It feels like reverting #2157 should be appropriate, if we **don't know** why it caused such an issue.

dai-shi (changes requested):

package.json
dbritto-dev: @dai-shi btw, I updated the exports to match jotai's and valtio's exports

dbritto-dev: @dai-shi but we can revert the changes

dai-shi: Yeah > https://github.com/pmndrs/zustand/pull/2170#discussion_r1379421218

package.json
dai-shi: @dbritto-dev We are waiting for the revert.

dbritto-dev: @dai-shi done

dai-shi: Ooops, sorry. I don't mean to revert #2157 fully.
Can you try what you think should work, and just avoid using `"./*"`.

dbritto-dev: In that we case we should apply the same change for each inner package until we can use the `./*`

dai-shi: Yes, that what I meant.

dbritto-dev: Sure, let me do it

dai-shi:
btw, @dbritto-dev do you understand why #2157 caused such an issue? It seems like TS picks different files for index and middleware, but I'm not sure how that could happen. It sounds like TS ignore "types" field, doesn't it?

dbritto-dev:
> btw, @dbritto-dev do you understand why #2157 caused such an issue? It seems like TS picks different files for index and middleware, but I'm not sure how that could happen. It sounds like TS ignore "types" field, doesn't it?

@dai-shi I understand the issue in the case of inner packages when you use "types" outside of "import", "module", "require" or "default" ts use by default that types ignoring where you are importing the packages when we added the new changes for the main package we fixed that but the inner packages still pointing to "d.ts" even if you are importing them from "esm" that's the conflict

dbritto-dev:
@dai-shi all yours

dai-shi:
> in the case of inner packages when you use "types" outside of "import", "module", "require" or "default" ts use by default that types ignoring where you are importing the packages;

So, the "types" line right under "exports" in v4.4.4 package.json was simply ignored? That explains it.

dbritto-dev:
@dai-shi would you mind taking care of these changes? I tested locally and works well

dai-shi:
Yes, it looks good.
I hope to get one or two more people to try it. https://github.com/pmndrs/zustand/discussions/2163#discussioncomment-7457555

dai-shi (approved):
The changes look good.

LarsArtmann:
Thank you for fixing my build :) 👍

sanduluca:
Was this released as I still have the issue with immer and `"zustand": "^5.0.3"`

image

dbritto-dev:
> Was this released as I still have the issue with immer and `"zustand": "^5.0.3"`
> 
> image

@sanduluca that doesn't look right to me, `addSound` method should be implemented keeping in mind `immer`

sanduluca:
In the image i was showing the that the immer is generating the error (before refactoring). 

But the point is still that it shows type errors with immer even with the example from docs.

image

<details><summary>Code and error message</summary>
<p>

```ts
type State = {
  count: number;
};

type Actions = {
  increment: (qty: number) => void;
  decrement: (qty: number) => void;
};

export const useCountStore = create<State & Actions>()(
  immer((set) => ({
    count: 0,
    increment: (qty: number) =>
      set((state) => {
        state.count += qty;
      }),
    decrement: (qty: number) =>
      set((state) => {
        state.count -= qty;
      }),
  }))
);
```
```bash
Argument of type 'StateCreator<State & Actions, [], [["zustand/immer", never]]>' is not assignable to parameter of type 'StateCreator<State & Actions, [], [never, unknown][]>'.
  Type 'StateCreator<State & Actions, [], [["zustand/immer", never]]>' is not assignable to type '{ $$storeMutators?: [never, unknown][] | undefined; }'.
    Types of property '$$storeMutators' are incompatible.
      Type '[["zustand/immer", never]] | undefined' is not assignable to type '[never, unknown][] | undefined'.
        Type '[["zustand/immer", never]]' is not assignable to type '[never, unknown][]'.
          Type '["zustand/immer", never]' is not assignable to type '[never, unknown]'.
            Type at position 0 in source is not compatible with type at position 0 in target.
              Type '"zustand/immer"' is not assignable to type 'never'.
```

image

</p>
</details>

dbritto-dev:
> In the image i was showing the that the immer is generating the error (before refactoring). 
> 
> 
> 
> But the point is still that it shows type errors with immer even with the example from docs.
> 
> 
> 
> image
> 
> 
> 
> <details><summary>Code, error message, package.json</summary>
> 
> <p>
> 
> 
> 
> ```ts
> 
> type State = {
> 
>   count: number;
> 
> };
> 
> 
> 
> type Actions = {
> 
>   increment: (qty: number) => void;
> 
>   decrement: (qty: number) => void;
> 
> };
> 
> 
> 
> export const useCountStore = create<State & Actions>()(
> 
>   immer((set) => ({
> 
>     count: 0,
> 
>     increment: (qty: number) =>
> 
>       set((state) => {
> 
>         state.count += qty;
> 
>       }),
> 
>     decrement: (qty: number) =>
> 
>       set((state) => {
> 
>         state.count -= qty;
> 
>       }),
> 
>   }))
> 
> );
> 
> ```
> 
> ```bash
> 
> Argument of type 'StateCreator<State & Actions, [], [["zustand/immer", never]]>' is not assignable to parameter of type 'StateCreator<State & Actions, [], [never, unknown][]>'.
> 
>   Type 'StateCreator<State & Actions, [], [["zustand/immer", never]]>' is not assignable to type '{ $$storeMutators?: [never, unknown][] | undefined; }'.
> 
>     Types of property '$$storeMutators' are incompatible.
> 
>       Type '[["zustand/immer", never]] | undefined' is not assignable to type '[never, unknown][] | undefined'.
> 
>         Type '[["zustand/immer", never]]' is not assignable to type '[never, unknown][]'.
> 
>           Type '["zustand/immer", never]' is not assignable to type '[never, unknown]'.
> 
>             Type at position 0 in source is not compatible with type at position 0 in target.
> 
>               Type '"zustand/immer"' is not assignable to type 'never'.
> 
> ```
> 
> 
> 
> image
> 
> 
> 
> </p>
> 
> </details> 

@sanduluca would you mind opening a discussion?

Ray5055:
Hi! Is there any updates on this matter? I'm facing similar issues however mine is with persist rather than immer

```
Argument of type 'StateCreator<AssetState, [], [["zustand/persist", AssetState]]>' is not assignable to parameter of type 'StateCreator<AssetState, [], [never, unknown][]>'.
  Type 'StateCreator<AssetState, [], [["zustand/persist", AssetState]]>' is not assignable to type '{ $$storeMutators?: [never, unknown][] | undefined; }'.
    Types of property '$$storeMutators' are incompatible.
      Type '[["zustand/persist", AssetState]] | undefined' is not assignable to type '[never, unknown][] | undefined'.
        Type '[["zustand/persist", AssetState]]' is not assignable to type '[never, unknown][]'.
          Type '["zustand/persist", AssetState]' is not assignable to type '[never, unknown]'.
            Type at position 0 in source is not compatible with type at position 0 in target.
              Type '"zustand/persist"' is not assignable to type 'never'.ts(2345)
```

PR #2193, opened 2023-11-13
refactor: revisit prettier config to be minimal

dai-shi:
The prettier config was defined back in some time, and prettier has been updated since then.
I think we should follow the default config as much as possible.


dai-shi:
@sandren Can you open a PR for jotai?
@Aslemammad Can you open a PR for valtio?

Aslemammad:
Hey @dai-shi, Yes, sure.

PR #2213, opened 2023-11-25
fix(vanilla): unexpected null state update behavior

wulimao49:
unexpected behavior when update state with [`null`]

[current logic here](https://github.com/pmndrs/zustand/blob/main/src/vanilla.ts#L76C28-L76C28)

the code above will cause something like
```typescript
const useStore = create<string | null>(() => 'default');

useStore.setState(null); // will actually run: Object.assign({}, 'default', null)

useStore.getState(); // get { "0": "d", "1": "e", "2": "f", "3": "a", "4": "u", "5": "l",  "6": "t" }
```

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -73,7 +73,7 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
     if (!Object.is(nextState, state)) {
       const previousState = state
       state =
-        replace ?? typeof nextState !== 'object'
+        replace ?? ( typeof nextState !== 'object' || nextState === null )
           ? (nextState as TState)
           : Object.assign({}, state, nextState)
       listeners.forEach((listener) => listener(state, previousState))
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -73,7 +73,7 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
     if (!Object.is(nextState, state)) {
       const previousState = state
       state =
-        replace ?? ( typeof nextState !== 'object' || nextState === null )
+        replace ?? (typeof nextState !== 'object' || nextState === null)
           ? (nextState as TState)
           : Object.assign({}, state, nextState)
       listeners.forEach((listener) => listener(state, previousState))
--- a/tests/vanilla/basic.test.ts
+++ b/tests/vanilla/basic.test.ts
@@ -112,6 +112,24 @@ it('can set the store without merging', () => {
   expect(getState()).toEqual({ b: 2 })
 })
 
+it('can set the object store to null', () => {
+  const { setState, getState } = createStore<{ a: number } | null>(() => ({
+    a: 1,
+  }))
+
+  setState(null)
+
+  expect(getState()).toEqual(null)
+})
+
+it('can set the non-object store to null', () => {
+  const { setState, getState } = createStore<string | null>(() => 'value')
+
+  setState(null)
+
+  expect(getState()).toEqual(null)
+})
+
 it('works with non-object state', () => {
   const store = createStore<number>(() => 1)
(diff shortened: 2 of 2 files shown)


src/vanilla.ts
```diff
@@ -73,7 +73,7 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
     if (!Object.is(nextState, state)) {
       const previousState = state
       state =
-        replace ?? typeof nextState !== 'object'
+        replace ?? ( typeof nextState !== 'object' || nextState === null )
```
dai-shi, line 76: Yeah, that makes sense.

dai-shi (commented):
Please check CI errors.

src/vanilla.ts
dbritto-dev, line 76: I would use `Object.prototype.toString` since typeof could return `object` in many cases. Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

```suggestion
        replace ?? Object.prototype.toString.call(nextState).slice(8, -1) === 'Object'
```

dbritto-dev (changes requested):

src/vanilla.ts
dai-shi: Sorry, but I don't prefer that. typeof should be good for our purpose and for edge cases, users can use `replace`.

dbritto-dev: @dai-shi in that case we need to add a new test case for this change

dai-shi: In any case, a test would be good. @wulimao49 would you be able to add one?

Wait, on second thought, this isn't an issue with `null` specifically.
Maybe, it's more reasonable to do this: `typeof state !== 'object' || state === null`
What do you guys think?

wulimao49: `typeof state !== 'object' || state === null` will make it impossible to set a object state to `null` except explicitly set `replace` to `true`.

both do it on `state` and `nextState` will have some kind of edge cases.

maybe we should make the merging logic configable? or just using `replace` is enough.

 yes i can set a test if we decide a certain solution on this.

dai-shi: Here's my decision. This doesn't cover edge cases. It's intentional.
Just pick either one please:
- `typeof nextState !== 'object' || nextState === null`
- `typeof state !== 'object' || state === null`

Which makes less error-prone?

dbritto-dev: @dai-shi in that case I'll go with the second one

dai-shi: @wulimao49 How about you?

wulimao49: @dai-shi i prefer the second one

dai-shi: Let's go with `typeof state !== 'object' || state === null`.
Can you create some test cases?

wulimao49: test cases added

src/vanilla.ts
```diff
@@ -73,7 +73,7 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
     if (!Object.is(nextState, state)) {
       const previousState = state
       state =
-        replace ?? typeof nextState !== 'object'
+        replace ?? (typeof nextState !== 'object' || nextState === null)
```
dai-shi, line 76: This is what we agreed, right?
```suggestion
        replace ?? (typeof state !== 'object' || state === null)
```

src/vanilla.ts
dai-shi: Yeah, that's too much. 99% users don't use this capability anyway.

I considered changing to `replace ?? defaultReplace`, and `const defalutReplace = typeof state !== 'object || state === null` on initialization. But, its capability is rather limited, and the behavior can be less predictable.

I think the current solution is a fair point.

PR #2236, opened 2023-12-09
[v5] breaking: make React 18 as minimal requirement

charkour:
## Related Issues or Discussions

- for #2138
- waiting on #2235 

## Check List

- [x] `yarn run prettier` for formatting code and docs

PR #2138, opened 2023-10-22
v5

dai-shi:
- [x] drop default exports #2238
- [x] drop deprecated features #2235
- [x] make React 18 as minimal requirement #2236
- [x] make use-sync-external-store a peer dependency #2237 #2301 #2302
- [x] make TS version requirement and tests in CI precise (follow Jotai) #2257
- [x] drop "module" condition (follow Jotai) #2270
- [x] use `"./*"` in package.json exports and change Node.js requirement #2272
- [x] drop UMD/SystemJS builds and simplify build configs #2287
- [x] drop getServerState #2300
- [x] separate react entry point #2303
- [x] drop es5 #2380 #2433
- [x] #2395
- [x] #2427
- [x] #2466
- [x] write v5 migration guide
- [x] #2580

PR #2237, opened 2023-12-09
[v5] breaking: make use-sync-external-store an optional peer dependency

charkour:
## Summary

- for #2138
- waiting on #2236

## Check List


package.json
```diff
@@ -251,22 +248,27 @@
...
     "shx": "^0.3.4",
     "typescript": "^5.3.2",
+    "use-sync-external-store": "^1.2.0",
     "vitest": "^0.34.6"
   },
   "peerDependencies": {
-    "immer": ">=9.0",
     "@types/react": ">=16.8",
-    "react": ">=16.8"
+    "immer": ">=9.0",
```
charkour, line 256: Yarn prefers to have these dependencies listed in alphabetical order which is why these changed.

package.json
```diff
@@ -251,22 +248,27 @@
...
     },
-    "@types/react": {
+    "immer": {
       "optional": true
     },
     "react": {
       "optional": true
+    },
+    "use-sync-external-store": {
+      "optional": true
```
charkour, line 271: I opted to make this optional because it is only used with `react` which is also optional.

dai-shi: Yes, that was my intention. I forgot react was already optional.

dai-shi (approved):
LGTM

dai-shi:
Oooops, we forgot to remove `use-sync-external-store` from react.ts.

PR #2235, opened 2023-12-09
[v5] breaking: drop deprecated features

charkour:
## Summary

- For #2138
- Waiting on #2238 

## Check List

- [x] `yarn run prettier` for formatting code and docs
- [x] Update documentation to remove references to dropped features

PR #2238, opened 2023-12-09
[v5] breaking: drop default exports

charkour:
## Summary

- For #2138

## Check List

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -87,11 +87,9 @@ function createCommonJSConfig(input, output, options) {
       esModule: false,
       outro: options.addModuleExport
         ? [
-            `module.exports = ${options.addModuleExport.default};`,
-            ...Object.entries(options.addModuleExport)
-              .filter(([key]) => key !== 'default')
-              .map(([key, value]) => `module.exports.${key} = ${value};`),
-            `exports.default = module.exports;`,
+            ...Object.entries(options.addModuleExport).map(
+              ([key, value]) => `module.exports.${key} = ${value};`,
+            ),
           ].join('\n')
         : '',
     },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -179,13 +179,12 @@ module.exports = function (args) {
     createCommonJSConfig(`src/${c}.ts`, `dist/${c}`, {
       addModuleExport: {
         index: {
-          default: 'react',
           create: 'create',
           useStore: 'useStore',
           createStore: 'vanilla.createStore',
         },
-        vanilla: { default: 'vanilla', createStore: 'createStore' },
-        shallow: { default: 'shallow', shallow: 'shallow$1' },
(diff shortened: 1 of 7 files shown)


rollup.config.js
```diff
@@ -87,11 +87,9 @@ function createCommonJSConfig(input, output, options) {
       esModule: false,
       outro: options.addModuleExport
         ? [
-            `module.exports = ${options.addModuleExport.default};`,
```
charkour, line 90: The changes in the rollup config result in a change in the `build/index.js` file. I'm confident the first line can be removed but less confident the last line can be. @dai-shi, please take a look and let me know what you find. 

```diff
- module.exports = react;
module.exports.create = create;
module.exports.useStore = useStore;
module.exports.createStore = vanilla.createStore;
- exports.default = module.exports;
```

src/context.ts
```diff
@@ -37,7 +37,7 @@ type WithoutCallSignature<T> = { [K in keyof T]: T[K] }
 /**
  * @deprecated Use `createStore` and `useStore` for context usage
  */
-function createContext<S extends StoreApi<unknown>>() {
+export function createContext<S extends StoreApi<unknown>>() {
```
charkour, line 40: This is a temporary change in order to keep this PR small that will be removed in #2235

src/shallow.ts
```diff
@@ -4,16 +4,4 @@ import { shallow } from './vanilla/shallow.ts'
 // export { shallow } from './vanilla/shallow.ts'
 // export { useShallow } from './react/shallow.ts'
```
dai-shi, line 5: Please use these lines.

src/context.ts
dai-shi: nice idea.

rollup.config.js
dai-shi: We don't need `addModuleExport` entirely. That was a workaround for default exports and a big motivation to drop it in v5.
Oh, this is hard without knowing the context.

dai-shi (changes requested):
well, i thought the first one would be easy, but it doesn't seem so.
please see the comments. let me know if it's unclear.

src/shallow.ts
charkour: I'll make this change. Originally, I hadn't used these lines to keep the PR small, but I will make this change.

rollup.config.js
charkour: Ah, that makes sense. I see why it can be removed now. I learned something new about CJS. Thanks!

src/shallow.ts
dai-shi: ah, sorry about the confusion.

charkour:
Daishi, I'll be away from my computer for awhile, but can review any more change requests you may have later today.

dai-shi (approved):
LGTM! It's a good first step!

PR #2276, opened 2023-12-29
Fix a code typo in persisting-store-data.md

reza55n:
The code inside the documentation is `storage` with no values, the correct form would be `storage: storage`

## Check List


docs/integrations/persisting-store-data.md
```diff
@@ -672,7 +672,7 @@ export const useBearStore = create<BearState>()(
     }),
     {
       name: 'food-storage',
-      storage,
+      storage: storage,
```
dai-shi, line 675: It's object shorthand. It think it's fairly common now.

PR #2277, opened 2023-12-29
feat: getInitialState

TkDodo:
This PR adds a new `getInitialState` method to the store api, which will always return the initial state that was used to create the store. The `persist` middleware also implements this method because it alters the initial state when restoring from a storage.

In React, this new method can then be used for `getServerSnaphot` in `useSyncExternalStore` to avoid avoid hydration mismatches when the state changes between the SSR and the first CSR. This could happen before when using the persist middleware, but can also happen whenever the store state was changed on the client only before completing the first CSR.

Fixes #1174

## Check List

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -271,8 +271,6 @@ const oldImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
     api,
   )
 
-  api.getServerState = () => configResult
-
   // a workaround to solve the issue of not storing rehydrated state in sync storage
   // the set(state) value would be later overridden with initial state by create()
   // to avoid this, we merge the state from localStorage into the initial state.
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -271,6 +271,8 @@ const oldImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
     api,
   )
 
+  api.getServerState = () => configResult
+
   // a workaround to solve the issue of not storing rehydrated state in sync storage
   // the set(state) value would be later overridden with initial state by create()
   // to avoid this, we merge the state from localStorage into the initial state.
@@ -424,6 +426,8 @@ const newImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
     get,
     api,
   )
+
+  api.getServerState = () => configResult
 
   // a workaround to solve the issue of not storing rehydrated state in sync storage
   // the set(state) value would be later overridden with initial state by create()
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -425,7 +425,7 @@ const newImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
     api,
   )
 
-  api.getServerState = () => configResult
+  api.getInitialState = () => configResult
 
   // a workaround to solve the issue of not storing rehydrated state in sync storage
   // the set(state) value would be later overridden with initial state by create()
--- a/src/react.ts
+++ b/src/react.ts
@@ -22,13 +22,19 @@ type ExtractState<S> = S extends { getState: () => infer T } ? T : never
 
 type ReadonlyStoreApi<T> = Pick<StoreApi<T>, 'getState' | 'subscribe'>
 
+type WithReact<S extends ReadonlyStoreApi<unknown>> = S & {
+  getServerState?: () => ExtractState<S>
+}
+
 let didWarnAboutEqualityFn = false
 
 const identity = <T>(arg: T): T => arg
 
-export function useStore<S extends StoreApi<unknown>>(api: S): ExtractState<S>
+export function useStore<S extends WithReact<StoreApi<unknown>>>(
+  api: S,
+): ExtractState<S>
 
-export function useStore<S extends StoreApi<unknown>, U>(
+export function useStore<S extends WithReact<StoreApi<unknown>>, U>(
   api: S,
   selector: (state: ExtractState<S>) => U,
 ): U
@@ -37,14 +43,14 @@ export function useStore<S extends StoreApi<unknown>, U>(
  * @deprecated Use `useStoreWithEqualityFn` from 'zustand/traditional'
  * https://github.com/pmndrs/zustand/discussions/1937
  */
-export function useStore<S extends StoreApi<unknown>, U>(
+export function useStore<S extends WithReact<StoreApi<unknown>>, U>(
   api: S,
   selector: (state: ExtractState<S>) => U,
   equalityFn: ((a: U, b: U) => boolean) | undefined,
 ): U
 
 export function useStore<TState, StateSlice>(
-  api: StoreApi<TState>,
+  api: WithReact<StoreApi<TState>>,
   selector: (state: TState) => StateSlice = identity as any,
   equalityFn?: (a: StateSlice, b: StateSlice) => boolean,
 ) {
@@ -69,7 +75,7 @@ export function useStore<TState, StateSlice>(
   return slice
 }
 
-export type UseBoundStore<S extends ReadonlyStoreApi<unknown>> = {
+export type UseBoundStore<S extends WithReact<ReadonlyStoreApi<unknown>>> = {
   (): ExtractState<S>
   <U>(selector: (state: ExtractState<S>) => U): U
   /**
--- a/src/react.ts
+++ b/src/react.ts
@@ -24,6 +24,8 @@ type ReadonlyStoreApi<T> = Pick<StoreApi<T>, 'getState' | 'subscribe'>
 
 let didWarnAboutEqualityFn = false
 
+const identity = <T>(arg: T): T => arg
+
 export function useStore<S extends StoreApi<unknown>>(api: S): ExtractState<S>
 
 export function useStore<S extends StoreApi<unknown>, U>(
@@ -43,7 +45,7 @@ export function useStore<S extends StoreApi<unknown>, U>(
 
 export function useStore<TState, StateSlice>(
   api: StoreApi<TState>,
-  selector: (state: TState) => StateSlice = api.getState as any,
+  selector: (state: TState) => StateSlice = identity as any,
   equalityFn?: (a: StateSlice, b: StateSlice) => boolean,
 ) {
   if (
--- a/src/react.ts
+++ b/src/react.ts
@@ -67,7 +67,7 @@ export function useStore<TState, StateSlice>(
   const slice = useSyncExternalStoreWithSelector(
     api.subscribe,
     api.getState,
-    api.getInitialState,
+    api.getServerState || api.getInitialState,
     selector,
     equalityFn,
   )
--- a/src/react.ts
+++ b/src/react.ts
@@ -104,6 +104,11 @@ const createImpl = <T>(createState: StateCreator<T, [], []>) => {
   const api =
     typeof createState === 'function' ? createStore(createState) : createState
 
+  if (!api.getServerState) {
+    const initialState = api.getState()
+    api.getServerState = () => initialState
+  }
+
   const useBoundStore: any = (selector?: any, equalityFn?: any) =>
     useStore(api, selector, equalityFn)
 
--- a/src/react.ts
+++ b/src/react.ts
@@ -22,17 +22,11 @@ type ExtractState<S> = S extends { getState: () => infer T } ? T : never
 
 type ReadonlyStoreApi<T> = Pick<StoreApi<T>, 'getState' | 'subscribe'>
 
-type WithReact<S extends ReadonlyStoreApi<unknown>> = S & {
-  getServerState?: () => ExtractState<S>
-}
-
 let didWarnAboutEqualityFn = false
 
-export function useStore<S extends WithReact<StoreApi<unknown>>>(
-  api: S,
-): ExtractState<S>
+export function useStore<S extends StoreApi<unknown>>(api: S): ExtractState<S>
 
-export function useStore<S extends WithReact<StoreApi<unknown>>, U>(
+export function useStore<S extends StoreApi<unknown>, U>(
   api: S,
   selector: (state: ExtractState<S>) => U,
 ): U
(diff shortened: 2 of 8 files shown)


src/react.ts
```diff
@@ -107,8 +107,13 @@ const createImpl = <T>(createState: StateCreator<T, [], []>) => {
       "[DEPRECATED] Passing a vanilla store will be unsupported in a future version. Instead use `import { useStore } from 'zustand'`.",
     )
   }
-  const api =
+  const api = (
     typeof createState === 'function' ? createStore(createState) : createState
+  ) as WithReact<StoreApi<T>>
```
TkDodo, line 112: @dai-shi I think `getServerState` should be required in `WithReact`:

https://github.com/pmndrs/zustand/blob/517524d2b05843c6a26d2713af08cc0ca1f62f3c/src/react.ts#L25-L27

but I'm not sure if that would be a breaking change ?

TkDodo:
okay this doesn't work - here's the reproduction depending on the preview build:

https://stackblitz.com/edit/stackblitz-starters-lxg5yy?file=app%2Fpage.tsx

I've traced it down to when we're calling

```
const initialState = api.getState()
```

the state already contains the hydrated one from localstorage. So the question is now - where do I best get access to capturing the actual initial state that was passed to `create` ?

debugging a bit more, it seems that the `persist` middleware needs to yield both the values from localstorage as well as the default value? I don't see `createStore` getting access to it when chaining `createStore(persist(...))` because `persist` basically "swallows" it ...

src/react.ts
dai-shi: Well, I think `WithReact` is for a different purpose. Anyway, the thing is that we don't want to attach `getServerState` to vanilla `api`, so that's why it doesn't require here.

src/react.ts
```diff
@@ -107,8 +107,13 @@ const createImpl = <T>(createState: StateCreator<T, [], []>) => {
...
+
+  const defaultState = api.getState()
+
+  api.getServerState = () => defaultState
```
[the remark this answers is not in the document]
TkDodo: it doesn't work because `getState()`, even at this point, already returns the state from localstorage. see:
- https://github.com/pmndrs/zustand/pull/2277#issuecomment-1871965063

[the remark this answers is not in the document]
dai-shi: oooops, I missed that comment.

dai-shi:
> the state already contains the hydrated one from localstorage.

Yeah, middleware can modify state during initialization.
I was thinking something like this:
```js
const initialState = ...
const useFooStore = create(persist((set) => ({
  ...initialState
}), { ... }))
useFooStore.getServerState = () => initialState
```

dai-shi:
Maybe, this would be more developer friendly:
```js

const useFooStore = create(
  (set) => ({ ... }),
  () => ({ ... }), // for server state
)
```

TkDodo:
But both approaches would be something that must be done in user-land, right?

if the persist middleware would somehow yield the serverState (or initialState), we could use that in the react adapter. I'm thinking that this should work ootb because without getServerSnapshot implemented, any state change between ssr and csr would trigger a hydration warning - this is independent of the persist middleware

dai-shi:
Yeah, so if we need ootb solution, vanilla store should know about react convention. It's reasonable as most Zustand users are React users.

If that's the case, we can simply define optional `getServerState` in api in vanilla.ts. It's up to middleware how to implement it.

dai-shi:
> we can simply define optional getServerState in api in vanilla.ts.

Let me add a commit.

TkDodo:
thanks @dai-shi 🙏 

I've added one more commit where the persist middleware actually sets `getServerState`: 6d046fe5

However, I still think we should bring back setting `getServerState` in the react `createImpl` in case `getServerState` hasn't been set already to avoid hydration errors from other updates that could happen to the store in-between (that are unrelated to persist middleware). What do you think?

src/middleware/persist.ts
```diff
@@ -271,6 +271,8 @@ const oldImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
     api,
   )
 
+  api.getServerState = () => configResult
+
```
TkDodo, line 275: not sure if this is also needed in `oldImpl` ?

src/react.ts
```diff
@@ -110,6 +104,11 @@ const createImpl = <T>(createState: StateCreator<T, [], []>) => {
   const api =
     typeof createState === 'function' ? createStore(createState) : createState
 
+  if (!api.getServerState) {
+    const initialState = api.getState()
+    api.getServerState = () => initialState
+  }
+
```
TkDodo, line 111: this is what I had in mind with my comment here:

- https://github.com/pmndrs/zustand/pull/2277#issuecomment-1872232908

TkDodo:
I'm still getting hydration errors ([latest stackblitz fork](https://stackblitz.com/edit/stackblitz-starters-xasjkd?file=package.json)), but I don't really understand it.

error is:
> Warning: Text content did not match. Server: "hydration-default" Client: "hydration-local"

debugging this at the `useSyncExternalStore` call however leads to this:

Screenshot 2023-12-29 at 18 32 04

`getServerState` returns the correct server value that we pass to `getServerSnapshot`, which means it should be used by react for the server render and the first client render.

Not really sure how this can still show a hydration error 🤔

src/middleware/persist.ts
dai-shi: We don't modify oldImpl. Just changing newImpl is fine.

src/react.ts
dai-shi: I don't think this helps, as you said `initialState` is not really "initial". We shouldn't modify `api` here anyways.

src/middleware/persist.ts
TkDodo: reverted here: 9c7b13e3

src/react.ts
TkDodo: so my thinking was that without this, we would get hydration errors on a simple manual state restore from somewhere else, like:

```
  bears: 0,
  increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
  removeAllBears: () => set({ bears: 0 }),
}))

// only on client
if (typeof window !== 'undefined') {
  const bears = window.localStorage.getItem('bears')
  useStore.setState(state => ({ ... state, bears }))
}
```

by "capturing" the initial state of `bears:0`, which was used to do the server-side rendering, we make sure that the first client side render will also render with `bears:0`.

The `persist` middleware has to do this too because otherwise, at this point, we don't know the initial state anymore. But doing it here also seems correct. Yes, when someone writes code like this, we could point them to the persist middleware, but the same thing could happen with a fast async restore (e.g. `localforage`) or really just calling `setState` on the client with something else (like a window global that was populated by the server or a third party script ...)

TkDodo:
found out what the issue is. It's because we're using `useSyncExternalStoreWithSelector`, and the default selector is `api.getState`, which will always yield the client state:

https://github.com/TkDodo/zustand/blob/2d868c68abfd76f9274135ba384394415d7f0905/src/react.ts#L46

I think the selector should just be an identity function, as it gets the snapshot passed (which is either the result from `api.getState` _or_ `api.getServerState`.

I think this should fix it: a985abe7 🤞

TkDodo:
yep, here is a working fork without hydration errors 🎉 

https://stackblitz.com/edit/stackblitz-starters-y74eij?file=app%2Fpage.tsx

dai-shi:
> I think this should fix it: [a985abe](https://github.com/pmndrs/zustand/commit/a985abe706ebdeb26c33bae83c548ce117848c18) 🤞

Nice. It's historical code and I have been wondering if we should use identity function.

TkDodo:
should I also change `traditional.ts` or is this also deprecated and shouldn't be touched ?

src/react.ts
dai-shi: Hmmmm, okay, if that's the case, I'd probably do it in vanilla and rename it to `getInitialState`. But, I'm not 100% sure if we go with it. Let me add a commit and see how it feels.

dai-shi:
> should I also change `traditional.ts` or is this also deprecated and shouldn't be touched ?

It's not deprecated. Let me work on it in a moment.

dai-shi (commented):
The code looks pretty good and feels comfortable.
Thanks so much for working on this. I think I finally understand getServerSnapshot in uSES.

dai-shi:
What's remaining from my perspective:
- [ ] check if the change doesn't break your example.
- [ ] update PR title and description.
- [ ] write minimal docs at least.

dbritto-dev:
These changes looks good I've been thinking on updating the testing guide since now we have a way to get the initial state.

dbritto-dev:
> These changes looks good I've been thinking on updating the testing guide since now we have a way to get the initial state.

BTW, if you need any help with this PR just ping me I would be happy to help.

charkour:
Hi @TkDodo and @dai-shi,

Thanks for the work on this PR. After doing my best to re-build my persist HTML mismatch reproduction using the latest commit and I still get the hydration error. It's possible the way I was loading the `zustand` package isn't working properly, but here is the reproduction for you to play around with. Thank you! I apologize if I'm confused. 

[Reproduction on StackBlitz](https://stackblitz.com/edit/stackblitz-starters-yxevax?file=package.json)
[Reproduction on Code Sandbox](https://codesandbox.io/p/devbox/next-js-x2ym23)

charkour:
To follow up, I actually think this fix _is_ working, but StackBlitz had cached the zustand 4.4.7. Sorry for the confusion!

Something went wrong with the Code Sandbox link, so I'll update that.

charkour:
While this change fixes the hydration mismatch error, it does introduce a flicker of the default state before initializing the persisted state which I would say is not ideal. My position is that the hydration mismatch error is a result of the zustand user not realizing `localStorage` isn't accessible on the server, but there are cases where the server and client both have access to persisted state (if the state is stored in an external database for example).

charkour:
Ah, the hydration issue still occurs when using the [Next.js /pages router](https://codesandbox.io/p/devbox/next-js-persist-forked-3cn2vz) but not the [Next.js /app router](https://codesandbox.io/p/devbox/next-js-persist-6n55r8).

I hope these reproductions help you during the testing! Thank you

TkDodo:
> it does introduce a flicker of the default state before initializing the persisted state which I would say is not ideal.

This is the same with the hydration error, because the generated html on the server will not match the localstorage state. If you want to avoid the flicker, the only chance is to block server side rendering completely. That is something the user can do right now, and react will get a better api for this in the future, the `postpone` api (see: https://x.com/acdlite/status/1741164234627563544) that will integrate with suspense boundaries.

But from the library perspective, if the user chooses to render the zustand state on the server and restore with the persist plugin on the client, the correct thing to do is to implement `getServerSnapshot`.

TkDodo:
@charkour the stackblitz reproduction you posted works for me without hydration mismatch - maybe you can also see this on this link?

https://stackblitzstartersyxevax-hofw--3000--a2aabdd9.local-credentialless.webcontainer.io/

Maybe you didn't `npm install` correctly or stackblitz didn't pick up the lib change correctly - but it works :)

TkDodo:
> Ah, the hydration issue still occurs when using the [Next.js /pages router](https://codesandbox.io/p/devbox/next-js-persist-forked-3cn2vz) but not the [Next.js /app router](https://codesandbox.io/p/devbox/next-js-persist-6n55r8).

Not sure what's up with the pages example, but when I inspect the source code of zustand, I can see the version _without_ this fix, while the app dir example has the correct version. There shouldn't be any difference between pages and app directory as far as this PR is concerned ...

TkDodo:
> What's remaining from my perspective:
> 
> * [ ]  check if the change doesn't break your example.
> * [ ]  update PR title and description.
> * [ ]  write minimal docs at least.

@dai-shi I can confirm that your changes are good, and I updated the PR description.

If someone can help with the docs, I'd appreciate it. Also, I think some tests would be good, especially for SSR. I think a minimal way how to trigger hydration warnings could be done with [this example](https://github.com/pmndrs/zustand/pull/2277#discussion_r1438561618), and another one where the persist plugin is used might also be valuable.

I probably won't have the time to add them this week. If someone wants to go ahead please let me know, otherwise I'll take a stab next week.

charkour:
Thanks for looking into the reproductions. Your explanation of the expected usage makes sense.

dai-shi:
So, it's working as expected? Only the remaining is docs, right?

charkour:
Yes, I believe it's working as expected, and docs should be written on how to avoid flickering if that's the user's requirement.

dbritto-dev:
Hi folks, this week I'll work on the docs

sewera:
I can review them, just ping me when you open a PR :)

TkDodo:
Sorry for letting this slide for so long. I added a test for hydration mismatches: 

28c45272

I verified that this test fails on main because the following errors is logged to the console:

```
Warning: Text content did not match. Server: "0" Client: "1"
    at div
    at Component (~/zustand/tests/ssr.test.tsx:93:23)
    at Suspense
Error: Text content does not match server-rendered HTML.
```

And there is no error no this PR 🎉 

Imo, this PR is good to go. I don't really know what docs changes would be needed, but I think @dbritto-dev will cover this area?

dbritto-dev:
@TkDodo yeah, I'll take care of docs

dbritto-dev:
@dai-shi @TkDodo we're good to go and merge these changes. I'll take of docs after that.

dbritto-dev:
@dai-shi sure

dai-shi:
haha, I thought you had some ideas.

As we discussed elsewhere, it might be nice if we had API docs. For now, it's `readme.md`. So, I will take care of readme in this PR.

What you can consider is something in docs. I think you have two options.
- Add a note about this new feature in `persist` doc.
- Create a new guide about SSR/Hydration (and maybe Next.js)

dbritto-dev:
@dai-shi I have some ideas just to know if you would like something in specific

dai-shi (approved):
This is a great improvement! Nice work @TkDodo and all!

PR #2294, opened 2024-01-16
Export the immer middleware like other middlewares

crazyurus:
## Summary

The middleware built into zustand is exported via

```ts
```

However, only the immer middleware needs to be exported via

```ts
import { immer } from 'zustand/middleware/immer'
```

I think a unified export method is better.  So this PR changes the import method of immer to the following

```ts
import { immer } from 'zustand/middleware'
```

At the same time, the existing import method is retained to make it compatible.

## Check List


dai-shi:
Unfortunately, it requires people who are not using immer to install immer with some bundlers. so, it's not acceptable.

PR #2300, opened 2024-01-20
[v5]: Drop `getServerState`

charkour:

For #2138

## Summary

After some discussion, we will drop `getServerState` in v5 because it is inaccessible from the userland, and `getInitialState` solves the problem. 

Please let me know if any updates should be made. Thank you

## Check List


dai-shi (approved):
LGTM! Thanks!

PR #2301, opened 2024-01-20
[v5]: do not depend on use-sync-external-store

dai-shi:
for #2138

Just forgot it in https://github.com/pmndrs/zustand/pull/2237#issuecomment-1901773891


charkour (approved):
Thanks for catching this. 

We need to keep use-sync-external-store in `traditional.ts` to support the selector API, correct?

dai-shi:
Yes, not for selector but for equalityFn.

dai-shi:
> Warning: The result of getServerSnapshot should be cached to avoid an infinite loop

I see...

dai-shi:
image

PR #2302, opened 2024-01-20
[v5]: refactor useMemoSelector

dai-shi:
for #2138 

the previous version has an issue with a corner case.


charkour:
What was the corner case that this solved? Would it be helpful to add a test to ensure it does not regress?

dai-shi:
Yes, I added: 97ff407
See also the commit history and find the test fails without the change.

charkour:
Thanks!

PR #2304, opened 2024-01-21
Update readme.md

nagendraallam:
Fixes #
- Updated readme.md to center status + updated colors


dai-shi:
Thanks for the suggestion, but the color is intentional and I don't think centering isn't very important. (I'd prefer left align.)

PR #2332, opened 2024-02-14
Relax types for `persist` middleware

rafaeelaudibert:
Right now, we're requiring new `StateStorage` implementations to forcefully return `void` from inside `setItem` and `removeItem`.

When using a library that returns the set value for `setItem` or `removeItem`, therefore returning  a `string`, it causes Typescript to fail, requiring some weird workarounds.

For example, when using `localforage` (https://github.com/localForage/localForage) one needs to do what we describe below. Notice the `void` keyword.

```typescript
const storage: StateStorage = {
   ...localForage,
   setItem: void localForage.setItem.bind(localForage)
}
```

Another, longer, alternative is

```typescript
const storage: StateStorage = {
   ...localForage,
   // Curly braces are required because we need to "return `void`"
   setItem: (name, value) => {
       localStorage.setItem(name, value)
  }
}
```

By changing the type implementation to ignore types - using unknown - we can simply use `localforage` - and similar libraries - as if we were using `window.localStorage`

## Check List

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -6,8 +6,8 @@ import type {
 
 export interface StateStorage {
   getItem: (name: string) => string | null | Promise<string | null>
-  setItem: (name: string, value: string) => void | Promise<void>
-  removeItem: (name: string) => void | Promise<void>
+  setItem: (name: string, value: string) => unknown | Promise<unknown>
+  removeItem: (name: string) => unknown | Promise<unknown>
 }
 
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
The change looks okay to me.

Does anyone want to give another review?

dbritto-dev:
LGTM

gideonkipkorir123:
This expression is not callable.
  Type 'StoreApi<UserStoreState>' has no call signatures.

PR #2360, opened 2024-02-26
fix: return type of persist migrate function changed to PersistedState

bobdepreziz:

Fixes #
Return type of migrate function  was not correct, should be PersisedState instead of S

## Check List


dai-shi (commented):
I think this is correct, but not 100% sure.
Can anyone review?

dai-shi (approved):
should be good.

dbritto-dev (approved):
LGTM

PR #2365, opened 2024-02-28
refactor: Switch to Object.hasOwn

onlined:
## Summary

We have `Object.hasOwn` and it looks more readable and intuitive than `Object.prototype.hasOwnProperty.call`.

## Check List

### DIFF
--- a/src/vanilla/shallow.ts
+++ b/src/vanilla/shallow.ts
@@ -39,7 +39,7 @@ export function shallow<T>(objA: T, objB: T) {
   }
   for (let i = 0; i < keysA.length; i++) {
     if (
-      !Object.prototype.hasOwnProperty.call(objB, keysA[i] as string) ||
+      !Object.hasOwn(objB, keysA[i] as string) ||
       !Object.is(objA[keysA[i] as keyof T], objB[keysA[i] as keyof T])
     ) {
       return false


dbritto-dev:
Since we are planning to use es2018 I don't know if we can continue with this change

onlined:
Why ES2018 but not ES2022? Is there a target browser compatibility ratio?

src/vanilla/shallow.ts
```diff
@@ -39,7 +39,7 @@ export function shallow<T>(objA: T, objB: T) {
   }
   for (let i = 0; i < keysA.length; i++) {
     if (
-      !Object.prototype.hasOwnProperty.call(objB, keysA[i] as string) ||
+      !Object.hasOwn(objB, keysA[i] as string) ||
```
dai-shi, line 42: I generally like the new syntax, but technically, we still support ie11.
https://github.com/pmndrs/zustand/blob/3c6dc2d12b0d2a2d5f403fe6ddad1053055ab72b/rollup.config.js#L107

So, let's make this change to v5 instead of v4. ie11 will break but it's probably okay.

dai-shi (changes requested):
change the base branch.

src/vanilla/shallow.ts
dbritto-dev: I'm down with that this should be targeted to v5

dbritto-dev:
@onlined would you mind creating a PR from fresh v5 branch to v5 branch instead of main branch?

src/vanilla/shallow.ts
dbritto-dev: @dai-shi @onlined in any case we need to keep in mind this -> https://caniuse.com/?search=hasOwn

onlined: Updated base branch.

dai-shi (approved):
Thanks.

I actually wonder if we should skip inherited properties. If we should compare all properties, `for...in` and `in` operator would do.

onlined:
@dai-shi I think it's more correct to check if the own properties and the prototypes are equal. Diving deep into inherited properties could cause unintuitive results.

dai-shi:
> Diving deep into inherited properties could cause unintuitive results.

Gotcha.

> check if the own properties and the prototypes are equal.

I assume we normally use plain objects. Maybe we can clarify in docs.

PR #2366, opened 2024-02-28
Enable usage of Corepack

onlined:
## Summary

Corepack chooses package manager to use with the field `packageManager`. For users using Yarn through Corepack, this is a real convenience.

## Check List


dai-shi (approved):
(I'd like to migrate to pnpm eventually for all z/j/v.)

PR #2374, opened 2024-02-29
refactor: Use for..of instead of C-style for

onlined:
## Check List

### DIFF
--- a/src/vanilla/shallow.ts
+++ b/src/vanilla/shallow.ts
@@ -37,10 +37,10 @@ export function shallow<T>(objA: T, objB: T) {
   if (keysA.length !== Object.keys(objB).length) {
     return false
   }
-  for (let i = 0; i < keysA.length; i++) {
+  for (const keyA of keysA) {
     if (
-      !Object.prototype.hasOwnProperty.call(objB, keysA[i] as string) ||
-      !Object.is(objA[keysA[i] as keyof T], objB[keysA[i] as keyof T])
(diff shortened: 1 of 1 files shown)


src/vanilla/shallow.ts
```diff
@@ -37,10 +37,10 @@ export function shallow<T>(objA: T, objB: T) {
   if (keysA.length !== Object.keys(objB).length) {
     return false
   }
-  for (let i = 0; i < keysA.length; i++) {
+  for (const keyA of keysA) {
```
dai-shi, line 40: We generally don't use for-of.
Oh, but we already use it in L28! Failed to review back then. Maybe we should change L28 and L17 with for-let.
And, we can use for-of from v5.
What do you think? @dbritto-dev

onlined: Why don't you use it? It's great for iterables. 

Why from v5? Is it IE11 again?

dbritto-dev: @dai-shi yeah, same code but let instead of const. btw, we can merge it on v4 and then on v5

dai-shi: Yes, we try _best_ to make v4 work on ie11.
Note that es2018 is esm target, but for cjs we target es5 maybe. I'm not sure if everything is transpiled into es5.

dai-shi: Wait, `for-of` is properly transpiled. So, this is fine.

dai-shi (approved):
for-of is transpiled into C-style for. So, this is okay.

PR #2395, opened 2024-03-08
[v5]: follow React "standard" way with breaking behavioral change

dai-shi:
For #2138 

I think useMemoSelector is a hack that should be avoided. We should generally follow how React works. If we need previous behavior, we should use `zustand/traditional` which is based on a library provided by React, instead of adding our own hack.

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -474,7 +474,7 @@ it('can set the store without merging', () => {
 
 it('only calls selectors when necessary', async () => {
   type State = { a: number; b: number }
-  const useBoundStore = create<State>(() => ({ a: 0, b: 0 }))
+  const useBoundStore = createWithEqualityFn<State>(() => ({ a: 0, b: 0 }))
   const { setState } = useBoundStore
   let inlineSelectorCallCount = 0
   let staticSelectorCallCount = 0
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -472,7 +472,45 @@ it('can set the store without merging', () => {
   expect(getState()).toEqual({ b: 2 })
 })
 
-it('only calls selectors when necessary', async () => {
+it('only calls selectors when necessary with static selector', async () => {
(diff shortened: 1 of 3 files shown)


tests/ssr.test.tsx
```diff
@@ -19,11 +19,9 @@ const useBearStore = create<BearStoreState & BearStoreAction>((set) => ({
 }))
 
 function Counter() {
-  const { bears, increasePopulation } = useBearStore(
-    ({ bears, increasePopulation }) => ({
-      bears,
-      increasePopulation,
-    }),
```
dai-shi, line 26: I would say it worked before by chance. This selector creates a new object every time, which isn't supported in v5. We should use a memoized selector, `zutand/traditional` or 👇 .

tests/basic.test.tsx
```diff
@@ -474,7 +474,7 @@ it('can set the store without merging', () => {
 
 it('only calls selectors when necessary', async () => {
   type State = { a: number; b: number }
-  const useBoundStore = create<State>(() => ({ a: 0, b: 0 }))
+  const useBoundStore = createWithEqualityFn<State>(() => ({ a: 0, b: 0 }))
```
dai-shi, line 477: This behavior isn't supported in v5. Use `zustand/traditional`.

dai-shi:
While this change may surprise users, v5 should live long with React 19+, so I believe this is a good move. And, major bump is only the chance to make this kind of changes.

tests/ssr.test.tsx
dai-shi: Or, was this test intentional to check the behavior with such selectors? @TkDodo

TkDodo: Hmm I think I copied this way if creating a store from somewhere. Definitely not intentional

dai-shi: Thanks for confirming. That was the reason I introduced `useMemoSelector`, without looking at the test code. I believe it is an unnecessary hack.

PR #2397, opened 2024-03-08
feat(slices) add slice module for creating namespaced slices

weeksie:
I use Zustand for a fairly large project with many different slices where using the currently documented approach would be too cumbersome and lead to possible clashes between slice properties. This approach, which I've been using for a while, has been effective at keeping the slices separate, however after upgrading Zustand I updated my slice code to fix some bugs in my implementation and to work better with the `devtools` middleware.

I figured I'd share/start a convo here if anybody was interested in this approach

## Summary
- New `slice` module for creating namespaced slices (somewhat similar to redux's `combineReducers`
- Exported `CreateSlice` type for defining slices
- Exported `slice` factory function for creating slices and attaching them to the main store
- Slice `set` function has different API than the default, removing the `replace` parameter in lieu of an optional `name` string
- An example gist: https://gist.github.com/weeksie/46a7e5fa1ddc481917155be856018dfd

Here's a screenshot of devtools from a project that uses a pretty big/complex store. I had only added name strings to a few of the existing calls, but even the default name is nice because they're scoped to the slice namespace.

image

Anyway, I just want to say thanks for creating Zustand, it's been a wonderful addition to my workflow and a delight to use.

## Check List

### DIFF
--- a/src/slice.ts
+++ b/src/slice.ts
@@ -1,62 +1,69 @@
-import { type StoreApi } from './vanilla';
-import { type NamedSet } from './middleware';
+import { type StoreApi } from './vanilla'
+import { type NamedSet } from './middleware'
 
-type FirstSetStateArg<T> = Parameters<StoreApi<T>['setState']>[0];
-export type SetState<T> = (partial: FirstSetStateArg<T>, name?: string) => void;
-export type GetState<T> = StoreApi<T>['getState'];
+type FirstSetStateArg<T> = Parameters<StoreApi<T>['setState']>[0]
+export type SetState<T> = (partial: FirstSetStateArg<T>, name?: string) => void
+export type GetState<T> = StoreApi<T>['getState']
 
-type StateObject = Record<string, any>;
+type StateObject = Record<string, any>
 
 type SliceKey<S extends StateObject> = {
-  [K in keyof S]: S[K] extends S ? K : never;
-}[keyof StateObject];
+  [K in keyof S]: S[K] extends S ? K : never
+}[keyof StateObject]
 
 export const sliceGet = <S extends StateObject, K extends SliceKey<S>>(
   get: GetState<S>,
-  key: K
+  key: K,
 ): GetState<S[K]> => {
-  return () => get()[key];
-};
+  return () => get()[key]
+}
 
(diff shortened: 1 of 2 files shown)


src/slice.ts
```diff
@@ -0,0 +1,69 @@
...
+export const sliceSet = <S extends StateObject, K extends SliceKey<S>>(
+  set: NamedSet<S>,
+  key: K,
+): SetState<S[K]> => {
+  /**
+   * differs from global set in that the `replace` option is removed in
+   * favor of a `name` optional arg.
+   */
+  return (partial: FirstSetStateArg<S[K]>, name?: string) => {
+    const fullName = `${key}.${name ?? 'set'}`
```
weeksie, line 30: I clearly missed the [docs](https://github.com/pmndrs/zustand?tab=readme-ov-file#logging-actions) on logging a full object to `devtools`. 

Perhaps something more like:

```typescript
    const type = `${key}.${name ?? 'set'}`;

    set(
      (state) => {
        const nextState =
          typeof partial === 'function'
            ? (partial as (s: S[K]) => S[K])(state[key])
            : partial;

        return {
          [key]: {
            ...state[key],
            ...nextState
          }
        } as Partial<S>;
      },
      false,
      { type, payload: partial },
```

That does limit the flexibility of the log, if for instance someone wanted to send something other than the change set as the payload.

dai-shi:
Thanks for your suggestion!
Our intention is to keep the core minimal and add more features in its ecosystem.
Would you be interested in creating `zustand-slice` or something?
I can invite you to https://github.com/zustandjs.

This approach seems new to me, but looks interesting.
It may or may not help, but see some old discussions: https://github.com/pmndrs/zustand/issues/178

This won't be merged, but I'll let it open for a while for some visibility.

pinkynrg:
This feature holds immense potential! Do you have any insights on its potential timeline for implementation, or whether it will indeed be implemented?

dai-shi:
> I can invite you to https://github.com/zustandjs.

Sent it.

> I'll publish as a separate package

It will be great!

fyi, recently, I tried something else for a slice pattern without namespacing. Here's my take: https://githhub.com/zustandjs/zustand-slices

PR #2404, opened 2024-03-12
Fix devtools warning in CLI

m-bp:

n/a

## Summary

The warning always shows in the CLI. This fixes it.

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -158,7 +158,11 @@ const devtoolsImpl: DevtoolsImpl =
     }
 
     if (!extensionConnector) {
-      if (import.meta.env?.MODE !== 'production' && enabled) {
+      if (
+        typeof window !== 'undefined' &&
+        import.meta.env?.MODE !== 'production' &&
+        enabled
+      ) {
         console.warn(
           '[zustand devtools middleware] Please install/enable Redux devtools extension',
         )


dai-shi:
What's the problem of showing the warning? If it doesn't help much, we should remove it entirely.

m-bp:
It makes no sense to show a warning related to a browser extension in the node console. I don't think it's wrong to show it in the browser (redux does it, right?). I just don't want that message to show in my console every time Next.js renders a page.

dai-shi:
> in the node console.

Well, you don't need to use the middleware technically.

> I don't think it's wrong

Does anyone else have good experience with the message in the browser?

dai-shi:
I forgot that our solution provided now is `enabled`.

Please specify `enabled: typeof window !== 'undefined'`

m-bp:
But how does it make sense to check for a browser plugin when there's no browser? Before the `enabled` flag was added there was a window check. No one would ever use `enabled: isDevelopment` vs `enabled: typeof window !== "undefined" && isDevelopment`. I know I can circumvent this but there's no point in showing EVERYONE who uses the devtools a warning for no reason unless they specifically disable it.

dai-shi:
I think we should consider removing the warning entirely, if the current solution doesn't fit well.
Can you modify the branch?

PR #2421, opened 2024-03-18
refactor: simplify SkipTwo type definition

Gaic4o:
## Summary

Just a small suggestion. 💙

Refactor the SkipTwo type definition to improve readability and reduce complexity. 

Previously, the type used a nested ternary operator, which made it harder to understand at a glance. The refactoring simplifies the conditional checks for arrays with a length of 0 or 1, always returning an empty array in these cases.

For arrays of length 3 or more, the type now correctly infers and returns the remaining elements `A` after skipping the first two. 

## Check List

### DIFF
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -18,19 +18,7 @@ declare module '../vanilla' {
 }
 
 type Write<T, U> = Omit<T, keyof U> & U
-type SkipTwo<T> = T extends { length: 0 }
-  ? []
-  : T extends { length: 1 }
-    ? []
-    : T extends { length: 0 | 1 }
-      ? []
-      : T extends [unknown, unknown, ...infer A]
-        ? A
-        : T extends [unknown, unknown?, ...infer A]
-          ? A
-          : T extends [unknown?, unknown?, ...infer A]
-            ? A
-            : never
+type SkipTwo<T> = T extends [unknown?, unknown?, ...infer A] ? A : []
 
 type WithImmer<S> = Write<S, StoreImmer<S>>
 


dai-shi:
https://tsplay.dev/WzqrEm

image

It doesn't return an empty array.

dai-shi:
https://tsplay.dev/mxE1XW

image

Gaic4o:
I was wondering if there was another solution, and tried setting the type. I ended up using the nested ternary operator, but I think it makes the type more flexible. What do you think of the code below?

```typescript
type IsEmptyArray<T> = T extends [] ? true : false;
type SkipTwo<T> = IsEmptyArray<T> extends true ? [] : T extends [unknown?, unknown?, ...infer A] ? A : [];
```

dai-shi:
That doesn't work: https://tsplay.dev/WJ3Ovw

image

Gaic4o:
I think the code you previously wrote is better code. I feel like there is no way now. Thank you. I learned a lot.

Gaic4o:
Hello @dai-shi, nice to meet you.
Recently, I reconsidered the SkipTwo logic.
Based on your feedback, I refactored the code as follows.
In this code, I still use nested operators, but each condition is clearly separated, can be managed independently, or is designed for scalability.
I would like to know what you think about this code. Thank you.

```javascript
type IsEmpty<T extends unknown[]> = T extends [] ? true : false;
type IsSingle<T extends unknown[]> = T extends [unknown?] ? true : false;
type IsDouble<T extends unknown[]> = T extends [unknown?, unknown?] ? true : false;
type SkipFirstTwo<T extends unknown[]> = T extends [unknown?, unknown?, ...infer Rest] ? Rest : never;
  
type SkipTwo<T extends unknown[]> = 
    IsEmpty<T> extends true ? [] :
    IsSingle<T> extends true ? [] :
    IsDouble<T> extends true ? [] :
    SkipFirstTwo<T>;
```

dai-shi:
Thanks for the challenge.
I think the single type alias as of now is better than multiple utils (your suggestion).
After all, this is a hack.

By the way, we will drop very old TS supports in Zustand v5. You might want to do another challenge with v5 then.

PR #2457, opened 2024-04-03
optimize: function type judgment

JinYuSha0:

optimize: 
[https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342](https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342)

## Summary
Although this problem may be an official typescript problem, this way of writing should be better than using as force conversion.

## Check List

- [ √] `yarn run prettier` for formatting code and docs

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -58,6 +58,10 @@ type CreateStoreImpl = <
   initializer: StateCreator<T, [], Mos>,
 ) => Mutate<StoreApi<T>, Mos>
 
+const isFunction = (arg: unknown): arg is Function => {
+  return typeof arg === 'function'
+}
+
 const createStoreImpl: CreateStoreImpl = (createState) => {
   type TState = ReturnType<typeof createState>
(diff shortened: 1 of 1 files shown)


src/vanilla.ts
```diff
@@ -67,10 +71,7 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
   const setState: StoreApi<TState>['setState'] = (partial, replace) => {
     // TODO: Remove type assertion once https://github.com/microsoft/TypeScript/issues/37663 is resolved
     // https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342
-    const nextState =
-      typeof partial === 'function'
-        ? (partial as (state: TState) => TState)(state)
-        : partial
+    const nextState = isFunction(partial) ? partial(state) : partial
```
dai-shi, line 74: It's our intention to avoid this. We hope TS fixes it in the future.

dai-shi:
Thanks for the suggestion, but it's as intented.

PR #2474, opened 2024-04-08
Update guides

dbritto-dev:

- Update guides
- Update demos

## Check List

### DIFF
--- a/docs/guides/nextjs.md
+++ b/docs/guides/nextjs.md
@@ -111,9 +111,9 @@ import { useStore } from 'zustand'
 
 import { type CounterStore, createCounterStore } from '@/stores/counter-store'
 
-export type CounterBoundStore = ReturnType<typeof createCounterStore>
+export type CounterStoreApi = ReturnType<typeof createCounterStore>
 
-export const CounterStoreContext = createContext<CounterBoundStore | undefined>(
+export const CounterStoreContext = createContext<CounterStoreApi | undefined>(
   undefined,
 )
 
@@ -124,7 +124,7 @@ export interface CounterStoreProviderProps {
 export const CounterStoreProvider = ({
   children,
 }: CounterStoreProviderProps) => {
-  const storeRef = useRef<CounterBoundStore>()
+  const storeRef = useRef<CounterStoreApi>()
   if (!storeRef.current) {
     storeRef.current = createCounterStore()
   }
@@ -204,9 +204,9 @@ import {
   initCounterStore,
 } from '@/stores/counter-store'
 
-export type CounterBoundStore = ReturnType<typeof createCounterStore>
+export type CounterStoreApi = ReturnType<typeof createCounterStore>
 
-export const CounterStoreContext = createContext<CounterBoundStore | undefined>(
+export const CounterStoreContext = createContext<CounterStoreApi | undefined>(
   undefined,
 )
 
@@ -217,7 +217,7 @@ export interface CounterStoreProviderProps {
 export const CounterStoreProvider = ({
   children,
 }: CounterStoreProviderProps) => {
-  const storeRef = useRef<CounterBoundStore>()
+  const storeRef = useRef<CounterStoreApi>()
   if (!storeRef.current) {
     storeRef.current = createCounterStore(initCounterStore())
   }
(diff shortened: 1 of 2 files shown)


docs/guides/nextjs.md
```diff
@@ -194,16 +196,18 @@ export const createCounterStore = (
...
 
 import {
   type CounterStore,
   createCounterStore,
   initCounterStore,
 } from '@/stores/counter-store'
 
-export const CounterStoreContext = createContext<StoreApi<CounterStore> | null>(
-  null,
+export type CounterBoundStore = ReturnType<typeof createCounterStore>
```
dai-shi, line 207: Can we avoid using "Bound" word? It's confusing people.
(As a side note, I think we should avoid "useBoundStore" in docs if possible.)

dbritto-dev: @dai-shi any suggestions? Since it's StoreApi<CounterStore> what's your suggestion?

dai-shi: Well, if I had an idea, I would have already changed `useBoundStore`.

dai-shi: On second thought, I don't think CounterBoundStore is actually "bound".
```suggestion
export type CounterStoreApi = ReturnType<typeof createCounterStore>
```

dbritto-dev: @dai-shi makes sense, I'll change it

dai-shi (approved):

dai-shi:
let's wait for a couple of days and see if @sewera can take a look.

dbritto-dev:
@sewera all yours

docs/guides/testing.md
```diff
@@ -57,10 +57,22 @@ this means your application logic does not need to be changed or mocked when wri
 
 The mock provided below will enable the relevant test runner to reset the zustand stores after each test.
```
sewera, line 58: This paragraph indicates that the next thing will be a mock of the store, but there is an unexpected _shared code_ section.

It isn't clear from the context, but is the _shared code_ shared between production and test code? If so, you can write a paragraph to explain the purpose. You can also drop the _shared code_ heading and keep everything in one section with a note that _shared code_ is the recommended way of dealing with Zustand in tests.

sewera (changes requested):
I didn't fully get the _shared code_ part, see my comment in the code

docs/guides/testing.md
dbritto-dev: @sewera the shared code part as I mentioned in the docs is just for testing purposes due to duplication of code on both test cases w/o context (create and createStore + context)

dbritto-dev: @sewera lmk if that makes sense, btw I can add a short description if that's necessary

sewera: If you could add it, I think it will read better :)

dbritto-dev:
@sewera all yours

sewera (approved):

docs/guides/testing.md
```diff
@@ -57,10 +57,25 @@ this means your application logic does not need to be changed or mocked when wri
...
 
-> **Warning:** In the following examples we are only supporting the curried version of create `(create()(...))`,
-> since it is supported on both JavaScript and TypeScript. If you are using the uncurried version of
-> `create(...)` you will need to update your hooks to use the curried version, or update the mocks in order
-> to support the uncurried version.
+### Shared code just for testing purposes
+
+This shared code was added to avoid code duplications in our demo since we use the same counter store
```
sewera, line 62:
```suggestion
This shared code was added to avoid code duplication in our demo since we use the same counter store
```

docs/guides/testing.md
```diff
@@ -57,10 +57,25 @@ this means your application logic does not need to be changed or mocked when wri
...
+creator for both implementations, with (`createStore`) and without `Context` API (`create`).
```
sewera, line 63:
```suggestion
creator for both implementations, with and without `Context` API — `createStore` and `create`, respectively.
```

sewera (changes requested):
Some minor wording changes

dbritto-dev:
@sewera done

sewera (approved):

PR #2481, opened 2024-04-12
fix: delete root keys with immer middleware

simlmx:

Fixes #2480

## Summary

Currently when calling `setState` with the `immer` middleware, `replace=false` is assumed by default. This means that the old state is merged with immer's `produce` output. I don't think there is any reason to merge those states: immer's `produce` already returns a smart version of the whole state where copies are used whenever possible. The merging is mostly equivalent to `replace=True`, *except* when keys are deleted: the old deleted keys get merged with the new state.

I'm also assuming that when using the `immer` middleware, we always want to do something like

```typescript
store.setState((state) => { state.x = 3})
```

and never
```typescript
store.setState({x: 3})
```
so I simplified the middleware consequently.

I added a test to reproduce the bug and some other simple tests.

**Someone with more experience with the `immer` middleware should double-check all this! There could very well be usage patterns that I'm not aware of.**

## Check List

@dai-shi @dbritto-dev Let me know if I can improve anything in the PR.

### DIFF
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -57,9 +57,19 @@ const immerImpl: ImmerImpl = (initializer) => (set, get, store) => {
   type T = ReturnType<typeof initializer>
 
   store.setState = (updater, replace, ...a) => {
-    const nextState = (
-      typeof updater === 'function' ? produce(updater as any) : updater
-    ) as ((s: T) => T) | T | Partial<T>
+    if (replace === false) {
+      throw new Error(
+        '`replace` must be `true` (or nullish) when using the `immer` middleware.',
+      )
+    }
+    // We always want to `replace` with `immer`.
+    replace = true
+    if (typeof updater !== 'function') {
+      throw new Error(
+        'You must pass a function to `setState` when using the `immer` middleware.',
+      )
+    }
+    const nextState = produce(updater as any) as (s: T) => T
 
     return set(nextState as any, replace, ...a)
   }
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -55,7 +55,7 @@ type ImmerImpl = <T>(
 const immerImpl: ImmerImpl = (initializer) => (set, get, store) => {
   type T = ReturnType<typeof initializer>
 
-  store.setState = (updater, ...a) => {
+  store.setState = (updater, _, ...a) => {
     const nextState = produce(updater as any) as (s: T) => T
 
     return set(nextState as any, /* replace */ true, ...a)
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -42,6 +42,7 @@ type StoreImmer<S> = S extends {
     ? {
         setState(
           nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
+          shouldReplace?: boolean | undefined,
           ...a: SkipTwo<A>
         ): Sr
       }
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -42,7 +42,6 @@ type StoreImmer<S> = S extends {
     ? {
         setState(
           nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
-          shouldReplace?: boolean | undefined,
           ...a: SkipTwo<A>
         ): Sr
       }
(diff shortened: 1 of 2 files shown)


src/middleware/immer.ts
```diff
@@ -57,9 +57,19 @@ const immerImpl: ImmerImpl = (initializer) => (set, get, store) => {
   type T = ReturnType<typeof initializer>
 
   store.setState = (updater, replace, ...a) => {
```
dbritto-dev, line 59: Let's remove `shouldReplace?: boolean | undefined` on line `45` and remove `replace` parameter since wouldn't be necessary

```suggestion
  store.setState = (updater, ...a) => {
```

src/middleware/immer.ts
```diff
@@ -57,9 +57,19 @@ const immerImpl: ImmerImpl = (initializer) => (set, get, store) => {
...
+    // We always want to `replace` with `immer`.
+    replace = true
+    if (typeof updater !== 'function') {
+      throw new Error(
+        'You must pass a function to `setState` when using the `immer` middleware.',
+      )
+    }
+    const nextState = produce(updater as any) as (s: T) => T
 
     return set(nextState as any, replace, ...a)
```
dbritto-dev, line 74: then set `replace` as `true` by default

```suggestion
    const nextState = produce(updater as any) as (s: T) => T

    return set(nextState as any, true, ...a)
```

dbritto-dev (changes requested):
Please review all comments

src/middleware/immer.ts
simlmx: Shouldn't we also remove `...a` and maybe change `nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void)` to `updater: (state: Draft<T>) => void`.

I'll be honest I'm very confused with the typing in these middlewares 😅

src/middleware/immer.ts
simlmx: Just to be sure: also remove the check on `updater`?

simlmx: I applied your suggestions and pushed a new commit but it breaks the typing, but I'm not sure why.

It's not happy with this line:
```typescript
return set(nextState as any, /* replace */ true, ...a)
```
(`Expected 1-2 arguments, but got 3.`)

If I put back the `replace` in this line the error goes away:
```typescript
 store.setState = (updater, replace, ...a) => {
```

But I'm really not sure why.

dbritto-dev: Let me look into that quickly

src/middleware/immer.ts
```diff
@@ -56,12 +55,10 @@ type ImmerImpl = <T>(
 const immerImpl: ImmerImpl = (initializer) => (set, get, store) => {
   type T = ReturnType<typeof initializer>
 
-  store.setState = (updater, replace, ...a) => {
-    const nextState = (
-      typeof updater === 'function' ? produce(updater as any) : updater
-    ) as ((s: T) => T) | T | Partial<T>
+  store.setState = (updater, ...a) => {
```
dbritto-dev, line 58:
```suggestion
  store.setState = (updater, _, ...a) => {
```

src/middleware/immer.ts
```diff
@@ -42,7 +42,6 @@ type StoreImmer<S> = S extends {
     ? {
         setState(
           nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
```
dbritto-dev, line 44:
```suggestion
          nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
          shouldReplace?: boolean | undefined,
```

dbritto-dev:
@simlmx we need to add a comment on immer middleware saying that `shouldReplace` parameter is completely ignore on immer middleware and will be set up to true by default

connormanning:
> I'm also assuming that when using the `immer` middleware, we always want to do something like
> 
> ```ts
> store.setState((state) => { state.x = 3})
> ```
> 
> and never
> 
> ```ts
> store.setState({x: 3})
> ```

Is this accurate?  I thought the vanilla version was still intended to be supported with `immer` if you *do* return a value:
- returning nothing applies any in-place `Draft` modifications
- returning a value applies the vanilla shallow merge behavior
- doing both is an error: `caught Error: [Immer] An immer producer returned a new value *and* modified its draft. Either return a new value *or* modify the draft`

This behavior makes sense to me in the context of a middleware: intercepting only the `void` returns.  For example you could use the immer model only for a few operations which operate on nested states, or another example being adding immer to an existing store for a few operations without having to rewrite all of the existing vanilla operations.

I'm no expert on this project so I'll defer to the maintainers, but I think #2495, which explicitly adds TS definitions to support the model described above, contradicts the assumptions in this PR.  Hopefully someone with a greater understanding of this library and its middleware can chime in.

dbritto-dev:
@connormanning when you use immer you return the whole state so you should merge the current state with the new state you should always replace the state since immer handle the changes, this PR set replace true by default

dbritto-dev:
@connormanning let's discuss the changes to immer middleware here

dai-shi:
oookay, I guess we have some design conflicts.
I guess the question if whether we should force `replace=true` with `immer` middleware.

As I generally prefer simpler api and implementation, my suggestion is:
```ts
        setState(
          nextStateOrUpdater: ((state: Draft<T>) => T | void),
          ...a: SkipTwo<A>
        ): Sr
```
But, this is a breaking change, so it should only come in v5.

And, my suggestion is, if this does't cover some use cases, use `produce` without the middleware. (we need to educate it in docs.)

dai-shi:
Or, maybe:
```ts
        setState(
          nextStateOrUpdater: T | ((state: Draft<T>) => T | void),
          ...a: SkipTwo<A>
        ): Sr
        ```

dbritto-dev:
@dai-shi yeah, I guess we need to implement this on v5

dbritto-dev:
Let's create a new PR if is necessary I'll close this one for now

dbritto-dev:
Let's create a new PR if is necessary I'll close this one for now

PR #2494, opened 2024-04-19
docs: add zustand-slices

dai-shi:
Add https://github.com/zustandjs/zustand-slices in the third-party lib list.


dbritto-dev:
LGTM

PR #2507, opened 2024-04-24
Simplifies vanilla `StoreApi`

ahrjarrett:
I will circle back later today and add comments explaining in more detail

## Summary
- removes `SetStateInternal` in favor of inline definition
- replaces `Partial` with a greedy version (lets us show users concrete types sooner)
- removes the type assertion on `partial` by fixing the variance hack elsewhere
- moves deprecated subset of the StoreApi to separate interface, to create some separation between them

## Check List


ahrjarrett:
Ah, [I see now](https://github.com/pmndrs/zustand/issues/710) why this doesn't work.

Closing, but I have a [couple ideas](https://tsplay.dev/mMPRQN) that I think might allow us to get around `T` being invariant. 

Will open a separate PR in the future when I have more time to explore them.

PR #2545, opened 2024-05-09
Migrate to pnpm

dai-shi:
following https://github.com/pmndrs/jotai/pull/2548


sewera:
Out of curiosity, why pnpm and not simply npm or latest yarn?

I found in my professional work that project linking for development is much more reliable in yarn than in pnpm, but my company migrated to pnpm due to better monorepo support.

dai-shi:
Yeah, I was/am a bit skeptical about the trickiness and the strictness (which I kind of like). But, the best thing for me is disk space efficiency as I have lots of similar projects.

sewera:
Fair point

PR #2557, opened 2024-05-16
Immer state creator should take 4 type params

ddetkovs:
## Summary
Immer middleware state creator should mimic the StateCreator type and pass through the U type.

## Check List


dbritto-dev:
@ddetkovs no, shouldn't

PR #2574, opened 2024-05-30
fix nextjs guide in docs

Prains:
Current pull request change import statement in next js use-case guide.

Since release of Next-15-rc and use of experimental-react-compiler next js cannot use zustand/vanilla package anymore, as it does not has any bindings to React itself and cannot trigger component rerenders anymore. 

React-compiler memoizes zustand/vannila store and it actions. Actions does not provoke rerender or changing of store anymore

### DIFF
--- a/docs/guides/nextjs.md
+++ b/docs/guides/nextjs.md
@@ -70,7 +70,7 @@ request.
 
 ```ts
 // src/stores/counter-store.ts
-import { createStore } from 'zustand/vanilla'
+import { createStore } from 'zustand'
 
 export type CounterState = {
   count: number


dbritto-dev:
@Prains create = createStore + useStore if you want something bound to React  but by default createStore don't do that

Prains:
@dbritto-dev I am aware of that. I followed Next js guide using next-15-rc and react-compiler. Importing createStore from 'zustand/vanilla' does not make store be changed and rerender component, since it's value being memoized. 

But importing store from 'zustand' directly helped me and solved memoization issue. For that reason i want to update docs, so anyone who will follow my path and will use react-compiler will not have any hard time solving this issue

dbritto-dev:
@Prains in any case we should revisit that when v5 is out

PR #2576, opened 2024-06-02
Install only stable versions on fresh installs

dbritto-dev:
## Summary

Avoid install react 19 on fresh installs

## Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -256,7 +256,7 @@
   "peerDependencies": {
     "@types/react": ">=16.8",
     "immer": ">=9.0.6",
-    "react": ">=16.8"
+    "react": ">=16.8, <19"
   },
   "peerDependenciesMeta": {
     "@types/react": {


package.json
```diff
@@ -256,7 +256,7 @@
   "peerDependencies": {
     "@types/react": ">=16.8",
     "immer": ">=9.0.6",
-    "react": ">=16.8"
+    "react": ">=16.8, <19"
```
dai-shi, line 259: What's the issue? This means it doesn't work with React 19, which is not true, right?

dbritto-dev: @daishi actually, this change is more related to only install stable versions on fresh installs and avoid other packages complain like testing library react

dai-shi: What is "fresh install"? Is it about developing Zustand itself?

I don't feel this change is acceptable, because we should support React 19.

> might require some few changes to match react 19 and their new rules

What are those?

dbritto-dev: @dai-shi fresh install is when you don't have node_modules folder and you run yarn/npm/pnpm install

dbritto-dev: @dai-shi yeah, for development

dbritto-dev: image

dai-shi: For the zustand repository, it's intentional, because we want to develop and run tests against `react@rc`. It's just a warning. So, bear with it until we have `react@19`.

charkour: The fresh install in development will install the react version listed in [devDependencies](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#devdependencies) which is listed as React 19 RC. peerDependencies affects userland and development.

dbritto-dev: @dai-shi 👆🏼

charkour: @dbritto-dev, I think my comment supports Daishi's suggestions.

Because zustand has React 19 RC specified in its devDeps, then we want those developing zustand to install React 19 RC, correct?

Having React 19 allows in peerDeps won't install React 19 for the user, unless they specify it in their application's deps or devDeps.

dbritto-dev: @charkour actually, on fresh install you would ended up installing react 19 rc even on userland

BTW, on development I'm not sure if you want to install react 19 rc. Sounds OK when you want to check if your package still working or doesn't break but I'm not sure if want to get react 19 rc while you are on development at least on v4

dbritto-dev: @dai-shi @charkour but up to you peeps, keep me in the loop or feel free to close this PR

charkour: > actually, on fresh install you would ended up installing react 19 rc even on userland

I'm sure it would install React 19 RC if someone clones pmndrs/zustand locally for development, but if a user installs zustand as a package, no React would be installed because it is listed as a `peerDependency` and not a `dependency`.
	
> BTW, on development I'm not sure if you want to install react 19 rc

I agree. Should we update the GitHub Actions to run a matrix with React 18 and React 19?

I agree that @dai-shi should decide. Thank you

dai-shi: Using 19-rc in devDependencies is intentional. I wish testing-library supports 19 soon, but as long as they are just warnings, it's fine.

dai-shi: And even in userland, fresh installing 19-rc is fine, because it's the package manager behavior, and React versioning suggests it.

PR #2586, opened 2024-06-05
useStore uses ReadonlyStoreApi

holic:

Fixes https://github.com/pmndrs/zustand/discussions/2581

## Summary

`UseBoundStore` types were relaxed in https://github.com/pmndrs/zustand/pull/1589 because the internals of this method did not use any of the setters, only getters. We can also do this for `useStore` types because its API only uses getters.

Context: I am building a library that returns a store API with only getter methods exposed and I would like folks to be able to pass this directly into Zustand's `useStore` React helper.

## Check List


dai-shi (approved):
LGTM
I think I can take care of the rest. Thanks for your contribution!

PR #2598, opened 2024-06-10
doc: added zustand-computed-state to external libraries

yasintz:
https://www.npmjs.com/package/zustand-computed-state


dai-shi:
Please run `pnpm run prettier`.

PR #2661, opened 2024-08-02
(docs) remove unintended change from default merge behaviour to identical but manual one

vorant94:

Fixes #

## Summary

targeted section of the README talks about integration with redux devtools. to accomplish this the dev needs to define third argument for `set` function. since arguments for `set` function are positional it means he/she needs to define the second, `replace`, argument as well

currently README suggests just to use `false` value for `replace` arg i.e. to override the default value by identical manual one, but defined on the dev side.

more clean way would be to use `undefined` at the second argument position that will make JS itself to use the default value, meaning moving control of it back lib authors, not dev

## Check List


dai-shi (approved):
LGTM. Thanks!

PR #2664, opened 2024-08-05
Fix return type for persist's partialize option

darnfish:
## Summary
Fix return type for persist's partialize option

## Check List

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -108,7 +108,7 @@ export interface PersistOptions<S, PersistedState = S> {
    *
    * @params state The state's value
    */
-  partialize?: (state: S) => PersistedState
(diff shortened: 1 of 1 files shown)


dbritto-dev:
@darnfish what's the issue?

src/middleware/persist.ts
```diff
@@ -108,7 +108,7 @@ export interface PersistOptions<S, PersistedState = S> {
    *
    * @params state The state's value
    */
-  partialize?: (state: S) => PersistedState
+  partialize?: (state: S) => Partial<PersistedState>
```
dai-shi, line 111: `PersistedState` would be `Partial<State>`, although it does't have to be.

dai-shi:
Closing as it's intentional. Feel free to open a discussion.

PR #2668, opened 2024-08-08
fix: "value" to "content" attr for og:image

ryota-murakami:
## Summary
I fixed the `og:iamge`
 attribute name to `content` according to the [Open Graph Protocol spec](https://ogp.me/#metadata). 

Thanks for always actively maintaining such a great library!

## Check List


dai-shi (approved):
👍

PR #2674, opened 2024-08-11
`pmndrs/docs` workflow

abernier:
calling reusable `pmndrs/docs/.github/workflows/build.yml@app-router`

Result here: https://pmndrs.github.io/zustand


.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
+name: Build documentation and deploy to GitHub Pages
+on:
+  push:
+    branches: ['docs']
```
abernier, line 4: change to default branch before merging

.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
...
+  workflow_dispatch:
+
+# Cancel previous run (see: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency)
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    uses: pmndrs/docs/.github/workflows/build.yml@app-router
```
abernier, line 14: change to `'@main'` once https://github.com/pmndrs/docs/pull/282 is merged

dai-shi (commented):
LGTM

Maybe we can remove ./bear.jpg file and points to ./docs/logo.jpg instead, if they are identical.

PR #2680, opened 2024-08-15
chore(deps): update dependencies

dai-shi:
for v4.5.5


package.json
```diff
@@ -203,15 +203,15 @@
   "homepage": "https://github.com/pmndrs/zustand",
   "packageManager": "pnpm@8.15.0",
   "dependencies": {
-    "use-sync-external-store": "1.2.0"
+    "use-sync-external-store": "1.2.2"
```
dai-shi, line 206: I didn't check the changelog. Is there a changelog somewhere?

charkour: Seems like there was only a version bump but no source code changes: https://npmdiff.dev/use-sync-external-store/1.2.0/1.2.2/

charkour: I checked with Renovate too and found the same change:

https://app.renovatebot.com/package-diff?name=use-sync-external-store&from=1.2.0&to=1.2.2

dai-shi: Thanks for checking!!

PR #2684, opened 2024-08-16
chore: bear.jpg in ./docs

dai-shi:
as it is the same file as `logo.jpg`.


dai-shi:
This needs to be merged to test it.

PR #2692, opened 2024-08-20
Persist documentation for Map & Set add storage and retrieval of version

melutovich:
--HG--
branch : discussion-2689-persist-doc-map-and-set-missing-version

Fixes # Discussion 2689

## Summary

Fixes getItem and setItem on the storage for the Map & Set persist example, so that respectfully those methods retrieves and sets the version when present.

## Check List

- [v] `pnpm run prettier` for formatting code and docs

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -783,17 +783,19 @@ interface BearState {
     getItem: (name) => {
       const str = localStorage.getItem(name);
       if (!str) return null;
-      const { state } = JSON.parse(str);
+      const { state, version } = JSON.parse(str);
       return {
         state: {
           ...state,
           transactions: new Map(state.transactions),
         },
+        version
       }
     },
     setItem: (name, newValue: StorageValue<BearState>) => {
       // functions cannot be JSON encoded
       const str = JSON.stringify({
+        ...newValue,
         state: {
           ...newValue.state,
           transactions: Array.from(newValue.state.transactions.entries()),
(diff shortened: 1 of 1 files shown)


docs/integrations/persisting-store-data.md
```diff
@@ -783,17 +783,19 @@ interface BearState {
...
           ...state,
           transactions: new Map(state.transactions),
         },
+        version
       }
     },
     setItem: (name, newValue: StorageValue<BearState>) => {
       // functions cannot be JSON encoded
       const str = JSON.stringify({
+        ...newValue,
```
dai-shi, line 798: This is good.

docs/integrations/persisting-store-data.md
```diff
@@ -783,17 +783,19 @@ interface BearState {
...
       const str = localStorage.getItem(name);
       if (!str) return null;
-      const { state } = JSON.parse(str);
+      const { state, version } = JSON.parse(str);
       return {
         state: {
           ...state,
           transactions: new Map(state.transactions),
         },
+        version
```
dai-shi, line 792: This makes `version: undefined` if version is undefined. We also need to use object spread here like `setItem`.

dai-shi (commented):
Thanks for opening up a PR! I think a small change is required. Thanks.

docs/integrations/persisting-store-data.md
melutovich: I've pushed another commit switching this as you've suggested.

dai-shi (approved):
Looks good to me! Thanks for your contribution.

PR #2696, opened 2024-08-23
Fix Immer type inference for `setState`

chrisvander:
Bases the type of `setState` off of `setState` instead of `getState`.

Fixes #1621

## Summary

This PR fixes the Immer middleware so that the type of `setState` is inferred off of the store's `setState` as opposed to `getState`. When using a middleware like [zustand-computed](https://github.com/chrisvander/zustand-computed), this would previously have the result of a differing `getState` generic would move over to `setState` unintentionally.

## Check List

### DIFF
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -1,6 +1,10 @@
 import { produce } from 'immer'
 import type { Draft } from 'immer'
-import type { StateCreator, StoreMutatorIdentifier } from '../vanilla.ts'
+import type {
+  StateCreator,
+  StoreApi,
+  StoreMutatorIdentifier,
+} from '../vanilla.ts'
 
 type Immer = <
   T,
@@ -18,7 +22,6 @@ declare module '../vanilla' {
 }
 
 type Write<T, U> = Omit<T, keyof U> & U
-type First<T> = T extends [infer A] ? A : never
 type SkipTwo<T> = T extends { length: 0 }
   ? []
   : T extends { length: 1 }
@@ -38,25 +41,24 @@ type WithImmer<S> = Write<S, StoreImmer<S>>
 type StoreImmer<S> = S extends {
   setState: infer SetState
 }
-  ? SetState extends {
-      (...a: infer A1): infer Sr1
-      (...a: infer A2): infer Sr2
-    }
-    ? {
-        setState(
-          nextStateOrUpdater:
-            | First<A1>
-            | Partial<First<A1>>
-            | ((state: Draft<First<A1>>) => void),
-          shouldReplace?: false,
-          ...a: SkipTwo<A1>
-        ): Sr1
-        setState(
-          nextStateOrUpdater: First<A2> | ((state: Draft<First<A2>>) => void),
-          shouldReplace: true,
-          ...a: SkipTwo<A2>
-        ): Sr2
+  ? SetState extends StoreApi<infer T>['setState']
+    ? SetState extends {
+        (...a: infer A1): infer Sr1
+        (...a: infer A2): infer Sr2
       }
+      ? {
+          setState(
+            nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
+            shouldReplace?: false,
+            ...a: SkipTwo<A1>
+          ): Sr1
+          setState(
+            nextStateOrUpdater: T | ((state: Draft<T>) => void),
+            shouldReplace: true,
+            ...a: SkipTwo<A2>
+          ): Sr2
+        }
+      : never
     : never
   : never
 
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -44,6 +44,9 @@ type StoreImmer<S> = S extends {
       (...a: infer A2): infer Sr2
     }
     ? {
+        // Ideally, we would want to infer the `nextStateOrUpdater` `T` type from the
+        // `A1` type, but this is infeasible since it is an intersection with
+        // a partial type.
         setState(
           nextStateOrUpdater:
             | SetStateType<A2>
--- a/src/middleware/immer.ts
+++ b/src/middleware/immer.ts
@@ -18,6 +18,7 @@ declare module '../vanilla' {
 }
 
 type Write<T, U> = Omit<T, keyof U> & U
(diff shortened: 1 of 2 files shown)


src/middleware/immer.ts
```diff
@@ -35,25 +39,26 @@ type SkipTwo<T> = T extends { length: 0 }
...
-          nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
-          shouldReplace?: false,
-          ...a: SkipTwo<A1>
-        ): Sr1
-        setState(
-          nextStateOrUpdater: T | ((state: Draft<T>) => void),
-          shouldReplace: true,
-          ...a: SkipTwo<A2>
-        ): Sr2
+  ? SetState extends StoreApi<infer T>['setState']
```
dai-shi, line 44: I'm not sure if this works because `StoreApi['setState']` is very complicated type. Also, `setState` can be modified by upstream middleware. Can we infer `T` from `A1` and `A2`?

chrisvander: It does infer correctly as the StoreApi<T> `setState` type is based off of the single generic passed into StoreApi. I think doing it this way bottles the complexity of the type and handles it appropriately. All the tests passed and it seems to resolve correctly. 

When I tried to base it off of A1 and A2, I would've needed to fight through the entire intersection type (e.g. `T | Partial<T> | ...`) in order to get T. That method feels more brittle than inferring through StoreApi.

Any situations you're thinking of specifically here? Like a modified `setState` type from another middleware?

chrisvander: Well, actually, if I used the A2 type, there would be no Partial to fight with. Just the object and the function type. One sec, I'll try that.

dai-shi: > Like a modified setState type from another middleware?

Yes.

chrisvander: Check out my latest commit and let me know what you think of that. The assumption I made is that `A2` will always have a first parameter that is `T | (...args: any[]) => any`. So, `T` can be found by excluding any function types. That should be more robust.

chrisvander: Inferring off of `A1` seemed non-trivial given that it is `T | Partial<T> | ...`, which I'm not sure if there's a method to derive `T` out of that.

dai-shi: Hmmm, interesting. I guess the approach is right though.

src/middleware/immer.ts
```diff
@@ -44,12 +45,17 @@ type StoreImmer<S> = S extends {
     }
     ? {
         setState(
-          nextStateOrUpdater: T | Partial<T> | ((state: Draft<T>) => void),
+          nextStateOrUpdater:
+            | SetStateType<A2>
+            | Partial<SetStateType<A2>>
+            | ((state: Draft<SetStateType<A2>>) => void),
```
dai-shi, line 51: Can we add a comment that ideally we want to infer types from `A1`.

dai-shi (changes requested):
this is too tricky and not 100% correct anyway. seems fine for now.
can you add a test please?

chrisvander:
> this is too tricky and not 100% correct anyway. seems fine for now. can you add a test please?

The simplest test would probably be to register an example middleware that modifies `getState` but not `setState` and see if it resolves correctly downstream. Does that sound reasonable?

chrisvander:
Check out that new test and let me know if it's sufficient. Pollution of the `StoreMutators` type should be isolated to the test file.

dai-shi:
Thanks for working on it.

> Pollution of the `StoreMutators` type should be isolated to the test file.

Hmm, but it made me think that it should be safer if the new test is put in a new test file. What do you think?

chrisvander:
> Thanks for working on it.
> 
> 
> 
> > Pollution of the `StoreMutators` type should be isolated to the test file.
> 
> 
> 
> Hmm, but it made me think that it should be safer if the new test is put in a new test file. What do you think?

I'd be okay with that, but it does make sense alongside other middleware tests. For most of the middleware it may be nice to have some contrived third party middleware examples to try things against. Outside the scope of this PR but thought I would mention.

tests/middlewareTypes.test.tsx
```diff
@@ -19,6 +19,27 @@ type CounterState = {
...
+  U = T,
+>(
+  f: StateCreator<T, [...Mps, ['org/example', A]], Mcs>,
+) => StateCreator<T, Mps, [['org/example', A], ...Mcs], U & A>
+
+type Write<T, U> = Omit<T, keyof U> & U
+type StoreModifyAllButSetState<S, A> = S extends {
+  getState: () => infer T
+}
+  ? Omit<StoreApi<T & A>, 'setState'>
```
dai-shi, line 34: I'm not sure if I understand this fully, but omitting setState may prevent inferring type from setState, no?
The reason the immer impl works is it's using `set` instead of `setState`, but we don't get types from `set`.
I'm a bit confused what's going on.

chrisvander: Sorry, just updated that. The goal is to use an example middleware that modifies the store state to add some `A` type to all *but* the `setState`, and then assert that it affected the types appropriately on both the inside of the Immer middleware and the outside.

dai-shi: yeah, but still it doesn't feel like a good example, because the implementation of the example middleware should include `delete api.setState`, but `immer` middleware infers types from `setState`.

chrisvander: Why would the implementation need to include `delete api.setState`? The example rewrites `setState`, it doesn't delete it. 

Do you agree that the type assertions are sufficient? Since this change is strictly on a type level, I didn't think implementation was necessary.

dai-shi: > it doesn't delete it.

Okay, I think I misread something then.

dai-shi: yeah, I think I forgot that `Write` is an intersection.
Never mind.

dai-shi (approved):
LGTM
Thanks for your contribution!

(Now, I wonder if we should cut an rc release only for this change. 🤔 )

chrisvander:
A minor bump may not hurt 😁

PR #2706, opened 2024-08-27
Add docs for APIs and Hooks

dbritto-dev:


dai-shi (commented):
LGTM

dbritto-dev:
@sewera 🙏🏽

sewera:
Review in progress, but I'll need some time, it's a big PR

dbritto-dev:
> Review in progress, but I'll need some time, it's a big PR

@sewera  thanks

dai-shi:
How about separating PRs and merging them one by one?

dbritto-dev:
@dai-shi sure

dbritto-dev:
@sewera any news?

docs/apis/create-store.md
```diff
@@ -0,0 +1,543 @@
...
+```
+
+For example, this line mutates the state from a past render:
+
+```ts
+personStore.getState().firstName = (e.target as any).value
+```
+
+The reliable way to get the behavior you’re looking for is to create a new object and pass it to
+`setPerson`. But here, you want to also copy the existing data into it because only one of the
```
sewera, line 455:
```suggestion
`setPerson`. But here you want to also copy the existing data into it because only one of the
```

docs/apis/create-store.md
```diff
@@ -0,0 +1,543 @@
...
+fields has changed:
+
+```ts
+personStore.getState().setPerson({
+  firstName: e.target.value, // New first name from the input
+})
+```
+
+> [!NOTE]
+> We don’t need to copy every property separately due to `set` function performs shallow
```
sewera, line 465:
```suggestion
> We don’t need to copy every property separately due to `set` function performing shallow
```

docs/apis/create-with-equality-fn.md
```diff
@@ -0,0 +1,618 @@
...
+        }}
+      />
+    </div>
+  )
+}
+```
+
+## Troubleshooting
+
+### I’ve updated the state, but the screen doesn’t update
```
sewera, line 442: You can link the "you should always use immutable updates" note to here to visualize what happens if you don't.

sewera (changes requested):
That's all I'll be able to contribute. I have a medical situation ongoing, so please extrapolate my suggestions to places I missed and move forward with this PR without me.

docs/apis/create-with-equality-fn.md
dbritto-dev: @sewera nice suggestion

docs/apis/create-store.md
```diff
@@ -0,0 +1,543 @@
...
+Arrays are mutable in JavaScript, but you should treat them as immutable when you store them in
+state. Just like with objects, when you want to update an array stored in state, you need to create
+a new one (or make a copy of an existing one), and then set state to use the new array.
+
+By default, `set` function performs a shallow merge. To update array values we should assign new
+values to ensure updates are applied correctly, and avoid unexpected behaviors. To completely
+replace the state with a new one, use the `replace` parameter set to `true`.
+
+> [!IMPORTANT]
+> We should prefer immutable operations like: `[...array]`, `concat(...)`, `filter(...)`,
```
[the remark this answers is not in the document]
dbritto-dev: I'll add a guide in the new learn section talking about that

PR #2717, opened 2024-09-01
docs: corrected example

EthanRBrown:
## Summary

Corrected store shape in "Selecting multiple state slices" to be consistent with the example (which use `nuts`, `honey`, and `treats` instead of `bears`).

## Check List

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -94,10 +94,7 @@ const useBearStore = create(set => ({
   nuts: 0,
   honey: 0,
   treats: {},
-  increaseNuts: () => set(state => ({ nuts: state.nuts + 1 })),
-  increaseHoney: () => set(state => ({ honey: state.honey + 1 })),
-  increaseTreats: (treat: string, count?: number) =>
-    set(state => ({ treats: { ...state.treats, [treat]: (state.treats[treat] ?? 0) + 1 } })),
+  ...
 }))
 
 // Object pick, re-renders the component when either state.nuts or state.honey change
@@ -114,7 +111,7 @@ const [nuts, honey] = useBearStore(
 const treats = useBearStore(useShallow((state) => Object.keys(state.treats)))
 ```
 
-For more control over re-rendering, you may provide any custom equality function.
+For more control over re-rendering, you may provide any custom equality function (this example requires the use of [`createWithEqualityFn`](./docs/migrations/migrating-to-v5.md#using-custom-equality-functions-such-as-shallow)).
 
 ```jsx
 const treats = useBearStore(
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -90,10 +90,14 @@ If you want to construct a single object with multiple state-picks inside, simil
...
-  removeAllBears: () => set({ bears: 0 }),
+const useBearStore = create(set => ({
+  nuts: 0,
+  honey: 0,
+  treats: {},
+  increaseNuts: () => set(state => ({ nuts: state.nuts + 1 })),
+  increaseHoney: () => set(state => ({ honey: state.honey + 1 })),
+  increaseTreats: (treat: string, count?: number) =>
+    set(state => ({ treats: { ...state.treats, [treat]: (state.treats[treat] ?? 0) + 1 } })),
 }))
```
dai-shi, line 101: Hi, thanks for the suggestion. I didn't realize this.
We don't want to increase lines readme.md (and possibly move detail/complete examples to docs),
so can we make this simpler??

```suggestion
const useBearStore = create((set) => ({ nuts: 0, honey: 0, treats: {} /* ... */ }))
```

dai-shi (changes requested):
Thanks.

---
I found another issue.

The following is incorrect. We need `createWithEqualityFn` to make it work.
```
For more control over re-rendering, you may provide any custom equality function.

const treats = useBearStore(
  (state) => state.treats,
  (oldTreats, newTreats) => compare(oldTreats, newTreats),
)
```

readme.md
EthanRBrown: Thanks, and I agree.  The update functions are unnecessary, so I removed them and added a `...`, which I think is clear enough.  I also added a link to what documentation I could find for `createWithEqualityFn`.

I'm afraid I don't have the bandwidth to go deeper and break this into its own file, etc.  Hopefully this will be sufficient for now.  Thanks for the feedback!

readme.md
```diff
@@ -90,10 +90,11 @@ If you want to construct a single object with multiple state-picks inside, simil
...
 
-const useBearStore = create((set) => ({
-  bears: 0,
-  increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
-  removeAllBears: () => set({ bears: 0 }),
+const useBearStore = create(set => ({
+  nuts: 0,
+  honey: 0,
+  treats: {},
+  ...
```
dai-shi, line 97:
```suggestion
  // ...
```

dai-shi (approved):
Cool! Thanks for your contribution!

PR #2745, opened 2024-09-18
docs: add plaintext language to code block to prevent layout breaks

87xie:

N/A

## Summary

Hi everyone, I noticed that the error message code block in the [v5 migration guideline](https://zustand.docs.pmnd.rs/migrations/migrating-to-v5#requiring-stable-selector-outputs)  doesn’t display correctly on the docs site, so I added the `plaintext` language to the code block to prevent layout breaks. 

Feel free to close this if it is unnecessary.

**Before**

image

**After**

image

## Check List

### DIFF
--- a/docs/migrations/migrating-to-v5.md
+++ b/docs/migrations/migrating-to-v5.md
@@ -105,7 +105,7 @@ const action = useMainStore((state) => {
 
 The error message will be something like this:
 
-```
+```plaintext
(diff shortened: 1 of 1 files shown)


dai-shi:
I'm not sure if this fixes the issue, but let's merge and see.

87xie:
It looks better now. Thanks for the review!

Issue #2741, opened 2024-09-14
[Feedback Request] Does Zustand v5-rc work for you?


dai-shi:
## How to Install

```bash
npm i zustand@next
```

## Migration Guide

https://github.com/pmndrs/zustand/blob/main/docs/migrations/migrating-to-v5.md

## How to respond

Add comments to this issue or add reactions to this issue comment.

👍 : I've tested v5-rc and it works fine for me.
👎 : I've tested v5-rc and found an issue. < Comment below!
👀 : I'll try it later.

pavitra-infocusp:
Is it intentional to push an RC release to the latest?
![image](https://github.com/user-attachments/assets/c3c641de-ba15-498b-a633-a33705528255)

Anyways, my app doesn't work. It crashed when I open the page :(
```
Unexpected Application Error!
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at checkForNestedUpdates (http://localhost:5173/node_modules/.vite/deps/chunk-HHZE2HUH.js?v=9df50b02:19659:19)
    at scheduleUpdateOnFiber (http://localhost:5173/node_modules/.vite/deps/chunk-HHZE2HUH.js?v=9df50b02:18533:11)
    at dispatchSetState (http://localhost:5173/node_modules/.vite/deps/chunk-HHZE2HUH.js?v=9df50b02:12403:15)
    at update (http://localhost:5173/node_modules/.vite/deps/@nextui-org_react.js?v=54cc688f:14580:11)
    at http://localhost:5173/node_modules/.vite/deps/@nextui-org_react.js?v=54cc688f:14583:7
    at commitHookEffectListMount (http://localhost:5173/node_modules/.vite/deps/chunk-HHZE2HUH.js?v=9df50b02:16915:34)
    at commitLayoutEffectOnFiber (http://localhost:5173/node_modules/.vite/deps/chunk-HHZE2HUH.js?v=9df50b02:17002:23)
    at commitLayoutMountEffects_complete (http://localhost:5173/node_modules/.vite/deps/chunk-HHZE2HUH.js?v=9df50b02:17980:17)
[...]

PR #2749, opened 2024-09-18
fix(docs): improve v5 migration guide

dai-shi:
https://github.com/pmndrs/zustand/issues/2741#issuecomment-2359569305

### DIFF
--- a/docs/migrations/migrating-to-v5.md
+++ b/docs/migrations/migrating-to-v5.md
@@ -98,15 +98,36 @@ For example, this may cause infinite loops.
 
 ```js
 // v4
-const action = useMainStore((state) => {
-  return state.action ?? () => {}
-})
+const [searchValue, setSearchValue] = useStore((state) => [
+  state.searchValue,
+  state.setSearchValue,
+])
 ```
 
 The error message will be something like this:
 
 ```plaintext
 Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
+```
+
+To fix it, `useShallow` should help return a stable reference.
+
+```js
+// v5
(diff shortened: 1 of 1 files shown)


docs/migrations/migrating-to-v5.md
```diff
@@ -109,6 +110,26 @@ The error message will be something like this:
 Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
 ```
 
+To fix it, `useShallow` should help return a stable reference.
```
sewera, line 113:
```suggestion
To fix it, use the `useShallow` hook, which will return a stable reference.
```

docs/migrations/migrating-to-v5.md
```diff
@@ -109,6 +110,26 @@ The error message will be something like this:
...
+```js
+// v5
+import { useShallow } from 'zustand/shallow'
+
+const [searchValue, setSearchValue] = useStore(
+  useShallow((state) => [state.searchValue, state.setSearchValue]),
+)
+```
+
+Here's another example that may cause infinite loops.
```
sewera, line 124: I don't really understand why there is "another example," but the comment says that it's v4. Didn't you mean to say "Here's a similar example in v4 that may cause infinite loops?"

sewera (changes requested):

docs/migrations/migrating-to-v5.md
dai-shi: I mean there are several cases that cause infinite loops.
I want to show typical two examples that are actually reported by users.

sewera (approved):

docs/migrations/migrating-to-v5.md
dai-shi: Feel free to make suggestions to clarify that.

PR #2777, opened 2024-10-08
Fix "previous" and "next" navigation links

chaficnajjar:
## Summary

The "next" link at the bottom of  https://zustand.docs.pmnd.rs/guides/prevent-rerenders-with-use-shallow points to https://zustand.docs.pmnd.rs/integrations/immer-middleware when it should be pointing to https://zustand.docs.pmnd.rs/guides/ssr-and-hydration instead (i.e. the next page in the list of links in the sidebar).

This PR updates all the `nav` numbers in order to correct the navigation order. If this is not the right solution, please feel free to edit this PR with a better one.

## Check List

### DIFF
--- a/docs/apis/create-store.md
+++ b/docs/apis/create-store.md
@@ -1,7 +1,7 @@
 ---
 title: createStore
 description: How to create vanilla stores
-nav: 202
+nav: 24
 ---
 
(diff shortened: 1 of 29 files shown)


dbritto-dev (approved):
LGTM

PR #2816, opened 2024-10-24
fix(shallow): iterable-like insensitive keys order comparison

dbritto-dev:
## Related Bug Reports or Discussions

Fixes #2808 

## Summary

https://github.com/pmndrs/zustand/discussions/2808

## Check List

- [x] `pnpm run prettier` for formatting code and docs

PR #2821, opened 2024-10-27
fix(shallow): iterable-like insensitive keys order comparison (alternate implementation)

dai-shi:
Using the same tests in #2816.


dai-shi:
@dbritto-dev How do you like this one?

PR #2831, opened 2024-11-02
chore(docs): update docs content

dbritto-dev:
## Summary

Update docs content

## Check List

### DIFF
--- a/docs/guides/how-to-reset-state.md
+++ b/docs/guides/how-to-reset-state.md
@@ -44,7 +44,7 @@ const useSlice = create<State & Actions>()((set, get) => ({
 Resetting multiple stores at once
 
 ```ts
-import type * as zustand from 'zustand'
+import type { StateCreator } from 'zustand'
 import { create: actualCreate } from 'zustand'
 
 const storeResetFns = new Set<() => void>()
@@ -56,7 +56,7 @@ const resetAllStores = () => {
 }
 
 export const create = (<T>() => {
-  return (stateCreator: zustand.StateCreator<T>) => {
+  return (stateCreator: StateCreator<T>) => {
     const store = actualCreate(stateCreator)
     const initialState = store.getInitialState()
     storeResetFns.add(() => {
@@ -64,7 +64,7 @@ export const create = (<T>() => {
     })
     return store
   }
-}) as typeof zustand.create
+}) as typeof actualCreate
 ```
 
 ## CodeSandbox Demo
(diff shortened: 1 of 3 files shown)


docs/guides/how-to-reset-state.md
```diff
@@ -44,8 +44,8 @@ const useSlice = create<State & Actions>()((set, get) => ({
 Resetting multiple stores at once
 
 ```ts
-import { create as _create } from 'zustand'
-import type { StateCreator } from 'zustand'
+import type * as zustand from 'zustand'
```
dai-shi, line 47: Personally, I'm not a big fan of this. Can we keep the previous style?

dbritto-dev: Sure

dbritto-dev:
@dai-shi all yours

docs/guides/testing.md
```diff
@@ -83,8 +83,9 @@ In the next steps we are going to setup our Jest environment in order to mock Zu
 
 ```ts
 // __mocks__/zustand.ts
-import * as zustand from 'zustand'
 import { act } from '@testing-library/react'
+import type * as zustand from 'zustand'
```
dai-shi, line 87: Can you do the same for others please?

dbritto-dev: In this case we use it to type the import from `requireActual` or

dbritto-dev: @dai-shi keep me in the loop

dai-shi: Oh, I see L91.

Can we do something like this? Feels a little more explicit.
```suggestion
import type * as ZustandExportedTypes from 'zustand'
```

dbritto-dev: @dai-shi sure

dbritto-dev:
@dai-shi all yours

dai-shi (approved):
Thanks. I'm not sure if it's common practice, so we can change it if someone has suggestions.

loddit:
I noticed that `xIsNext` might not be necessary as an independent state ([React State Design Principle](https://react.dev/learn/thinking-in-react#step-3-find-the-minimal-but-complete-representation-of-ui-state)). The Board component can derive `player` from squares, for example:
```
const player = (squares.filter(v => ['O', 'X'].includes(v)).length % 2) ? 'O' : 'X'
```
was it designed this way for a specific reason?

PR #2836, opened 2024-11-05
docs: fix typo in tutorial-tic-tac-toe.md

ytraddan:

There is a  mistake in the store name

Fixes #
It's should be `useGameStore`  as seen in the code block above

wrong : `useSquaresStore` declares a `squares` state that's initially set to that array. Each entry in the
correct : `useGameStore` declares a `squares` state that's initially set to that array. Each entry in the

## Check List


dbritto-dev (approved):
LGTM

dbritto-dev:
@ytraddan thanks

PR #2850, opened 2024-11-11
test: add jest-dom matchers and eslint plugin

kretajak:
## Summary

Testing Library provides nice set of custom matchers that are more declarative and easier to understand. Along with `@testing-library/jest-dom` there is an accompanying `eslint-plugin-jest-dom` which enforce using these selectors. 

## Check List


dai-shi (approved):
LGTM

PR #2880, opened 2024-12-04
chore(deps): update dev dependencies

dai-shi:
for v5.0.2


package.json
```diff
@@ -114,40 +114,40 @@
...
+    "react-dom": "19.0.0-rc.1",
     "redux": "^5.0.1",
-    "rollup": "^4.24.3",
+    "rollup": "^4.28.0",
     "rollup-plugin-esbuild": "^6.1.1",
     "shelljs": "^0.8.5",
     "shx": "^0.3.4",
-    "typescript": "^5.6.3",
+    "typescript": "^5.7.2",
     "use-sync-external-store": "^1.2.2",
```
goosewobbler, line 149: React 19 just got released, this needs to be updated as I don't think 1.2.2 supports R19

dai-shi: This is just a dev dependency, and the peer dependency is specified as a range, so it shouldn't be the actual problem.

We will update it for the next release anyway.

dai-shi: Actually, "devDependecies" is not included in the published package.json.
See: https://unpkg.com/browse/zustand@5.0.2/package.json

goosewobbler: Ah ok, ignore me.  I doubt this is the issue indeed.  I'll wait for the next release before trying R19 again.

dai-shi: > I doubt this is the issue indeed. I'll wait for the next release before trying R19 again.

So, do you have an actual issue? The next release doesn't change it. Please report your issue.

goosewobbler: I just fixed the issue, one of my tests was throwing `TypeError: Cannot read properties of null (reading 'useSyncExternalStore')`, turned out to be a dependency issue, fixed by adding zustand to devDeps...

PR #2883, opened 2024-12-06
chore(docs): Adds import statements to devtools code example

atticoos:
## Summary

Came to this page to double check where I import `devtools` from, noticed it missing from the code sample

## Check List


dai-shi (approved):
thx!

PR #2888, opened 2024-12-09
chore(.nvmrc): specify node version v18

sukvvon:
## Summary

* add `.nvmrc` node version 18 for [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth)
* update `actions/setup-node` to use `node-version-file` instead of `node-version`

## Check List


dai-shi:
For now, we don't use nvmrc. "engines" should be the single source of truth, but I'm afraid it's not maintained. Let me see what we can do.

sukvvon:
> For now, we don't use nvmrc. "engines" should be the single source of truth, but I'm afraid it's not maintained. Let me see what we can do.

@dai-shi The engines field specified in the package.json indicates that this library requires Node.js version 12.20.0 or higher for users. Based on the workflows, I inferred that the development environment uses Node.js version 18 or higher, which is why I added the .nvmrc file.

dai-shi:
It was my misunderstanding. You want the source of truth for developing zustand. Never mind, let's leave it for now. You can develop it with node 20 or something.

sukvvon:
> It was my misunderstanding. You want the source of truth for developing zustand. Never mind, let's leave it for now. You can develop it with node 20 or something.

@dai-shi I think nvm is widely used, and [the official Node.js documentation also recommends using nvm when installing on macOS first](https://nodejs.org/en/download/package-manager). Having the Node.js version specified provides more convenience for developers in terms of DX, as it eliminates the need to worry about which version to use when downloading node_modules and starting development. Whether it’s Node 18 or 20, it saves time spent on making that decision.

if you dont care about node version 18 or 20 etc..., I will change the Node version to 20 inside the .nvmrc file.

dai-shi:
Let's wait and see how others think.
We generally avoid having dot files.
I use nvm but none of my libraries have .nvmrc.
I'm more comfortable with having a guide in CONTRIBUTING.md.
(nvm is widely used, but not for everyone. and other people start to want their own config file, eventually.)

dbritto-dev:
Sorry, but no. Let's keep it simple, nvm is widely used but not the first choice

dai-shi:
Feel free to send a PR to improve CONTRIBUTING.md.

PR #2890, opened 2024-12-10
docs(CONTRIBUTING.md): change 'pnpm' to 'pnpm install' for installing dependencies

sukvvon:
## Summary

* The correct command is not `pnpm` but `pnpm install`

## Check List

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,7 +65,7 @@ you can use this suggested workflow:
 
 - Fork this repository;
 - Create a new feature branch based on the `main` branch;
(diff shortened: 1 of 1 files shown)


dbritto-dev:
Same as `yarn install`, you can use `yarn` or `yarn install`

sukvvon:
@dbritto-dev but we are using `pnpm` not `yarn`, `yarn` is right command to install dependencies but `pnpm` is wrong, if we want it, we have to write `pnpm install` or `pnpm i`.

dbritto-dev:
@sukvvon in that case looks good to me, let me merge it

PR #2898, opened 2024-12-13
chore(.eslintrc.json): remove prettier in extends

sukvvon:
## Summary

* remove `prettier` in eslintrc.json's extends because we already use [`plugin:prettier/recommended`, which includes `prettier`](https://github.com/prettier/eslint-plugin-prettier?tab=readme-ov-file#installation)
* 
## Check List


dai-shi (approved):
Thanks!

PR #2900, opened 2024-12-14
Rewrite the documentation about why the typescript types can not be inferred

GuichiZhao:
As discussed in  https://github.com/pmndrs/zustand/discussions/2895

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -26,65 +26,51 @@ const useBearStore = create<BearState>()((set) => ({
 
   <br/>
 
-**TLDR**: Because state generic `T` is invariant.
+**TLDR**: Because we just provide the information for the returned type but no information for the parameter type.
 
 Consider this minimal version `create`:
 
-```ts
+``` ts
 declare const create: <T>(f: (get: () => T) => T) => T
-
-const x = create((get) => ({
+const x = create(() => ({
   foo: 0,
-  bar: () => get(),
+  bar: ''
 }))
-// `x` is inferred as `unknown` instead of
-// interface X {
-//   foo: number,
-//   bar: () => X
+// const x: {
+//     foo: number;
+//     bar: string;
 // }
 ```
-
-Here, if you look at the type of `f` in `create`, i.e. `(get: () => T) => T`, it "gives" `T` via return (making it covariant), but it also "takes" `T` via `get` (making it contravariant). "So where does `T` come from?" TypeScript wonders. It's like that chicken or egg problem. At the end TypeScript, gives up and infers `T` as `unknown`.
-
-So, as long as the generic to be inferred is invariant (i.e. both covariant and contravariant), TypeScript will be unable to infer it. Another simple example would be this:
+Here, if we do not provide a `get` param for `f` function, `x` can be inferred properly. However, if we provide the `get` function
 
 ```ts
-const createFoo = {} as <T>(f: (t: T) => T) => T
-const x = createFoo((_) => 'hello')
-```
-
-Here again, `x` is `unknown` instead of `string`.
-
-  <details>
-    <summary>More about the inference (just for the people curious and interested in TypeScript)</summary>
-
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -26,65 +26,54 @@ const useBearStore = create<BearState>()((set) => ({
...
 
 Consider this minimal version `create`:
 
 ```ts
 declare const create: <T>(f: (get: () => T) => T) => T
-
-const x = create((get) => ({
+const x = create(() => ({
   foo: 0,
-  bar: () => get(),
```
dai-shi, line 38: I think this is required to reproduce the problem. Otherwise, it would have been easier.

dai-shi (commented):
I'll leave @devanshj for detailed reviews.

docs/guides/typescript.md
GuichiZhao: Bring back `bar: () => get()` and add some text about how the ts behavior related to real-world scenario

devanshj:
The problem with this PR is that it makes it look as if TypeScript is doing it's best and is correct whereas it can do better. I've opened [an issue in TypeScript](https://github.com/microsoft/TypeScript/issues/60929) to know which feature will enable the inference we want.

Once we know that we can just link to that issue. I don't think we need this explanation of how TypeScript's current inference algorithm works as it's not that complicated. Or maybe we can keep it. I'll let @dai-shi decide that.

hnhattien:
## jdada

hnhattien:
Good

GuichiZhao:
> The problem with this PR is that it makes it look as if TypeScript is doing it's best and is correct whereas it can do better. I've opened [an issue in TypeScript](https://github.com/microsoft/TypeScript/issues/60929) to know which feature will enable the inference we want.
> 
> Once we know that we can just link to that issue. I don't think we need this explanation of how TypeScript's current inference algorithm works as it's not that complicated now. Or maybe we can keep it. I'll let @dai-shi decide that.

After reading the https://github.com/microsoft/TypeScript/issues/49618 and https://github.com/microsoft/TypeScript/issues/60922 as your issue referred, I admit that there are some point I am missing, it **does** relate to the "chicken and egg" problem

Let me approve this:
There can be an even simpler case 
``` typescript
function test<T>(fn: (prev: T) => T) { }
test((prev) => ({ a: 1 })); // T is inferred as "unknown"
```
My initial point:
The reason why `T` is `unknown` is that there is no hint for `prev`

And the point can be easily broken if we try `NoInfer`

``` typescript
function test<T>(fn: (prev: NoInfer<T>) => T) { }
test((prev) => ({ a: 1 })); // T is inferred as "unknown"
```
We already tell the engine that never cares about `prev` when you did the inference, but it still does not work, so the root case has nothing to do with the "no type information"

As the typescript team pointed out
``` typescript
function test<T>(fn: (prev: NoInfer<T>) => T) { }
test((prev) => {
  return prev ? 1:0
});
```
if the output type relates to input type,  typescript will refuse to do anything
it seems to me that the MS team believes such usage is an anti-pattern anyway.

> That said, I'm not really sure what the endgoal is. Any "improvement" we make here is just going to sow a bunch of "TypeScript is inconsistent, therefore has bug" reports because people will wonder why
https://github.com/microsoft/TypeScript/issues/49618#issuecomment-1162391274




maybe we should consider avoiding such usage altogether
``` typescript
declare const create: <T>() => T

const x = create(() => ({
  foo: 0,
  bar: () => {
    x.setState({foo:x.getState().foo+1})
  },
}))

// OR
const y=create(()=>({
  foo: 0,
})) 

const bar=() => {
  y.setState({foo:y.getState().foo+1})
}
```
Or simply use https://zustand.docs.pmnd.rs/middlewares/combine

@devanshj

dai-shi:
As I wrote somewhere, Zustand's API design is JavaScript first, not TypeScript first. We are not changing the API anytime soon. The current typing is hacky, but it's been doing well. Moving forward, I think what we should do is to recommend `combine` or `zustand-slices` or some other slicing utils.

dai-shi:
we would like to keep the advanced ts guide as is. thanks for your suggestion.

PR #2904, opened 2024-12-15
chore: pkg.pr.new

Aslemammad:
Adding pkg.pr.new to zustand!


Aslemammad:
I think we should add the pkg.pr.new app right now!

.github/workflows/cr.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - run: corepack enable
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'pnpm'
+      - run: pnpm install
+      - run: pnpm build
+      - run: pnpm dlx pkg-pr-new publish './dist' --compact --template './examples/*'
```
dai-shi, line 17: Can you add a newline at the end?

BTW, we may want a simpler example like `examples/counter`. `examples/demo` might be too big for testing.
If you don't have time, let me know so that I can take it.

dai-shi (commented):
I have been waiting for it!

.github/workflows/cr.yml
sukvvon: @dai-shi 

> Can you add a newline at the end?

In this case, i can fix it in new PR ASAP

dai-shi: We do have this anyway: https://github.com/pmndrs/jotai/pull/2862
Let's see if @Aslemammad responds in 24 hours.

@sukvvon Would you be interested in creating a simple counter app in `examples/counter`?

sukvvon: @dai-shi  i knew that counter app is already exist in `examples/demo`, could you explain your purpose?

dai-shi: demo has many dependencies for visual effects. for testing purpose with pkg.pr.new, a simpler counter example would be easier.

btw, we also have an issue related with examples in valtio: https://github.com/pmndrs/valtio/discussions/1012

sukvvon: > BTW, we may want a simpler example like examples/counter. examples/demo might be too big for testing.

@dai-shi do you wanna separating it?

Aslemammad: feel free to tackle it @sukvvon i'll meanwhile finish this PR reviews and jotai's PR too.

dai-shi:
This is nice. I didn't like the show toggle.

image

dai-shi:
Looks good.

image

.github/workflows/cr.yml
sukvvon: > @sukvvon Would you be interested in creating a simple counter app in examples/counter?

In this case, I can write the logic, but it’s hard for me to clearly understand what is desired stylistically, so I think it would be better for @dai-shi to handle it. 😄

dai-shi: No problem. I can handle it, but I don't do any styling.

PR #2912, opened 2024-12-18
chore(eslint): migrate to flat config and simplify

sukvvon:
## Summary

* Mirgrate legacy `.eslintrc` to `eslint.config.js`
* Simplify from the existing configuration

## Check List


dai-shi:
This is nice! I wasn't sure when all plugins are ready. They weren't when I last tried.
Are you sure all configs are migrated? No additions or removals?

sukvvon:
> This is nice! I wasn't sure when all plugins are ready. They weren't when I last tried. Are you sure all configs are migrated? No additions or removals?

@dai-shi i checked all plugins. if it's not i will keep my eyes on this

sukvvon:
> This is nice! I wasn't sure when all plugins are ready. They weren't when I last tried. Are you sure all configs are migrated? No additions or removals?

@dai-shi i checked all plugins. if it's not i will keep my eyes on this, i migrate with no additions and removals

dai-shi:
> I'll keep the current config.

Do you mean to keep using eslint 8?

sukvvon:
@dbritto-dev Tools like `oxc` and `biome` could be considered as alternatives, but since they are still under development, it might be too early to use them. I believe `eslint` is still more practical at this point, with abundant references, widespread adoption, and ease of application. For now, as we are currently using `eslint`, it would be better to focus on supporting `eslint` and consider simplifying later.

dbritto-dev:
I wouldn't put much effort on eslint is pretty slow and now are better options, apart from that biome is ready for production

dai-shi:
Our ecosystem still depends on eslint plugins. Maybe, eslint-plugin-react-compiler is one of the biggest hurdles. Let's stick with eslint.

dbritto-dev:
@dai-shi do we need this change to support `react-compiler` plugin?

dai-shi:
react-compiler is already included. eslint 8 will no longer be supported. This change is what I wanted to do when eslint 9 was released, but at that point something wasn't ready.

dai-shi (commented):
Let's do some practices that I learned with https://github.com/jotaijs/jotai-valtio/pull/11.
The primary goal is to simplify configs and prefer defaults.

- Use `"lint": "eslint ."` in package.json and move linter configs to config file
- Remove `eslint-config-prettier` package
- Can we use `eslint-import-resolver-typescript` package to simplify eslint.config.js?
- Can we use `eslint.config.mjs` and use ESM syntax?
- Let's remove `globals` packages. The added commend in `examples/demo/src/utils/copy-to-clipboard.js` seems good.
- Remove `examples/*/eslint.config.*` and eslint packages from `examples/*/package.json`
- No need to use`eslint-plugin-prettier` package, as we don't configure stylistic rules.
- Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

sukvvon:
> Let's do some practices that I learned with [jotaijs/jotai-valtio#11](https://github.com/jotaijs/jotai-valtio/pull/11). The primary goal is to simplify configs and prefer defaults.
> 
> * Use `"lint": "eslint ."` in package.json and move linter configs to config file
> * Remove `eslint-config-prettier` package
> * Can we use `eslint-import-resolver-typescript` package to simplify eslint.config.js?
> * Can we use `eslint.config.mjs` and use ESM syntax?
> * Let's remove `globals` packages. The added commend in `examples/demo/src/utils/copy-to-clipboard.js` seems good.
> * Remove `examples/*/eslint.config.*` and eslint packages from `examples/*/package.json`
> * No need to use`eslint-plugin-prettier` package, as we don't configure stylistic rules.
> * Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

@dai-shi I'll do it step by step

sukvvon:
* [x] Use "lint": "eslint ." in package.json and move linter configs to config file
* [ ] Remove eslint-config-prettier package
* [ ] Can we use eslint-import-resolver-typescript package to simplify eslint.config.js?
* [x] Can we use eslint.config.mjs and use ESM syntax?
* [ ] Let's remove globals packages. The added commend in examples/demo/src/utils/copy-to-clipboard.js seems good.
* [x] Remove examples/*/eslint.config.* and eslint packages from examples/*/package.json
* [ ] No need to useeslint-plugin-prettier package, as we don't configure stylistic rules.
* [ ] Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

dai-shi:
> so in my opinion, it's good to remain `eslint.config.js` in `examples/demo` but change file extension `js` to `mjs` that you want.

Good point. I missed it. Yeah, let's keep the file with `.mjs`.

dai-shi:
> we need `eslint-config-prettier`, `eslint-plugin-prettier`, because in `package.json` we have config prettier and we should reflect it in eslint, if we do not, we don't know why file is fixed and we can't see error in prettier

I'm not sure if I understand the problem. What will be the problem for example? I didn't see any issues in https://github.com/jotaijs/jotai-valtio/pull/11. Do you mean we should fix that too?

sukvvon:
> > so in my opinion, it's good to remain `eslint.config.js` in `examples/demo` but change file extension `js` to `mjs` that you want.
> 
> Good point. I missed it. Yeah, let's keep the file with `.mjs`.

in `examples/demo` in package.json

> > we need `eslint-config-prettier`, `eslint-plugin-prettier`, because in `package.json` we have config prettier and we should reflect it in eslint, if we do not, we don't know why file is fixed and we can't see error in prettier
> 
> I'm not sure if I understand the problem. What will be the problem for example? I didn't see any issues in [jotaijs/jotai-valtio#11](https://github.com/jotaijs/jotai-valtio/pull/11). Do you mean we should fix that too?

@dai-shi 

`eslint-config-prettier` prevents conflicts between ESLint and Prettier, while `eslint-plugin-prettier` integrates Prettier rules into ESLint for formatting checks.

So if we use them, we can see Prettier error with Eslint

sukvvon:
> I'm not sure if I understand the problem. What will be the problem for example? I didn't see any issues in https://github.com/jotaijs/jotai-valtio/pull/11. Do you mean we should fix that too?

@dai-shi we have to use it. if we use, DX will increase

sukvvon:
* [x] Use `"lint": "eslint ."` in package.json and move linter configs to config file
* [ ] Remove `eslint-config-prettier` package
* [ ] Can we use `eslint-import-resolver-typescript` package to simplify `eslint.config.js`?
* [x] Can we use eslint.config.mjs and use ESM syntax?
* [x] ~~Let's remove globals packages. The added commend in `examples/demo/src/utils/copy-to-clipboard.js` seems good.~~
* [x] ~~Remove` examples/*/eslint.config.*` and eslint packages from `examples/*/package.json`~~
* [ ] No need to use `eslint-plugin-prettier` package, as we don't configure stylistic rules.
* [ ] Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

examples/demo/eslint.config.js
```diff
@@ -0,0 +1,44 @@
+import eslint from '@eslint/js'
+import react from 'eslint-plugin-react'
+import reactHooks from 'eslint-plugin-react-hooks'
+import reactRefresh from 'eslint-plugin-react-refresh'
+import globals from 'globals'
+
+export default [
+  {
+    ignores: ['dist'],
+  },
```
sukvvon, line 10: Add ignores

examples/demo/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "lint": "eslint src --ext js,jsx --report-unused-disable-directives --max-warnings 0",
+    "lint": "eslint .",
```
sukvvon, line 9: Concise eslint script

examples/demo/eslint.config.js
```diff
@@ -0,0 +1,44 @@
...
+  eslint.configs.recommended,
+  react.configs.flat.recommended,
+  react.configs.flat['jsx-runtime'],
+  {
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        ...globals.es2020,
+      },
```
sukvvon, line 19: Add globals, it's related `.eslintrc.cjs` that is removed

package.json
```diff
@@ -65,11 +65,11 @@
     "build:traditional": "rollup -c --config-traditional",
     "postbuild": "pnpm patch-d-ts && pnpm copy && pnpm patch-old-ts && pnpm patch-esm-ts",
     "prettier": "prettier \"*.{js,json,md}\" \"{examples,src,tests,docs}/**/*.{js,jsx,ts,tsx,md,mdx}\" --write",
-    "eslint": "eslint --no-eslintrc --c .eslintrc.json --fix '*.{js,ts}' '{src,tests}/**/*.{ts,tsx}'",
+    "eslint": "eslint '*.{js,mjs,ts}' '{src,tests}/**/*.{ts,tsx}' --fix ",
     "test": "pnpm run '/^test:.*/'",
     "test:format": "prettier '*.{js,json,md}' '{examples,src,tests,docs}/**/*.{js,jsx,ts,tsx,md,mdx}' --list-different",
     "test:types": "tsc --noEmit",
-    "test:lint": "eslint  --no-eslintrc --c .eslintrc.json '*.{js,ts}' '{src,tests}/**/*.{ts,tsx}'",
+    "test:lint": "eslint '*.{js,mjs,ts}' '{src,tests}/**/*.{ts,tsx}'",
```
sukvvon, line 72: Add `mjs` because we have `eslint.config.mjs` file

eslint.config.mjs
```diff
@@ -0,0 +1,139 @@
...
+import reactCompiler from 'eslint-plugin-react-compiler'
+import reactHooks from 'eslint-plugin-react-hooks'
+import testingLibrary from 'eslint-plugin-testing-library'
+import globals from 'globals'
+import tseslint from 'typescript-eslint'
+
+export default tseslint.config(
+  {
+    ignores: ['**/dist/'],
+  },
```
sukvvon, line 17: Add ignores

sukvvon:
> I'm not sure if I understand the problem. What will be the problem for example? I didn't see any issues in [jotaijs/jotai-valtio#11](https://github.com/jotaijs/jotai-valtio/pull/11). Do you mean we should fix that too?

https://github.com/jotaijs/jotai-valtio/pull/11 In this PR, i thought you want to remove `eslint-plugin-prettier`, `eslint-config-prettier`. code in config file is right.

dai-shi:
> `eslint-config-prettier` prevents conflicts between ESLint and Prettier, while `eslint-plugin-prettier` integrates Prettier rules into ESLint for formatting checks.
> 
> So if we use them, we can see Prettier error with Eslint

I don't think it's recommended anymore or not a trend. We shouldn't see prettier errors through eslint.

dai-shi:
> I think we need to add `globals` because eslint team recommended it and it's existed in `examples/demo/.eslintrc.cjs` that is removed

Can you give me a link that eslint team recommends it?
I'm not trying to completely follow `.eslintrc.cjs` config in this PR. Maybe I should take it over because the requirement isn't precise.

sukvvon:
> > `eslint-config-prettier` prevents conflicts between ESLint and Prettier, while `eslint-plugin-prettier` integrates Prettier rules into ESLint for formatting checks.
> > So if we use them, we can see Prettier error with Eslint
> 
> I don't think it's recommended anymore or not a trend. We shouldn't see prettier errors through eslint.

## with prettier plugin

image

## without prettier plugin

image

sukvvon:
> > I think we need to add `globals` because eslint team recommended it and it's existed in `examples/demo/.eslintrc.cjs` that is removed
> 
> Can you give me a link that eslint team recommends it? I'm not trying to completely follow `.eslintrc.cjs` config in this PR. Maybe I should take it over because the requirement isn't precise.

https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables

@dai-shi Link here!

dai-shi:
@sukvvon Thanks for the info!
Hmmmm, I can't really explain why, but I strongly feel like dropping `eslint-plugin-prettier` and `globals`. I might be wrong, but for now, I would like to push this idea. I'll revisit if I get more feedback on it from others later.

sukvvon:
@dai-shi 

I understand. However, I believe that libraries exist to enhance the convenience of developers, their users. Furthermore, they exist because they are necessary. The libraries we create are no exception. We are using ESLint and Prettier, and I believe that following what they provide or recommend (or aim for) will help us reduce maintenance costs in the future. Additionally, since we cannot predict what side effects may occur, not using them might impose unexpected costs on us. I believe that following common practices is the best way to minimize our costs.

dai-shi:
These two points have different arguments.

- `eslint-plugin-prettier`: I don't think the editor should warn stylistic errors. It feels bad DX. Devs shouldn't care about styles. This can be controversial, but I want to disable it for now. I will reconsider it if many people submit PRs with wrong format and the reason is because of eslint not warning it.

- `globals`: I dislike the idea to automatically add some globals, unless it's too many. I don't think we use too many of them. How many errors do we see by removing `globals`?

dai-shi:
On second thought, using `globals` itself isn't that bad. It's just that we should use the config strictly selectively.

sukvvon:
> eslint-config-prettier prevents conflicts between ESLint and Prettier, while eslint-plugin-prettier integrates Prettier rules into ESLint for formatting checks.

If we are 100% certain that ESLint and Prettier will not conflict, it might not be necessary. Are you suggesting that we don’t integrate ESLint and Prettier?

eslint.config.mjs
```diff
@@ -0,0 +1,139 @@
...
+  eslint.configs.recommended,
+  tseslint.configs.recommended,
+  react.configs.flat['jsx-runtime'],
```
dai-shi, line 20: Can we add this?
```suggestion
  react.configs.flat.recommended,
  react.configs.flat['jsx-runtime'],
```

sukvvon: Sure!, i was thinking about this

sukvvon: my purpose was migration

eslint.config.mjs
```diff
@@ -0,0 +1,139 @@
...
+  importPlugin.flatConfigs.errors,
+  importPlugin.flatConfigs.warnings,
+  {
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        ...globals['shared-node-browser'],
+        ...globals.node,
+        ...globals.es2015,
```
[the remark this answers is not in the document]
sukvvon: i agree with you

examples/demo/eslint.config.js
```diff
@@ -0,0 +1,44 @@
...
+        sourceType: 'module',
+      },
+    },
+    plugins: {
+      'react-hooks': reactHooks,
+      'react-refresh': reactRefresh,
+    },
+    settings: {
+      react: {
+        version: '18.2',
```
dai-shi, line 31: Does this work?

```suggestion
        version: 'detect',
```

dai-shi:
> my purpose was migration

Sorry, it wasn't clear. As you have good knowledge, I think it's a good timing to refine it. So, my purpose isn't just migration.

sukvvon:
> > my purpose was migration
> 
> Sorry, it wasn't clear. As you have good knowledge, I think it's a good timing to refine it. So, my purpose isn't just migration.



> Let's convert to draft while the work is in progress.

I agree with you 👍

examples/demo/eslint.config.js
sukvvon: 100% works

dai-shi:
> Are you suggesting that we don’t integrate ESLint and Prettier?

Yes.

I wasn't sure if we need eslint-config-prettier, but as I read the docs, recommended rules don't have stylistic rules, so I assume we don't need eslint-config-pretiter. I might be wrong, but we can disable stylistic rules selectively if we find any issues.

sukvvon:
> > Are you suggesting that we don’t integrate ESLint and Prettier?
> 
> Yes.
> 
> I wasn't sure if we need eslint-config-prettier, but as I read the docs, recommended rules don't have stylistic rules, so I assume we don't need eslint-config-pretiter. I might be wrong, but we can disable stylistic rules selectively if we find any issues.

@dai-shi Are you referring to the recommended configuration of `eslint-plugin-prettier`? Are you suggesting removing `eslint-config-prettier` and keeping only `eslint-plugin-prettier`, or are you suggesting removing both `eslint-config-prettier` and `eslint-plugin-prettier`?

dai-shi:
> Are you referring to the `recommended` configuration of `eslint-plugin-prettier`?
> 
> Are you suggesting removing `eslint-config-prettier` and keeping only `eslint-plugin-prettier`, or are you suggesting removing both `eslint-config-prettier` and `eslint-plugin-prettier`?

Sorry for the confusion. Let's remove both `eslint-config-prettier` and `eslint-plugin-prettier`. I'm pretty sure we don't need them.

dbritto-dev:
At this point biome seems a good option 😅

dai-shi:
> At this point biome seems a good option 😅

We don't use biome now whatever you say 😄 , but I'm curious how you can configure eslint-plugin-react-compiler with biome. I might use it in my other projects, if that's fairly easy.

sukvvon:
image

I think when we remove `eslint-plugin-prettier` and `eslint-config-prettier`, `tslib` remove

sukvvon:
I will roll back to the version where it was removed, only in the eslint.config.mjs.

dai-shi:
tslib seems a peer dependency of `@rollup/plugin-typescript`, so we need to install it  explicitly after removing `eslint-*-prettier`.

dbritto-dev:
> > At this point biome seems a good option 😅
> 
> We don't use biome now whatever you say 😄 , but I'm curious how you can configure eslint-plugin-react-compiler with biome. I might use it in my other projects, if that's fairly easy.

@dai-shi sure, I'll create a PR for that so we can compare them (eslint vs biome)

dai-shi:
@dbritto-dev You don't need to create a PR. I just need to see how it can be configured. Maybe one of your projects?

dbritto-dev:
@dai-shi sure

sukvvon:
* [x] Use `"lint": "eslint ."` in package.json and move linter configs to config file
* [x] Remove `eslint-config-prettier` package
* [x] Can we use `eslint-import-resolver-typescript` package to simplify `eslint.config.js`?
* [x] Can we use eslint.config.mjs and use ESM syntax?
* [x] ~~Let's remove globals packages. The added commend in `examples/demo/src/utils/copy-to-clipboard.js` seems good.~~
* [x] ~~Remove` examples/*/eslint.config.*` and eslint packages from `examples/*/package.json`~~
* [x] No need to use `eslint-plugin-prettier` package, as we don't configure stylistic rules.
* [ ] Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

@dai-shi I think we almost done!

sukvvon:
> Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

@dai-shi Let's do this step by step

sukvvon:
> Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

@dai-shi I gonna do this until today

sukvvon:
* [x] Use `"lint": "eslint ."` in package.json and move linter configs to config file
* [x] Remove `eslint-config-prettier` package
* [x] Can we use `eslint-import-resolver-typescript` package to simplify `eslint.config.js`?
* [x] Can we use `eslint.config.mjs` and use ESM syntax?
* [x] ~~Let's remove globals packages. The added commend in `examples/demo/src/utils/copy-to-clipboard.js` seems good.~~
* [x] ~~Remove` examples/*/eslint.config.*` and eslint packages from `examples/*/package.json`~~
* [x] No need to use `eslint-plugin-prettier` package, as we don't configure stylistic rules.
* [x] Remove some custom rules that don't cause "errors" in the current zustand/jotai/valtio code. (we keep import order rules.)

@dai-shi Done!

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+        typescript: {
+          alwaysTryTypes: true,
+          project: 'tsconfig.json',
+        },
+      },
+    },
+    rules: {
+      eqeqeq: 'error',
+      curly: ['warn', 'multi-line', 'consistent'],
+      'no-console': 'off',
```
sukvvon, line 59:
```suggestion
      'no-console': 'warn',
```
@dai-shi Since `no-console` is not configured by `eslint.configs.recommended`, it is an unnecessary setting, so I was planning to remove it. However, if you want to avoid the use of `console.log` in the code, how about changing its setting `off` to `warn` to improve code quality?

dai-shi: we allow `console.log`.

sukvvon: Ok, i remove it

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+  tseslint.configs.recommended,
+  react.configs.flat.recommended,
+  react.configs.flat['jsx-runtime'],
+  {
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        ...globals.node,
+        ...globals.es2015,
+      },
```
dai-shi, line 26: I'm still concerning this, but I might be misunderstand something, so I'll take it over.

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+      parser: tseslint.parser,
+      parserOptions: {
+        ecmaVersion: 2018,
+        sourceType: 'module',
+        ecmaFeatures: {
+          jsx: true,
+        },
+      },
```
dai-shi, line 34: As we use `tseslint.config`, can it be just default?

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+    },
+    plugins: {
+      import: importPlugin,
```
dai-shi, line 37: I didn't know, but does this mean that if we use custom rules, we can't use `importPlugin.flatConfigs.recommended`, can we?

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+      },
+      'import/extensions': ['.js', '.jsx', '.ts', '.tsx'],
+      'import/parsers': {
+        '@typescript-eslint/parser': ['.ts', '.tsx'],
+      },
+      'import/resolver': {
+        typescript: {
+          alwaysTryTypes: true,
+          project: 'tsconfig.json',
+        },
```
dai-shi, line 53: How about `typescript: true`?

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+      'react-hooks': reactHooks,
+    },
+    settings: {
+      react: {
+        version: 'detect',
+      },
+      'import/extensions': ['.js', '.jsx', '.ts', '.tsx'],
+      'import/parsers': {
+        '@typescript-eslint/parser': ['.ts', '.tsx'],
+      },
```
dai-shi, line 48: What would be the default?

eslint.config.mjs
```diff
@@ -0,0 +1,127 @@
...
+      'import/resolver': {
+        typescript: {
+          alwaysTryTypes: true,
+          project: 'tsconfig.json',
+        },
+      },
+    },
+    rules: {
+      eqeqeq: 'error',
```
dai-shi, line 57: Isn't it `error` by default?

dai-shi (commented):
Thanks for working on it! Looks much better.

eslint.config.mjs
sukvvon: https://eslint.org/docs/latest/rules/eqeqeq

@dai-shi `eqeqeq` is not included in `eslint.configs.recommended`, so it needs to be explicitly configured as it is now.

eslint.config.mjs
sukvvon: https://github.com/import-js/eslint-plugin-import/blob/main/config/flat/recommended.js

@dai-shi I considered using `importPlugin.flatConfigs.recommended`, but since there are settings we would need to set `off` and others that would be overridden, I decided to configure it as it is now. I felt it would be better to avoid having `off` settings. The behavior remains the same as before.”

eslint.config.mjs
sukvvon: @dai-shi I’ll change it as you suggested. 👍

eslint.config.mjs
dai-shi: that also makes sense. i'll see how it goes. thanks for your suggestion.

eslint.config.mjs
sukvvon: * https://github.com/import-js/eslint-plugin-import/blob/main/config/recommended.js#L24-L27
* https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/configs/base.ts#L13-L16
* https://github.com/jsx-eslint/eslint-plugin-react/blob/master/index.js#L41-L44
* https://github.com/jsx-eslint/eslint-plugin-react/blob/master/index.js#L82-L87

@dai-shi The properties specified in the link are all merged, so they are the same as the ones currently set in `eslint.config.mjs`. Since it is a duplicate configuration, I will remove it.

eslint.config.mjs
sukvvon: * default [`import/parsers`](https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#importparsers) is same with our file
* default [`import/extensions`](https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#importextensions) is `js` (but we set react in shared config so default is `js`, `jsx`), but we are using `typescript` without `importPlugin.flatConfigs.typescript`, so we have to add `ts`, `tsx`. I think may be we can combine [`importPlugin.flatConfigs.typescript`](https://github.com/import-js/eslint-plugin-import/blob/main/config/typescript.js) with out custom configs like [this link](https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#typescript). I try it

dai-shi: I don't specify the extensions in https://github.com/jotaijs/jotai-valtio/pull/11. I think there must be some defaults we can rely on.

eslint.config.mjs
sukvvon: image

this is jotal-valtio code, as you see, eslint don't know where `console` comes from. so we have to sepecify with `globals`

dai-shi: thanks for confirming. there might be some misconfiguration in my editor. i'll look into it!

sukvvon: * https://github.com/sindresorhus/globals/blob/main/data/browser.mjs
* https://github.com/sindresorhus/globals/blob/main/data/node.mjs
  * https://github.com/sindresorhus/globals/blob/main/data/nodeBuiltin.mjs
  * https://github.com/sindresorhus/globals/blob/main/data/commonjs.mjs
* https://github.com/sindresorhus/globals/blob/main/data/es2015.mjs

@dai-shi These code snippets will help you understand.

dai-shi: I don't think I'm misunderstanding what it is. I'm just against to it.

dai-shi: > As you can see, this is the https://github.com/jotaijs/jotai-valtio/pull/11. ESLint does not know the source of console.

Yeah, I can reproduce it in eslint.config.js. But, if I write console.log in src/index.ts, it doesn't complain. I expect it to complain without `globals`. Can you help dig into it? It feels like `extensions` works for something and doesn't work for something else.

sukvvon: @dai-shi 

I think `js` and `jsx` rely on `globals` setting for recognizing global variables, whereas `ts` and `tsx` already have global type support provided by TypeScript, making the setting unnecessary. I also tested it, and the error only occurred in `js`.

So, I think in the top-level `eslint.config.mjs`, our main logic in the `ts` and `tsx` files doesn’t need `globals`, and since the purpose of the `js` file is for configuration, setting only `globals.node` should be enough.

sukvvon: > image
> @dai-shi As you can see, this is the [`jotal-valtio`](https://github.com/jotaijs/jotai-valtio/pull/11). ESLint does not know the source of `console`. Therefore, we need to inform ESLint of its source through `globals`.

@dai-shi Also this was a`js` file

dai-shi: Hmm, alright, that's not what I originally proposed, but if it's already configured, I wouldn't rather disable it.

We basically use ts/tsx only. So,

> since the purpose of the `js` file is for configuration, setting only `globals.node` should be enough.

Yeah, and we don't use `console` in config files, so `globals.node` isn't necessary. (If we really need it, we can reconsider or use inline config.)

sukvvon: @dai-shi I gonna remove `globals`

eslint.config.mjs
sukvvon:
```suggestion
      parserOptions: {
        ecmaVersion: 2018,
      },
```

@dai-shi I think we can change this like this

dai-shi: looks good. what's the default `ecmaVersion` if we don't specify?

sukvvon: @dai-shi It seems that `ecmaVersion` does not have a default value as of now

sukvvon:
```suggestion
```

https://eslint.org/docs/latest/use/configure/language-options#specifying-javascript-options

@dai-shi In this document `ecmaVersion` default value is `latest`

dai-shi: yeah, we are transpiling anyway. let's use the default `latest` for now until we really find any issues.

sukvvon: https://github.com/import-js/eslint-plugin-import/blob/main/config/recommended.js#L24-L27

@dai-shi However, `importPlugin.flatConfigs.recommended` sets `ecmaVersion: 2018`, so if we follow this, it seems we can set `ecmaVersion` accordingly.

sukvvon: > yeah, we are transpiling anyway. let's use the default `latest` for now until we really find any issues.

@dai-shi We can set it to `ecmaVersion: 2018`, but do you want to use `latest` for now until any issues arise?

dai-shi: > However, `importPlugin.flatConfigs.recommended` sets `ecmaVersion: 2018`,

If that's the case, I change my mind from https://github.com/pmndrs/zustand/pull/2912#discussion_r1897397153, and suggest to use `importPlugin.flatConfigs.recommended` and make some rules `off`.

sukvvon: > > However, `importPlugin.flatConfigs.recommended` sets `ecmaVersion: 2018`,
> 
> If that's the case, I change my mind from [#2912 (comment)](https://github.com/pmndrs/zustand/pull/2912#discussion_r1897397153), and suggest to use `importPlugin.flatConfigs.recommended` and make some rules `off`.

@dai-shi Ok, i'll reflect your suggest

eslint.config.mjs
sukvvon: > I don't specify the extensions in [jotaijs/jotai-valtio#11](https://github.com/jotaijs/jotai-valtio/pull/11). I think there must be some defaults we can rely on.

@dai-shi I’ll look into it.

sukvvon: @dai-shi 

```js
'import/extensions': ['.js', '.jsx', '.ts', '.tsx'],
'import/parsers': {
  '@typescript-eslint/parser': ['.ts', '.tsx'],
},
```

I believe that `import/resolver: { typescript: true }` works as it utilizes `eslint-import-resolver-typescript`, which seems to cover the functionality of the above settings. Therefore, these configurations can likely be omitted while still maintaining functionality. However, this is just a hypothesis and not a confirmed fact. Hence, it’s better to follow the recommendations in this [link](https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#typescript) when configuring.

dai-shi: To be honest, the readme in eslint-plugin-import repo doesn't seem well documented, and I don't trust it.

> However, this is just a hypothesis and not a confirmed fact.

Even if the hypothesis is broken, we only see CI errors, right? Note that our code is basically all TypeScript and can ignore .js in the worst case.

sukvvon: @dai-shi I understand, i gonna remove it

dai-shi: Thanks. I might be wrong, but I'd like to start minimal and fix it as we find actual issues.

sukvvon:
@dai-shi We're ready to merge

eslint.config.mjs
sukvvon: @dai-shi Alright, let’s keep an eye on it.

package.json
dai-shi: I would like `"eslint ."` here too, and also for `--fix` above.

dai-shi (commented):
Other than the comment, it looks good to me.

I still wonder if we should have eslint for demo, but let's put it aside for this one.

Can you update Jotai and Valtio PRs too?

sukvvon:
> Other than the comment, it looks good to me. Thanks so much for your continued effort!
> 
> (I still wonder if we should have eslint for demo, but let's put it aside for this one.)

> Can you update Jotai and Valtio PRs too?

sukvvon:
> Can you update Jotai and Valtio PRs too?

@dai-shi Of course, I’ll do the same for Jotai and Valtio!

package.json
sukvvon: @dai-shi The reason I configured it this way is simply to download `./pnpm-lock.yaml` and run ESLint. Since the dependencies `@react-three/drei`, `@react-three/fiber`, and `@react-three/postprocessing`, etc... are not included within it, errors occur.

However, I understand your intention, so after this PR is merged, I will create a new PR to address it accordingly.

dai-shi: Ohhh, now I remember that. Sorry about the repeated asking.
Hm, I don't know honestly what solution I like. Definitely we don't want monorepo. So, the current solution seems good.

I would probably try adding `examples/` in `ignores` in eslint.config.mjs. How you tried before?

dai-shi (approved):
This should be good, but one last comment.

package.json
sukvvon: @dai-shi 

> I would probably try adding `examples/` in `ignores` in eslint.config.mjs. How you tried before?

Oh, I hadn’t thought of that approach, that sounds great, i gonna reflect your idea 👍

sukvvon: @dai-shi 

> Hm, I don't know honestly what solution I like. Definitely we don't want monorepo. So, the current solution seems good.

After it is merged, I will think of a better way to utilize `eslint` for the `examples`.

dai-shi (approved):
Looks good to me! Thanks for your contribution!

dai-shi:
> After this PR is merged, I will think of a better way to utilize `eslint` for the `examples`.

Please finish up zustand and valtio PRs first.

PR #2918, opened 2024-12-19
chore(.github/workflows): improve workflow with pnpm dlx and caching

sukvvon:
## Summary

* add `cache-dependency-path` in `actions/setup-node`
* add `pnpm/action-setup

## Check List


.github/workflows/cr.yml
```diff
@@ -7,11 +7,12 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - run: corepack enable
+      - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 20
           cache: 'pnpm'
-      - run: pnpm install
+          cache-dependency-path: '**/pnpm-lock.yaml'
```
dai-shi, line 15: Default works, so we don't need it. Let's remove it from other `*.yml`.

.github/workflows/cr.yml
```diff
@@ -7,11 +7,12 @@ jobs:
...
+      - run: pnpm install --frozen-lockfile
```
dai-shi, line 16: frozen-lockfile is the default in CI, so let's remove it. https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

sukvvon:
@dai-shi Are you saying that explicitly specifying caching in cr.yml might not be necessary because caching is already handled in other actions?

dai-shi:
No, it works without specifying it because it's the default behavior.

sukvvon:
```yml
name: Publish Any Commit

on: [push, pull_request]

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: pnpm/action-setup@v4
      - uses: actions/setup-node@v4
        with:
          node-version: 20
          cache: 'pnpm'
      - run: pnpm install
      - run: pnpm build
      - run: pnpm dlx pkg-pr-new publish './dist' --compact --template './examples/*'
```

@dai-shi i understood, you want to simplify like this? by including only what is absolutely necessary.

dai-shi:
Yes!

sukvvon:
@dai-shi cool, i'll do it all

dai-shi (approved):
👏 Please update the PR description.

sukvvon:
@dai-shi Done!

dai-shi:
Thanks for your work for all three repos. I like it.

PR #2924, opened 2024-12-29
chore(eslint): change 'dist' area in 'ignores', sort rules

sukvvon:
## Summary

* Change `dist` area in `ignores`.
  * We can just ignore the `./dist`.
* Sort `rules` by the order `plugins` were added and by name.
  * It is to make it more predictable for the user and improve readability.

## Check List

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -69,13 +69,13 @@ export default tseslint.config(
           pathGroupsExcludedImportTypes: ['builtin'],
         },
       ],
+      '@typescript-eslint/no-explicit-any': 'off',
       '@typescript-eslint/no-unused-vars': [
         'warn',
         { argsIgnorePattern: '^_', varsIgnorePattern: '^_' },
       ],
-      '@typescript-eslint/no-explicit-any': 'off',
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -10,7 +10,7 @@ import tseslint from 'typescript-eslint'
 
 export default tseslint.config(
   {
-    ignores: ['**/dist/', 'examples/**'],
+    ignores: ['dist/**', 'examples/**'],
```
dai-shi, line 13: Actually, I don't think `**` is necessary.
```suggestion
    ignores: ['dist/', 'examples/'],
```

sukvvon: @dai-shi That sounds good! 👍

dai-shi (commented):
Can you apply this change to Valtio and Jotai PRs?

sukvvon:
> Can you apply this change to Valtio and Jotai PRs?

Sure! in Valtio PR, I already applied it

dai-shi:
Thanks for your contributions to all the PRs!

PR #2929, opened 2025-01-01
Update auto-generating-selectors.md

vorant94:

Fixes #

## Summary

fix typo in semicolon placement

## Check List


vorant94:
i thought this leading semicolon is just a typo, but now i see that it is added as a auto-fix by running prettier...

i heard about using leading semicolon for immediately-invoked functions, when you want to concatenate multiple JS files, but i don't think it is relevant here (not IIF, no JS file concatenation, simply code snippet in Markdown)

interesting prettier reminisce of the old days

PR #2931, opened 2025-01-02
chore(example/starter): update package manager to pnpm@9.15.2

jacob-lcs:
## Summary

This pull request includes a small change to the `examples/starter/package.json` file. The change specifies the package manager to be used for the project. So that the `pnpm-lock.yaml` file will not change when execute `pnpm i`.

* [`examples/starter/package.json`](diffhunk://#diff-0f8ffb6f82d2a180f9426495e2faf686abe5e5bb83ff7034eb442f0ab9f81b98R11): Added `"packageManager": "pnpm@9.15.2"` to specify the use of `pnpm` version 9.15.2 as the package manager.

## Check List

### DIFF
--- a/examples/starter/package.json
+++ b/examples/starter/package.json
@@ -8,6 +8,7 @@
     "build": "tsc && vite build",
     "preview": "vite preview"
   },
+  "packageManager": "pnpm@9.15.2",
   "dependencies": {
     "zustand": "^5.0.2",
     "react": "^18.3.1",


dai-shi:
Thanks for the suggestion, but it's our intention not to specify package manager in examples. Examples don't include lock files.

dai-shi:
And, noticed there is currently `./examples/demo/pnpm-lock.yaml` which is not intentional. @sukvvon Can we remove it?

jacob-lcs:
@dai-shi  Yes, `./examples/starter` also has a `pnpm-lock.yaml`

sukvvon:
@dai-shi When we modify code inside the examples directory or perform checks in dev mode, having the `pnpm-lock.yaml` file speeds up the installation process and ensures that the same dependency versions and environment are maintained. This plays an important role in keeping the collaboration consistent by ensuring that developers work in the same environment. I completely understand your intent, but since the `pnpm-lock.yaml` file has already been committed to Git, I don’t think there is a need to remove it. Additionally, it is generally not recommended to remove the `pnpm-lock.yaml` file.

dai-shi:
> since the `pnpm-lock.yaml` file has already been committed to Gi

I consider that it's committed by mistake. I'm sorry about my misreview.

> This plays an important role in keeping the collaboration consistent by ensuring that developers work in the same environment.

I've never experienced the inconsistency problem, actually. If there would be in the future, let's consider to include it.

The rationale is that we don't assume any specific package manager for examples.

> Additionally, it is generally not recommended to remove the pnpm-lock.yaml file.

I understand it's for the library code. Examples are more like docs in our case.

sukvvon:
@dai-shi It's not recommandation, but i think It's possible removing `pnpm-lock.yaml`, because it's just `examples`. Is the goal to remove something just because you think it’s unnecessary?

dai-shi:
Yes, it's confusing as they are not pnpm projects. (and that's why this PR is proposed.)
If we care about the consistency, I'm fine to remove caret `^` in package.json `dependencies`. (which isn't as close as lock files, though.)

sukvvon:
@dai-shi I understood what you want, i'll create a related PR and test it

dai-shi:
Thanks so much. I count on you for all three repos z/j/v.

PR #2933, opened 2025-01-03
chore(.github/workflows): change 'node-version' to 'lts/*' in 'actions/setup-node'

sukvvon:
## Summary

* Change `node-version` to `lts/*` to always use the latest LTS version of Node.js.

## Check List

### DIFF
--- a/.github/workflows/compressed-size-action.yml
+++ b/.github/workflows/compressed-size-action.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: lts/*
+          node-version: 'lts/*'
           cache: 'pnpm'
       - uses: preactjs/compressed-size-action@v2
         with:
--- a/.github/workflows/compressed-size-action.yml
+++ b/.github/workflows/compressed-size-action.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
           cache: 'pnpm'
       - uses: preactjs/compressed-size-action@v2
         with:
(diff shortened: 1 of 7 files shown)


dbritto-dev (approved):

.github/workflows/compressed-size-action.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
```
dbritto-dev, line 13: What about using `lts/*` instead? so we always get the latest lts version

```suggestion
          node-version: lts/*
```

sukvvon: @dbritto-dev Sounds good 👍

dbritto-dev:
@dai-shi all yours

dai-shi (commented):
I was writing those two comments previously, but it's already resolved!

dai-shi (approved):
https://github.com/actions/setup-node

> The node-version input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.

Yeah, `lts/*` seems good.

dai-shi:
Can you also send two PRs to jotai and valtio?

sukvvon:
> Can you also send two PRs to jotai and valtio?

I'll do it all

.github/workflows/test-multiple-versions.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: lts/*
```
dai-shi, line 41:
```suggestion
          node-version: `lts/*`
```

dai-shi (changes requested):
a leftover?

dai-shi (approved):
Thanks for your contribution as always!

PR #2935, opened 2025-01-05
feat(types): Make ExtractState public

vorant94:

Fixes https://github.com/pmndrs/zustand/discussions/2928

## Summary

after asking about how to infer state type outside of state declaration it happens that there is a dedicated but internal only type helper for it called `ExtractState`

this PR makes this helper public, moves it to `vanilla.ts` since its not react-specific and adds a short docs paragraph with its usage example

## Check List


dai-shi (approved):
Perfect!

PR #2938, opened 2025-01-06
docs(create.md): fix state selector syntax in troubleshooting example

HoberMin:

Fixes #

## Summary

Fixed incorrect state selector syntax in the troubleshooting example of create.md. The original code had a circular reference issue where person and setPerson were trying to reference themselves. Updated the selectors to correctly access the state and setPerson from the store state.

## Check List


dbritto-dev (approved):
LGTM

PR #2939, opened 2025-01-06
chore(rollup): migrate commonjs to esmodule

sukvvon:
## Summary

* LIke `eslint.config.mjs`, migrate `rollup.config.js` to `rollup.config.mjs`

## Check List

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -1,14 +1,14 @@
-/* eslint-disable no-undef */
-const path = require('path')
-const alias = require('@rollup/plugin-alias')
-const resolve = require('@rollup/plugin-node-resolve')
-const replace = require('@rollup/plugin-replace')
-const typescript = require('@rollup/plugin-typescript')
-const { default: esbuild } = require('rollup-plugin-esbuild')
+/*global process*/
+import path from 'path'
+import alias from '@rollup/plugin-alias'
+import resolve from '@rollup/plugin-node-resolve'
+import replace from '@rollup/plugin-replace'
+import typescript from '@rollup/plugin-typescript'
+import esbuild from 'rollup-plugin-esbuild'
 
 const extensions = ['.js', '.ts', '.tsx']
 const { root } = path.parse(process.cwd())
-const entries = [
+const _entries = [
   { find: /.*\/vanilla\/shallow\.ts$/, replacement: 'zustand/vanilla/shallow' },
   { find: /.*\/react\/shallow\.ts$/, replacement: 'zustand/react/shallow' },
   { find: /.*\/vanilla\.ts$/, replacement: 'zustand/vanilla' },
@@ -50,7 +50,9 @@ function createESMConfig(input, output) {
     output: { file: output, format: 'esm' },
     external,
     plugins: [
-      alias({ entries: entries.filter((e) => !e.find.test(input)) }),
+      alias({
+        entries: _entries.filter((entry) => !entry.find.test(input)),
+      }),
       resolve({ extensions }),
       replace({
         ...(output.endsWith('.js')
@@ -78,7 +80,9 @@ function createCommonJSConfig(input, output) {
     output: { file: output, format: 'cjs' },
     external,
     plugins: [
-      alias({ entries: entries.filter((e) => !e.find.test(input)) }),
+      alias({
+        entries: _entries.filter((e) => !e.find.test(input)),
+      }),
       resolve({ extensions }),
       replace({
         'import.meta.env?.MODE': 'process.env.NODE_ENV',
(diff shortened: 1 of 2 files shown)


rollup.config.mjs
```diff
@@ -1,14 +1,14 @@
...
+import alias from '@rollup/plugin-alias'
+import resolve from '@rollup/plugin-node-resolve'
+import replace from '@rollup/plugin-replace'
+import typescript from '@rollup/plugin-typescript'
+import esbuild from 'rollup-plugin-esbuild'
 
 const extensions = ['.js', '.ts', '.tsx']
 const { root } = path.parse(process.cwd())
-const entries = [
+const _entries = [
```
dai-shi, line 11: Don't use `_` prefix, which is reserved for "unused" variables.

sukvvon: @dai-shi i'll fix it

dai-shi:
we should make sure that the resulting `dist` doesn't change a single byte with this change.

sukvvon:
> we should make sure that the resulting `dist` doesn't change a single byte with this change.

@dai-shi Based on my observation, the size of the `dist` folder has not changed. The previous `dist` and the current `dist` code are identical. Do you know a reliable way to accurately check the size of the `dist` folder?

dai-shi:
> Do you know a reliable way to accurately check the size of the `dist` folder?

Checking size in't enough. `diff -r` should do.

sukvvon:
@dai-shi 

> > Do you know a reliable way to accurately check the size of the `dist` folder?
> 
> Checking size in't enough. `diff -r` should do.

```bash
diff -r before_dist after_dist
```

I executed the above command and confirmed that there are no changes.

dai-shi (approved):
Looks good. Let's do the same for valtio and jotai.

rollup.config.mjs
```diff
@@ -1,14 +1,14 @@
...
+import alias from '@rollup/plugin-alias'
+import resolve from '@rollup/plugin-node-resolve'
+import replace from '@rollup/plugin-replace'
+import typescript from '@rollup/plugin-typescript'
+import esbuild from 'rollup-plugin-esbuild'
 
 const extensions = ['.js', '.ts', '.tsx']
 const { root } = path.parse(process.cwd())
-const entries = [
+const aliasEntries = [
```
dai-shi, line 11: Looking at other two PRs, I wonder if we can revert this.

I also wonder why we return `[]` for the script. I'm sure I did it. 😓

dai-shi: > I also wonder why we return `[]` for the script. I'm sure I did it. 😓

#2942

sukvvon: @dai-shi I thought it was intentional.

sukvvon: @dai-shi I reflected https://github.com/pmndrs/zustand/pull/2942

dai-shi: Thanks. I should merge #2942 first and then ask you to check the diff again.

sukvvon: @dai-shi I already checked this case, i'm sure this fine

dai-shi: Well, technically after merging #2942, we should check it again. I'm sure it's fine, but let's make it sure.

dai-shi:
@sukvvon Reusing a PR is better than closing and reopening it. Hope you find out the way next time.

sukvvon:
@dai-shi 

> @sukvvon Reusing a PR is better than closing and reopening it. Hope you find out the way next time.

It was my mistake. I will do better next time. 🙇🏻‍♂️

sukvvon:
@dai-shi Thanks your advice

rollup.config.mjs
masungmin-dev: Hello, @dai-shi , @sukvvon 
I think There’s a side effect from the changes in version 5.0.3(with #2942)

when installing zustand using **npm alias install**, type references are not loaded properly.

It appears that this issue is happening in the script alias “patch-d-ts” ran with rollup.config's exported "entries" has been exported recently #2942 
(which modifies the relative paths in "dist" to point to the entry)

If the scripts and type definitions in the built output don’t use relative paths, using zustand with npm alias install becomes hard.

I would like to know if there was a specific reason for this process or can exclude

dai-shi: Hi, would you like to open a new discussion? Commenting on old PRs is hardly noticeable.

The use of entry points is intentional and all common in zustand, jotai and valtio. I think there was an issue with relative paths.

What's your problem with npm alias install, and how can we reproduce it with minimal steps?

masungmin-dev: Hi @dai-shi 
I’ve solved this issue. It appears that the problem was caused by using npm alias install together with a tsconfig.json that sets moduleResolution to "node", which prevented proper path resolution.

To reproduce the issue, when using zustand vanilla’s createStore along with middleware (subscribeWithSelector), the issue occurred in middleware.d.ts where type references from zustand/vanilla could not be properly resolved.

To fix this, I changed the moduleResolution to a more flexible mode, although it likely won’t work in "node" mode.

Thank you for your response 😀

PR #2958, opened 2025-01-15
Rename zustood to zustand-x

zbeyens:
We've renamed the package from `@udecode/zustood` to `zustand-x` a while back.


dai-shi (approved):
👍

PR #2987, opened 2025-01-30
feat(devtool): inferred action type

alimertcakar:
## Summary
image

With devtools() middleware, we can use Redux devtools extension to debug fired actions. But actions are shown as either "anonymous" or by the custom name we provided in set() function like: `set(foo, undefined, "state/setCount")`.

It is really tedious to provide a action name for every `set` call. Wouldn't it be nice to have the option to have it default to a more meaningful name?

Now we can do that by passing `inferActionName: true,` to devtools options. It finds the the function name which called `set()` by reading the current stack.

This works for firefox,chrome and edge. (There isn't a redux devtools extension on Safari anyways)
In node.js it's displayed as `Object.functionName` instead of `functionName`,  I don't think it matters as it's tests only.

## Check List


tests/devtools.test.tsx
```diff
@@ -189,6 +189,36 @@ describe('When state changes...', () => {
...
+          setCount: (newCount: number) => {
+            set({ count: newCount })
+          },
+        }),
+        options,
+      ),
+    )
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    api.getState().setCount(10)
```
alimertcakar, line 213: I don't know how to type this, I couldn't find a similar usage in this test

dai-shi:
This looks nice! I don't think we need `inferActionName` option, if this works reliably. Otherwise, I'd be a bit hesitant to merge.

tests/devtools.test.tsx
dai-shi: It should be able to do with `create<T>()(...)`.

alimertcakar:
> This looks nice! I don't think we need `inferActionName` option, if this works reliably. Otherwise, I'd be a bit hesitant to merge.

Seems to work reliably on the mock project I created with a simple store. 
I didn't test it on more complex projects. I can't get pnpm link --global zustand to work. `Failed to resolve entry for package "zustand". The package may have incorrect main/module/exports specified in its package.json.`

Is there a guide on local development? I'd like to meddle with zustand in some real life projects if possible

dai-shi:
I'm not sure what's wrong, but you can use pkg-pr-new  https://github.com/pmndrs/zustand/pull/2987#issuecomment-2625403458 if you don't find a local solution.

dai-shi (changes requested):
Remove `inferActionName` option.

dai-shi:
@alimertcakar Are you still around?

dai-shi:
@alimertcakar I hope you have a chance to look at it.

dai-shi:
@dbritto-dev Alright, I'll wait for it.

dbritto-dev (approved):

dbritto-dev:
@dai-shi all yours

dai-shi (approved):
LGTM

PR #2989, opened 2025-01-31
feat: add syncStore middleware for cross-tab state synchronization

fariborz0015:
## Description:

I have implemented a new middleware called `syncStore` which allows for synchronization of state between different tabs or windows in a browser using the `BroadcastChannel` API. If the browser does not support `BroadcastChannel`, it falls back to `localStorage` for state synchronization.  

### Key features:
- **Cross-tab synchronization**: Using `BroadcastChannel`, the state will be automatically synchronized between multiple open tabs of the same origin.
- **Fallback support**: If `BroadcastChannel` is not supported by the browser, it will use `localStorage` as a fallback.
- **Customizable sync fields**: You can specify which state fields you want to sync across tabs by using the `fields` option.
- **Updated documentation**: The README has been updated with instructions on how to use this new middleware, and example usage has been provided.
- **Main export update**: The middleware has been added to the central export file for easy access.

### Example Usage:

```tsx
type TestState = {
  count: number
  user: string
  increment: () => void
}
const useStore = create(
  syncStore<TestState>({
    name: 'test-storage', // Unique storage key
    fields: ['count'], // Only 'count' will be synced across tabs
    storage: sessionStorage, // Optional: You can specify the storage to be used
  })((set) => ({
    count: 0,
    user: 'guest', // This field is NOT synced
    increment: () => set((s) => ({ count: s.count + 1 })),
  }))
);
```

### Testing:

I have manually tested the functionality of the middleware to ensure it works correctly across multiple tabs and in both supported and unsupported browsers (with fallbacks to `localStorage`). The state synchronization works as expected in both scenarios.

However, I am relatively new to writing unit tests and had some challenges in creating comprehensive tests for this middleware. As a result, I was unable to write tests that cover every edge case.

**If this PR is not accepted due to the lack of tests, I would greatly appreciate your help in fixing and improving the tests.**

### Notes:
- The `syncStore` middleware uses the `BroadcastChannel` API for real-time synchronization. If a browser does not support `BroadcastChannel`, it defaults to `localStorage`.
- I have included all necessary changes to the README and other documentation.


dai-shi:
Hi, thanks for the suggestion.
We basically don't add new middleware in this repo.
Would you create a new package as a third-party library? and list it in [third-party libs](https://github.com/pmndrs/zustand/blob/main/docs/integrations/third-party-libraries.md).
If you are interested, I can invite you to https://github.com/zustandjs.

PR #2994, opened 2025-02-04
docs: Remove unnecessary void operator

Sunjae95:
## Summary
Removed the unnecessary void operator from the <button> element's onClick handler.

The void operator is not needed since incrementCount does not require explicit suppression of its return value.
Improves code readability and maintainability by eliminating unnecessary operations.

## Check List

- [ ] `pnpm run fix:format` for formatting code and docs


dai-shi (approved):
Looks okay to me

PR #2995, opened 2025-02-04
docs: strict null check in createContext example

jroitgrund:

N/A

## Summary

~~This PR edits the 'context usage' example from the docs to use state, rather than a ref, to hold the zustand store.~~
This PR edits the 'context usage' example from the docs to use a strict mull check, rather than `!ref.current`, so that the react compiler recognizes the ref is immutable and optimizes the component.

~~Both useState and useRef are correct in practice: useState will only render once since we never set the state after initialization, and useRef is safe to use since the component never needs to rerender when the ref value changes.~~

~~However, lint rules will flag useRef as a mistake, because we are passing a ref value to a child component, which in general can cause stale components since changing refs doesn't trigger re-renders.~~

## Check List

### DIFF
--- a/docs/previous-versions/zustand-v3-create-context.md
+++ b/docs/previous-versions/zustand-v3-create-context.md
@@ -111,19 +111,20 @@ Discussion: https://github.com/pmndrs/zustand/discussions/1276
 Here's the new context usage with v4 API.
 
 ```jsx
-import { createContext, useContext, useState } from 'react'
+import { createContext, useContext, useRef } from 'react'
 import { createStore, useStore } from 'zustand'
 
 const StoreContext = createContext(null)
 
 const StoreProvider = ({ children }) => {
-  const [store] = useState(() =>
-    createStore((set) => ({
+  const storeRef = useRef()
+  if (storeRef.current === null) {
+    storeRef.current = createStore((set) => ({
       // ...
     }))
-  )
+  }
   return (
-    <StoreContext.Provider value={store}>
+    <StoreContext.Provider value={storeRef.current}>
       {children}
     </StoreContext.Provider>
   )


dai-shi:
Is it eslint-plugin-react-hooks or eslint-plugin-react-compiler?
I heard that useRef for lazy initialization will be supported by the compiler.

jroitgrund:
It's react-compiler

```
│   62:39  error  Ref values (the `current` property) may not be accessed during render.
│ (https://react.dev/reference/react/useRef)  react-compiler/react-compiler
```

dai-shi:
Can you look for official info from the compiler team that recommends useState over useRef for lazy initialization?

jroitgrund:
I'd say it falls under https://react.dev/learn/referencing-values-with-refs#differences-between-refs-and-state

> You shouldn’t read (or write) the current value during rendering.

I think they didn't want to special case 'useRef but you never update .current', and they would rather recommend useState without ever callling the setter for immutable data which is used during render.

dai-shi:
I would like to see if it's officially unsupported with the compiler. Maybe we should wait for the compiler to be out of the experimental phase.

jroitgrund:
Looks like it's being discussed, I asked a question here: https://github.com/facebook/react/pull/30843#issuecomment-2633740433

josephsavona:
The compiler does support this pattern, but you have to specifically check that against null, ie `if (ref.current == null) { /* lazy init */ }`

dai-shi:
> The compiler does support this pattern, but you have to specifically check that against null, ie `if (ref.current == null) { /* lazy init */ }`

Cool. Does it work with `=== undefined` too? (Or, `=== null` if it's initialized with `null`)

docs/previous-versions/zustand-v3-create-context.md
```diff
@@ -111,20 +111,19 @@ Discussion: https://github.com/pmndrs/zustand/discussions/1276
...
 ```jsx
-import { createContext, useContext, useRef } from 'react'
+import { createContext, useContext, useState } from 'react'
 import { createStore, useStore } from 'zustand'
 
 const StoreContext = createContext(null)
 
 const StoreProvider = ({ children }) => {
-  const storeRef = useRef()
-  if (!storeRef.current) {
```
dai-shi, line 121: Please revert changes and change this to
```js
  if (storeRef.current === null) {
```

ahmadconnectedmotion:
Yes, that's right

dai-shi:
> Does it work with `=== undefined` too? (Or, `=== null` if it's initialized with `null`)

As I tried in the playground. `== null` and `=== null` works, but `=== undefined` does not.

jroitgrund:
Happy to make this change, but curious as to the rationale. Do refs provide a benefit over state? state provides:
- a built-in API for lazy initialisation, rather than a manual check
- the 'correct' way to consume values at render-time (see https://github.com/pmndrs/zustand/pull/2995#issuecomment-2633707224)

dai-shi:
The rationale is that `useRef` should be more lightweight than `useState`. We don't need "state" that changes over time, but just a stable reference.

Having that said, I'm fine to recommend `useState` as best practice with React Compiler. I think there should be a dedicated guide, not in docs/previous-versions/zustand-v3-create-context.md.
This document is about the migration from v3, so I would like to avoid changes. So, for now, I would like to change it as https://github.com/pmndrs/zustand/pull/2995#discussion_r1982615037. Please note that I may change my mind in the future.

jroitgrund:
Pushed, edited description + title.

PR #2998, opened 2025-02-05
Update nextjs.md

theperfectpunk:
add mandatory argument for useRef

Fixes #

## Summary

Two code snippets in the documentation have useRef hook without initialValue mandatory argument, added null as the initialValue for both of the hooks

## Check List


dai-shi (commented):
Nice catch! Do we have other places in docs that need to be fixed too?

theperfectpunk:
> Nice catch! Do we have other places in docs that need to be fixed too?

updated one more useRef in the testing section

dai-shi (approved):
Thank you!

PR #3003, opened 2025-02-10
fix: partial state update behaviour

waynetee:
## Summary

Currently `setState(state)` does nothing if `state` is same as the previous state. However this is not the case for partial state updates. Right now subscribers are called even if the partial state update leaves the state unchanged.

This PR aligns the behavior of partial state updates with full state updates. If the "update" leaves the state unchanged, subscribers will not be called. This greatly improves performance in scenarios when state is left unchanged.

Note that this does change the behaviour of Zustand, so feedback is welcome.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -70,13 +70,28 @@ const createStoreImpl: CreateStoreImpl = (createState) => {
       typeof partial === 'function'
         ? (partial as (state: TState) => TState)(state)
         : partial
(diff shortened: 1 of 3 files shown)


dai-shi:
This is designed as intended.
If you need such behavior, our recommendation is to use function update.
If you really need this behavior as default, our suggestion is to develop a middleware as a third-party library. (I've actually suggested it once, but it hasn't happen until now.)

PR #3012, opened 2025-02-15
chore(package.json): upgrade pnpm from v8 to v9

sukvvon:
### Changes
* Upgraded pnpm from v8 to the stable v9

### Reason for not upgrading to pnpm 10
* Compatibility concerns with various libraries
* pnpm 10 was recently released, and its stability has not been fully verified yet
* For now, we are applying the stable pnpm 9, and we will consider upgrading to pnpm 10 once its stability is confirmed

## Check List


dai-shi (approved):
LGTM

Let's do it for all three repos.

PR #3043, opened 2025-03-11
chore(eslint.config.mjs): update 'eslint-plugin-react-hooks' config to use 'recommended-latest'

sukvvon:
## Summary

* We can use recommended-latest instaed of adding in rules in eslint-plugin-react-hooks 5.2.0 latest version.
  * https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/README.md#flat-config-eslintconfigjsts

## Check List

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -17,10 +17,10 @@ export default tseslint.config(
   tseslint.configs.recommended,
   react.configs.flat.recommended,
   react.configs.flat['jsx-runtime'],
+  reactHooks.configs['recommended-latest'],
   {
     plugins: {
       'react-compiler': reactCompiler,
-      'react-hooks': reactHooks,
     },
     settings: {
       react: {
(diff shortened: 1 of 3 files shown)


dai-shi:
Can we do the same for the react compiler?

https://github.com/jotaijs/jotai-valtio/blob/eab0f401353f0fc4eda466b307d4e4bb97903519/eslint.config.js#L17-L18

sukvvon:
> Can we do the same for the react compiler?
> 
> https://github.com/jotaijs/jotai-valtio/blob/eab0f401353f0fc4eda466b307d4e4bb97903519/eslint.config.js#L17-L18

@dai-shi 

https://github.com/facebook/react/blob/main/compiler/packages/eslint-plugin-react-compiler/src/index.ts

`reactCompiler.configs.recommended` sets `error` not `warn`, do you wanna set `react-compiler` to `error` not `warn`?

dai-shi:
Yeah, I think all warnings are resolved or excluded already so `error` should be fine.

sukvvon:
> Yeah, I think all warnings are resolved or excluded already so `error` should be fine.

OK, i'll reflect your opinion

sukvvon:
@dai-shi Done ✅

dai-shi (approved):
Thank you.

PR #3044, opened 2025-03-12
ci(.github/workflows): add concurrency

sukvvon:
## Summary

* Add `concurrency` in action
  * The `concurrency` setting is needed to prevent duplicate runs when a PR is modified or commits are pushed multiple times, ensuring only the latest run is executed. Multiple executions can lead to resource wastage, and since only the latest state matters, canceling previous runs and keeping the latest one improves efficiency and reduces resource consumption.

## Check List

### DIFF
--- a/.github/workflows/compressed-size-action.yml
+++ b/.github/workflows/compressed-size-action.yml
@@ -2,6 +2,10 @@ name: Compressed Size
 
 on: [pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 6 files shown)


dai-shi:
Thanks, but no thanks. I don't prefer cancelling.

sukvvon:
@dai-shi I saw in concurrency in `cr.yml`, do you want to remove it?

dai-shi:
I already asked if it can be removed, but it seemed like it's required.

PR #3054, opened 2025-03-15
add hydrate method to store for state initialization

sirarifarid:
N/A

## Summary
This PR introduces a new method to hydrate the store with server fetched data, similar to `useHydratedAtoms`. This feature is something we really need in Zustand to work with SC and client store. While we can achieve this with context, it still remains bound within that scope.

**Quick look** 
code-snapshot

## Check List


dai-shi:
Thanks for the suggestion. I think this topic is important. This PR isn't acceptable for several reasons as is, but before that, I would like to understand the problem better.

Can you please create a minimal reproduction with StackBlitz and open a new discussion?

PR #3059, opened 2025-03-16
ci(.github/workflows): add 'test.yml' and rename job

sukvvon:
## Summary

* Changes job name in some files
* Add `test.yml`

## Check List

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,5 +17,8 @@ jobs:
           node-version: 'lts/*'
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm run test
+      - run: pnpm run test:format
+      - run: pnpm run test:types
+      - run: pnpm run test:lint
+      - run: pnpm run test:spec
       - run: pnpm run build # we don't have any other workflows to test build


dai-shi:
> * Change job name `lint` to `test`

Sounds good.

>   * We can reduce action time in `link-and-type.yml`

I don't have any issues with time.

>   * We can see cause immediately in PR when `lint-and-type` action fail

It doesn't feel very necessary.

What could be a drawback?

sukvvon:
> > * We can reduce action time in `link-and-type.yml`
> 
> I don't have any issues with time.

> > * We can see cause immediately in PR when `lint-and-type` action fail
> 
> It doesn't feel very necessary.
> 
> What could be a drawback?

sukvvon:
> > * We can reduce action time in `link-and-type.yml`
> 
> I don't have any issues with time.

> > * We can see cause immediately in PR when `lint-and-type` action fail
> 
> It doesn't feel very necessary.
> 
> What could be a drawback?

sukvvon:
> > * We can reduce action time in `link-and-type.yml`
> 
> I don't have any issues with time.

After checking the execution time, it turns out there is no reduction.

> > * We can see cause immediately in PR when `lint-and-type` action fail
> 
> It doesn't feel very necessary.
> 
> What could be a drawback?

It is not strictly necessary. However, by making it consistent with the test script in `package.json` and running it in parallel, we can immediately see where an issue occurs—whether in `test:format`, `test:lint`, or `test:types`—which improves the PR experience. Additionally, since these scripts are independent, they can be executed separately in parallel.

dai-shi (approved):
I don't think it's super necessary, but if there's no drawback, I don't have a huge objection. At least, if we have `fail-safe: false`, it should be fine.

Please send the same one to the other two repos.

sukvvon:
@dai-shi Done.

.github/workflows/lint-and-type.yml
```diff
@@ -7,8 +7,12 @@ on:
     types: [opened, synchronize]
 
 jobs:
-  lint:
+  test:
```
dai-shi, line 10: Actually, this is conflicting with `.github/workflows/test-multiple-versions.yml`. So, either revert it or name something different.

dai-shi (changes requested):

.github/workflows/lint-and-type.yml
sukvvon: @dai-shi 

I don’t think there is a conflict just because the `job` names are the same, as they are different actions. Are you saying that the names seem to overlap?

dai-shi: I mean, it conflicts when I configure the github branch protection. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

sukvvon, line 10:
```suggestion
  quality_check:
```

@dai-shi How about this?

sukvvon, line 10:
```suggestion
  check_qualify:
```

@dai-shi Or this?

dai-shi: Hmm, not a big fan.

How about changing `.github/workflows/test-multiple-versions.yml` to have `test_multiple_versions` and `.github/workflows/test-multiple-builds.yml` to have `test_multiple_builds`?

sukvvon: @dai-shi Sounds good

dai-shi: go ahead.

sukvvon: * `.github/workflows/test-multiple-versions.yml`: `version`
* `.github/workflows/test-multiple-builds.yml`: `build`
* `.github/workflows/test-old-typescripts.yml`: `typescript`

@dai-shi How about this?

.github/workflows/test-multiple-versions.yml
```diff
@@ -7,7 +7,7 @@ on:
     types: [opened, synchronize]
 
 jobs:
-  test:
+  test_spec:
```
dai-shi, line 10: oh, we have spec here. should we move this to `lint-and-type.yml` and maybe rename it to `test.yaml`?

.github/workflows/lint-and-type.yml
dai-shi: I would like to keep the similarity with yml file names.

.github/workflows/test-multiple-versions.yml
sukvvon: @dai-shi How about merging `lint-and-type.yml` and `test.yml` into `ci.yml`?

dai-shi: I meant to merge into `test.yml` which is equivalent to `pnpm test`.

sukvvon: > I meant to merge into `test.yml` which is equivalent to `pnpm test`.

Oh, i misunderstood it, i gonna reflect your opinon

dai-shi (approved):
LGTM
Thanks!

dai-shi (changes requested):
@sukvvon Let's avoid matrix in test.yml. Sorry for going back.

.github/workflows/test-multiple-versions.yml
```diff
@@ -7,20 +7,7 @@ on:
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 'lts/*'
-          cache: 'pnpm'
-      - run: pnpm install
-      - run: pnpm run build # we don't have any other workflows to test build
```
dai-shi, line 20: Let's leave this comment in `test.yml`.

.github/workflows/test.yml
```diff
@@ -17,6 +17,5 @@ jobs:
           node-version: 'lts/*'
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm run test:format
-      - run: pnpm run test:types
-      - run: pnpm run test:lint
+      - run: pnpm run test
```
dai-shi, line 20: Hm, how about running test:* sequentially? I think it's easier to know which one stops.

dai-shi (approved):
LGTM

PR #3074, opened 2025-04-03
docs: update readme TS sample to be less verbose and match TS docs

alm0st907:
Will look and update the PR if I find them.

Fixes #

## Summary

In the readme, the typescript sample is very verbose and inconsistent with the actual typescript documentation. I think the consistency and conciseness are important for a first time experience when getting started. 

[Link to docs for comparison.](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md)

## Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,27 +466,16 @@ Basic typescript usage doesn't require anything special except for writing `crea
 
 ```ts
 import { create } from 'zustand'
-import { devtools, persist } from 'zustand/middleware'
-import type {} from '@redux-devtools/extension' // required for devtools typing
 
 interface BearState {
   bears: number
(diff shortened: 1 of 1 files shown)


alm0st907:
@dbritto-dev I understand readme changes are very low-pri, but it would be nice to either clarify what is wrong with this contribution or update the contributing.md rules to say they will not be accepted. As far as I am aware based on the contribution rules there is nothing wrong with this PR and it was made in good faith.

dbritto-dev:
@alm0st907 sure, we try to keep readme as it is right now except for few exceptions:
- the api changed
- the imports changed

Everything else is not accepted due to we have specific guides for those

PR #3082, opened 2025-04-15
fix: handle hardware acceleration disabled scenario in demo site

B1u3B01t:
This PR improves the demo site's error handling when hardware acceleration is disabled in the browser. The changes include: as discussed in #3071 

- Added a fallback UI

The fallback scene now shows:
- A static image of the Zustand ogimage

This fixes the issue where the demo site would break without any feedback when hardware acceleration is disabled.

Testing:
- fallback works when hardware acceleration is disabled

Fixes #

## Check List

### DIFF
--- a/examples/demo/src/components/Scene.jsx
+++ b/examples/demo/src/components/Scene.jsx
@@ -149,24 +149,26 @@ function Effects() {
 
 function FallbackScene() {
   return (
-    <div style={{
-      position: 'absolute',
-      top: 0,
-      left: 0,
-      width: '100%',
-      height: '100%',
-      display: 'flex',
-      alignItems: 'center',
-      justifyContent: 'center',
-      background: '#010101'
-    }}>
-      <img 
-        src="/ogimage.jpg" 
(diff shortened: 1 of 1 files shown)


examples/demo/src/components/Scene.jsx
```diff
@@ -147,49 +147,85 @@ function Effects() {
...
 export default function Scene() {
+  const [error, setError] = useState(null)
+
+  if (error) {
+    return <FallbackScene />
+  }
+
   return (
-    <Canvas>
+    <Canvas onError={(e) => setError(e)}>
```
dai-shi, line 184: Does this work?
```suggestion
    <Canvas onError={setError}>
```

dai-shi:
Please fix code format.

examples/demo/src/components/Scene.jsx
B1u3B01t: Yes. Will make the change

dai-shi (approved):
The code looks good to me. Thanks for your contribution!

PR #3088, opened 2025-04-20
test(vanilla): extract common 'spy' setup into 'beforeEach'

sukvvon:
## Summary

* Extracted duplicated `spy` initialization into `beforeEach` for reuse.
* This refactoring improves test readability and makes maintenance easier.

## Check List

### DIFF
--- a/tests/vanilla/subscribe.test.tsx
+++ b/tests/vanilla/subscribe.test.tsx
@@ -1,10 +1,15 @@
-import { describe, expect, it, vi } from 'vitest'
+import { type Mock, beforeEach, describe, expect, it, vi } from 'vitest'
 import { subscribeWithSelector } from 'zustand/middleware'
 import { createStore } from 'zustand/vanilla'
 
 describe('subscribe()', () => {
+  let spy: Mock
+
+  beforeEach(() => {
+    spy = vi.fn()
+  })
+
   it('should not be called if new state identity is the same', () => {
-    const spy = vi.fn()
     const initialState = { value: 1, other: 'a' }
     const { setState, subscribe } = createStore(() => initialState)
 
@@ -14,7 +19,6 @@ describe('subscribe()', () => {
   })
 
   it('should be called if new state identity is different', () => {
-    const spy = vi.fn()
     const initialState = { value: 1, other: 'a' }
     const { setState, getState, subscribe } = createStore(() => initialState)
 
@@ -24,7 +28,6 @@ describe('subscribe()', () => {
   })
 
   it('should not be called when state slice is the same', () => {
-    const spy = vi.fn()
     const initialState = { value: 1, other: 'a' }
     const { setState, subscribe } = createStore(
       subscribeWithSelector(() => initialState),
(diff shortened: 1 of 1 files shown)


tests/vanilla/subscribe.test.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { type Mock, beforeEach, describe, expect, it, vi } from 'vitest'
 import { subscribeWithSelector } from 'zustand/middleware'
 import { createStore } from 'zustand/vanilla'
 
 describe('subscribe()', () => {
+  let spy: Mock
+
+  beforeEach(() => {
+    spy = vi.fn()
+  })
```
dbritto-dev, line 10:
```suggestion
  let spy: Mock = vi.fn()

  afterEach(() => {
    vi.mockReset()
  })
```

dbritto-dev (changes requested):

tests/vanilla/subscribe.test.tsx
sukvvon: @dbritto-dev 

I think you are suggesting me using `afterEach` with `vi.resetAllMocks`, does it?

dbritto-dev: @sukvvon updated

sukvvon: @dbritto-dev 

While the approach you suggested also works correctly, I chose to use `spy = vi.fn()` inside `beforeEach` for the sake of consistency and clarity across tests.

The reasons are as follows:

* Using `beforeEach` creates a new instance for each test, preventing unintended shared state.
* It also makes the test’s intent and flow clearer, which helps with maintenance and debugging.
* Additionally, it offers more flexibility if different mock implementations need to be injected per test in the future.

dbritto-dev: @sukvvon from what I saw it makes no different from the original implementation. I do not see how creating a global scope variable helps with the points you are talking. In fact if you want to use a global scope variable then you should avoid re-assigning the spy and just reset the spy when is needed if that's not the case then the original implementation is enough

dbritto-dev:
The current works well and there's no benefit from these changes

tests/vanilla/subscribe.test.tsx
sukvvon: @dbritto-dev @dai-shi 

I simply wanted to extract and reuse the repeatedly used code for the sake of consistency

dai-shi: In this case, smaller scope is actually more readable to me.

You may want to open a new discussion for suggestions before opening a new PR.

sukvvon: > In this case, smaller scope is actually more readable to me.

@dai-shi 

I understood, thanks

dbritto-dev-seatpick: @dai-shi I'm down with you I don't any benefits from these changes and smaller scope is better for readability

sukvvon: > @dai-shi I'm down with you I don't any benefits from these changes and smaller scope is better for readability

@dbritto-dev 

I understood, thanks

PR #3108, opened 2025-05-05
feat(vanilla): update shallow compare function and tests

dbritto-dev:

Fixes #3107

## Check List


src/vanilla/shallow.ts
```diff
@@ -49,6 +52,9 @@ export function shallow<T>(valueA: T, valueB: T): boolean {
   if (Object.is(valueA, valueB)) {
     return true
   }
+  if (type(valueA) !== type(valueB)) {
+    return false
+  }
```
steaks, line 57: Clever! I didn't know about the `Object.prototype.toString` trick. I think the block below be removed

```
if (
    typeof valueA !== 'object' ||
    valueA === null ||
    typeof valueB !== 'object' ||
    valueB === null
  ) {
    return false
  }
  ```

Comment (dbritto-dev) [reply src/vanilla/shallow.ts:57]: actually, we need it

File: src/vanilla/shallow.ts [hunk 2]
```diff
@@ -1,6 +1,9 @@
 const isIterable = (obj: object): obj is Iterable<unknown> =>
   Symbol.iterator in obj
 
+const type = <T>(obj: T): string =>
+  Object.prototype.toString.apply(obj).slice(8, -1).toLowerCase()
+
```
dai-shi, line 6: I don't think this is very nice and acceptable. If we want to distinguish an array, use `Array.isArray()`.

tests/vanilla/shallow.test.tsx
```diff
@@ -172,6 +172,25 @@ describe('shallow', () => {
...
+  const obj = { 0: 'foo', 1: 'bar' }
+  const arr = ['foo', 'bar']
+  const set = new Set(['foo', 'bar'])
+  const map = new Map([
+    [0, 'foo'],
+    [1, 'bar'],
+  ])
+
+  it('compares different data structures', () => {
+    expect(shallow<unknown>(obj, arr)).toBe(false)
```
dai-shi, line 185: As @charkour pointed out, `toBe(true)` in this case can be reasonable.

Otherwise, if we generalize this pattern, it's `shallow(iterable, nonIterable).toBe(false)`. Do we want it? I'm open for it.

dai-shi (changes requested):
There's an implementation problem and a design mismatch.

src/vanilla/shallow.ts
steaks: Oh I see what it's doing now. Sounds good

src/vanilla/shallow.ts
dai-shi: Here's one solution:

```diff
diff --git a/src/vanilla/shallow.ts b/src/vanilla/shallow.ts
index 749e99b..09f594b 100644
--- a/src/vanilla/shallow.ts
+++ b/src/vanilla/shallow.ts
@@ -57,6 +57,9 @@ export function shallow<T>(valueA: T, valueB: T): boolean {
   ) {
     return false
   }
+  if (Array.isArray(valueA) !== Array.isArray(valueB)) {
+    return false
+  }
   if (!isIterable(valueA) || !isIterable(valueB)) {
     return compareEntries(
       { entries: () => Object.entries(valueA) },
```

tests/vanilla/shallow.test.tsx
```diff
@@ -172,6 +172,25 @@ describe('shallow', () => {
...
+    expect(shallow<unknown>(obj, set)).toBe(false)
+    expect(shallow<unknown>(obj, map)).toBe(false)
```
dai-shi, line 187:
```js
  const obj = { 0: 'foo', 1: 'bar' }
  const map = new Map([[0, 'foo'], [1, 'bar']])
```
Are they shallow equal, or not? (It feels like our current implementation doesn't handle it very well.)

tests/vanilla/shallow.test.tsx
steaks: Are `Set / Map / Array` iterable and `Object` non iterable?

```
const isIterable = (obj: object): obj is Iterable<unknown> =>
  Symbol.iterator in obj
  
isIterable({}) //false
isIterable([]) //true
isIterable(new Map()) //true
isIterable(new Set()) // true
```

dai-shi: I think the `shallow` implementation should be simple enough to tell the "spec". That's Zustand's philosophy.

I'm not sure how much we should add special cases. Checking if iterable seems reasonable, checking `.prototype` should be okay.

dai-shi: Okay, let's be strict about it with Object.getPrototypeOf.
See: dbd5ecd
I also refactored a bit to imply the spec.

src/vanilla/shallow.ts
dai-shi: https://github.com/pmndrs/zustand/pull/3108#discussion_r2075038796

tests/vanilla/shallow.test.tsx
```diff
@@ -172,6 +172,25 @@ describe('shallow', () => {
...
+    expect(shallow<unknown>(arr, set)).toBe(false)
+    expect(shallow<unknown>(arr, map)).toBe(false)
+    expect(shallow<unknown>(set, map)).toBe(false)
```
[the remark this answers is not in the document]
dai-shi: https://github.com/pmndrs/zustand/pull/3108#discussion_r2075038796

tests/vanilla/shallow.test.tsx
dai-shi: https://github.com/pmndrs/zustand/pull/3108#discussion_r2075038796

dbritto-dev:
@dai-shi @steaks looks good to me

tests/vanilla/shallow.test.tsx
steaks: This feels intuitive and clean to me.

It may be worth adding a guidance in the [troubleshooting section](https://github.com/pmndrs/zustand/blob/main/docs/apis/shallow.md#troubleshooting). Maybe something like this?

### Objects with different prototypes

The `shallow` function checks whether the two objects have the same prototype. If their prototypes are referentially different, shallow will return false. This comparison is done using:

```ts
Object.getPrototypeOf(a) === Object.getPrototypeOf(b)
```

Objects created with the object initializer (`{}`) or with `new Object()` inherit from `Object.prototype` by default. However, objects created with `Object.create(proto)` inherit from the proto you pass in—which may not be `Object.prototype.`

```ts
const obj1 = Object.create({ foo: "bar" }); // prototype is {foo: "bar"}
const obj2 = {};                            // prototype is Object.prototype

shallow(obj1, obj2); // false
```

dbritto-dev:
@dai-shi all yours

tests/vanilla/shallow.test.tsx
dai-shi: Yes, I knew that gotcha.

@dbritto-dev What do you think about adding @steaks suggestion to docs?

dai-shi (approved):
Remaining: update docs https://github.com/pmndrs/zustand/pull/3108#discussion_r2076552297

dbritto-dev:
@dai-shi sure, I'll take care of docs

tests/vanilla/shallow.test.tsx
dbritto-dev: @dbritto-dev let's add it

dai-shi:
> @dai-shi sure, I'll take care of docs

Good. Please add it to this PR, because I want to merge it at the same time.

dbritto-dev:
@dai-shi all yours

docs/apis/shallow.md
```diff
@@ -25,6 +25,7 @@ const equal = shallow(a, b)
   - [Comparing Maps](#comparing-maps)
 - [Troubleshooting](#troubleshooting)
   - [Comparing objects returns `false` even if they are identical.](#comparing-objects-returns-false-even-if-they-are-identical)
+  - [Comparing objects with different prototypes]()
```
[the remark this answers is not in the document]
dbritto-dev: @dai-shi done

dai-shi (approved):
I'll merge it when I prepare the next release.

PR #3109, opened 2025-05-06
Docs: fix misinformation about Date persistence with JSONStorage

sviande:
## Documentation example error

The example in the documentation doesn't work for Date object

## Summary

The updated code uses `this` to access to the Date object directly, the var `value` contains the value of `this[key].toJSON()`

From [the MDN docs for JSON.stringify replacer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter)

> As a function, it takes two parameters: the key and the value being stringified. The object in which the key was found is provided as the replacer's this context.

And
>All [Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) objects implement the toJSON() method, which returns a string (the same as calling toString()). Thus, they will be serialized as strings. Similarly, [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) objects implement toJSON(), which returns the same as [toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString).

## Check List

- [ X] `pnpm run fix` for formatting and linting code and docs

### DIFF
--- a/docs/integrations/persisting-store-data.md
+++ b/docs/integrations/persisting-store-data.md
@@ -461,9 +461,9 @@ const storage = createJSONStorage(() => sessionStorage, {
     }
     return value
   },
-  replacer: (key, value) => {
-    if (value instanceof Date) {
-      return { type: 'date', value: value.toISOString() }
+  replacer: function (this, key, value) {
+    if (this[key] instanceof Date) {
+      return { type: 'date', value: this[key].toISOString() }
     }
     return value
   },


dbritto-dev:
@sviande I don't see the changes that you talked about in the PR and I don't see any misinformation, even the current implementation is aligned to MDN docs

sviande:
The code in the current example doesn't work as intended.

As I understand it, the code example should serialize a Date object to {type: 'date', value: '2025-05-06T15:16:39+00:00'}. When we retrieve the store, it should read from {type: 'date', value: '2025-05-06T15:16:39+00:00'} and return a Date object.
```ts
const store = {start: new Date('2025-05-06T15:16:39+00:00')};
const savedJSON = {'start': {'type': 'date', 'value': '2025-05-06T15:16:39+00:00'}};
const loaded = {start: Date('2025-05-06T15:16:39+00:00')};
```
However, the example behavior never gets a Date instance into the variable `value`; instead its a string, and you never get back a Date object.
```ts
const store = {start: new Date('2025-05-06T15:16:39+00:00')};
const savedJSON = {'start': '2025-05-06T15:16:39+00:00'};
const loaded =  {start: '2025-05-06T15:16:39+00:00'};
```

In the fixed example, the code checks the type of the original object instead of the serialized value.
```ts
const store = {start: new Date('2025-05-06T15:16:39+00:00')};
const savedJSON = {'start': {'type': 'date', 'value': '2025-05-06T15:16:39+00:00'}};
const loaded = {start: Date('2025-05-06T15:16:39+00:00')};
```

I hope this is clearer.

dbritto-dev:
@sviande would you mind creating a minimal repro on stackblitz?

sviande:
Test: 

1. load page view the current date
2. click on `set and save date`
3. reload page the date should be the same as previous

From the current documentation: the reload failed and when you check into the localStorage the JSON is not as expected
https://stackblitz.com/edit/p2hayvvo-thewfirs?file=src%2Findex.tsx

From the fix works as expected
https://stackblitz.com/edit/p2hayvvo-pwush1ry?file=src%2Findex.tsx

dbritto-dev:
@sviande thanks, got it seems using `this` is a hacky way to do it by default `replacer` only accept `key` and `value` so the usage even if works is not acceptable

sviande:
Ok what do you consider hacky ? 
It's in the documentation this is specially provided into replacer context.
>As a function, it takes two parameters: the key and the value being stringified. **The object in which the key was found is provided as the replacer's this context**.

I saw your update at least there no error in the doc. thanks.

dbritto-dev:
> Ok what do you consider hacky ? It's in the documentation this is specially provided into replacer context.
> 
> > As a function, it takes two parameters: the key and the value being stringified. **The object in which the key was found is provided as the replacer's this context**.
> 
> I saw your update at least there no error in the doc. thanks.

I consider hacky something that is not documented but could work, even on MDN there's no examples that use `this`

sviande:
Was linked to #2403

dbritto-dev:
@sviande yeah, it would be a problem for migrations

PR #3111, opened 2025-05-06
feat: add devtools.cleanup() method

maxam2017:
## Summary
When a store is wrapped in a context provider, remounting can cause duplicate instances in Redux DevTools; this patch adds a `devtools.cleanup()` method, allowing users to manually unsubscribe and clean up tracked connections.

## Check List

## Note
related to #2254


dai-shi:
This sounds reasonable to me. Can anyone give a deep review?

dai-shi:
@dbritto-dev What do you think?

dbritto-dev:
@dai-shi LGTM but up to you

src/middleware/devtools.ts
```diff
@@ -146,6 +149,22 @@ const extractConnectionInformation = (
   return { type: 'tracked' as const, store, ...newConnection }
 }
 
+const removeStoreFromTrackedConnections = (
+  name: string | undefined,
+  store: string | undefined,
+) => {
+  if (store === undefined) return
```
dai-shi, line 156: If `store` is undefined, we don't do anything.
So, is the goal of this PR is to avoid memory leaks?

dai-shi (commented):
I had a closer look, but it looks good.
One concern, which is not this PR's fault, is that the code around connection info isn't very readable. It took me while to get it.

We would like to refactor in the future, and for that, can you add a test that depends on `removeStoreFromTrackedConnections`?

maxam2017:
> I had a closer look, but it looks good. One concern, which is not this PR's fault, is that the code around connection info isn't very readable. It took me while to get it.
> 
> We would like to refactor in the future, and for that, can you add a test that depends on `removeStoreFromTrackedConnections`?

@dai-shi 
I'd like to test `removeStoreFromTrackedConnections` by verifying that the store is actually removed from the `trackedConnections` Map. However, since `trackedConnections` is a private variable within the module, I'm not sure how to properly test this.

Do you have any suggestions for how to test this functionality properly?
Thanks for your guidance!

maxam2017:
> Actually, I don't. What would happen is memory leaks, but testing memory leaks isn't very straightforward. Testing internal variables isn't nice either. (It's only testing implementation details.)
> 
> Actually, I have an idea. If we fail to remove from tracked connections, after cleanup, when we try another connection, it will fail to connect again. So, we should be able to test it from outside. Somewhat different, but kind of.

Here's the updated patch: 383d932. Let me know if anything needs to be adjusted!

dai-shi (approved):
LGTM
Thanks for your contribution!

dbritto-dev:
@dai-shi I'll take care of docs

PR #3117, opened 2025-05-10
chore(vitest): rename config to .ts for ts/tsx consistency

sukvvon:

Fixes #

## Summary

* Rename `vitest.config.mts` to `vitest.config.ts` for consistency with the project’s use of .ts/.tsx files.

## Check List


dbritto-dev:
Vitest config has nothing to do with ts/tsx you could even use .mts and use pure js

sukvvon:
@dbritto-dev 

https://vitest.dev/config/#configuring-vitest

According to the official Vitest documentation, `vitest.config.ts` overrides the configuration from `vite.config.ts` and is used independently for Vitest-specific settings. It also clearly states that Vitest supports all conventional JS and TS extensions, meaning `.ts` is fully supported for the configuration file.

The `.mts` extension is specifically necessary only when TypeScript files are directly executed by Node in an ESM context, but in the case of Vitest, which uses `esbuild` to load the config, the extension doesn’t affect the configuration behavior. `.ts` works just as well as `.mts,` and switching to `.ts` would improve consistency with the rest of the codebase.

Therefore, there is no need to use `.mts `for the Vitest config file, and `.ts` is sufficient.

dai-shi:
For consistency, mts seems more reasonable as the project default is commonjs.

PR #3118, opened 2025-05-12
docs: clarify and improve dynamic scoped store usage in useStoreWithEqualityFn demo

sabertobihwy:

No related issue. This is a standalone improvement to the documentation examples.

## Summary

This PR improves the example under the section **"Using dynamic scoped (non-global) vanilla stores in React"**.

- Adds an explicit context type to `createContext`, enabling stronger type checking
- Replaces `stores === undefined` with `stores === null`, matching the context default 
- Optimizes `useCallback` to directly memoize the store factory result
- Fixes minor bug: parameters not `tab-${index}` but `${index}`
- Updates the example to use **multiple `CounterStoresProvider` instances** for true per-`Tabs` state isolation
- Adjusts the accompanying description to clarify the example above

## Check List

### DIFF
--- a/docs/hooks/use-store-with-equality-fn.md
+++ b/docs/hooks/use-store-with-equality-fn.md
@@ -689,7 +689,7 @@ Next, we need a way to manage and access these stores throughout our app. We’l
 for this.
 
 ```tsx
-const CounterStoresContext = createContext(null)
+const CounterStoresContext = createContext<Map<string, ReturnType<typeof createCounterStore>> | null >(null)
 
 const CounterStoresProvider = ({ children }) => {
   const [stores] = useState(
@@ -714,12 +714,12 @@ const useCounterStore = <U,>(
 ) => {
   const stores = useContext(CounterStoresContext)
 
-  if (stores === undefined) {
+  if (stores === null) {
     throw new Error('useCounterStore must be used within CounterStoresProvider')
   }
 
   const getOrCreateCounterStoreByKey = useCallback(
-    () => createCounterStoreFactory(stores),
+    createCounterStoreFactory(stores),
     [stores],
   )
 
@@ -738,7 +738,7 @@ counter.
 function Tabs() {
   const [currentTabIndex, setCurrentTabIndex] = useState(0)
   const counterState = useCounterStore(
-    `tab-${currentTabIndex}`,
+    `${currentTabIndex}`,
     (state) => state,
   )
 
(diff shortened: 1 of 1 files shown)


sabertobihwy:
Hi, thank you for the feedback — I now understand my mistake was calling `createFactory(stores)` immediately inside `useCallback`.
I'll revert to using:
 `const getOrCreateCounterStoreByKey = useCallback(
  (key: string) => createCounterStoreFactory(stores!)(key),
  [stores],
)
`
If I may, I'd still like to share some thoughts on my changes, just for discussion:
First is about `useCounterStore`, given the code:

* Regarding the useContext typing: you're right, adding an explicit type and avoiding undefined isn't necessary. My intent was just to align it with other examples in other three sections that do include explicit types — I understand it's more about consistency preference than correctness.
*  The function `getOrCreateCounterStoreByKey`  in the example has a minor issue `() => createCounterStoreFactory(stores)`  that should be changed to `(key: string) => createCounterStoreFactory(stores!)(key)`

Second is in `Tabs`, there is a minor problem since we have already passed 'tab-' in `useStoreWithEqualityFn`, so in `useCounterStore` we cannot pass redundant prefix
image

Lastly, about Provider changes, I noticed
+  the first section uses a single global store.
+ The second introduces a global Map to dynamically create multiple stores.
+ The third one uses Context.Provider to isolate a single store per scope.
+ So naturally, the fourth scenario would be using Context.Provider to isolate a Map, giving each scope its own set of stores. However, in the final demo, it still uses a single provider. I think wrapping each <Tabs /> with its own provider would better demonstrate store isolation per scope  — just a thought!

sabertobihwy:
@dbritto-dev 
> Hi, thank you for the feedback — I now understand my mistake was calling `createFactory(stores)` immediately inside `useCallback`. I'll revert to using: `const getOrCreateCounterStoreByKey = useCallback( (key: string) => createCounterStoreFactory(stores!)(key), [stores], ) ` If I may, I'd still like to share some thoughts on my changes, just for discussion: First is about `useCounterStore`, given the code:
> 
> * Regarding the useContext typing: you're right, adding an explicit type and avoiding undefined isn't necessary. My intent was just to align it with other examples in other three sections that do include explicit types — I understand it's more about consistency preference than correctness.
> * The function `getOrCreateCounterStoreByKey`  in the example has a minor issue `() => createCounterStoreFactory(stores)`  that should be changed to `(key: string) => createCounterStoreFactory(stores!)(key)`
> 
> Second is in `Tabs`, there is a minor problem since we have already passed 'tab-' in `useStoreWithEqualityFn`, so in `useCounterStore` we cannot pass redundant prefix image
> 
> Lastly, about Provider changes, I noticed
> 
> * the first section uses a single global store.
> * The second introduces a global Map to dynamically create multiple stores.
> * The third one uses Context.Provider to isolate a single store per scope.
> * So naturally, the fourth scenario would be using Context.Provider to isolate a Map, giving each scope its own set of stores. However, in the final demo, it still uses a single provider. I think wrapping each  with its own provider would better demonstrate store isolation per scope  — just a thought!

dbritto-dev:
@sabertobihwy the only minor issue that I saw is `() => createCounterStoreFactory(stores)` that should be changed to `(key: string) => createCounterStoreFactory(stores!)(key)` apart from that I don't see the benefits that you are talking about

- Redundant prefix I don't think so, the example use a key and prefix makes debugging easier
- The fourth scenario doesn't need of multiple providers scoped in this context means that only exists within the provider could be one provider or multiple providers
- The check `stores === undefined` is basically to throw an error when you are trying to `useCounterStore` outside of `CounterStoresProvider` has nothing to do with the initial value of `CounterStoresContext`

sabertobihwy:
@dbritto-dev this is what I'm talking about "redundant prefix"
image

PR #3123, opened 2025-05-17
test(*): remove 'vitest' imports

sukvvon:

Fixes #

## Check List


dai-shi:
Thanks, but no thanks. I prefer the current explicit approach.

sukvvon:
@dai-shi 

How about remame `tests/setup.ts` to `vitest.setup.ts`?

dbritto-dev:
> Thanks, but no thanks. I prefer the current explicit approach.

I'm down with you @dai-shi the global setup exists to support an easy migration from Jest

dai-shi:
I'd prefer having setup files in tests.

PR #3127, opened 2025-05-21
chore(package.json): add 'test:coverage' script

sukvvon:

Fixes #

## Check List


dai-shi:
No thanks. We don't need to run it. If you want do `npx vitest run --coverage`.

PR #3129, opened 2025-05-21
test(vanilla/shallow): add test for pure iterable with different values returns false

sukvvon:

Fixes #

### AS-IS

image

### TO-BE

image

## Check List


dai-shi (approved):
LGTM

PR #3130, opened 2025-05-21
test(vanilla/basic): add test for 'getInitialState'

sukvvon:

Fixes #

### AS-IS

image

### TO-BE

image

## Check List

### DIFF
--- a/tests/vanilla/basic.test.ts
+++ b/tests/vanilla/basic.test.ts
@@ -73,7 +73,7 @@ it('can get the store', async () => {
 it('can get the initial state', () => {
   const initial = { value: 1 }
   const store = createStore(() => initial)
-store.setState({ value: 2 })
+  store.setState({ value: 2 })
   expect(store.getInitialState()).toBe(initial)
 })
 
--- a/tests/vanilla/basic.test.ts
+++ b/tests/vanilla/basic.test.ts
@@ -70,6 +70,13 @@ it('can get the store', async () => {
   expect(store.getState().getState2().value).toBe(1)
 })
 
+it('can get the initial state', () => {
+  const initial = { value: 1 }
+  const store = createStore(() => initial)
+
+  expect(store.getInitialState()).toBe(initial)
+})
+
 it('can set the store', async () => {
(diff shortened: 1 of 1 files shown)


tests/vanilla/basic.test.ts
```diff
@@ -70,6 +70,13 @@ it('can get the store', async () => {
   expect(store.getState().getState2().value).toBe(1)
 })
 
+it('can get the initial state', () => {
+  const initial = { value: 1 }
+  const store = createStore(() => initial)
+
```
dai-shi, line 76: Let's add this so that it tests if getInitialState does return the initial one.
```suggestion
store.setState({ value: 2 })
```

tests/vanilla/basic.test.ts
```diff
@@ -70,6 +70,13 @@ it('can get the store', async () => {
   expect(store.getState().getState2().value).toBe(1)
 })
 
+it('can get the initial state', () => {
+  const initial = { value: 1 }
+  const store = createStore(() => initial)
+store.setState({ value: 2 })
```
dai-shi, line 76:
```suggestion
  store.setState({ value: 2 })
```

tests/vanilla/basic.test.ts
```diff
@@ -70,6 +70,13 @@ it('can get the store', async () => {
   expect(store.getState().getState2().value).toBe(1)
 })
 
+it('can get the initial state', () => {
+  const initial = { value: 1 }
+  const store = createStore(() => initial)
+  store.setState({ value: 2 })
+  expect(store.getInitialState()).toBe(initial)
```
sukvvon, line 77:
```suggestion
  const initial = { value: 1 }
  const store = createStore(() => initial)
  
  store.setState({ value: 2 })
  
  expect(store.getInitialState()).toBe(initial)
```

How about adding newline?

dai-shi: It's minority, but I usually don't add many lines.

dai-shi (approved):

PR #3134, opened 2025-05-23
refactor(vanilla): resolve todo by removing type assertion using inferred type predicates

2yunseong:
Fixes #

## Summary
Refer to the TypeScript issue [37663: T | (() => T)](https://github.com/microsoft/TypeScript/issues/37663). (you can see the issue in the previous TODO comment too)

Before TypeScript 5.5, type inference with `typeof` was not possible in this case.  

However, in TS 5.5+, this is now supported thanks to [Inferred Type Predicates](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#inferred-type-predicates).

image

And I used it to resolve the todo.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -57,6 +57,8 @@ type CreateStoreImpl = <
   initializer: StateCreator<T, [], Mos>,
 ) => Mutate<StoreApi<T>, Mos>
 
+const isFunction = (x: unknown) => typeof x === 'function'
+
 const createStoreImpl: CreateStoreImpl = (createState) => {
   type TState = ReturnType<typeof createState>
   type Listener = (state: TState, prevState: TState) => void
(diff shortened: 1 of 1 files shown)


src/vanilla.ts
```diff
@@ -57,19 +57,16 @@ type CreateStoreImpl = <
...
+const isFunction = (x: unknown) => typeof x === 'function'
+
 const createStoreImpl: CreateStoreImpl = (createState) => {
   type TState = ReturnType<typeof createState>
   type Listener = (state: TState, prevState: TState) => void
   let state: TState
   const listeners: Set<Listener> = new Set()
 
   const setState: StoreApi<TState>['setState'] = (partial, replace) => {
-    // TODO: Remove type assertion once https://github.com/microsoft/TypeScript/issues/37663 is resolved
```
dai-shi, line 67: This issue isn't closed.

src/vanilla.ts
```diff
@@ -57,19 +57,16 @@ type CreateStoreImpl = <
...
-    // https://github.com/microsoft/TypeScript/issues/37663#issuecomment-759728342
-    const nextState =
-      typeof partial === 'function'
-        ? (partial as (state: TState) => TState)(state)
-        : partial
+    const nextState = isFunction(partial) ? partial(state) : partial
```
dai-shi, line 69: This is not the goal.
It should be this:
```suggestion
    const nextState = typeof partial === 'function' ? partial(state) : partial
```

dai-shi:
Thanks. Let's keep eyes on the upstream issue.

PR #3147, opened 2025-06-07
Avoid inferring action type when explicit action name is provided

vchirikov:
## Summary

https://github.com/pmndrs/zustand/pull/2987 Added call `findCallerName()` which tries to infer action name from error's callstack even when user explicitly provided action name (`nameOrAction` or `anonymousActionType`). I think it's much better to call `findCallerName` ONLY if user didn't provide anything we can use.

## Check List


dai-shi (approved):
LGTM

IDisposable:
I just came here to make the exact same comment. :shipit:

PR #3151, opened 2025-06-11
refactor(middleware): replace export * with explicit named and type exports

sukvvon:

Fixes #

## Summary

* Aligned export format with [`middleware.ts`](https://github.com/pmndrs/zustand/blob/main/src/middleware.ts)
* Even if we use export `*` syntax, we still only export `useShallow` in [`react/shallow.ts`](https://github.com/pmndrs/zustand/blob/main/src/react/shallow.ts) and `shallow` in [`vanilla/shallow.ts`](https://github.com/pmndrs/zustand/blob/main/src/vanilla/shallow.ts)

## Check List

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,5 +1,16 @@
-export * from './middleware/redux.ts'
-export * from './middleware/devtools.ts'
-export * from './middleware/subscribeWithSelector.ts'
-export * from './middleware/combine.ts'
-export * from './middleware/persist.ts'
+export { redux } from './middleware/redux.ts'
+export {
+  devtools,
+  type DevtoolsOptions,
+  type NamedSet,
+} from './middleware/devtools.ts'
+export { subscribeWithSelector } from './middleware/subscribeWithSelector.ts'
+export { combine } from './middleware/combine.ts'
+export {
+  persist,
+  createJSONStorage,
+  type StateStorage,
+  type StorageValue,
+  type PersistStorage,
+  type PersistOptions,
+} from './middleware/persist.ts'


dai-shi:
This is somewhat intentional. I think we should have more explicit control on what to export from the package. I would like to change `middleware.ts` to follow `shallow.ts` style. I thought it would be a breaking change, but maybe not. Can you try it please?

src/middleware.ts
```diff
@@ -1,5 +1,16 @@
...
+export { subscribeWithSelector } from './middleware/subscribeWithSelector.ts'
+export { combine } from './middleware/combine.ts'
+export {
+  persist,
+  createJSONStorage,
+  type StateStorage,
+  type StorageValue,
+  type PersistStorage,
+  type PersistOptions,
+} from './middleware/persist.ts'
```
sukvvon, line 16: > This is somewhat intentional. I think we should have more explicit control on what to export from the package. I would like to change `middleware.ts` to follow `shallow.ts` style. I thought it would be a breaking change, but maybe not. Can you try it please?

@dai-shi 

I reflected your opinion. How does it look like?

dai-shi: Sorry for the delay. Yeah, this looks good. This keeps everything as is, right?

sukvvon: @dai-shi 

image

Only the changes mentioned above have been made.

```ts
// shallow.d.ts
export { shallow } from 'zustand/vanilla/shallow';
export { useShallow } from 'zustand/react/shallow';

// esm/shallow.d.mts
export { shallow } from 'zustand/vanilla/shallow';
export { useShallow } from 'zustand/react/shallow';
```

It is the same as `shallow`.

PR #3159, opened 2025-06-23
chore: update dev dependencies

dai-shi:
for v5.0.6


dai-shi:
@sukvvon If you have a chance, would you help looking into the build issue please? (we should probably fix jotai and valtio too.)

sukvvon:
> @sukvvon If you have a chance, would you help looking into the build issue please? (we should probably fix jotai and valtio too.)

@dai-shi 

https://github.com/rollup/plugins/pull/1653

The above PR has been merged, and as a result, `@rollup/plugin-typescript` has been updated from version `12.1.2` to `12.1.3`. Therefore, I expect it would work if we simply keep this library at version `12.1.2`. Would you like to give it a try?

I will make a PR in J and V.

dai-shi:
> Therefore, I expect it would work if we simply keep this library at version `12.1.2`. Would you like to give it a try?

That's the last resort. I would like to upgrade rollup and @rollup/plugin-typescript.

dai-shi:
It looks like https://github.com/rollup/plugins/pull/1880 will revert the change, so let's do your suggestion.

sukvvon:
> It looks like [rollup/plugins#1880](https://github.com/rollup/plugins/pull/1880) will revert the change, so let's do your suggestion.

@dai-shi 

Thanks!

PR #3160, opened 2025-06-23
docs(README): remove 'omit' in 'Overwriting state' section

sukvvon:

Fixes #

## Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,11 +129,7 @@ const useFishStore = create((set) => ({
   salmon: 1,
   tuna: 2,
   deleteEverything: () => set({}, true), // clears the entire store, actions included
-  deleteTuna: () =>
-    set((state) => {
-      const { tuna, ...rest } = state
-      return rest
-    }, true),
+  deleteTuna: () => set(({ tuna, ...rest }) => rest, true),
 }))
 ```
 


dai-shi (changes requested):
If we were to change the README, it would be nicer to avoid the library at all. Can you try?

README.md
```diff
@@ -125,7 +125,7 @@ const treats = useBearStore(
 The `set` function has a second argument, `false` by default. Instead of merging, it will replace the state model. Be careful not to wipe out parts you rely on, like actions.
 
 ```jsx
-import { omit } from 'es-toolkit/compat'
+import omit from 'es-toolkit/compat/omit'
```
sukvvon: @dai-shi 

How does it look like?

dai-shi: No. I mean to avoid using the external library at all.

dbritto-dev: I'm down with you @dai-shi, we should remove the external lib in this case and replace it with `Object.fromEntries(...)` + `Object.entries()` + `.filter()`

README.md
```diff
@@ -125,7 +125,16 @@ const treats = useBearStore(
...
+function omit(obj, keys) {
+  const result = { ...obj }
+
+  for (let i = 0; i < keys.length; i++) {
+    const key = keys[i]
+    delete result[key]
+  }
+
+  return result
+}
```
sukvvon: @dai-shi, @dbritto-dev 

How does it look like?

GeoDaoyu: It looks like not FP, perhaps it should be like this:
```js
const omit = keysToOmit => obj =>
  Object.fromEntries(
    Object.entries(obj).filter(([key]) => !keysToOmit.includes(key))
  );
```

sukvvon: @GeoDaoyu 

```ts
function omit(obj, keys) {
  return Object.fromEntries(
    Object.entries(obj).filter(([key]) => !keys.includes(key))
  )
}
```

I understand what you’re saying, but I’ve already taken browser compatibility into consideration for now.

dai-shi: Please don't create a util function. Just do it in the store action in a simple way.

README.md
```diff
@@ -125,13 +125,15 @@ const treats = useBearStore(
...
   salmon: 1,
   tuna: 2,
   deleteEverything: () => set({}, true), // clears the entire store, actions included
-  deleteTuna: () => set((state) => omit(state, ['tuna']), true),
+  deleteTuna: () =>
+    set((state) => {
+      const { tuna, ...rest } = state
+      return rest
+    }, true),
 }))
```
sukvvon: @dai-shi 

I reflected your opinion.

dai-shi: Cool. I'm not sure if we prefer this, but we could also do this.
```js
set(({ tuna, ...rest }) => rest, true);
```

@dbritto-dev thoughts?

sukvvon: @dai-shi 

Sounds good. I reflected your opinion.

dbritto-dev: @dai-shi I'm down with you

dai-shi (approved):
Thanks!!

PR #3163, opened 2025-06-30
chore(vitest.config): enable 'typecheck' and 'coverage'

sukvvon:

Fixes #

## Check List

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -20,10 +20,12 @@ export default defineConfig({
       : ['default'],
     setupFiles: ['tests/setup.ts'],
     coverage: {
+      enabled: true,
       include: ['src/**/'],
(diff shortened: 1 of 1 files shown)


dai-shi:
I would like coverage to be opt-in, and I guess typecheck is covered by test:types.

sukvvon:
> I would like coverage to be opt-in, and I guess typecheck is covered by test:types.

@dai-shi 

Since `test:types` already covers type checking, I think we can exclude `typecheck`.
As for coverage, I think with this single setting, we can get both test results and coverage reports.
It could also eliminate the need to run something like `pnpm run test:spec --coverage`.

dai-shi:
we prefer `pnpm run test:spec --coverage` for now. We may change our mind in the future.

Thanks for the suggestion anyway!

PR #3169, opened 2025-07-05
chore: inserted a link to the social network Twitter

Olexandr88:

Fixes #

## Check List


dai-shi:
Thanks, but we don't tweet about Zustand in that account. So, let's not add it here.

Olexandr88:
Ok

PR #3171, opened 2025-07-09
docs: add currying for createStore

soyboi:
I hope this is a correct change, but I was reading the docs, and if I didn't misunderstand, we always need to do currying when creating a store with typescript. This seemed to be missing in the code example used in auto generating selectors.

### DIFF
--- a/docs/guides/auto-generating-selectors.md
+++ b/docs/guides/auto-generating-selectors.md
@@ -99,7 +99,7 @@ interface BearState {
   increment: () => void
 }
 
-const store = createStore<BearState>((set) => ({
+const store = createStore<BearState>()((set) => ({
   bears: 0,
   increase: (by) => set((state) => ({ bears: state.bears + by })),
   increment: () => set((state) => ({ bears: state.bears + 1 })),


dai-shi:
I'm fine to add it as it's TS code.
But, @dbritto-dev are you working on something else?

dbritto-dev:
@dai-shi in the new docs under apis/middlewares we talked about that, we do have ts examples. BTW, everything else use js examples.

docs/guides/auto-generating-selectors.md
```diff
@@ -99,7 +99,7 @@ interface BearState {
   increment: () => void
 }
 
-const store = createStore<BearState>((set) => ({
+const store = createStore<BearState>()((set) => ({
```
dai-shi, line 102: This is a TS example, so it makes sense to have `()`.
Or, do you have a PR/branch that will be merged later? @dbritto-dev

dbritto-dev: @dai-shi in that case we can merge this one

Issue #132, opened 2020-08-05
Using getter to calculate computed values


filipesmedeiros:
Hey!

Is it ok to use the getter to return a computed value in a function? Since using it in a value doesn't work?

Example:

```typescript
const [useAuth] = create((set, get) => ({
    user: {username: undefined, authLevel: 0},
    // isSignedIn: !!get().user.username // doesn't work because it's executed before the store is created, so get() returns undefined right?
    isSignedIn: () => get().user.username // should work, just call isSignedIn() instead of using as a value, in the components?
})}
```

Also, I feel like computed values are a cool features for state management libraries to have out of the box, are there planes to add it? Do you guys need help with it somehow?

shahruz:
I usually would just do something like `const isSignedIn = useAuth(state => !!state.user.username);` in the consuming component for this.

filipesmedeiros:
Right! That seems prettier. But that only works for component use right? If I want to separate some logic while calculating various layers of computed values in a store (in more complex cases) that doesn't work anymore. For example:

```typescript
const [useStore] = create((set, get) => {
    value: 30,
    value1: get().value + 60,
    value2: get().value1 / 4,
    value3: get().value2 * 1000
})
```

Logically, it might be better if `value3` depends on `value2` directly than if you had to repeat the computations you did for the other ones, right?

jordanmkoncz:
I was also curious about the recommended approach for selectors which derive/compute values from state (and possibly from other selectors themselves). Reading about Recoil and how it handles this using [selectors](https://recoiljs.org/docs/basic-tutorial/selectors) (also see https://www.youtube.com/watch?v=_ISAA_Jt9kI), I couldn't see anything analogous in Zustand which has the same optimisations.

drcmda:
sure you can use getters like that.

as for recoil and atoms, it's a slightly different paradigm. we have something coming up that's concentrating on that approach.

dai-shi:
@drcmda I don't think getters work because of `Object.assign()`. Once you `setState({ user })`, `signedIn` becomes a boolean value. I didn't try it, so correct me if I'm wrong.

jordanmkoncz:
Thanks @dai-shi for your response.

>b) Chaining computed values (aka atoms). For b), we are planning a new project coming soon, and it's outside the scope of zustand.

[...]

Issue #108, opened 2020-04-28
Correct approach to get derived values from state


ha3:
Hello,

In my store, I have several methods that calculates same derived values from state. I am looking for an optimized way to achive that.

Right now, I can do this:

```
getDerivedValue: () => {
  return get().listA[get().indexB].value;
}
...
get().getDerivedValue();
```

But I prefer something like this:

```
get derivedValue () {
  return get().listA[get().indexB].value;
}
...
get().derivedValue;
```

Unfortunately, using getter gives only the inital state, returning value doesn't update with state changes. My computations are not that expensive, but it would be much better to access the value as it was a variable then calling it as a function. My concern is a bit aesthetic but anyway, it would be nice to have something like Mobx's [computed property](https://mobx.js.org/refguide/computed-decorator.html). Is there a workaround for using getter or could you provide an alternative to achieve this?

Thank you.

universse:
Hi, maybe you can use a selector

```js
function useDerivedState () {
  return useStore(state => state.listA[state.indexB].value)
}
```

Then in your component

```js
  const derivedState = useDerivedState()
```

dai-shi:
The custom hook works if you are using React component.

If the original question is about non-React solution,
The first one seems reasonable.
I don't think we can support object getters unless there's fundamental requirement.

For performance, you could subscribe to the store and somehow cache the derived values.

Closing this. If you need further discussion, please feel free to reopen.

dai-shi:
> but I guess that's what you were referring to when you said you don't think we can support object getters?

Exactly. Those object getters are evaluated only at the creation time.
After several months from my previous comment, I'm more confident that zustand does/will not support the mobx style computed properties natively, likewise redux doesn't.

> Any other thoughts?

https://github.com/cmlarsen/zustand-middleware-computed-state
As far as I see its source code, this seems the right approach in zustand.

clementroche:
> Came across this and surprised to not see the topic of derived state anywhere in the readme. Would love to see how others are handling it!
> 
> I was hoping to be able to do something like this:
> 
> ```js
> const useStore = create((set) => ({
>   first: 'Daishi',
>   last: 'Kato',
>   get fullName() {
>     return `${this.first} ${this.last}`;
>   },
> }));
> ```
> 
> but I guess that's what you were referring to when you said you don't think we can support object getters?
> 
[...]

PR #3188, opened 2025-07-21
Introduction docs

mindplay-dk:
Two things on the "Getting started" page:

1. The JS example doesn't really work with TS - this page didn't even mention TypeScript, which is probably the default for most users today. I don't think we shouldn't leave TS users to search for the most basic TS patterns that work.

2. Computed state is a [whole](https://github.com/pmndrs/zustand/issues/132) [debacle](https://github.com/pmndrs/zustand/issues/108) - let's put the two most basic, most idiomatic patterns front and center.

The full docs are linked for both of these examples, but honestly this is probably enough for 90% of users.

Both of these things are almost certainly unavoidable in the *majority* of projects, so hopefully this gets everyone going *fast*. 🙂

### DIFF
--- a/docs/getting-started/introduction.md
+++ b/docs/getting-started/introduction.md
@@ -53,6 +53,57 @@ const useStore = create((set) => ({
 }))
 ```
 
+<details>
+<summary><strong>Using TypeScript?</strong></summary>
+
+> #### TypeScript with Inference
+> 
+> To get **type-inference** in TypeScript, your property initializations need to be separated from your methods - use `combine` to do that:
+> 
+> ```ts
+> import { create } from 'zustand'
+> import { combine } from 'zustand/middleware'
+> 
+> const useStore = create(
+>   combine(
+>     {
+>       bears: 0,
+>     },
+>     (set) => ({
+>       increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
+>       removeAllBears: () => set({ bears: 0 }),
+>       updateBears: (newBears: number) => set({ bears: newBears }),
+>     })
+>   )
+> )
+> ```
+>
+> #### TypeScript with Explicit Types
+> 
+> If you prefer **explicit type-hinting**, you don't need `combine` - instead, just do this:
+>
+> ```ts
+> import { create } from 'zustand'
+> 
+> interface BearStore {
+>   bears: number
+>   increasePopulation(): void
+>   removeAllBears(): void
+>   updateBears(newBears: number): void
+> }
+> 
+> const useStore = create<BearStore>((set) => ({
+>   bears: 0,
+>   increasePopulation: () => set((state) => ({ bears: state.bears + 1 })),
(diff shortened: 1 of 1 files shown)


dbritto-dev:
@mindplay-dk as far as I know we want to keep intro as it is right now, that's why we do have TypeScript guide. I'd like to improve TS guide rather than changing the intro.

mindplay-dk:
How can I put this.

Your documentation isn't great.

It's not that the writing is bad, really - it's that there is way too much of it, it's too detailed, and there is no progression.

Most people don't want to read an entire novel about all sort of advanced TypeScript patterns - the large majority of us need two or three things at the most, but the stuff that's on the landing page is *definitely* not enough for any real world use case.

You make it look simpler than it is, and people run off in 10 different directions. Every zustand codebase I've seen looks different - everyone comes up with different solutions to the same basic problems.

You've seen those two very long threads about computed state - the number of (weird, crazy, broken) solutions people come up with. What I'm trying to tell you is, it's the same out here in the wild. I assume because no one has been taught the basics, because there is nowhere (in official docs) to learn the basics.

I'm trying to help here.

Really, I'm trying to help my coworkers first, but none of this is specific to my coworkers, or anyone using a state manager.

If you don't want a landing page that covers the basics, would you consider putting it somewhere else?

I mean, I could go and put it all in a blog post, but no one will read it.

dai-shi:
I agree the TypeScript guide is too advanced.

@dbritto-dev You have work-in-progress doc revamp, right?
We should discuss based on it. Hope you can push it soon.

mindplay-dk:
What do you think about the issue of computed state?

No one commented on that.

There's been two very long threads about it - I spent an awful lot of time on this myself, before finally realizing that Zustand plays nice with React primitives for that, and doesn't actually need to do anything in this area.

I just don't think it's obvious at all.

The "getting started" guide shows you state and updates and says "that's it!" - I really think that laying out a few breadcrumbs to guide new users on their way could go a long way here...

mindplay-dk:
And regarding TypeScript, the example in "getting started" doesn't give you working types in TS.

No one thinks this is a problem? 🙂

dbritto-dev-seatpick:
@mindplay-dk well, in that case we need to rewrite the whole getting started guide

mindplay-dk:
> well, in that case we need to rewrite the whole getting started guide

the "whole" guide is 2 paragraphs and 2 code samples.

hardly a monumental effort? I probably did most of the leg work in this PR. 🤷‍♂️

I only kept those two sections collapsed because I figured you want this to *look* simple - like, that's part of your "brand"?

but like... *“Everything should be made as simple as possible, but not simpler.”*

I just don't think *reality* is quite as simple as you want it to appear here. 😌

If it were up to me, I would just unfold those two sections - or even put the JS and TS samples side by side, front and center...

You still have your pretty landing page for your "branding", if that's the concern?

PR #3209, opened 2025-08-13
feat: add zustand playground

Triumph-light:
PR Description:

Hi @dai-shi   👋
As a heavy Zustand user and React developer, I built a Playground for testing and prototyping state management in my projects. Since Zustand is my go-to state library, I thought this could be valuable for the community.
It serves two main purposes:
* Real-time debugging: Quickly test the library's latest features
* Lower barrier to entry: Help new users understand the library visually

Key features:
* Self-contained in /playground (no impact on core project)
* Zero dependencies

I'll maintain it alongside library updates

Happy to adjust the structure or location as needed. Thanks for your great work!😆

### DIFF
--- a/examples/demo/src/components/Details.jsx
+++ b/examples/demo/src/components/Details.jsx
@@ -3,6 +3,7 @@ export default function Details() {
     <>
       <nav className="nav">
         <a href="https://zustand.docs.pmnd.rs/">Documentation</a>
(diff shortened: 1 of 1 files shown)


dai-shi:
We can't guarantee to keep it in the future, because there are many playground solutions, one of which we generally prefer is stackblitz.

But, I'm fine to add yours for now and see how it goes.

@dbritto-dev Thoughts?

dbritto-dev:
@dai-shi  I don't see any benefits on adding another playground to the list

Triumph-light:
> @dai-shi I don't see any benefits on adding another playground to the list

@dbritto-dev Thank you for reviewing this PR. I completely understand your perspective. 

Personally, I find this lightweight approach offers better experience for quickly testing small demos since tools like StackBlitz and CodeSandbox require dependency installation first, which makes initial loading slower. 

This is just my personal observation - those tools are probably better suited for more complex cases. 

Regardless of whether this gets merged or not, I'm truly happy and honored to have participated in this discussion. Really appreciate the opportunity!

dbritto-dev-seatpick:
@Triumph-light but they're a bunch of alternatives to that even codepen, jsbin, jsfiddle, etc

dai-shi:
We used to have a playground link (codesandbox, which is no longer available and deleted) in our readme, so it's nice to have one.
livecodes.io is also pretty lightweight.

PR #3242, opened 2025-09-30
docs: updated basic usage of the TypeScript Guide

yuraBezh:

Fixes #

## Summary
The main goal of the PR is to make the Typescript guide more beginner-friendly, add/extend examples.  

## Check List

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -5,16 +5,30 @@ nav: 7
 
 ## Basic usage
 
-The difference when using TypeScript is that instead of writing `create(...)`, you have to write `create<T>()(...)` (notice the extra parentheses `()` too along with the type parameter) where `T` is the type of the state to annotate it. For example:
+When you create a Store with TypeScript, you should use a small variation of the `create` function call. Instead of `create(...)`, write `create<T>()(...)`.
+
+> [!NOTE]
+>
+> - The `<T>` is the type describing your state.
+> - The extra pair of parentheses.
+
+Two different jobs:
+
+1. The first call `create<T>()` is only for typing. The `<T>` tells TypeScript: “My store will look like type `T`.”
+2. The second call `(...)` is where you give Zustand the function that builds the store (`initial state` + `actions`).
+
+So you’re really saying: “Create a typed store of type `T`, and here is the function that creates everything necessary.”
 
 ```ts
 import { create } from 'zustand'
 
+// Describe the shape of your state and actions
 interface BearState {
   bears: number
   increase: (by: number) => void
 }
 
+// Pass the interface as a generic type <BearState>
 const useBearStore = create<BearState>()((set) => ({
   bears: 0,
   increase: (by) => set((state) => ({ bears: state.bears + by })),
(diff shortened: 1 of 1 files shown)


dai-shi:
The current TypeScript guide is somewhat advanced. Instead of making it beginner friendly, I think we should separate into to docs. @dbritto-dev What do you think?

dbritto-dev:
@dai-shi I'd like something like this -> https://react.dev/learn/typescript

dai-shi:
> @dai-shi I'd like something like this -> https://react.dev/learn/typescript

Not sure what exactly you are suggesting, but I take that we would have two docs:
1. TypeScript beginner guide
2. TypeScript advanced guide

dbritto-dev:
@dai-shi yeah, two docs

yuraBezh:
@dai-shi @dbritto-dev 
Thank you for the review and thoughts. 
I got both your points! I was working on the second, let's call it "TypeScript beginner guide" these days. The PR will be ready in 20 mins.

Regarding the first commit, IMO, the updates are also sensible. Not because it makes the advanced guide simpler, but because now it is easier to read and catch the main ideas instead of spending extra time/effort to understand slightly obscure code (without a simple explanation). So the plan was not to change the ideas that were provided. But to polish the existing code and formulations.

yuraBezh:
@dai-shi @dbritto-dev
So I made changes regarding the "Typescript Beginner Guide" in the new PR.
Please, check it: https://github.com/pmndrs/zustand/pull/3246

Let me know. Will we continue working on the current PR?

dbritto-dev (approved):
LGTM

dbritto-dev:
@dai-shi any thoughts?

docs/guides/typescript.md
```diff
@@ -5,16 +5,30 @@ nav: 7
...
+>
+> - The `<T>` is the type describing your state.
+> - The extra pair of parentheses.
+
+Two different jobs:
+
+1. The first call `create<T>()` is only for typing. The `<T>` tells TypeScript: “My store will look like type `T`.”
+2. The second call `(...)` is where you give Zustand the function that builds the store (`initial state` + `actions`).
+
+So you’re really saying: “Create a typed store of type `T`, and here is the function that creates everything necessary.”
```
dai-shi: I don't think this is technically correct, and it's a bit confusing. Can we remove this?

devanshj:
My humble suggestion would be to not do this "dumbing down" of the TypeScript deep dives. They are meant for user who want to learn something. By dumbing down and removing words like "invariant" and replacing it with "2 opposite ways" you're taking away an opportunity to learn something new by googling it. You've also removed the most important bit - the `withError` example along with the link to the TypeScript issue - that explains why the currying is needed. These deep dives are already hidden in the collapsibles so your Average Joe won't see them anyway. I don't mind a separate beginner friendly section/document but let's keep these advanced bits as it is.

Just a suggestion, I'll let @dai-shi and maintainers decide how we want to proceed forward.

docs/guides/typescript.md
yuraBezh: Yes, agreed. It's removed

dai-shi:
> I don't mind a separate beginner friendly section/document but let's keep these advanced bits as it is.

Thanks. Yeah, that's what I felt.

> Just a suggestion, I'll let @dai-shi and maintainers decide how we want to proceed forward.

I mean, we can decide it at last, but your input is very helpful.

dai-shi:
Let's rename this guide to `./docs/guides/advanced-typescript.md` or something, and only change the title.
```diff
- title: TypeScript Guide
+ title: Advanced TypeScript Guide
```
Please open a new PR.

PR #3087, opened 2025-04-18
fix(package): react-native condition

dai-shi:
following https://github.com/pmndrs/jotai/pull/3042

related discussions:
- #1967 
- #3089 
- #3090

PR #3243, opened 2025-10-01
feat: add browser condition to package.json exports for better React Native Web and Metro support

fkreinh:
## Summary

This PR adds the `browser` condition to the package.json `exports` field to improve compatibility with React Native Web and Metro bundler (React Native 0.79+).

## Motivation

With React Native 0.79, Metro introduced stable support for package.json `exports` field resolution as part of the modern ES Module resolution strategy. This change ensures that Zustand properly resolves modules in React Native Web environments and Metro bundler configurations.

According to the [React Native 0.79 release notes](https://reactnative.dev/blog/2025/04/08/react-native-0.79#metro-faster-startup-and-package-exports-support), Metro now matches different conditions depending on the platform:
- For native platforms: `react-native` condition
- For web exports: `browser` condition  
- For server exports: `node`, `react-server`, and `workerd` conditions

## Changes

- Added `browser` condition to the root export (`.`) in package.json
- Added `browser` condition to the wildcard export (`./*`) in package.json
- Both conditions point to the CommonJS builds (`./index.js` and `./*.js`)

## Benefits

1. **Better React Native Web support**: Ensures proper module resolution when using Zustand in React Native Web projects
2. **Metro 0.82+ compatibility**: Aligns with Metro's ES Module resolution strategy
3. **Expo compatibility**: Works seamlessly with Expo's Metro configuration (SDK 53+)
4. **Consistent behavior**: Maintains the same module resolution as the existing `react-native` condition

## Testing

The changes maintain backward compatibility as they only add a new condition without modifying existing ones. The resolution order in the exports map ensures that:
1. `react-native` condition is checked first for React Native environments
2. `browser` condition is checked for web/browser environments
3. `import` condition is checked for ESM imports
4. `default` condition is the fallback

## References

- [React Native 0.79 - Metro package exports support](https://reactnative.dev/blog/2025/04/08/react-native-0.79#metro-faster-startup-and-package-exports-support)
- [Expo Metro Config - ES Module resolution](https://docs.expo.dev/versions/v53.0.0/config/metro/#es-module-resolution)
- [Node.js package.json exports documentation](https://nodejs.org/api/packages.html#conditional-exports)

## Checklist

- [x] Follows conventional commit spec
- [x] Changes are focused and minimal
- [x] Maintains backward compatibility
- [x] No breaking changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,10 @@
         "types": "./index.d.ts",
         "default": "./index.js"
       },
+      "browser": {
+        "types": "./index.d.ts",
+        "default": "./index.js"
+      },
       "import": {
         "types": "./esm/index.d.mts",
         "default": "./esm/index.mjs"
@@ -42,6 +46,10 @@
     },
     "./*": {
       "react-native": {
+        "types": "./*.d.ts",
+        "default": "./*.js"
+      },
+    "browser": {
         "types": "./*.d.ts",
         "default": "./*.js"
       },


dai-shi:
I'm not sure how this helps.

> * For web exports: `browser` condition

Why doesn't it fallback to `default` condition?

I'm hesitant to make `browser` to limit to only CJS, because modern browser supports ESM.

fkreinh:
Metro 0.82+ does not fall back to `default` for web builds — it requires a `browser` condition. Without it, Metro resolves to the ESM entry (`import`), which triggers `import.meta` errors in React Native Web.  
([React Native 0.79 release notes – Metro package exports support](https://reactnative.dev/blog/2025/04/08/react-native-0.79#metro-faster-startup-and-package-exports-support))

The reason `browser` points to CJS instead of ESM is that Metro still expects a CommonJS-compatible build for React Native Web. If we point `browser` to ESM, Metro resolves `./esm/index.mjs` and fails with:  

```
Cannot use 'import.meta' outside a module
```
([Expo Metro config – ES Module resolution](https://docs.expo.dev/versions/v53.0.0/config/metro/#es-module-resolution))

Modern bundlers and browsers that want ESM will resolve through the `import` condition instead.  
([Node.js package.json exports documentation](https://nodejs.org/api/packages.html#conditional-exports))

In short, this addition aligns Zustand’s exports with Metro’s platform-specific resolution strategy and prevents runtime errors in Expo/React Native Web.

**Compatibility / No Breaking Changes**  
This change only *adds* the `browser` condition; it does not alter existing ones.  
- Node.js ignores `browser` and continues resolving `import`/`default` as before.  
- Webpack, Vite, and Rollup prioritize `import` for ESM, so they won’t be affected.  
- The only environments impacted are React Native Web builds with Metro, which now resolve correctly.  

Therefore, this should not break any existing consumers of Zustand.

dai-shi:
I had a closer look, and found this in a tricky situation.
As I believe "browser" condition could be used by native browsers, we'd prefer it ESM rather than CJS.
What's tricky is that RNW picks "browser" condition without "react-native" condition. To me, it should be "react-native" AND "browser".
So, my hope was 6a92d29 should work...

https://nodejs.org/api/packages.html#conditional-exports says:

> "browser" - any web browser environment.

as community convention, so I don't think it's only for react-native.

dai-shi:
Related: #3087

dai-shi:
backlink: https://github.com/pmndrs/jotai/discussions/3041#discussioncomment-14840711

fkreinh:
@dai-shi this config should make this PR not necessary on RN web. 
Closing this PR.

https://gist.github.com/Kudo/39041f31f3a055442f9200540e8da649

```js
module.exports = function (api) {
  api.cache(true);
  return {
    presets: [
      [
        'babel-preset-expo',
        {
          unstable_transformImportMeta: true,
        },
      ],
    ],
  };
};
```

dai-shi:
Yeah, that's probably only the workaround as of now.

If anyone is interested, please file an issue in the expo (or metro?) repo.
cc @byCedric

PR #3246, opened 2025-10-02
docs: created the new TypeScript Beginner Guide

yuraBezh:
## Summary
Created the new TypeScript Beginner Guide. This is the result of the [previous PR discussions](https://github.com/pmndrs/zustand/pull/3242)

## Check List

### DIFF
--- a/docs/guides/beginner-typescript.md
+++ b/docs/guides/beginner-typescript.md
@@ -86,12 +86,8 @@ function ResetZoo() {
   return (
     <div>
       <div>{bears}</div>
-      <button onClick={() => increase(5)}>
-        Increase by 5
-      </button>
-      <button onClick={reset}>
-        Reset
-      </button>
+      <button onClick={() => increase(5)}>Increase by 5</button>
+      <button onClick={reset}>Reset</button>
     </div>
   )
 }
@@ -163,16 +159,20 @@ interface BearState {
 
 const useBearStore = create<BearState>()(() => ({
   bears: 2,
-  food: 10
+  food: 10,
 }))
 
 // In components, you can use both stores safely
 function MultipleSelectors() {
   const { bears, food } = useBearStore(
-    useShallow((state) => ({ bears: state.bears, food: state.food }))
+    useShallow((state) => ({ bears: state.bears, food: state.food })),
   )
 
-  return <div>We have {food} units of food for {bears} bears</div>
+  return (
+    <div>
+      We have {food} units of food for {bears} bears
+    </div>
+  )
 }
 ```
 
@@ -361,13 +361,11 @@ function Zoo() {
 
   return (
     <div>
-      <div>{bears} bears and {fish} fish</div>
-      <button onClick={addBear}>
-        Add bear
-      </button>
-      <button onClick={addFish}>
-        Add fish
-      </button>
+      <div>
+        {bears} bears and {fish} fish
+      </div>
+      <button onClick={addBear}>Add bear</button>
+      <button onClick={addFish}>Add fish</button>
     </div>
   )
 }


docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,446 @@
...
+// Define types for state & actions
+interface BearState {
+  bears: number
+  food: string
+  increase: (by: number) => void
+  feed: (food: string) => void
+}
+
+// Create store with explicit generic <BearState>
+export const useBearStore = create<BearState>((set) => ({
```
dai-shi: Shouldn't TypeScript guide (for beginner or not) show `create<T>()(...)` usage?

yuraBezh: Good catch, that’s a valid point. Added this part

docs/guides/beginner-typescript.md
dai-shi, line 1: Can the file name be `docs/guides/beginner-typescript.md` or `docs/guides/basic-typescript.md`?

yuraBezh: Sure, done

yuraBezh: Also updated the title according to the "Advanced TypeScript Guide" title format

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,474 @@
...
+  food: string
+  increase: (by: number) => void
+  feed: (food: string) => void
+}
+
+// Fix the type with create<BearState>()
+const createBearStore = create<BearState>()
+
+// Initialize the store separately
+export const useBearStore = createBearStore((set) => ({
```
dai-shi: We don't recommend separating this. The extra parens `()`  only exists because of TS limitation.

Please stick with `create<T>()(...)` pattern throughout the guide.

```suggestion
export const useBearStore = create<BearState>()((set) => ({
```

docs/guides/beginner-typescript.md
dai-shi: We should always tell `create<T>()(...)` for beginners. `create<T>(...)` works for some cases, but it's for advanced users.

docs/guides/beginner-typescript.md
yuraBezh: Got it. Fixed

docs/guides/beginner-typescript.md
yuraBezh: @dai-shi  I switched to create<T>()(...). Also updated examples to make them more consistent. Could you please take a look?

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,490 @@
...
+- Zustand ExtractState: a convenience helper that does the same thing but reads more clearly in a Zustand context.
+
+Both produce the same result. Zustand just added `ExtractState` because `ReturnType<typeof store.getState>` looked a bit “too TypeScript-heavy”.
+
+```ts
+import { ExtractState } from 'zustand'
+
+type BearState = ExtractState<typeof useBearStore>
+// or
+type BearSnapshot = ReturnType<typeof useBearStore.getState>
```
dai-shi: Should we only show the first method? Isn't it enough for beginners?

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,490 @@
...
+}
+
+const bearCreator: StateCreator<BearState> = (set) => ({
+  bears: 0,
+  increase: () => set((s) => ({ bears: s.bears + 1 })),
+})
+
+const bearStoreCreator = devtools(bearCreator)
+
+export const useBearStore = create<BearState>()(bearStoreCreator)
```
dai-shi: I don't prefer this separation either. Let's suggest `create<T>()(devtools(...))` pattern in the guide for all middleware.

dai-shi (dismissed):
Sorry for the delay. Thanks for your patience.

docs/guides/beginner-typescript.md
yuraBezh: I agree, let's make it more straightforward

docs/guides/beginner-typescript.md
yuraBezh: @dai-shi Sure! 
I unified the view of the middleware usage with `create<T>()(devtools(...))` pattern and without redundant store creators. Please review.

dai-shi:
(sorry for the delay. will have a look when i get time.)

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+
+function BearFeeder() {
+  // Destructure multiple fields if needed
+  const { food, feed } = useBearStore()
+  return (
+    <button onClick={() => feed("berries")}>
+      Feed bears {food}
+    </button>
+  )
+}
```
dai-shi: Do we need to show this usage in this guide? Usually, people get confused with this usage because it will rerender with other (hidden) value changing.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+```
+
+### Multiple Stores
```
dai-shi: This section seems too long for its contribution and the rarity of the use case. Could you reduce the example code? Or, maybe just move this section closer to the end of guide.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+}
+```
+
+### Extracting Types
+
+Zustand provides a built-in helper called `ExtractState`. This is useful for tests, utility functions, or component props.
+It returns the full type of your store’s state and actions without having to manually redefine them.
+
+```ts
+import { ExtractState } from 'zustand'
```
dai-shi: To be compatible with `verbatimModuleSyntax`.
```suggestion
import type { ExtractState } from 'zustand'
```

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+
+type BearState = ExtractState<typeof useBearStore>
+```
+
+Extracting the Store type:
+
+```ts
+// store.ts
+import { create, ExtractState } from 'zustand'
```
dai-shi:
```suggestion
import { create, type ExtractState } from 'zustand'
```

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+}
+
+logBearState(useBearStore.getState())
+```
+
+### Selectors
+
+#### Multiple Selectors
+
+Sometimes you need more than one property. Returning an object from the selector gives you both.
```
dai-shi: But, this would be easy to lead to extra rerenders. We don't recommend returning an object.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+}
+
+const useBearStore = create<BearState>()(() => ({
+  bears: 2,
+  food: 10
+}))
+
+// In components, you can use both stores safely
+function MultipleSelectors() {
+  const { bears, food } = useBearStore()
```
dai-shi: Same; Usually, we don't do this.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,484 @@
...
+      </button>
+      <button onClick={() => useBearStore.setState({ location: "mountains" })}>
+        Change location
+      </button>
+    </>
+  )
+}
+```
+
+### Additional API
```
dai-shi: It might be better to describe this with the previous hooks, except for useShallow, as they are used with the pair.

dai-shi (commented):
It depends on if this guide will be the first doc _or_ it will be read after README. It feels a bit repetitive in the latter case. Even for the former case, we should tweak slightly and show important features first and less common features later. See comments below for example.

docs/guides/beginner-typescript.md
yuraBezh: I also felt this. Let's simplify the example

docs/guides/beginner-typescript.md
yuraBezh: I would move this to the end of the guide. I think it's useful to highlight for them the 'multiple stores' usage for the future

docs/guides/beginner-typescript.md
yuraBezh: @dai-shi Thank you for the review. 
I moved them to the "Hooks" section and removed the "Additional API" documentation block. 
The "createWithEqualityFn" section is now a standalone documentation section.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,486 @@
...
+### Extracting Types
+
+Zustand provides a built-in helper called `ExtractState`. This is useful for tests, utility functions, or component props.
+It returns the full type of your store’s state and actions without having to manually redefine them.
+
+```ts
+import type { ExtractState } from 'zustand'
+
+type BearState = ExtractState<typeof useBearStore>
+```
```
dai-shi: Feels a bit repetitive with the following example.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,486 @@
...
+
+const useBearStore = create<BearState>()(() => ({
+  bears: 2,
+  food: 10
+}))
+
+// In components, you can use both stores safely
+function MultipleSelectors() {
+  const { bears, food } = useBearStore(
+    useShallow((state) => ({ bears: state.bears, food: state.food }))
```
dai-shi: If we show `useShallow` usage here, you want to `import` it.

docs/guides/beginner-typescript.md
yuraBezh: Agree, it's probably leftover from the previous version

docs/guides/beginner-typescript.md
yuraBezh: It makes sense to me

dbritto-dev:
@yuraBezh I wouldn't add hooks section due to we do have those in apis, hooks and middlewares sections I feel that a bit redundant

yuraBezh:
@dbritto-dev @dai-shi 
My understanding was that this guide should serve as a lightweight entry point summarizing the essentials.
That’s why the hooks section was included - to keep the overview complete and self-contained.

We already have separate, detailed sections for APIs, Hooks, and Middlewares - each with its own in-depth article.
If we want to emphasize a single source of truth, I can add links to the corresponding documentation pages. Or I can remove the "Hooks" section if that feels more consistent.

That’s just my view, though - happy to adjust based on what the team decides.

dai-shi:
I'm with @dbritto-dev on this.
While having TS beginner guide is a great idea, I don't assume people would only read this guide, so I would like to avoid duplication if it's not TS specific.
Maybe you can add links to other docs.

yuraBezh:
@dbritto-dev @dai-shi 
Thank you for the feedback. 
I’ve removed the "Hooks" section and added API documentation links.

dbritto-dev:
@yuraBezh thanks, I'd to improve the troubleshooting section in those docs but I need gather some cases from discussions first hehe

yuraBezh:
> @yuraBezh thanks, I'd to improve the troubleshooting section in those docs but I need gather some cases from discussions first hehe

@dbritto-dev @dai-shi 
Is there anything else I should do for this PR, or are we good?
If the Troubleshooting section isn’t ready yet, I can remove the last commit with related links for now and open a small PR with links once Troubleshooting is finalized.

BTW, I see some failing checks - not sure if they’re related to the PR changes or not. All tests passed locally.

dai-shi:
I'm waiting for @dbritto-dev review.

Not sure about CI either. Maybe if we retrigger it, it may pass.

dai-shi:
> Not sure about CI either. Maybe if we retrigger it, it may pass.

Maybe it doesn't. Not sure how it happens. We need to figure it out.

dbritto-dev:
@dai-shi could it be `@redux-devtools/extension`?

dai-shi:
> @dai-shi could it be `@redux-devtools/extension`?

Seems like it's due to the node version. In CI, the `lts` version is automatically picked.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,380 @@
...
+  feed: (food) => set(() => ({ food })),
+}))
+```
+
+### Using the Store in Components
+
+Inside components, you can read state and call actions. Selectors `(s) => s.bears` subscribe to only what you need.
+This reduces re-renders and improves performance. JS can do this too, but with TS your IDE autocompletes state fields.
+
+```ts
```
dai-shi:
```suggestion
```tsx
```

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,380 @@
...
+  return <h1>{bears} bears around</h1>
+}
+```
+
+### Resetting the Store
+
+Resetting is useful after logout or “clear session”. We use `typeof initialState` to avoid repeating property types.
+TypeScript updates automatically if `initialState` changes. This is safer and cleaner compared to JS.
+
+```ts
```
dai-shi:
```suggestion
```tsx
```

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,380 @@
...
+
+#### Multiple Selectors
+
+Sometimes you need more than one property. Returning an object from the selector lets you access multiple fields at once.
+However, directly destructuring properties from that object can cause unnecessary re-renders.
+To avoid this, it’s recommended to wrap the selector with `useShallow`, which prevents re-renders when the selected values remain shallowly equal.
+This is more efficient than subscribing to the whole store. TypeScript ensures you can’t accidentally misspell `bears` or `food`.
+See the [API documentation](https://zustand.docs.pmnd.rs/hooks/use-shallow) for more details on `useShallow`.
+
+```ts
```
dai-shi:
```suggestion
```tsx
```

docs/guides/beginner-typescript.md
dai-shi: By the way, we could also do this. 
```ts
const [bears, food] = useBearStore(useShallow((s) => [s.bears, s.food]));
```
I guess people like objects over arrays.

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,380 @@
...
+  return <div>We have {food} units of food for {bears} bears</div>
+}
+```
+
+#### Derived State with Selectors
+
+Not all values need to be stored directly - some can be computed from existing state. You can derive values using selectors.
+This avoids duplication and keeps the store minimal. TypeScript ensures `bears` is a number, so math is safe.
+
+```ts
```
dai-shi:
```suggestion
```tsx
```

docs/guides/beginner-typescript.md
```diff
@@ -0,0 +1,380 @@
...
+// or
+const bears = useBearStore((s) => ({ bears: s.bears }), shallow)
+```
+
+### Multiple Stores
+
+You can create more than one store for different domains. For example, `BearStore` manages bears and `FishStore` manages fish.
+This keeps state isolated and easier to maintain in larger apps. With TypeScript, each store has its own strict type - you can’t accidentally mix bears and fish.
+
+```ts
```
dai-shi:
```suggestion
```tsx
```

dai-shi (approved):
LGTM

PR #3252, opened 2025-10-07
Filter out specific actions from being sent to Redux DevTools

totorototo:
# Action Blacklist Feature for Zustand DevTools Middleware

This enhancement adds support for filtering out specific actions from being sent to Redux DevTools, similar to the `actionBlacklist` feature in Redux DevTools.

## Overview

The `actionBlacklist` option allows you to prevent certain actions from appearing in the Redux DevTools extension. This is useful for:

- Hiding sensitive or private actions
- Reducing noise in the DevTools timeline
- Improving debugging experience by focusing on relevant actions
- Filtering out frequently called actions that are not important for debugging

### Array-based Filtering

Filter out specific action types by providing an array of action names:

```typescript
import { devtools } from 'zustand/middleware'

const useStore = create()(
  devtools(
    (set) => ({
      count: 0,
      increment: () =>
        set((state) => ({ count: state.count + 1 }), false, 'increment'),
      secretAction: () =>
        set((state) => ({ count: state.count * 2 }), false, 'secretAction'),
    }),
    {
      name: 'counter-store',
      actionBlacklist: ['secretAction'], // This action won't appear in DevTools
    },
  ),
)
```

### Function-based Filtering

For more complex filtering logic, provide a function that receives the action object and returns a boolean:

```typescript
const useStore = create()(
  devtools(
    (set) => ({
      // ... your state and actions
    }),
    {
      name: 'user-store',
      actionBlacklist: (action) => {
        // Filter out password-related and debug actions
        return (
          action.type.toLowerCase().includes('password') ||
          action.type.startsWith('DEBUG_')
        )
      },
    },
  ),
)
```

### With Store Option

The filtering also works when using the `store` option for multiple stores:

```typescript
const useOrderStore = create()(
  devtools(
    (set) => ({
      // ... your state and actions
    }),
    {
      name: 'app-stores',
      store: 'orders',
      actionBlacklist: ['internalCleanup'], // Filtered before store prefix is added
    },
  ),
)
```

### `actionBlacklist`

**Type:** `string[] | ((action: { type: string }) => boolean)`

**Description:** Defines which actions should be filtered out from Redux DevTools.

#### Array Format

- **Type:** `string[]`
- **Description:** Array of action type strings to filter out
- **Example:** `['reset', 'secretAction', 'internalUpdate']`

#### Function Format

- **Type:** `(action: { type: string }) => boolean`
- **Parameters:**
  - `action`: The action object containing at least a `type` property
- **Returns:** `true` if the action should be filtered out, `false` otherwise
- **Example:** `(action) => action.type.startsWith('private')`

## Behavior

- Actions are filtered **before** any store prefixes are added (when using the `store` option)
- Filtered actions still execute normally; they just don't appear in DevTools
- Anonymous actions can be filtered by targeting the `anonymousActionType` or the default 'anonymous' type
- The filtering is applied only to the DevTools output, not to the actual state changes

### Filtering Debug Actions

```typescript
actionBlacklist: (action) => action.type.startsWith('DEBUG_')
```

### Filtering Sensitive Operations

```typescript
actionBlacklist: ['updatePassword', 'setApiKey', 'loginWithToken']
```

### Filtering High-Frequency Actions

```typescript
actionBlacklist: (action) =>
  ['mousemove', 'scroll', 'typing'].includes(action.type)
```

### Case-Insensitive Filtering

```typescript
actionBlacklist: (action) => action.type.toLowerCase().includes('secret')
```

## Compatibility

This feature is compatible with all existing DevTools middleware options and works with:

- Single stores and multiple stores
- Named and unnamed connections
- Store prefixes
- All existing DevTools features (time travel, action replay, etc.)

The filtered actions are excluded only from the DevTools display but continue to function normally in your application.

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -77,6 +77,7 @@ export interface DevtoolsOptions extends Config {
   enabled?: boolean
   anonymousActionType?: string
   store?: string
+  actionBlacklist?: string[] | ((action: { type: string }) => boolean)
 }
 
 type Devtools = <
@@ -178,7 +179,7 @@ const findCallerName = (stack: string | undefined) => {
 const devtoolsImpl: DevtoolsImpl =
   (fn, devtoolsOptions = {}) =>
   (set, get, api) => {
-    const { enabled, anonymousActionType, store, ...options } = devtoolsOptions
+    const { enabled, anonymousActionType, store, actionBlacklist, ...options } = devtoolsOptions
 
     type S = ReturnType<typeof fn> & {
       [store: string]: ReturnType<typeof fn>
@@ -203,6 +204,17 @@ const devtoolsImpl: DevtoolsImpl =
     const { connection, ...connectionInformation } =
       extractConnectionInformation(store, extensionConnector, options)
 
+    const isActionBlacklisted = (action: { type: string }): boolean => {
+      if (!actionBlacklist) return false
+      if (Array.isArray(actionBlacklist)) {
+        return actionBlacklist.includes(action.type)
+      }
+      if (typeof actionBlacklist === 'function') {
+        return actionBlacklist(action)
+      }
+      return false
+    }
+
     let isRecording = true
     ;(api.setState as any) = ((state, replace, nameOrAction: Action) => {
       const r = set(state, replace as any)
@@ -218,6 +230,12 @@ const devtoolsImpl: DevtoolsImpl =
           : typeof nameOrAction === 'string'
             ? { type: nameOrAction }
             : nameOrAction
+      
+      // Check if action should be filtered out
+      if (isActionBlacklisted(action)) {
+        return r
+      }
+      
       if (store === undefined) {
         connection?.send(action, get())
         return r
--- a/tests/devtools.test.tsx
+++ b/tests/devtools.test.tsx
@@ -2520,3 +2520,53 @@ describe('cleanup', () => {
     )
   })
 })
+
+describe('actionBlacklist', () => {
+  it('should filter out blacklisted actions (array)', async () => {
+    const options = {
+      name: 'test-filter',
+      enabled: true,
+      actionBlacklist: ['secretAction'],
+    }
+    const api = createStore(
+      devtools(() => ({ count: 0 }), options),
+    )
+
+    // Normal action should be sent
+    api.setState({ count: 1 }, false, 'increment')
+    const [connection] = getNamedConnectionApis(options.name)
+    expect(connection.send).toHaveBeenLastCalledWith(
+      { type: 'increment' },
+      { count: 1 },
+    )
+
+    // Blacklisted action should be filtered out
+    connection.send.mockClear()
+    api.setState({ count: 2 }, false, 'secretAction')
+    expect(connection.send).not.toHaveBeenCalled()
+  })
+
+  it('should filter out blacklisted actions (function)', async () => {
+    const options = {
+      name: 'test-func-filter',
+      enabled: true,
+      actionBlacklist: (action: { type: string }) => action.type.startsWith('private'),
+    }
+    const api = createStore(
+      devtools(() => ({ count: 0 }), options),
+    )
+
+    // Normal action should be sent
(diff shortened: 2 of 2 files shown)


dai-shi:
Thanks for the suggestion.

> similar to the `actionBlacklist` feature in Redux DevTools.

Could you put the link to the feature in Redux DevTools?

src/middleware/devtools.ts
```diff
@@ -203,6 +205,17 @@ const devtoolsImpl: DevtoolsImpl =
     const { connection, ...connectionInformation } =
       extractConnectionInformation(store, extensionConnector, options)
 
+    const isActionBlacklisted = (action: { type: string }): boolean => {
+      if (!actionBlacklist) return false
```
dai-shi: Just stylistic preference.

```suggestion
      if (!actionBlacklist) {
        return false
      }
```

Or:
```ts
const isActionBlacklisted = (action: { type: string }): boolean =>
  !actionBlacklist ? false
  : Array.isArray(actionBlacklist) ? actionBlacklist.includes(action.type)
  : typeof actionBlacklist === 'function' ? actionBlacklist(action)
  : false
```
but maybe it's not very readable.

Hm, I don't think we need `!actionBlacklist ? false` at all.

So, it can be:
```suggestion
```

dai-shi (commented):
The code looks reasonable in general.

totorototo:
> Thanks for the suggestion.
> 
> > similar to the `actionBlacklist` feature in Redux DevTools.
> 
> Could you put the link to the feature in Redux DevTools?

here it is: 

https://github.com/reduxjs/redux-devtools/blob/main/extension/docs/API/Arguments.md#actionsdenylist--actionsallowlist

dai-shi:
Let's use the same terminology `actionsDenylist` and same signature `string | string[]` then.

src/middleware/devtools.ts
```diff
@@ -203,6 +213,17 @@ const devtoolsImpl: DevtoolsImpl =
...
+    const isActionDenylisted = (action: { type: string }): boolean => {
+      if (!actionsDenylist) return false
+      if (Array.isArray(actionsDenylist)) {
+        return actionsDenylist.includes(action.type)
+      }
+      if (typeof actionsDenylist === 'function') {
+        return actionsDenylist(action)
+      }
+      return false
+    }
```
dai-shi: Do you want to support plain string list and function filter as opposed to be compatible with Redux one (regex string and array of regex string)?

totorototo: right, it should be fully compatible with Redux DevTools' standard configuration.

dai-shi (commented):
Oh, do we just need to passthrough it? That's sweet.

Remaining:
- fix CI error (format?)
- update README

dai-shi:
Thanks. Please run prettier, which should be easier.

> update README

Oops, I meant the PR description. But, fixing the docs is great!

src/middleware/devtools.ts
```diff
@@ -218,6 +218,7 @@ const devtoolsImpl: DevtoolsImpl =
           : typeof nameOrAction === 'string'
             ? { type: nameOrAction }
             : nameOrAction
+
```
dai-shi:
```suggestion
```

dai-shi:
Hm, merging main causes the CI failure? Could you check it? Let me know if you need any help.
Also, please update PR description to focus on docs and tests.

totorototo:
> Hm, merging main causes the CI failure? Could you check it? Let me know if you need any help. Also, please update PR description to focus on docs and tests.

Sure, I will!

totorototo:
tests passed locally. linter / formatter issue from time to time.

dai-shi:
We need to figure out how CIs are failing.
It seems unrelated to this PR change, but it's related with devtools. Maybe you have some insight?

dai-shi:
> We need to figure out how CIs are failing. It seems unrelated to this PR change, but it's related with devtools. Maybe you have some insight?

https://github.com/pmndrs/zustand/pull/3246#issuecomment-3471257012

PR #3269, opened 2025-10-18
chore: add support for depandabot

dbritto-dev:
## Check List

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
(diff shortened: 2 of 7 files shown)


dai-shi:
@dbritto-dev Can you add some reference links and your motivation to the PR description?

dbritto-dev:
@dai-shi time to time we use to update the packages to keep everything up to date but task can be automated with a depandabot setup so we don't need to worry about updating packages.

For instance: https://github.com/pmndrs/zustand/pull/3267

.github/depandabot.yml
```diff
@@ -0,0 +1,16 @@
...
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+
+  - package-ecosystem: 'docker'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+
```
dai-shi, line 12: Is it correct that we don't use these?

dai-shi:
> For instance: #3267

It's kind of rare case. I'm not sure if I like the bot opens PRs. If I don't, I'll revert. We will see.

dbritto-dev:
@dai-shi for updating packages and do housekeeping I think is fine in any case you can close the PRs.

dbritto-dev:
@dai-shi I removed the docker config but npm is ok that covers: `npm`, `pnpm` and `yarn`

Ref: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-

dai-shi:
> you can close the PRs.

If I close everything, there's no point of having it in the first place. Anyway, we'll see. It might be helpful to notice it.

dbritto-dev:
@dai-shi in any case you can read more about depandabot here: https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories

PR #3307, opened 2025-11-29
chore: publish.yml

dai-shi:
with trusted publisher

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,6 +6,7 @@ on:
 
 permissions:
   id-token: write
+  contents: read
 
 jobs:
   publish:


.github/workflows/publish.yml
```diff
@@ -4,6 +4,9 @@ on:
   release:
     types: [published]
 
+permissions:
+  id-token: write
```
dbritto-dev, line 8:
```suggestion
  id-token: write
  contents: read
```

dai-shi: > The critical requirement is the id-token: write permission, which allows GitHub Actions to generate OIDC tokens. 

I tested this in https://github.com/wakujs/waku-jotai and it works fine.

dai-shi: https://github.com/actions/checkout?tab=readme-ov-file#recommended-permissions

After reading that, it looks like the recommended default as it was the previous behavior.

dai-shi (commented):
should be good

PR #3320, opened 2025-12-09
docs: add memoized selector usage with `useStore`

oleksandr-danylchenko:

Fixes #

## Summary
When I went through the ["Initialize state with props"](https://zustand.docs.pmnd.rs/guides/initialize-state-with-props) guide and the [Extracting context logic into a custom hook](https://zustand.docs.pmnd.rs/guides/initialize-state-with-props#extracting-context-logic-into-a-custom-hook) section specifically, it wasn't clear to me that using `useShallow` is allowed for memoizing the selector's output.
I only discovered that it's a valid pattern in the https://github.com/pmndrs/zustand/discussions/2203. Therefore, it would be nice to specify such an option explicitly.

## Check List

### DIFF
--- a/docs/getting-started/comparison.md
+++ b/docs/getting-started/comparison.md
@@ -49,14 +49,13 @@ type State = {
   count: number
 }
 
-type Actions = {
-  increment: (qty: number) => void
-  decrement: (qty: number) => void
+type Action = {
+  type: 'increment' | 'decrement'
+  qty: number
 }
 
-type Action = {
-  type: keyof Actions
-  qty: number
+type Actions = {
+  dispatch: (action: Action) => void
 }
 
 const countReducer = (state: State, action: Action) => {
--- a/docs/guides/slices-pattern.md
+++ b/docs/guides/slices-pattern.md
@@ -94,7 +94,7 @@ export const useBoundStore = create((...a) => ({
 
 Adding middlewares to a combined store is the same as with other normal stores.
 
-Adding `persist` middleware to our `useBoundStore`:
+Adding [`persist` middleware](../integrations/persisting-store-data.md) to our `useBoundStore`:
 
 ```js
 import { create } from 'zustand'
--- a/docs/guides/tutorial-tic-tac-toe.md
+++ b/docs/guides/tutorial-tic-tac-toe.md
@@ -1,7 +1,7 @@
 ---
 title: 'Tutorial: Tic-Tac-Toe'
 description: Building a game
-nav: 0
+nav: 1
 ---
 
 # Tutorial: Tic-Tac-Toe
--- a/docs/integrations/third-party-libraries.md
+++ b/docs/integrations/third-party-libraries.md
@@ -16,6 +16,7 @@ This can be done using third-party libraries created by the community.
 
 - [@colorfy-software/zfy](https://colorfy-software.gitbook.io/zfy/) — 🧸 Useful helpers for state management in React with Zustand.
 - [@csark0812/zustand-expo-devtools](https://github.com/csark0812/zustand-expo-devtools) — 🧭 Connect Zustand to Redux DevTools in Expo + React Native using the official Expo DevTools plugin system.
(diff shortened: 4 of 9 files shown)


docs/guides/initialize-state-with-props.md
```diff
@@ -126,6 +126,34 @@ function CommonConsumer() {
 }
 ```
 
+### Optionally use memoization for stable selector outputs
```
dai-shi: Hm, I don't think we call `useShallow` a memoization solution. It doesn't memoize the function. Can you come up with other titles?

oleksandr-danylchenko: @dai-shi, the "memoized selector" term is used relatively extensively on the `useShallow` documentation page:

https://github.com/pmndrs/zustand/blob/bf07ccd2aea0d1ff5fe4028641f0aeb192bf587f/docs/hooks/use-shallow.md?plain=1#L3
https://github.com/pmndrs/zustand/blob/bf07ccd2aea0d1ff5fe4028641f0aeb192bf587f/docs/hooks/use-shallow.md?plain=1#L10
https://github.com/pmndrs/zustand/blob/bf07ccd2aea0d1ff5fe4028641f0aeb192bf587f/docs/hooks/use-shallow.md?plain=1#L36-L37
https://github.com/pmndrs/zustand/blob/bf07ccd2aea0d1ff5fe4028641f0aeb192bf587f/docs/hooks/use-shallow.md?plain=1#L41

oleksandr-danylchenko: I [rephrased](https://github.com/pmndrs/zustand/compare/4b249bd25272d9c073e7b690abd5f9ade4ac610e..8113292136da0db26719c3516333812345417a7b) the title to say "Optionally use **_memoized selector_** for stable outputs"

dai-shi: Oh, good point.
It's not memoizing the function, but memoizing (or caching) the function output.
Let me see @dbritto-dev 's thoughts. Please wait for that.

dbritto-dev: @oleksandr-danylchenko "memoized selector" != "stable outputs", memoizing selector are just a micro optimization we should focus on stable outputs rather than stable/memoized selectors

oleksandr-danylchenko: @dbritto-dev, unfortunately  I'm not sure that I understand what needs to be changed in the PR 🤔 

According to the [docs](https://zustand.docs.pmnd.rs/hooks/use-shallow#returns):
> `useShallow` returns a **_memoized version of a selector_** function _**using a shallow comparison for memoization**_.

Then, that shallow comparison provides an ability to "stabilize" the memoized selector's output. Judging by an example on that page:
```ts
const names = useBearFamilyMealsStore(
    useShallow((state) => Object.keys(state)),
 )
```
We can say that brand-new arrays created on the state's update are compared not by their reference, but by their value.

oleksandr-danylchenko: Nevertheless, I agree that, unlike [`reselect`](https://reselect.js.org/), `useShallow` doesn't prevent the selector's body from evaluation when the state changes.

dbritto-dev: yeah, we should talk about stable results instead of memoized selectors

dbritto-dev:
@oleksandr-danylchenko LGTM

docs/guides/initialize-state-with-props.md
dai-shi: > yeah, we should talk about stable results instead of memoized selectors

@dbritto-dev Yeah, so my suggestion is to change the existing docs first. Could you open a PR?

Then, @oleksandr-danylchenko can update this PR accordingly.

dbritto-dev: @dai-shi sure, let me work on that

dai-shi (approved):
Let me merge this and @dbritto-dev can replace the terminology all at once.

PR #3321, opened 2025-12-09
chore(eslint): enable 'testing-library', 'jest-dom', 'vitest' plugins and fix violations

sukvvon:

*  relates https://github.com/pmndrs/jotai/pull/3190

## Summary

- Enable `testing-library`, `jest-dom`, `vitest` ESLint plugins for test files
- Add missing assertions to 19 type-only tests to fix `vitest/expect-expect` rule
- Rename `children` to `childItems` to fix `testing-library/no-node-access` false positive
- Add `eslint-disable-next-line` for `this.props.children` (React standard pattern)

## Check List

### DIFF
--- a/docs/guides/initialize-state-with-props.md
+++ b/docs/guides/initialize-state-with-props.md
@@ -44,10 +44,10 @@ export const BearContext = createContext<BearStore | null>(null)
 
 ```tsx
 // Provider implementation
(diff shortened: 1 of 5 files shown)


dai-shi (approved):
LGTM
Thanks!

(just curious if there's anything we could back port to Valtio?)

PR #3333, opened 2025-12-26
Fix Devtools action dispatching

msanchezdev:

Fixes #3081 

## Summary

Currently, we only seem to be able to dispatch actions from the devtools when using the `redux` middleware along with the `devtools` middleware (havent tested that case).

This PR adds support for exposing the store actions so they can be dispatched through the Redux Devtools without having to create reducers or anything else.

image

## Check List

- [ ] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,12 +157,16 @@
     "vitest": "^4.0.14"
   },
   "peerDependencies": {
+    "@redux-devtools/utils": "^3.1.1",
     "@types/react": ">=18.0.0",
     "immer": ">=9.0.6",
     "react": ">=18.0.0",
     "use-sync-external-store": ">=1.2.0"
   },
   "peerDependenciesMeta": {
+    "@redux-devtools/utils": {
+      "optional": true
+    },
     "@types/react": {
       "optional": true
     },
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -1,4 +1,5 @@
 import type {} from '@redux-devtools/extension'
+import type { ActionCreatorObject } from '@redux-devtools/utils';
 
 import type {
   StateCreator,
@@ -72,11 +73,44 @@ type StoreDevtools<S> = S extends {
     }
   : never
 
-export interface DevtoolsOptions extends Config {
+type Join<Prefix extends string, K extends string> = Prefix extends ""
+  ? K
+  : `${Prefix}${K}`;
+
+type NextPrefix<Prefix extends string, K extends string> = Prefix extends ""
+  ? `${K}.`
+  : `${Prefix}${K}.`;
+
+type FunctionPaths<T, Prefix extends string = ""> = {
+  [K in keyof T & string]: T[K] extends (...args: any[]) => any
+    ? Join<Prefix, K>
+    : T[K] extends Record<string, any>
+      ? FunctionPaths<T[K], NextPrefix<Prefix, K>> extends infer P
+        ? P extends string
+          ?
+            | P
+            | `${Prefix}${K}.*`
+            | `${Prefix}${K}.**`
+          : never
+        : never
+      : never;
+}[keyof T & string];
+
+type ActionCreatorsMask<T extends Record<string, any>> = {
+  [K in FunctionPaths<T> | '*' | '**' | (string & {})]?: boolean | (
+    // Allow overriding leafs
+    K extends '*' | '**' | `${string}.*` | `${string}.**`
+      ? never
+      : (...args: any[]) => void
+  );
+};
+
+export interface DevtoolsOptions<T extends Record<string, unknown> = {}> extends Omit<Config, 'actionCreators'> {
   name?: string
   enabled?: boolean
   anonymousActionType?: string
   store?: string
+  actionCreators?: ActionCreatorsMask<T>;
 }
 
 type Devtools = <
@@ -86,7 +120,7 @@ type Devtools = <
   U = T,
 >(
   initializer: StateCreator<T, [...Mps, ['zustand/devtools', never]], Mcs, U>,
-  devtoolsOptions?: DevtoolsOptions,
+  devtoolsOptions?: DevtoolsOptions<T & {}>,
 ) => StateCreator<T, Mps, [['zustand/devtools', never], ...Mcs]>
 
 declare module '../vanilla' {
@@ -131,7 +165,7 @@ const extractConnectionInformation = (
   extensionConnector: NonNullable<
     (typeof window)['__REDUX_DEVTOOLS_EXTENSION__']
   >,
-  options: Omit<DevtoolsOptions, 'enabled' | 'anonymousActionType' | 'store'>,
+  options: Omit<DevtoolsOptions, 'enabled' | 'anonymousActionType' | 'store' | 'actionCreators'>,
 ) => {
   if (store === undefined) {
     return {
@@ -200,9 +234,6 @@ const devtoolsImpl: DevtoolsImpl =
       return fn(set, get, api)
     }
 
-    const { connection, ...connectionInformation } =
-      extractConnectionInformation(store, extensionConnector, options)
-
     let isRecording = true
     ;(api.setState as any) = ((state, replace, nameOrAction: Action) => {
       const r = set(state, replace as any)
@@ -254,6 +285,57 @@ const devtoolsImpl: DevtoolsImpl =
     }
 
     const initialState = fn(api.setState, get, api)
+    let actionCreators: ActionCreatorObject[] = [];
+    let evalAction: (action: string, _actionCreators: ActionCreatorObject[]) => void | undefined;
+
+    if (options.actionCreators) {
+      try {
+        evalAction = require('@redux-devtools/utils').evalAction;
+        actionCreators = getActionsArray(
+          initialState as Record<string, unknown>,
+          options.actionCreators as ActionCreatorsMask<{}>
+        );
+        // override to pass it to the extension connector, any is ok, we dont care about the type anymore
+        options.actionCreators = actionCreators as any;
+      } catch {
+        console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use actionCreators in devtools middleware');
+      }
+    }
+    
+    let usingReduxMiddleware = typeof (api as any).dispatch === 'function';
+    if (!usingReduxMiddleware) {
+      (api as any).dispatchFromDevtools = true;
+      (api as any).dispatch = (payload: {
+        type: string,
+        args: unknown[] | Record<string, unknown>
+      }) => {
+        if (!evalAction) {
+          console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use the action dropdown in the devtools');
+          return;
+        }
+
+        if (!('type' in payload)) {
+          console.error('[zustand devtools middleware] Payload must have a "type" property which should match an action creator');
+          return;
+        }
+
+        const action = actionCreators.find((action) => action.name === payload.type);
+        if (!action) {
+          console.error(`[zustand devtools middleware] Action type "${payload.type}" not found in actionCreators`);
+          return;
+        }
+
+        if(Array.isArray(payload.args)) {
+          return action.func(...payload.args);
+        }
+        
+        return action.func(...action.args.map((arg) => (payload.args as Record<string, unknown> ?? {})[arg]));
+      }
+    }
+
+    // We connect after extracting action creators from the initial state
+    const { connection, ...connectionInformation } = extractConnectionInformation(store, extensionConnector, options);
+
     if (connectionInformation.type === 'untracked') {
       connection?.init(initialState)
     } else {
@@ -301,52 +383,96 @@ const devtoolsImpl: DevtoolsImpl =
       }
     ).subscribe((message: any) => {
       switch (message.type) {
-        case 'ACTION':
+        case 'ACTION': {
+          // When using the action dropdown we get an object with the selected action
           if (typeof message.payload !== 'string') {
-            console.error(
-              '[zustand devtools middleware] Unsupported action format',
-            )
-            return
-          }
-          return parseJsonThen<{ type: unknown; state?: PartialState }>(
-            message.payload,
-            (action) => {
-              if (action.type === '__setState') {
-                if (store === undefined) {
-                  setStateFromDevtools(action.state as PartialState)
-                  return
-                }
-                if (Object.keys(action.state as S).length !== 1) {
-                  console.error(
-                    `
-                    [zustand devtools middleware] Unsupported __setState action format.
-                    When using 'store' option in devtools(), the 'state' should have only one key, which is a value of 'store' that was passed in devtools(),
-                    and value of this only key should be a state object. Example: { "type": "__setState", "state": { "abc123Store": { "foo": "bar" } } }
-                    `,
-                  )
-                }
-                const stateFromDevtools = (action.state as S)[store]
-                if (
-                  stateFromDevtools === undefined ||
-                  stateFromDevtools === null
-                ) {
-                  return
-                }
-                if (
-                  JSON.stringify(api.getState()) !==
-                  JSON.stringify(stateFromDevtools)
-                ) {
-                  setStateFromDevtools(stateFromDevtools)
-                }
-                return
-              }
+            if (!evalAction) {
+              console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use the action dropdown in the devtools');
+              return;
+            }
 
+            if (usingReduxMiddleware) {
+              // When using the redux plugin, we dispatch the action ourselves.
+              const action = evalAction(message.payload, actionCreators as any[]);
               if (!(api as any).dispatchFromDevtools) return
               if (typeof (api as any).dispatch !== 'function') return
-              ;(api as any).dispatch(action)
-            },
-          )
+              return (api as any).dispatch(action);
+            } 
 
+            // When not using the redux plugin, action creators are expected
+            // to cause the state change themselves.
+            if (!(api as any).dispatchFromDevtools) return
+            return evalAction(message.payload, actionCreators as any[]);
+          }
+
+          let action: { type: string, state?: PartialState, [key: string]: unknown };
+          try {
+            action = JSON.parse(message.payload);
+          } catch {
+            let errorMessage = `[zustand devtools middleware] Malformed JSON. When dispatching custom actions, please format the payload as a JSON string.
+              Examples:`;
+            if (usingReduxMiddleware) {
+              errorMessage += `
+              - { "type": "increment", "amount": 1 }
+              - { "type": "reset" }
+              `;
+            } else {
+              errorMessage += `
+              - { "type": "increment" }
+              - { "type": "increment", "args": [1] }
+              If you have args with default values or want to be \`undefined\`, prefer using the object syntax so you can omit them.
+              - { "type": "increment", "args": { "amount": 1 } }
+              `;
+            }
+            console.error(errorMessage.replace(/\n\s+/g, '\n'));
+            return;
+          }
+
+          if (action.type === '__setState') {
+            if (store === undefined) {
+              if (!('state' in action)) {
+                console.warn(
+                  `
+                  [zustand devtools middleware] Calling __setState without a state property.
+                  This may not be what you intended, you should explicitly set the state you want to set.
+                  Example: { "type": "__setState", "state": { "foo": "bar" } }
+                  `.replace(/\n\s+/g, '\n'),
+                );
+                return;
+              }
+              setStateFromDevtools(action.state as PartialState)
+              return
+            }
+            if (Object.keys(action.state as S).length !== 1) {
+              console.error(
+                `
+                [zustand devtools middleware] Unsupported __setState action format.
+                When using 'store' option in devtools(), the 'state' should have only one key, which is a value of 'store' that was passed in devtools(),
+                and value of this only key should be a state object. Example: { "type": "__setState", "state": { "abc123Store": { "foo": "bar" } } }
+                `.replace(/\n\s+/g, '\n'),
+              )
+            }
+            const stateFromDevtools = (action.state as S)[store]
+            if (
+              stateFromDevtools === undefined ||
+              stateFromDevtools === null
+            ) {
+              return
+            }
+            if (
+              JSON.stringify(api.getState()) !==
+              JSON.stringify(stateFromDevtools)
+            ) {
+              setStateFromDevtools(stateFromDevtools)
+            }
+            return
+          }
+
+          if (!(api as any).dispatchFromDevtools) return
+          if (typeof (api as any).dispatch !== 'function') return
+          ;(api as any).dispatch(action)
+          return;
+        }
         case 'DISPATCH':
           switch (message.payload.type) {
             case 'RESET':
@@ -428,4 +554,66 @@ const parseJsonThen = <T>(stringified: string, fn: (parsed: T) => void) => {
     )
   }
   if (parsed !== undefined) fn(parsed as T)
+}
+
+function maskToRegex(mask: string): RegExp {
+  const escaped = mask
+    .replace(/\./g, '\\.')
+    .replace(/\*\*|\*/g, '.*');
+
+  return new RegExp(`^${escaped}$`);
+}
+
+function getActionsArray<T extends Record<string, any>>(actionCreators: T, mask: ActionCreatorsMask<T>): ActionCreatorObject[] {
+  const getActions = require('@redux-devtools/utils').getActionsArray;
+  const flat = getActions(actionCreators);
+  const actions = new Map<string, ActionCreatorObject>();
+  const customActions = new Map<string, ActionCreatorObject>(
+    getActions(mask).map((action: ActionCreatorObject) => [action.name, action])
+  );
+
+  // Sort matchers by specificity
+  const matchers = Object.entries(mask ?? {})
+    .map(([key, picked]) => ({
+      matcher: maskToRegex(key),
+      picked,
+      key,
+    })).toSorted((a, b) => {
+      const aIsDoubleWildcard = a.key.includes('**');
+      const bIsDoubleWildcard = b.key.includes('**');
+      if (aIsDoubleWildcard !== bIsDoubleWildcard) {
+        return aIsDoubleWildcard ? -1 : 1;
+      }
+      
+      const aIsSingleWildcard = a.key.includes('*') && !aIsDoubleWildcard;
+      const bIsSingleWildcard = b.key.includes('*') && !bIsDoubleWildcard;
+      if (aIsSingleWildcard !== bIsSingleWildcard) {
+        return aIsSingleWildcard ? -1 : 1;
+      }
+      
+      return a.key.length - b.key.length;
+    });
+
+  for (const action of flat) {
+    for (const { matcher, picked } of matchers) {
+      if (matcher.test(action.name)) {
+        if (picked === false) {
+          actions.delete(action.name);
+          continue;
+        }
+
+        actions.set(action.name, customActions.get(action.name) ?? {
+          name: action.name,
+          func: typeof picked === 'function' ? picked : action.func,
+          args: action.args,
+        });
+      }
+    }
+  }
+
+  for (const [name, action] of customActions.entries()) {
+    actions.set(name, action);
+  }
+
+  return Array.from(actions.values());
 }
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -302,10 +302,10 @@ const devtoolsImpl: DevtoolsImpl =
       }
     }
     
-    let usingReduxMiddleware = typeof (api as any).dispatch === 'function';
+    const usingReduxMiddleware =
+      (api as any).dispatchFromDevtools && typeof (api as any).dispatch === 'function';
     if (!usingReduxMiddleware) {
-      (api as any).dispatchFromDevtools = true;
-      (api as any).dispatch = (payload: {
+      (api as any).__dispatch = (payload: {
         type: string,
         args: unknown[] | Record<string, unknown>
       }) => {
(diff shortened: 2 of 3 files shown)


msanchezdev:
I have to create tests now, but you can give this a look @dai-shi

src/middleware/devtools.ts
```diff
@@ -72,11 +70,12 @@ type StoreDevtools<S> = S extends {
     }
   : never
 
-export interface DevtoolsOptions extends Config {
+export interface DevtoolsOptions extends Omit<Config, 'actionCreators'> {
   name?: string
   enabled?: boolean
   anonymousActionType?: string
   store?: string
+  actionCreators?: boolean | (<A, P extends any[] = any[]>(...args: P) => A) | Record<string, boolean |(<A, P extends any[] = any[]>(...args: P) => A)>;
```
dai-shi: Zustand philosophy is primitive, simple, small.

I think this should be simplified much further.

src/middleware/devtools.ts
```diff
@@ -254,6 +250,18 @@ const devtoolsImpl: DevtoolsImpl =
...
 
     const initialState = fn(api.setState, get, api)
+    if (options.actionCreators === true) {
+      options.actionCreators = getActionsArray(initialState) as any;
+    } else if (options.actionCreators === false) { 
+      options.actionCreators = [] as any;
+    } else if (options.actionCreators && !Array.isArray(options.actionCreators)) {
+      options.actionCreators = getActionsArray(options.actionCreators) as any;
+    }
+    const actionCreators = (options.actionCreators ?? []) as unknown as ActionCreatorDefinition[];
```
dai-shi: Too many `any`s. I think it's also related with the simple api.

src/middleware/devtools.ts
```diff
@@ -254,6 +250,18 @@ const devtoolsImpl: DevtoolsImpl =
     }
 
     const initialState = fn(api.setState, get, api)
+    if (options.actionCreators === true) {
```
dai-shi: I think the default is to allow top-level actions only, but I'm not very sure.

dai-shi (commented):
Thanks for working on it!
The addition of `actionCreators` option seems reasonable.

Not a big fan of the complexity and feels like it could be more primitive.

I have also a mixed feeling about inlining redux-devtools-utils. At least, it's hard to review and I don't think I need to review redux-devtools-utils part.
Could you create a version with using `redux-devtools-utils` as a peer dependency? I think we could dynamic import it, so only users who use `actionCreators` depend on it.

src/middleware/devtools.ts
dai-shi: 🤔 
```ts
actionCreators?: boolean | { [dotNotationProp: string]: boolean } | ActionCreatorDefinition[]
```
or
```ts
actionCreators?: 'top-level-funcs' | 'all-nested-funcs' | ActionCreatorDefinition[]
```

src/middleware/devtools.ts
```diff
@@ -269,6 +277,28 @@ const devtoolsImpl: DevtoolsImpl =
...
+    if (!(api as any).dispatchFromDevtools) {
+      (api as any).dispatchFromDevtools = true;
+      (api as any).dispatch = (payload: { type: string }, ...args: any[]) => {
+        console.log('payload', payload, 'args', args);
+        if (!('type' in payload)) {
+          console.error('[zustand devtools middleware] Payload must have a "type" property which should match an action creator');
+          return;
+        }
+
+        const { type: actionType, ...actionArgs } = payload;
```
dai-shi: I'm confused, isn't payload something like this?
```js
  payload: {
    args: ["123"],
    name: "addTab",
    rest: "[]",
    selected: 0
  },
```

feels like I'm missing some specs.

src/middleware/devtools.ts
msanchezdev: Hmm, I was cleaning up the code and was doing like this:
```ts

type ActionCreator = <A, P extends any[] = any[]>(...args: P) => A;
type ActionCreatorMask = {
  [key: string]: boolean | ActionCreator | ActionCreatorMask;
};

export interface DevtoolsOptions extends Omit<Config, 'actionCreators'> {
  /* ... */
  actionCreators?: boolean | ActionCreatorMask;
}
```

Still too complex? Idea was to allow passing a mask of functions to expose, dot notation could work.

msanchezdev: I lean more towards the first version you mentioned

src/middleware/devtools.ts
msanchezdev: Was a bit unsure about exposing the actions by default, but I think it might be a good call.

src/middleware/devtools.ts
msanchezdev: The devtools itself uses internally the format:
```ts
{
  type: "ACTION",
  payload: {
    args: ["123"],
    name: "addTab",
    rest: "[]",
    selected: 0
  }
}
```

But that's the payload when you select an action from the dropdown,. When you write the payload yourself you are free to give it any shape and I think we dont need to be constrained to the redux way in this case (doesn't feel natural or intuitive).

I suggest for example:
```ts
function increment(amount: number) {}
// would be called with
{
  type: "increment",
  args: {
    amount: 20
  }
}
```

Now that you mention it I am spreading instead of taking the  `args` property

src/middleware/devtools.ts
dai-shi: As we use dot notation internally IIUC, I think the public api can be with dot notation too.

src/middleware/devtools.ts
dai-shi: Ah, I don't mean `true` by default. I mean if it's `true` the default should be only exposing top-level functions.

src/middleware/devtools.ts
dai-shi: Yeah, if it's our control, I prefer `args: [20]`.

msanchezdev: Alright, I think supporting both is okay, I am more a fan of using descriptive names and we already have the name of the arguments.

dai-shi: I don't love supporting both, just one solution should be good.
actions are defined by users. If they want named args, they could create such actions.

msanchezdev: As we chatted it looked like there was a misunderstanding before, in the end we left support for both as it's in the hands of the users to use one or the other, whatever they are comfortable with, We just restrict on the shape being 
```ts
{
  // name of the action creator
  type: "increment",
  // args to pass the function being called
  // can be either object of named args
  args: {
    amount: 25
  },
  // or array of positional args
  // args: [25]
}
```

package.json
```diff
@@ -157,6 +157,7 @@
     "vitest": "^4.0.14"
   },
   "peerDependencies": {
+    "@redux-devtools/utils": "^3.1.1",
```
dai-shi, line 160: Please also add this to peerDependenciesMeta with optional: too.

src/middleware/devtools.ts
```diff
@@ -72,11 +73,44 @@ type StoreDevtools<S> = S extends {
...
+          ?
+            | P
+            | `${Prefix}${K}.*`
+            | `${Prefix}${K}.**`
+          : never
+        : never
+      : never;
+}[keyof T & string];
+
+type ActionCreatorsMask<T extends Record<string, any> = Record<string, any>> = {
```
dai-shi: Do we need the default?
```suggestion
type ActionCreatorsMask<T extends Record<string, any>> = {
```

src/middleware/devtools.ts
```diff
@@ -1,4 +1,5 @@
 import type {} from '@redux-devtools/extension'
+import type { ActionCreatorObject } from '@redux-devtools/utils';
```
dai-shi, line 2: Doesn't this require all devtools users to install the utils lib? I'd like to make it opt-in. Could we use `import()` syntax to avoid it and if it's not installed it's just `any`?

src/middleware/devtools.ts
```diff
@@ -254,6 +285,32 @@ const devtoolsImpl: DevtoolsImpl =
...
 
     const initialState = fn(api.setState, get, api)
+    let actionCreators: ActionCreatorObject[] = [];
+    let evalAction =
+      (action: string, _actionCreators: ActionCreatorObject[]) =>  {
+        if (typeof action !== 'string') {
+          console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use actionCreators in devtools middleware');
+          return;
+        }
+        return new Function('return ' + action)();
```
dai-shi: Doesn't this lead to security vulnerabilities? Could we keep JSON.parse for this case?

src/middleware/devtools.ts
```diff
@@ -429,3 +518,128 @@ const parseJsonThen = <T>(stringified: string, fn: (parsed: T) => void) => {
...
+  const escaped = mask
+    .replace(/\./g, '\\.')
+    .replace(/\*\*|\*/g, '.*');
+
+  return new RegExp(`^${escaped}$`);
+}
+
+function getActionsArray<T extends Record<string, any>>(actionCreators: T, mask?: ActionCreatorsMask<T>): ActionCreatorObject[] {
+  if (Array.isArray(actionCreators)) return actionCreators;
+  if (typeof actionCreators !== 'object') return [];
```
dai-shi: I don't think we need to support these cases.

src/middleware/devtools.ts
```diff
@@ -254,6 +285,32 @@ const devtoolsImpl: DevtoolsImpl =
...
+        evalAction = require('@redux-devtools/utils').evalAction;
+      } catch {
+        console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use actionCreators in devtools middleware');
+      }
+      actionCreators = getActionsArray(
+        initialState as Record<string, unknown>,
+        options.actionCreators as ActionCreatorsMask<{}> ?? {}
+      );
+      // override to pass it to the extension connector, any is ok, we dont care about the type anymore
+      options.actionCreators = actionCreators as any;
```
dai-shi: Should these lines be in the try-catch block above?

src/middleware/devtools.ts
```diff
@@ -429,3 +518,128 @@ const parseJsonThen = <T>(stringified: string, fn: (parsed: T) => void) => {
...
+            actions.set(action.name, 
+              customActions.get(action.name) ?? {
+              name: action.name,
+              func: action.func,
+              args: action.args,
+            });
+          } else {
+            actions.delete(action.name);
+          }
+          continue next_action;
```
dai-shi: I don't love the use of loop with label. Could you take another approach?

src/middleware/devtools.ts
```diff
@@ -429,3 +518,128 @@ const parseJsonThen = <T>(stringified: string, fn: (parsed: T) => void) => {
...
+        });
+      }
+    }
+  }
+
+  return Array.from(actions.values());
+}
+
+// Adapted from https://github.com/fahad19/get-params/blob/master/index.js#L2-L28
+function getParams(func: (...args: any[]) => any): string[] {
```
dai-shi: Leftover?

dai-shi (changes requested):
Thanks for working on this! Left some comments.

Once we settled on the implementation, please add tests.

package.json
msanchezdev: Done

src/middleware/devtools.ts
msanchezdev: Done

src/middleware/devtools.ts
msanchezdev: Can't use JSON.parse as we are not dealing with JSON, we are dealing with Javascript code. This was part of the error initially, we were trying to parse as JSON. As this is the same approach the official devtools use and only meant for development, I think it'd be okay.

src/middleware/devtools.ts
msanchezdev: Yeah, removed

src/middleware/devtools.ts
msanchezdev: Actually could just be a break

src/middleware/devtools.ts
msanchezdev: Looks like I hadn't pushed the commit with the removal

src/middleware/devtools.ts
msanchezdev: It's a type import, it will be any if not installed, still I can remove it and just use `any[]` or inline it

src/middleware/devtools.ts
msanchezdev: Yeah

src/middleware/devtools.ts
dai-shi: Does this work?
```ts
type ActionCreatorObject = import('@redux-devtools/utils').ActionCreatorObject;
```
And, uninstalling '@redux-devtools/utils' should still compile.

src/middleware/devtools.ts
dai-shi: > As this is the same approach the official devtools

Please add a link to the official solution in the code comment.

dai-shi: On second thought, I don't think `new Function('return ' + action)()` is good. It's "eval" and we can't 100% assume that production code doesn't include the `devtools` middleware. I'd suggest to keep using JSON.parse here and let users to input JSON object rather than JS object in "custom action" field in the devtools extension.

src/middleware/devtools.ts
```diff
@@ -254,6 +285,57 @@ const devtoolsImpl: DevtoolsImpl =
...
+      (action: string, _actionCreators: ActionCreatorObject[]) =>  {
+        if (typeof action !== 'string') {
+          console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use actionCreators in devtools middleware');
+          return;
+        }
+        return new Function('return ' + action)();
+      }
+    
+    // Not using the redux plugin
+    if (!(api as any).dispatchFromDevtools && options.actionCreators) {
```
dai-shi: Should it be this?
```suggestion
    if (!(api as any).dispatch && options.actionCreators) {
```

msanchezdev: Done

src/middleware/devtools.ts
msanchezdev: Both are technically the same, though note that ActionCreatorObject is not used on any exported symbol, thus this only serves for typechecking on the library itself, doesn't show up in any built file nor reaches end users.

dai-shi: Please make sure that users can use `devtools` without installing `@redux-devtools/utils`, and typescript code compiles.

src/middleware/devtools.ts
```diff
@@ -254,6 +285,57 @@ const devtoolsImpl: DevtoolsImpl =
...
+          options.actionCreators as ActionCreatorsMask<{}>
+        );
+        // override to pass it to the extension connector, any is ok, we dont care about the type anymore
+        options.actionCreators = actionCreators as any;
+      } catch {
+        console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use actionCreators in devtools middleware');
+      }
+    }
+    
+    let usingReduxMiddleware = typeof (api as any).dispatch === 'function';
```
dai-shi: Can this be const?

dai-shi: I might have misguided you, but I think what we want here is this:
```suggestion
    const usingReduxMiddleware = (api as any).dispatchFromDevtools && typeof (api as any).dispatch === 'function'
```

Sorry about my confusion.

src/middleware/devtools.ts
```diff
@@ -254,6 +285,57 @@ const devtoolsImpl: DevtoolsImpl =
...
+    if (!usingReduxMiddleware) {
+      (api as any).dispatchFromDevtools = true;
```
dai-shi: This is the flag set from the `redux` middleware. It means like "allow dispatching from devtools middleware".

src/middleware/devtools.ts
```diff
@@ -254,6 +285,57 @@ const devtoolsImpl: DevtoolsImpl =
...
+          console.warn('[zustand devtools middleware] Please install @redux-devtools/utils to use the action dropdown in the devtools');
+          return;
+        }
+
+        if (!('type' in payload)) {
+          console.error('[zustand devtools middleware] Payload must have a "type" property which should match an action creator');
+          return;
+        }
+
+        const action = actionCreators.find((action) => action.name === payload.type);
```
dai-shi, line 322: Should we allow using `actionCreators` if the `redux` middleware is used?
I don't think we should modify `dispatch`.

dai-shi (commented):
I don't think we need to change the behavior at all for the `redux` middleware usage (at least in this PR). Or, I may misunderstand something.

src/middleware/devtools.ts
msanchezdev: Yes, as we are trying to be compatible with Redux Devtools not adding something new. Actions creators is what lets users select options from the actions dropdown and dispatch them by filling out the properties in the form.

What this PR does is just adding support for the dropdown itself as previously one could only dispatch "Custom action" with their JS/JSON code.

Expected and current behavior is as follows:
- For custom actions, JS will always error as being malformed and show example of the right pattern depending on the type of store being used.
- If @redux-devtools/utils is not installed it will warn whenever trying to dispatch actions from the devtools. **Except for custom actions on a Redux store which should always be dispatchable.**

```yaml
# Behavior Matrix
With @redux-devtools/utils:
  Normal Store:
    Custom Action:
      JSON: Success
      JS: Errors malformed
    Action Creator: Success
  Redux Store:
    Custom Action:
      JSON: Success
      JS: Errors malformed
    Action Creator: Success

Without @redux-devtools/utils:
  Normal Store:
    Custom Action:
      JSON: Warns required
      JS: Errors malformed
    Action Creator: Warns required
  Redux Store:
    Custom Action:
      JSON: Success
      JS: Errors malformed
    Action Creator: Warns required
```

src/middleware/devtools.ts
msanchezdev: I think we can remove the flag if it doesnt make any difference anymore unless it's set in other cases not just with the redux middleware being present, or it carries any other meaning.

src/middleware/devtools.ts
msanchezdev: I dont see value in dispatchFromDevtools being present anymore as we always support dispatching from devtools now, be it having or not having `redux` middleware

src/middleware/devtools.ts
dai-shi: > What this PR does is just adding support for the dropdown itself as previously one could only dispatch "Custom action" with their JS/JSON code.

Is this also for `redux` middleware? How can we support dropdown for it?

src/middleware/devtools.ts
msanchezdev: As mentioned in the previous comment, as the type definition or the import never get to dist files code compiles. At least in the tests I have run. Only way I see this not compiling would be with a locally linked version of the repo and updating the package.json to import from the .ts file instead of the dist files.

So far, all works as expected.

src/middleware/devtools.ts
msanchezdev: With the `redux` middleware you can:
1. Use the mask to expose functions in the state. (Though, I don't think this will be common in `redux` stores)
2. Do it the redux way and pass your actions creators to the actionsCreators option ([Redux devtools docs dont have good example](https://github.com/zalmoxisus/redux-devtools-extension/commit/477e69d8649dfcdc9bf84dd45605dab7d9775c03#diff-182cb140f8a0fd8bc37bbdcdad07bbadb9aebeb2d1b8ed026acd6132f2c88ce8R5)). It's just passing functions that return redux actions (thus the name of actionCreators).
3. Create custom action creators that update the state themselves.
```ts
const useCounterStore = create()(devtools(redux((action, state) => {
  switch(action.type) {
    case 'increment':
    return { ...state, count: state.count + (state.amount ?? 1) };
    case 'decrement':
    return { ...state, count: state.count - (state.amount ?? 1) };
    default:
    return { ...state };
  }
},{
  count: 0,
  reset: () => {
    useCounterStore.setState(useCounterStore.getInitialState());
  }
}), {
  actionCreators: {
    // Option 1: Use mask
    reset: true,
    // Option 2: Pass your action creators, probably defined elsewere
    increment: (amount = 1) => ({ type: 'increment', amount }),
    decrement: (amount = 1) => ({ type: 'increment', amount }),
    // Option 3: Mutate yourself
    power: (n = 2) => {
      const state = useCounterStore.getState()
      useCounterStore.setState(state.count ** n);
    }
  }
}))
```

src/middleware/devtools.ts
dai-shi: Let me try then.

dai-shi: Ah, you are right. the *.d.ts file doesn't have this import. My oversight.

https://stackblitz.com/edit/znhw2fdb?file=src%2Findex.tsx works fine with pnpm. confirmed.

src/middleware/devtools.ts
dai-shi: I think you misunderstand something about the flag. It's to tell from redux middleware to devtools middleware.
Do not change that convention in this PR.

src/middleware/devtools.ts
dai-shi: We don't need to support dropdown for redux middleware.

msanchezdev: > We don't need to support dropdown for redux middleware.

Are you sure we don't? I think not supporting it for redux middleware will confuse users that expect the devtools to support it.

Remember the drop down was initially a Redux thing, we are just patching to support normal Zustand store as well. It's not like we are adding something specific for Zustand.

I would like to know the reasoning behind wanting to remove support for action creators for Redux middleware. Redux is the one that brought actionCreators in the first place. What if someone is already using actionCreators in their workflow. (Maybe in some hacky way as it was a bit broken in our side)

src/middleware/devtools.ts
msanchezdev: I know it's to tell from redux middleware to devtools middleware. I'm using the flag to detect if redux middleware is being used.

Well, I was using it until I noticed it was no longer needed. I read very carefully through the code and noticed there was no single place where the flag was set to false, which would have changed things.

The only place the flag is used is for checking if the redux middleware is being used and a check in the Redux Devtools receive message subscription. Notice that if the flag stays, the check will have to be removed for this to work properly and the flag would be reduced to only informational data.

EDIT: Well, actually I am making a check to show different examples for malformed custom actions. That'd be the only valid place left.

If you want to keep the flag I would have to update the code with the changes mentioned above, let me know how you would like me to proceed. I am just explaining my reasoning on the changes, it was not a misunderstanding but careful inspection of the code. Unless it was....

dai-shi: > there was no single place where the flag was set to false, which would have changed things.

It's for the case that `.dispatch` is defined by a user, not by the middleware. You can't find it from the code.

src/middleware/devtools.ts
dai-shi: I think our `redux` middleware is pretty different from the original Redux, and I assume `redux` middleware is not widely used, which might be wrong.

> What if someone is already using actionCreators in their workflow.

So, my point is to keep the current behavior as much as possible if actionCreators option isn't used. And, that should be easily reviewed with code (as our test coverage is limited). To fix "a bit broken behavior", it should be a different PR, and maybe a different release. Please don't mix.

> Are you sure we don't?

Can we support `actionCreators: true` with the `redux` middleware? I thought it's just `dispatch(action)` which wouldn't be very useful.
I mean, what would be the Redux way for the dropdown? Shouldn't we list the actions that we pass to dispatch?

msanchezdev: > I think our `redux` middleware is pretty different from the original Redux

In what way do you mean it's different? According to my tests so far they behave the same, at least in what this PR concerns:
- Dispatching JSON for custom actions
- Dispatching actions from action creators dropdown

> I assume `redux` middleware is not widely used, which might be wrong.

I can't really guarantee it either. I'm just trying to keep behavior the same for stores with `redux` middleware. So that someone used to using the Redux Devtools doesn't find anything weird.

> Can we support `actionCreators: true` with the `redux` middleware?  I thought it's just `dispatch(action)` which wouldn't be very useful.

Yes, the example above shows how it works, but now looking more closely at the `redux` middleware it seems unlikely that a `redux` middleware store will have functions in the state apart from `dispatch` function. I guess we can only accept the mask if not using `redux` middleware.

> I mean, what would be the Redux way for the dropdown? Shouldn't we list the actions that we pass to dispatch?

That's exactly how redux handles it, you pass the list of action creator functions, which is the **Option 2** in the example above which is the redux way. **Option 3** would be the extra added by us, users would not need to instead of expanding the reducer for a test case just using users can mutate the state directly using the store's `setState()` function or whatever they may want to do.

> So, my point is to keep the current behavior as much as possible if actionCreators option isn't used.

This summarizes this PR in terms of functionality, the bold text is the change I would be doing. Let me know if this is okay or any other comments.

| with `redux` middleware | Custom Action| Action Creators |
|--------|--------|--------|
| Before | Only if JSON formatted, all actions + `__setState` | Errors with "unsupported format" |
| Now | Same + hint if malformed | Redux style action creators, **mask not supported** |

| without `redux` middleware | Custom Action| Action Creators |
|--------|--------|--------|
| Before | Only if JSON formatted, only `__setState` supported | Errors with "unsupported format" |
| Now | Same + hint if malformed + action creators | Works as expected |

src/middleware/devtools.ts
msanchezdev: > > there was no single place where the flag was set to false, which would have changed things.
> 
> It's for the case that `.dispatch` is defined by a user, not by the middleware. You can't find it from the code.

Yes I thought so, but then, what do you think would be the best approach here? 
- Error if `dispatch` is defined in the state?
- Use a different property like `__dispatch` if not using redux middleware? _Tested this and work without problems, uploaded in last commit: https://github.com/pmndrs/zustand/pull/3333/changes/cb6e9b9a6debbacf3925b755293e3ea94efd7e35_

src/middleware/devtools.ts
dai-shi: Thanks for the detailed response. That helps.

> In what way do you mean it's different?

That's actually a good question. I probably mean, Zustand's `redux` doesn't use action creators.

> That's exactly how redux handles it

You mean the original "redux". Okay, that's my oversight.
I'm still curious how it handles it with and without redux toolkit.

> Let me know if this is okay or any other comments.

My concern is the code level. I would like to see changes gradually.

So, while this is not a final decision, my request is, for "with redux middleware" case, we keep previous behavior even if it feels buggy (we'll tackle separately). I believe the code change will be much simpler, but I'm not 100% sure.

msanchezdev: > Zustand's `redux` doesn't use action creators.

It's actually there in the typing (I think it's when the `@redux/extension` package ins installed or something like that). But it will not work without this PR because of the lack of `evalAction`.

> I'm still curious how it handles it with and without redux toolkit

Not really sure about that, haven't used Redux much, just saw that Redux's composeWithDevtools accepts the actionCreators **object or array**. (We normalized it to just accept the object in this PR for simplicity, notice the Omit<...> in the `DevtoolsOptions` type).

> My concern is the code level. I would like to see changes gradually.

Hmm, code here is basically complete (not sure if you saw my last commit). It now matches the tables above. (Still haven't removed support for action creators mask when using redux middleware, but it may be just a check to exit early after calling their `getActionsArray` in our `getActionsArray`)

msanchezdev: > So, while this is not a final decision, my request is, for "with redux middleware" case, we keep previous behavior even if it feels buggy (we'll tackle separately). I believe the code change will be much simpler, but I'm not 100% sure.

Can you point it out on the table, I think it may help me have the real idea on what you would prefer.

Previous behavior was just to error when action creators are called (This didn't prevent the user from populating the action creators drop down)

dai-shi: | with `redux` middleware | Custom Action| Action Creators |
|--------|--------|--------|
| Before | Only if JSON formatted, all actions + `__setState` | Errors with "unsupported format" |
| Now | Same | Same |

| without `redux` middleware | Custom Action| Action Creators |
|--------|--------|--------|
| Before | Only if JSON formatted, only `__setState` supported | Errors with "unsupported format" |
| Now | Same + hint if malformed + action creators | Works as expected |


Again, I may change my mind after some reviews. Maybe, I should try forking your branch and see how it goes.

src/middleware/devtools.ts
dai-shi: Actually, I'm not sure why we need `.dispatch` or `__dispatch` for the cases without `redux` middleware.

msanchezdev: Now that you mention that, I think I can inline it in the ACTION message instead. Good catch, I was just trying to use the same approach that was already used. Will try it tomorrow.

src/middleware/devtools.ts
msanchezdev: if you can try the changes yourself would be great. I don't think leaving the Json parsing error as before is a good approach for custom actions. I'm okay with reverting the unsupported message for the drop down when not using redux.

src/middleware/devtools.ts
dai-shi: Sounds good. I'll see what I can try after that.

src/middleware/devtools.ts
dai-shi: > when not using redux.

Do you mean when using `redux`?

---

Updated:

| with `redux` middleware | Custom Action| Action Creators |
|--------|--------|--------|
| Before | Only if JSON formatted, all actions + `__setState` | Errors with "unsupported format" |
| Now | Nice error if it's not JSON. If it's JSON, they are actions for `.dispatch` or `__setState` | Error mentioning to use Custom Action in JSON format |

| without `redux` middleware | Custom Action| Action Creators |
|--------|--------|--------|
| Before | Only if JSON formatted, only `__setState` supported | Errors with "unsupported format" |
| Now | Nice error if it's not JSON. If it's JSON, they are actions specified with `actionCreators` or `__setState`, otherwise error gracefully | Drop down just works |

msanchezdev: Great, then we just have to confirm on the messages. the only one missing is action creators with redux middleware. The rest is already implemented

dai-shi: I'll wait for your refactor https://github.com/pmndrs/zustand/pull/3333#discussion_r2660576091, if you are going to do. Then, I'd jump in the code and confirm on the messages. After that, what's remaining is tests.

dai-shi:
@msanchezdev are you still around? last message: https://github.com/pmndrs/zustand/pull/3333#discussion_r2660815212

PR #3336, opened 2025-12-27
fix(persist): prevent race condition during concurrent rehydrate calls

Niyaz-Mazhitov:
## Summary

  Fixes race condition in persist middleware when multiple `rehydrate()` calls happen concurrently.

  ### Problem

  When using async storage, if `rehydrate()` is called multiple times before the first hydration completes, all hydrations run in parallel and results are applied in unpredictable order. This causes:
  - Stale state being applied over newer state
  - `onFinishHydration` callbacks firing multiple times
  - Inconsistent state after hydration

  ### Solution

  Added a `hydrationVersion` counter that:
  - Increments at the start of each `hydrate()` call
  - Is checked after each async step (getItem, migrate, before applying state)
  - Aborts the hydration if a newer one has started

  This ensures only the most recent `rehydrate()` call applies its state.

  ## Test plan

  - [x] Added test: multiple concurrent `rehydrate()` calls (only last one wins)
  - [x] Added test: async migration abort when newer hydration starts
  - [x] Added test: state changes during hydration behavior
  - [x] All existing persist tests pass (42 tests)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ If you would like to contribute by fixing an open issue or developing a new feat
 
 ##### Core
 
-1. Run `npm install` to install dependencies.
+1. Run `pnpm install` to install dependencies.
 2. Create failing tests for your fix or new feature in the [`tests`](./tests/) folder.
 3. Implement your changes.
 4. Run `pnpm run build` to build the library. _(Pro-tip: `pnpm run build-watch` runs the build in watch mode)_
--- a/docs/integrations/third-party-libraries.md
+++ b/docs/integrations/third-party-libraries.md
@@ -16,7 +16,7 @@ This can be done using third-party libraries created by the community.
 
 - [@colorfy-software/zfy](https://colorfy-software.gitbook.io/zfy/) — 🧸 Useful helpers for state management in React with Zustand.
 - [@csark0812/zustand-expo-devtools](https://github.com/csark0812/zustand-expo-devtools) — 🧭 Connect Zustand to Redux DevTools in Expo + React Native using the official Expo DevTools plugin system.
-- [@csark0812/zustand-getters](https://github.com/csark0812/zustand-getters) — 🔄 Make JavaScript object getters reactive in Zustand stores - define derived values with `get propertyName()` and they automatically trigger subscription updates when accessed.
+- [@csark0812/zustand-getters](https://github.com/csark0812/zustand-getters) — 🔄 Make JavaScript object getters reactive in Zustand stores — define derived values with `get propertyName()` and they automatically trigger subscription updates when accessed.
 - [@davstack/store](https://www.npmjs.com/package/@davstack/store) — A zustand store factory that auto generates selectors with get/set/use methods, supports inferred types, and makes global / local state management easy.
 - [@dhmk/zustand-lens](https://github.com/dhmk083/dhmk-zustand-lens) — Lens support for Zustand.
 - [@hpkv/zustand-multiplayer](https://github.com/hpkv-io/zustand-multiplayer/tree/main/packages/zustand-multiplayer) — HPKV multiplayer middleware for building realtime collaborative applications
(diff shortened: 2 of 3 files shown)


Niyaz-Mazhitov:
If needed, I can provide a minimal reproduction sandbox.
Let me know if you'd like additional tests or adjustments, happy to iterate.

src/middleware/persist.ts
```diff
@@ -275,21 +284,31 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
...
-            return [false, deserializedStorageValue.state] as const
+            return [false, deserializedStorageValue.state, false] as const
           }
         }
-        return [false, undefined] as const
+        return [false, undefined, false] as const
       })
       .then((migrationResult) => {
-        const [migrated, migratedState] = migrationResult
+        const [migrated, migratedState, aborted] = migrationResult
```
dai-shi: I think I understand the problem, but there's a little fear of mine that this implementation gets complicated more and more. 😱

Niyaz-Mazhitov: You're right, the tuple with 3 elements adds complexity. I can simplify this by removing the aborted flag from the tuple and just checking currentVersion !== hydrationVersion at the start of each .then():

Would you prefer this simpler approach? Happy to update the PR.

dai-shi: Oh, does that work? It sounds like a better approach.

Niyaz-Mazhitov: Done! Simplified the implementation:

  - Removed aborted flag from tuple
  - Now using early return + !migrationResult check instead
  - Tuple stays as original [migrated, state]

  All 42 persist tests passing.

Niyaz-Mazhitov:
Thanks a lot for the review and for looping in another reviewer!
I’ve updated the implementation to use the simpler version-check approach we discussed.
If you’d like me to adjust anything further or add more tests / comments, I’m happy to do so.

src/middleware/persist.ts
```diff
@@ -301,6 +320,10 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
         }
       })
       .then(() => {
+        // Abort if a newer hydration has started
+        if (currentVersion !== hydrationVersion) {
+          return
+        }
```
dai-shi, line 316: I think this is only required, and above this doesn't need the `currentVersion !== hydrationVersion` check, no?

src/middleware/persist.ts
```diff
@@ -318,7 +341,10 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
         finishHydrationListeners.forEach((cb) => cb(stateFromStorage as S))
       })
       .catch((e: Error) => {
-        postRehydrationCallback?.(undefined, e)
+        // Only call error callback if this hydration wasn't superseded
+        if (currentVersion === hydrationVersion) {
+          postRehydrationCallback?.(undefined, e)
+        }
```
dai-shi, line 337: Let's keep the same pattern:
```ts
        // Abort if a newer hydration has started
        if (currentVersion !== hydrationVersion) {
          return
        }
        postRehydrationCallback?.(undefined, e)
```

dai-shi (commented):
simplify further?

src/middleware/persist.ts
Niyaz-Mazhitov: Do you mean this check (in the third .then()) is also unnecessary?

If so, I'm concerned that between the second and third .then(), a new hydration could start, and we'd call finishHydrationListeners for a stale hydration.

Could you clarify which checks you'd like to keep?

dai-shi: Oh, I missed the third one, which has side effect. Yeah, I think I mean to drop the first two checks.

Niyaz-Mazhitov: Done! Removed the first two checks (after getItem and after async migrate). Kept the ones with side effects.

src/middleware/persist.ts
```diff
@@ -289,6 +294,10 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
         return [false, undefined] as const
       })
       .then((migrationResult) => {
+        // Abort if a newer hydration has started
+        if (!migrationResult || currentVersion !== hydrationVersion) {
```
dai-shi: We no longer need it, I guess.
```suggestion
        if (currentVersion !== hydrationVersion) {
```

Niyaz-Mazhitov: !migrationResult is still needed for the case when version mismatch occurs but no migrate function is provided the first .then() returns undefined in that case.

Or should we add a return after the console.error?

dai-shi: We have a return at L294.

Niyaz-Mazhitov: Ah, you're right! After console.error, it falls through to L294. So migrationResult is always a tuple. I'll remove !migrationResult.

dai-shi: oops, I merged main before you commit.

dai-shi (approved):
LGTM
Now, it's readable. Thanks for your contribution.

I'll merge this PR when I prepare a new release. Maybe next year. Anyone can feel free to give a review until then.

Niyaz-Mazhitov:
Thank you for the thorough review and patience! 🙏 Learned a lot from this process. Happy to contribute more in the future! 🚀

dbritto-dev:
@Niyaz-Mazhitov @dai-shi LGTM

dbritto-dev (approved):

PR #3340, opened 2025-12-30
Update updating-state.md, add With Limu demo

fantasticsoul:
add `With Limu` demo

Fixes #

## Check List


dai-shi:
Thanks for your suggestion. Unfortunately, we don't add library usage that can be just used instead of immer. If you are interested, feel free to open a PR to add the item in https://github.com/pmndrs/zustand/blob/main/docs/integrations/third-party-libraries.md.

dbritto-dev:
@fantasticsoul give a shot to https://www.npmjs.com/package/zustand-mutable

PR #3348, opened 2026-01-08
chore: fix type error for vitest config

grigoriy-reshetniak:
## Summary
Vitest config throws 

>TS9037: Default exports can't be inferred with --isolatedDeclarations.
vitest.config.mts(4, 1): Move the expression in default export to a variable and add a type annotation to it.

- [x] `pnpm run test` passed but with changes I've mentioned here https://github.com/pmndrs/zustand/discussions/3347

### DIFF
--- a/docs/integrations/third-party-libraries.md
+++ b/docs/integrations/third-party-libraries.md
@@ -16,7 +16,7 @@ This can be done using third-party libraries created by the community.
 
 - [@colorfy-software/zfy](https://colorfy-software.gitbook.io/zfy/) — 🧸 Useful helpers for state management in React with Zustand.
 - [@csark0812/zustand-expo-devtools](https://github.com/csark0812/zustand-expo-devtools) — 🧭 Connect Zustand to Redux DevTools in Expo + React Native using the official Expo DevTools plugin system.
-- [@csark0812/zustand-getters](https://github.com/csark0812/zustand-getters) — 🔄 Make JavaScript object getters reactive in Zustand stores - define derived values with `get propertyName()` and they automatically trigger subscription updates when accessed.
+- [@csark0812/zustand-getters](https://github.com/csark0812/zustand-getters) — 🔄 Make JavaScript object getters reactive in Zustand stores — define derived values with `get propertyName()` and they automatically trigger subscription updates when accessed.
 - [@davstack/store](https://www.npmjs.com/package/@davstack/store) — A zustand store factory that auto generates selectors with get/set/use methods, supports inferred types, and makes global / local state management easy.
 - [@dhmk/zustand-lens](https://github.com/dhmk083/dhmk-zustand-lens) — Lens support for Zustand.
 - [@hpkv/zustand-multiplayer](https://github.com/hpkv-io/zustand-multiplayer/tree/main/packages/zustand-multiplayer) — HPKV multiplayer middleware for building realtime collaborative applications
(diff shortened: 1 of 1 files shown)


dai-shi:
> Vitest config throws

How can I reproduce this? I've never seen it.

grigoriy-reshetniak:
This is what static type check in IntelliJ IDEA do for `vitest.config.mts` because it's not excluded in `tsconfig.json`, it looks like this:
image

This doesn't impact functionality of config in any way.
 
Excluding `vitest.config.mts` would solve this too.

dai-shi:
Thanks.

Changing like this reproduced it:
```diff
diff --git a/tsconfig.json b/tsconfig.json
index 8780345..d560466 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -21,6 +21,6 @@
       "zustand/*": ["./src/*.ts"]
     }
   },
-  "include": ["src/**/*", "tests/**/*"],
+  "include": ["src/**/*", "tests/**/*", "vitest.config.mts"],
   "exclude": ["node_modules", "dist"]
 }
```

I'd look for alternative solutions.

grigoriy-reshetniak:
Okay, thank you for checking this.

dai-shi:
How does this help your editor?

```diff
diff --git a/package.json b/package.json
index ead83b2..e21b940 100644
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "fix:lint": "eslint . --fix",
     "test": "pnpm run \"/^test:.*/\"",
     "test:format": "prettier . --list-different",
-    "test:types": "tsc --noEmit",
+    "test:types": "tsc && tsc -p tsconfig.configs.json",
     "test:lint": "eslint .",
     "test:spec": "vitest run",
     "patch-d-ts": "node --input-type=module -e \"import { entries } from './rollup.config.mjs'; import shelljs from 'shelljs'; const { find, sed } = shelljs; find('dist/**/*.d.ts').forEach(f => { entries.forEach(({ find, replacement }) => { sed('-i', new RegExp(' from \\'' + find.source.slice(0, -1) + '\\';$'), ' from \\'' + replacement + '\\';', f); }); sed('-i', / from '(\\.[^']+)\\.ts';$/, ' from \\'\\$1\\';', f); });\"",
diff --git a/tsconfig.configs.json b/tsconfig.configs.json
new file mode 100644
index 0000000..485ae30
--- /dev/null
+++ b/tsconfig.configs.json
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "isolatedDeclarations": false,
+    "declaration": false
+  },
+  "files": ["vitest.config.mts"]
+}
```

grigoriy-reshetniak:
@dai-shi it doesn't. Intellij doesn't care about `test:types` part, it's parsing `tsconfig.json` and passively applying it against all files _mostly_ ignoring `include` and `exclude` options ( I was wrong when I though excluding would help).

My original idea when I observed this just to make config file to follow same rules as other `.ts` files in the project (and it looked like a minor change), but after discovering that even adding to `exclude` doesn't help it seems like Intellij issue rather that zustand one to me.

PS:
If I try to manually configure IDE to use `tsconfig.configs.json` it's either starts to check only vitest.config.mts or, if I expand the include part it's applying `"isolatedDeclarations": false,` to everything. The only way I've made Intellij automatically pick up additional config and resolve it was by adding `reference` to `tsconfig.json`, but that's not acceptable.

dai-shi:
Yeah, there's no way for Intellij to read the special `tsconfig.configs.json` file.


> but after discovering that even adding to `exclude` doesn't help it seems like Intellij issue rather that zustand one to me.

It feels so to me too.

dai-shi:
I'm actually fine to rename vitest.config.mts to vitest.config.mjs, if @sukvvon is happy with it.

sukvvon:
I think keeping `.mts` is better than renaming to `.mjs`. This appears to be an IntelliJ-specific issue with user settings, not a zustand project issue. Since `vitest.config.mts` is not included in `tsconfig.json`'s `include` option, `pnpm run test:types` works correctly.

PR #3349, opened 2026-01-10
test: remove eslint-disable for no-unused-expressions

kamja44:
This PR removes the `/* eslint @typescript-eslint/no-unused-expressions: off */` directive from `tests/middlewareTypes.test.tsx`.
To satisfy the linter, unused expressions in the tests have been wrapped with the `void` operator.

## Changes

- Remove global eslint disable comment in `tests/middlewareTypes.test.tsx`
- Add `void` operator to unused expressions to prevent linting errors

## Contribution Checklist

- [x] Run `pnpm run fix:format` to format the code.
- [x] Run `pnpm run build` to build the library.
- [x] Run `pnpm run test` and ensure that they pass.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ If you would like to contribute by fixing an open issue or developing a new feat
 
 ##### Core
 
-1. Run `npm install` to install dependencies.
+1. Run `pnpm install` to install dependencies.
 2. Create failing tests for your fix or new feature in the [`tests`](./tests/) folder.
 3. Implement your changes.
 4. Run `pnpm run build` to build the library. _(Pro-tip: `pnpm run build-watch` runs the build in watch mode)_


tests/middlewareTypes.test.tsx
```diff
@@ -46,15 +44,15 @@ describe('counter state spec (no middleware)', () => {
...
+      void (useBoundStore((s) => s.count) * 2)
+      void useBoundStore((s) => s.inc)()
+      void (useBoundStore().count * 2)
+      void useBoundStore().inc()
+      void (useBoundStore.getState().count * 2)
+      void useBoundStore.getState().inc()
       return <></>
     }
-    TestComponent
+    void TestComponent
```
dai-shi, line 55: This looks good.

tests/middlewareTypes.test.tsx
```diff
@@ -46,15 +44,15 @@ describe('counter state spec (no middleware)', () => {
...
-      useBoundStore().count * 2
-      useBoundStore().inc()
-      useBoundStore.getState().count * 2
-      useBoundStore.getState().inc()
+      void (useBoundStore((s) => s.count) * 2)
+      void useBoundStore((s) => s.inc)()
+      void (useBoundStore().count * 2)
+      void useBoundStore().inc()
+      void (useBoundStore.getState().count * 2)
+      void useBoundStore.getState().inc()
```
dai-shi: Do you need this change?

kamja44: Sorry for the late response!

Yes, this change is necessary.

Without the `void` operators, ESLint throws 77 errors with the `@typescript-eslint/no-unused-expressions` rule. Since the goal of this PR is to remove the `eslint-disable` comment, we need to fix the actual linting issues.

I tested it by removing the `void` operators:
```bash
$ pnpm run test:lint
✖ 77 problems (77 errors, 0 warnings)
```

The alternative would be to keep the `eslint-disable` comment, but that defeats the purpose of this PR.

dai-shi: I just tried it. This seems to be enough:
```ts
      const TestComponent = () => {
        void (useBoundStore((s) => s.count) * 2)
        useBoundStore((s) => s.inc)()                                         
        void (useBoundStore().count * 2)                                      
        useBoundStore().inc()                                                 
        void (useBoundStore.getState().count * 2)                             
        useBoundStore.getState().inc()                                        
        return <></>                                                          
      }                                                                       
```

kamja44: Thank you for the advice! I've updated the code to use `void` only for value and property access expressions, and removed `void` from function calls. This makes the code cleaner while still passing the linting rules.

**Changes:**
- Kept `void` for: `void (count * 2)`, `void useBoundStore().count`, `void TestComponent`
- Removed `void` from: function calls like `inc()`, `dispatch()`, `setState()`, `subscribe()`

The linting now passes successfully:
```bash
$ pnpm run test:lint
✔ No errors
```

kamja44:
> LGTM Thanks for your contribution!

Thank you for the review and for approving the changes! I appreciate your guidance on making the code cleaner.

PR #3351, opened 2026-01-11
chore: fix vitest type definitions

grigoriy-reshetniak:
Addressing one of the FIXME comments

## Summary
Fixed path in `tsconfig.json`, [`setup.ts`](https://github.com/pmndrs/zustand/blob/main/tests/setup.ts) was using this declarations already.

#### Before
`npx tsc --noEmit --skipLibCheck false` returned 10 errors, 2 of them were related to `jest-dom` types.

#### After
`npx tsc --noEmit --skipLibCheck false` returns 8 errors, all of them related to `@redux-devtools/extension`.

## Check List

It seems to me that errors for `@redux-devtools/extension` cannot be addressed without fixes on their side.

### DIFF
--- a/tests/ssr.test.tsx
+++ b/tests/ssr.test.tsx
@@ -31,86 +31,83 @@ function Counter() {
   return <div>bears: {bears}</div>
 }
 
-describe.skipIf(!React.version.startsWith('18'))(
-  'ssr behavior with react 18',
-  () => {
-    it('should handle different states between server and client correctly', async () => {
-      const { hydrateRoot } =
-        await vi.importActual<typeof import('react-dom/client')>(
-          'react-dom/client',
-        )
+describe('ssr behavior with react 18+', () => {
+  it('should handle different states between server and client correctly', async () => {
+    const { hydrateRoot } =
+      await vi.importActual<typeof import('react-dom/client')>(
+        'react-dom/client',
+      )
 
-      const view = renderToString(
+    const view = renderToString(
+      <React.Suspense fallback={<div>Loading...</div>}>
+        <Counter />
+      </React.Suspense>,
+    )
+
+    const container = document.createElement('div')
+    document.body.appendChild(container)
+    container.innerHTML = view
+
+    expect(container).toHaveTextContent(/bears: 0/)
+
+    await act(async () => {
+      hydrateRoot(
+        container,
         <React.Suspense fallback={<div>Loading...</div>}>
           <Counter />
         </React.Suspense>,
       )
-
-      const container = document.createElement('div')
-      document.body.appendChild(container)
-      container.innerHTML = view
-
-      expect(container).toHaveTextContent(/bears: 0/)
-
-      await act(async () => {
-        hydrateRoot(
-          container,
-          <React.Suspense fallback={<div>Loading...</div>}>
-            <Counter />
-          </React.Suspense>,
-        )
-      })
-
-      expect(screen.getByText('bears: 1')).toBeInTheDocument()
-      document.body.removeChild(container)
     })
-    it('should not have hydration errors', async () => {
-      const useStore = create(() => ({
-        bears: 0,
-      }))
 
-      const { hydrateRoot } =
-        await vi.importActual<typeof import('react-dom/client')>(
-          'react-dom/client',
-        )
+    expect(screen.getByText('bears: 1')).toBeInTheDocument()
+    document.body.removeChild(container)
+  })
+  it('should not have hydration errors', async () => {
+    const useStore = create(() => ({
+      bears: 0,
+    }))
 
-      const Component = () => {
-        const bears = useStore((state) => state.bears)
-        return <div>bears: {bears}</div>
-      }
+    const { hydrateRoot } =
+      await vi.importActual<typeof import('react-dom/client')>(
+        'react-dom/client',
+      )
 
-      const view = renderToString(
+    const Component = () => {
+      const bears = useStore((state) => state.bears)
+      return <div>bears: {bears}</div>
+    }
+
+    const view = renderToString(
+      <React.Suspense fallback={<div>Loading...</div>}>
+        <Component />
+      </React.Suspense>,
+    )
+
+    const container = document.createElement('div')
+    document.body.appendChild(container)
+    container.innerHTML = view
+
+    expect(container).toHaveTextContent(/bears: 0/)
+
+    const consoleMock = vi.spyOn(console, 'error')
+
+    const hydratePromise = act(async () => {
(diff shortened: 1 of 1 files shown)


dai-shi:
@sukvvon can review quickly?

tsconfig.json
```diff
@@ -6,14 +6,14 @@
...
-    "skipLibCheck": true /* FIXME remove this once vite fixes it */,
+    "skipLibCheck": true /* FIXME remove this once redux-devtools/extension fixes it */,
     "allowImportingTsExtensions": true,
     "noUncheckedIndexedAccess": true,
     "exactOptionalPropertyTypes": true,
     "verbatimModuleSyntax": true,
     "declaration": true,
     "isolatedDeclarations": true,
-    "types": ["@testing-library/jest-dom"],
+    "types": ["@testing-library/jest-dom/vitest"],
```
sukvvon, line 16: While `/vitest` does fix jest-related errors with `--skipLibCheck false`, `skipLibCheck: true` is already set, so it doesn't affect actual usage. The https://www.npmjs.com/package/@testing-library/jest-dom#with-vitest also recommend `@testing-library/jest-dom`.

docs/guides/testing.md
```diff
@@ -151,7 +151,7 @@ afterEach(() => {
 
 ```ts
 // setup-jest.ts
-import '@testing-library/jest-dom'
+import '@testing-library/jest-dom/vitest'
```
sukvvon: For Jest, it should be `@testing-library/jest-dom` without `/vitest`.

docs/guides/testing.md
```diff
@@ -261,7 +261,7 @@ afterEach(() => {
 
 ```ts
 // setup-vitest.ts
-import '@testing-library/jest-dom'
+import '@testing-library/jest-dom/vitest'
```
sukvvon, line 264: 👍

tsconfig.json
```diff
@@ -6,14 +6,14 @@
     "esModuleInterop": true,
     "module": "esnext",
     "moduleResolution": "bundler",
-    "skipLibCheck": true /* FIXME remove this once vite fixes it */,
+    "skipLibCheck": true /* FIXME remove this once redux-devtools/extension fixes it */,
```
sukvvon, line 9: 👍

docs/guides/testing.md
grigoriy-reshetniak: My bad, updated.

grigoriy-reshetniak:
Hello @sukvvon.
Thank you for your feedback. I’ve addressed one of the comments, but I’d like to clarify another point.
>skipLibCheck: true is already set, so it doesn't affect actual usage

That’s true, however this change reduces the overall number of type errors and removes `vitest` from the list of reasons for keeping this flag enabled. Also, I've checked link you provided and it's a basic example. I was looking for any argument in `vitest` issues _against_ subpath type augmentation, but haven't found any. Also, `vitest` itself export augmented types, so it doesn't look to me that they discourage their usage.

If you still feel that this change should be dropped, please let me know whether you’d prefer me to adjust it or proceed in a different way.

sukvvon:
@gnapse Could you clarify the intended usage for `tsconfig.json` types with Vitest?

The [official docs](https://www.npmjs.com/package/@testing-library/jest-dom#with-vitest) show `@testing-library/jest-dom` (without `/vitest`) in `tsconfig.json`, while the setup file uses `@testing-library/jest-dom/vitest`.

However, when running `npx tsc --noEmit --skipLibCheck false`:

| `tsconfig.json` types | Errors |
|----------------------|--------|
| `@testing-library/jest-dom` | 10 (8 redux-devtools + 2 jest-dom) |
| `@testing-library/jest-dom/vitest` | 8 (redux-devtools only) |

The 2 jest-dom errors resolved by `/vitest`:
- `Cannot find type definition file for 'jest'`
- `Cannot find name 'expect'`

Is it recommended to use `/vitest` subpath in `tsconfig.json` types as well, or should it remain as the base path?

tsconfig.json
dbritto-dev, line 16: The change should be this

```suggestion
    "types": ["vitest/globals", "@testing-library/jest-dom"],
```

dbritto-dev (changes requested):

tsconfig.json
grigoriy-reshetniak: This way only one of two errors if fixed.

> node_modules/.pnpm/@testing-library+jest-dom@6.9.1/node_modules/@testing-library/jest-dom/types/jest.d.ts:1:23 - error TS2688: Cannot find type definition file for 'jest'.

still occurs. Shall I proceed with this change anyway?

dbritto-dev: We should only need Vitest global and jest-dom jest types

gnapse:
> @gnapse Could you clarify the intended usage for `tsconfig.json` types with Vitest?

@sukvvon not sure why you mentioned me. I have nothing to do with this, do I? Maybe a mistake while typing someone else's name?

tsconfig.json
sukvvon: @dbritto-dev About adding `vitest/globals`, we decided maintaining current explicit approach in #3123

sukvvon:
> Hello @sukvvon. Thank you for your feedback. I’ve addressed one of the comments, but I’d like to clarify another point.
> 
> > skipLibCheck: true is already set, so it doesn't affect actual usage
> 
> That’s true, however this change reduces the overall number of type errors and removes `vitest` from the list of reasons for keeping this flag enabled. Also, I've checked link you provided and it's a basic example. I was looking for any argument in `vitest` issues _against_ subpath type augmentation, but haven't found any. Also, `vitest` itself export augmented types, so it doesn't look to me that they discourage their usage.
> 
> If you still feel that this change should be dropped, please let me know whether you’d prefer me to adjust it or proceed in a different way.

@grigoriy-reshetniak Let's go with the current changes. Our goal is to reduce errors, and this PR achieves that.

sukvvon:
@dai-shi 👌

sukvvon:
> > @gnapse Could you clarify the intended usage for `tsconfig.json` types with Vitest?
> 
> @sukvvon not sure why you mentioned me. I have nothing to do with this, do I? Maybe a mistake while typing someone else's name?

@gnapse I mentioned you because i thought you're a jest-dom maintainer. This is about the `/vitest` subpath in `@testing-library/jest-dom`, so I thought you might have some insight. Thanks anyway.

dai-shi:
Thanks everyone.

grigoriy-reshetniak:
Thank you everyone for taking the time to look into this.

PR #3352, opened 2026-01-11
test(middlewareTypes): replace 'void' with 'expectTypeOf' for type assertions

sukvvon:

N/A

## Summary

Replace `void` operator patterns with explicit type assertions using `expectTypeOf` in `middlewareTypes.test.tsx` for better type verification.

- Replace `void (expression)` with `expectTypeOf(expression).toEqualTypeOf<T>()`
- Replace deprecated `toMatchTypeOf` with `toExtend` for explicit type checking
- Replace loose matchers (`toBeNumber`, `toBeVoid`, `toBeBoolean`) with stricter `toEqualTypeOf<T>()`
- Wrap `setState`, `subscribe`, `persist.hasHydrated` with `expectTypeOf` for type verification
- Remove redundant `expect(useBoundStore).toBeDefined()` assertions

## Check List


tests/middlewareTypes.test.tsx
```diff
@@ -44,16 +44,15 @@ describe('counter state spec (no middleware)', () => {
       inc: () => set({ count: get().count + 1 }, false),
     }))
     const TestComponent = () => {
```
dai-shi, line 46: `expectTypeOf` looks better, but do we need `TestComponent` then?

dai-shi: I think the original purpose of `TestComponent` is to make it look like an actual use case. Now, with `expectTypeOf`, it no longer holds. So, I think it's better to eliminate `TestComponent`.

sukvvon: `TestComponent` is needed. The argument to `expectTypeOf` is evaluated at runtime, so hook calls (`useBoundStore()`, `useBoundStore((s) => s.count)`) must be inside a component. Removing it causes "Invalid hook call" error.

sukvvon: https://github.com/pmndrs/jotai/blob/main/tests/react/types.test.tsx

We already use same pattern in Jotai.

dai-shi: Oh, it's required. never mind then.

PR #3362, opened 2026-01-18
chore: improve typing in devtools middleware

grigoriy-reshetniak:
## Summary
While going after another FIXME comment I've noticed repetitive `api as any` in devtools middleware. Also fixed some `as any` on the way.

## Check List
- [x] `npx tsc --noEmit` to check ts errors

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,6 +1,6 @@
 import eslint from '@eslint/js'
 import vitest from '@vitest/eslint-plugin'
-import { defineConfig } from 'eslint/config'
+import { defineConfig, globalIgnores } from 'eslint/config'
 import importPlugin from 'eslint-plugin-import'
 import jestDom from 'eslint-plugin-jest-dom'
 import react from 'eslint-plugin-react'
@@ -9,9 +9,7 @@ import testingLibrary from 'eslint-plugin-testing-library'
 import tseslint from 'typescript-eslint'
 
 export default defineConfig(
-  {
-    ignores: ['dist/', 'examples/', 'website/', 'coverage/'],
-  },
+  globalIgnores(['dist/', 'examples/', 'website/', 'coverage/']),
   eslint.configs.recommended,
   importPlugin.flatConfigs.recommended,
   tseslint.configs.recommended,


src/middleware/devtools.ts
```diff
@@ -26,6 +26,11 @@ type Message = {
   state?: any
 }
 
+type WithDispatch = {
+  dispatch: (...args: any[]) => any
```
dai-shi: Can we avoid `any` here?

src/middleware/devtools.ts
```diff
@@ -179,6 +184,7 @@ const devtoolsImpl: DevtoolsImpl =
   (fn, devtoolsOptions = {}) =>
   (set, get, api) => {
     const { enabled, anonymousActionType, store, ...options } = devtoolsOptions
+    const apiWithDispatch = api as typeof api & WithDispatch
```
dai-shi: This increases bundle size. We hope only type improvements without affecting the result of bundling.

src/middleware/devtools.ts
grigoriy-reshetniak: Updated

src/middleware/devtools.ts
grigoriy-reshetniak: Removed this in favor of type guard. Now bundle size became smaller than before.

grigoriy-reshetniak:
Hello @dai-shi.
Thank you for your review!
> If we were to do api as typeof api & WithDispatch, should WithDispatch have optional properties?

You right, but I already removed this line for other reason.

src/middleware/devtools.ts
```diff
@@ -26,6 +26,15 @@ type Message = {
   state?: any
 }
 
+type WithDispatch = {
+  dispatch: (...args: unknown[]) => void
+  dispatchFromDevtools: true
+}
+
+const hasDispatch = (api: unknown): api is WithDispatch =>
```
dai-shi: Actually, the name is confusing. This is not to check if it has the function, but if devtools should dispatch.
```suggestion
const shouldDispatchFromUs = (api: unknown): api is WithDispatch =>
```

src/middleware/devtools.ts
```diff
@@ -26,6 +26,15 @@ type Message = {
...
+  typeof (api as WithDispatch).dispatch === 'function' &&
+  (api as WithDispatch).dispatchFromDevtools
```
dai-shi: Let's keep the original order:
```suggestion
  (api as WithDispatch).dispatchFromDevtools &&
  typeof (api as WithDispatch).dispatch === 'function'
```

src/middleware/devtools.ts
```diff
@@ -26,6 +26,15 @@ type Message = {
   state?: any
 }
 
+type WithDispatch = {
+  dispatch: (...args: unknown[]) => void
+  dispatchFromDevtools: true
```
dai-shi:
```suggestion
  dispatchFromDevtools: unknown
```

should be more precise.

src/middleware/devtools.ts
grigoriy-reshetniak: Done, but the name is slightly different.

src/middleware/devtools.ts
grigoriy-reshetniak: Done

src/middleware/devtools.ts
grigoriy-reshetniak: Done

src/middleware/devtools.ts
```diff
@@ -26,6 +26,15 @@ type Message = {
   state?: any
 }
 
+type WithDispatch = {
+  dispatch: (...args: unknown[]) => void
+  dispatchFromDevtools: unknown
+}
+
+const shouldDispatchFromDevtools = (api: unknown): api is WithDispatch =>
+  (api as WithDispatch).dispatchFromDevtools === true &&
```
dai-shi: Let's not change the logic.
```suggestion
  (api as WithDispatch).dispatchFromDevtools &&
```

grigoriy-reshetniak: Sorry, but this is required since dispatchFromDevtools is `unknown` now. The intent of `shouldDispatchFromDevtools` not only check if fields are existing, but that `dispatchFromDevtools` is `true`. If this is not done it won't fully substitute original behavior.

Also, leaving `(api as WithDispatch).dispatchFromDevtools` intact throws a TS error:
> Type unknown is not assignable to type boolean

dai-shi: We didn't check if it's `true`, only if it's truthy. That's the original behavior. Do some workaround for the TS error. You can move it back to `true` from `unknown` if there's no workaround.

grigoriy-reshetniak: Updated. I picked `!!` over `Boolean()` because double negation already has been used in the repository.

dai-shi: Ah, I see. It was because of the type guard.

dai-shi (approved):
LGTM
Thanks for your contribution!

grigoriy-reshetniak:
Thank you for guiding me through this.

PR #3371, opened 2026-01-23
fix(immer): Proper typing for immer middleware in combination with slices 

wheerd:
## Summary

For the example from the docs for slices, but with immer:

```
const createBearSlice: StateCreator<
  BearSlice & FishSlice,
  [['zustand/immer', never]],
  [],
  BearSlice
> = (set) => ({
  bears: 0,
  addBear: () => set((state) => ({ bears: state.bears + 1 })),
  eatFish: () => set((state) => ({ fishes: state.fishes - 1 })),
})
```

Currently, this will create a type error.

## Check List

### DIFF
--- a/docs/guides/tutorial-tic-tac-toe.md
+++ b/docs/guides/tutorial-tic-tac-toe.md
@@ -915,8 +915,8 @@ component to update the game. Pass `xIsNext`, `currentSquares` and `handlePlay` 
 export default function Game() {
   const history = useGameStore((state) => state.history)
   const setHistory = useGameStore((state) => state.setHistory)
-  const currentMove = useGameStore((state) => state.currentMove)
-  const setCurrentMove = useGameStore((state) => state.setCurrentMove)
+  const xIsNext = useGameStore((state) => state.xIsNext)
+  const setXIsNext = useGameStore((state) => state.setXIsNext)
   const currentSquares = history[history.length - 1]
 
   function handlePlay(nextSquares) {
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -73,7 +73,7 @@ export interface PersistOptions<
    * Combining `createJSONStorage` helps creating a persist storage
    * with JSON.parse and JSON.stringify.
    *
-   * @default createJSONStorage(() => localStorage)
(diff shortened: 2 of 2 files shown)


wheerd:
I'm not sure where and how to add the tests: This is an adapted version of the example form the docs for slices that uses immer for the slices. Without this change, this gives a type error because T != U. At runtime it works, because the full store is avaialable to the slices anyway.

```typescript
import { type StateCreator, create } from 'zustand'
import { immer } from 'zustand/middleware/immer'

interface BearSlice {
  bears: number
  addBear: () => void
  eatFish: () => void
}

interface FishSlice {
  fishes: number
  addFish: () => void
}

const createBearSlice: StateCreator<BearSlice & FishSlice, [['zustand/immer', never]], [], BearSlice> = set => ({
  bears: 0,
  addBear: () => set(state => ({ bears: state.bears + 1 })),
  eatFish: () => set(state => ({ fishes: state.fishes - 1 }))
})

const createFishSlice: StateCreator<BearSlice & FishSlice, [['zustand/immer', never]], [], FishSlice> = set => ({
  fishes: 0,
  addFish: () => set(state => ({ fishes: state.fishes + 1 }))
})

const useBoundStore = create<BearSlice & FishSlice>()((...a) => ({
  ...immer(createBearSlice)(...a),
  ...immer(createFishSlice)(...a)
}))
```

dai-shi:
Sorry, I meant to add something to https://github.com/pmndrs/zustand/blob/main/tests/middlewareTypes.test.tsx.

> Without this change, this gives a type error

If that's the case, I'm pretty sure that we can add the test.

wheerd:
@dai-shi I added a test for this, can you please review it?

dai-shi:
Thanks! I'll check it later.
Would anyone else like to review?

honuuk:
@wheerd @dai-shi 
I tried this on main, and the test run stops due to a TS compile error.
On this PR branch, it runs as expected.

One small suggestion: the test name could be a bit more specific

```ts
describe('single middleware with sliced store', () => {
  it('immer with slices where slice type differs from store (#3371)', () => {
```

PR #3385, opened 2026-02-12
Update build workflow reference to use latest version

dbritto-dev:


dai-shi:
Thanks for the catch.

PR #3391, opened 2026-02-22
fix(persist): use latest state in post-rehydration callback

Shohjahon-n:
                                                                                                                                                
  Fixes #<issue-number>                                                                                                                         
                                                                                                                                                
  ## Summary                                                                                                                                    
                                                                                                                                                
  - Use current store state (`get()`) for the post-rehydration callback in async hydration flow.                                                
  - Prevent stale callback payload when state changes again during hydration.                                                                   
  - Add an async regression test that mutates state during hydration and verifies callback receives the latest merged state.                    
                                                                                                                                                
  ## Check List
                                                                                                                                                
### DIFF
--- a/docs/learn/guides/beginner-typescript.md
+++ b/docs/learn/guides/beginner-typescript.md
@@ -145,7 +145,7 @@ Sometimes you need more than one property. Returning an object from the selector
 However, directly destructuring properties from that object can cause unnecessary re-renders.
 To avoid this, it’s recommended to wrap the selector with `useShallow`, which prevents re-renders when the selected values remain shallowly equal.
 This is more efficient than subscribing to the whole store. TypeScript ensures you can’t accidentally misspell `bears` or `food`.
-See the [API documentation](https://zustand.docs.pmnd.rs/hooks/use-shallow) for more details on `useShallow`.
+See the [API documentation](../../reference/hooks/use-shallow.md) for more details on `useShallow`.
 
 ```tsx
 import { create } from 'zustand'
@@ -209,7 +209,7 @@ function TotalFood() {
 This middleware separates initial state and actions, making the code cleaner.
 TS automatically infers types from the state and actions, no interface needed.
 This is different from JS, where type safety is missing. It’s a very popular style in TypeScript projects.
-See the [API documentation](https://zustand.docs.pmnd.rs/middlewares/combine) for more details.
+See the [API documentation](../../reference/middlewares/combine.md) for more details.
 
 ```ts
 import { create } from 'zustand'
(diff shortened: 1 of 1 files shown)


src/middleware/persist.ts
```diff
@@ -314,12 +314,7 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
         if (currentVersion !== hydrationVersion) {
           return
         }
-        // TODO: In the asynchronous case, it's possible that the state has changed
-        // since it was set in the prior callback. As such, it would be better to
-        // pass 'get()' to the 'postRehydrationCallback' to ensure the most up-to-date
-        // state is used. However, this could be a breaking change, so this isn't being
-        // done now.
```
dai-shi, line 321: So, do you think this is a breaking change, or not?

dai-shi (commented):
Thanks for working on it. Let's take time to discuss when we should land this.

src/middleware/persist.ts
Shohjahon-n: I don't think it's a breaking change — the callback signature is the same, just state now reflects the actual store state instead of a stale snapshot. The TODO itself acknowledged this was the intended direction. It only affects the async path, sync flow is untouched.

dai-shi: > the callback signature is the same

We consider a behavioral breaking change seriously. async usage might be smaller than sync one, but it's not too small.

dai-shi: It's from https://github.com/pmndrs/zustand/pull/1689. Tagging @coffeebeats in case they are around...

dai-shi: On second thought, this can be considered a "fix", right?
So, I wonder why the comment was added back then. There might be some breaking use cases in their mind.

src/middleware/persist.ts
```diff
@@ -314,12 +314,7 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
...
-        postRehydrationCallback?.(stateFromStorage, undefined)
+        postRehydrationCallback?.(get() ?? stateFromStorage, undefined)
```
dai-shi: Do we need `?? stateFromStorage`?

src/middleware/persist.ts
coffeebeats: Hey. It's been a while since I looked at this, but if I recall correctly, I was unsure whether changing the state that gets passed to `postRehydrationCallback` qualified as a breaking change. I think I was concerned that consumers might somehow be depending on the stale values being passed in. More than that, however, I likely wanted to avoid unnecessary changes in https://github.com/pmndrs/zustand/pull/1692.

Regardless, I agree that based on the intent of the code, the fix proposed here is the way to go.

dai-shi: @coffeebeats Thanks for your reply! Glad to hear your comment. Okay, let's consider this as a fix.

src/middleware/persist.ts
dai-shi: @Shohjahon-n ?

dai-shi: Let's see if tests pass.

```suggestion
        postRehydrationCallback?.(get(), undefined)
```

PR #3405, opened 2026-02-27
fix(readme) : comparison documentaion link

chaesunbak:

Fixes #3404 

Link to "Difference between zustand and other state management libraries for React" documentation led to a 404 page.

## Check List


dai-shi:
Does it fix #3404 right?

chaesunbak:
> Does it fix #3404 right?

YES!! I missed that. I'll add it to the PR right away.

PR #3414, opened 2026-03-04
fix(devtools): correct redux devtools config type extension

grigoriy-reshetniak:

Fixes https://github.com/pmndrs/zustand/discussions/3413

## Summary
Regression was introduced because type was changed from `any` to `object` and (quite self-descriptive)
>Object literal may only specify known properties

which became a Type error when trying to pass `serialize` option to `devtools` middleware.

## Check List


pedroapfilho:
Is this the reason why the zustand store is not showing up on the `redux-devtools`?

dbritto-dev:
@pedroapfilho these are ts changes so shouldn't affect how `redux-devtools` work, file an issue is something is not working on your side

RaiMX:
`stateSanitizer` is gone. Sanitization doesn't work

PR #3431, opened 2026-03-18
perf: avoid redundant Map creation in compareEntries for Map inputs

Acadbek:
## Summary

`compareEntries` was creating unnecessary `new Map()` instances even when both
inputs were already `Map` objects. This adds a fast path that iterates directly
over the existing Maps without allocating new ones.

**Before:**
```ts
const mapA = valueA instanceof Map ? valueA : new Map(valueA.entries())
const mapB = valueB instanceof Map ? valueB : new Map(valueB.entries())
```

**After:**
```ts
if (valueA instanceof Map && valueB instanceof Map) {
  // iterate directly, no allocation
}
// non-Map path unchanged
```

## Check List

- [✅ ] `pnpm run fix` for formatting and linting code and docs

### DIFF
--- a/src/vanilla/shallow.ts
+++ b/src/vanilla/shallow.ts
@@ -13,8 +13,20 @@ const compareEntries = (
   valueA: { entries(): Iterable<[unknown, unknown]> },
   valueB: { entries(): Iterable<[unknown, unknown]> },
 ) => {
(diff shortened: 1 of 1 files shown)


dai-shi:
I don't think it creates unnecessary new Map() instances.

PR #3441, opened 2026-03-29
fix(persist): handle storage write errors gracefully

mahmoodhamdi:
## Summary

When `storage.setItem` throws synchronously (e.g., `QuotaExceededError` when localStorage quota is exceeded, or `SecurityError` in private browsing mode), the error propagates from `setState` even though the in-memory state was already successfully updated via the underlying `set` call.

This causes the application to crash on state updates when persistent storage is unavailable or full, despite the store state being correctly updated in memory.

## Changes

- Wrapped `setItem()` calls in `api.setState` and the initializer's `set` wrapper with try-catch
- Logs a warning on storage write failure, consistent with the existing pattern used when storage is initially unavailable (line 211-214)
- The `setItem()` call during hydration is intentionally left unwrapped, as errors there are already handled by the existing `.catch()` block and reported via `onRehydrateStorage`

## Tests

- Added test for `QuotaExceededError` handling when calling `api.setState`
- Added test for `QuotaExceededError` handling when calling `set` from the store initializer
- Verified in-memory state is correctly updated despite storage failure
- All 198 tests pass, TypeScript check clean, lint clean

### DIFF
--- a/src/middleware/persist.ts
+++ b/src/middleware/persist.ts
@@ -230,13 +230,27 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
 
   api.setState = (state, replace) => {
     savedSetState(state, replace as any)
-    return setItem()
+    try {
+      return setItem()
+    } catch (e) {
+      console.warn(
+        `[zustand persist middleware] Unable to update item '${options.name}', the given storage threw an error.`,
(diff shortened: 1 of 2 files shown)


dbritto-dev:
@mahmoodhamdi this should be a custom storage

PR #3443, opened 2026-03-29
refactor(devtools): remove duplicate module augmentation

mahmoodhamdi:
## Summary

The `StoreMutators` module augmentation for `'zustand/devtools'` is declared twice in `src/middleware/devtools.ts` — once at line 15 and an identical copy at line 101. Both declare the exact same interface member. Removed the duplicate.

## Changes

- Removed the second `declare module '../vanilla'` block (was identical to the first)
- No behavioral change — TypeScript merges duplicate `declare module` blocks, so this is purely a cleanup

## Verification

- All 196 tests pass
- TypeScript check clean
- ESLint clean


dbritto-dev:
thanks for your contribution

PR #3463, opened 2026-04-12
Add note about function naming for storee

JanadaJayathilaka:
Clarify naming convention for store function.

Fixes #

## Check List


dbritto-dev:
@JanadaJayathilaka that's a React convention not a Zustand convention so at this point you need to know how hooks work

PR #3469, opened 2026-04-16
fix(devtools): support Firefox/Safari stack format in findCallerName

SBolsec:

no existing issue — built on top of #2987

## Summary

PR #2987 added `findCallerName` to infer action types from the call stack, but the regex `/.+ (.+) .+/` only matches V8-style frames (`at funcName (source:line:col)`). 
SpiderMonkey (Firefox) and JavaScriptCore (Safari) use a different format — `funcName@source:line:col` — with no spaces, so the regex silently falls back to "anonymous".
See [MDN: Error.stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack#description) for the format differences across engines.

Changes:
- Add a second regex for the Gecko/JSC stack format (`/^([^@]+)@/`)
- Extract both regexes to module-level constants so they are compiled once, not on every `setState` call
- Export `findCallerName` and add a unit test covering the Firefox stack format

## Check List

### DIFF
--- a/src/middleware/devtools.ts
+++ b/src/middleware/devtools.ts
@@ -166,7 +166,14 @@ const removeStoreFromTrackedConnections = (
   }
 }
 
-const findCallerName = (stack: string | undefined) => {
+// V8 (Chrome/Edge/Node): "at <name> (<source>)"
+const v8StackLineRe = /.+ (.+) .+/
+// SpiderMonkey (Firefox) / JavaScriptCore (Safari): "<name>@<source>"
+const geckoStackLineRe = /^([^@]+)@/
+
+export const findCallerName = (
+  stack: string | undefined,
+): string | undefined => {
   if (!stack) return undefined
   const traceLines = stack.split('\n')
   const apiSetStateLineIndex = traceLines.findIndex((traceLine) =>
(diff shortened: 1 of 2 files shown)


src/middleware/devtools.ts
```diff
@@ -166,15 +166,26 @@ const removeStoreFromTrackedConnections = (
...
   const apiSetStateLineIndex = traceLines.findIndex((traceLine) =>
     traceLine.includes('api.setState'),
   )
   if (apiSetStateLineIndex < 0) return undefined
   const callerLine = traceLines[apiSetStateLineIndex + 1]?.trim() || ''
-  return /.+ (.+) .+/.exec(callerLine)?.[1]
+  return (
+    v8StackLineRe.exec(callerLine)?.[1] ||
+    geckoStackLineRe.exec(callerLine)?.[1] ||
+    undefined
```
dai-shi: Can we remove `|| undefined`?

dai-shi (commented):
The code looks good to me otherwise.

src/middleware/devtools.ts
SBolsec: Done, I also dropped the export to keep the `findCallerName` function internal and updated the test

dai-shi: Nice!

dai-shi (approved):
The code looks good to me.

dbritto-dev (approved):

PR #3442, opened 2026-03-29
test: expand React subscribe test coverage

mahmoodhamdi:
## Summary

The React-level `tests/subscribe.test.tsx` only had a single test that checked `typeof subscribe === 'function'`. Meanwhile, the vanilla counterpart (`tests/vanilla/subscribe.test.tsx`) has 9 comprehensive tests.

This expands the React subscribe tests to cover actual subscription behavior through the `create()` API.

## Tests Added

- State identity check (same reference → no notification)
- State change notification with previous/current state
- `subscribeWithSelector`: slice-based subscriptions
- `subscribeWithSelector`: custom equality functions
- `subscribeWithSelector`: `fireImmediately` option
- Unsubscription
- Multiple subscribers
- Selective unsubscription (one listener removed, others remain)

Goes from **1 test → 12 tests** for React subscribe behavior.

## Verification

- All 207 tests pass
- TypeScript check clean
- ESLint clean
- Prettier clean

PR #3471, opened 2026-04-20
test(middleware/immer): add runtime tests for immer middleware

KimHyeongRae0:

N/A — coverage-only test additions. No source changes.

## Summary

The immer middleware (`src/middleware/immer.ts`) has no dedicated runtime test file. `tests/middlewareTypes.test.tsx` only exercises it at the type level, leaving the `store.setState` override (lines 77–83) entirely uncovered at runtime. Other middleware (`devtools`, `persist`, `subscribeWithSelector`) already have dedicated runtime test files.

This mirrors the pattern accepted in #3442 (vanilla had subscribe coverage, React did not → add the missing runtime tests).

### Tests added (`tests/immer.test.tsx`)

- `should mutate state via an immer draft (function updater)` — covers the `produce(updater)` path
- `should merge state when given a non-function updater` — covers the `: updater` path
- `should replace the entire state when replace flag is true` — covers `replace` flag propagation
- `should produce nested state updates via a draft without mutating the previous state` — verifies immer's structural sharing
- `should work with vanilla createStore` — covers the vanilla surface

### Coverage delta for `src/middleware/immer.ts`

|              | Before | After |
| ------------ | -----: | ----: |
| Statements   | 71.42% |  100% |
| Branches     |     0% |  100% |
| Functions    | 66.66% |  100% |
| Lines        | 66.66% |  100% |

Suite total: 207 → 212 tests, all passing.

## Check List

### DIFF
--- a/tests/immer.test.tsx
+++ b/tests/immer.test.tsx
@@ -1,0 +1,74 @@
+import { describe, expect, it } from 'vitest'
+import { create } from 'zustand'
+import { immer } from 'zustand/middleware/immer'
+import { createStore } from 'zustand/vanilla'
+
+describe('immer middleware', () => {
+  it('should mutate state via an immer draft (function updater)', () => {
+    type State = { count: number; inc: () => void }
+    const useBoundStore = create<State>()(
+      immer((set) => ({
+        count: 0,
+        inc: () =>
+          set((state) => {
+            state.count += 1
+          }),
+      })),
+    )
+
+    useBoundStore.getState().inc()
+    expect(useBoundStore.getState().count).toBe(1)
+  })
+
+  it('should merge state when given a non-function updater', () => {
+    type State = { count: number; name: string }
+    const useBoundStore = create<State>()(
+      immer(() => ({ count: 0, name: 'zustand' })),
+    )
+
+    useBoundStore.setState({ count: 5 })
+    expect(useBoundStore.getState()).toEqual({ count: 5, name: 'zustand' })
+  })
+
+  it('should replace the entire state when replace flag is true', () => {
+    type State = { a: number; b?: number }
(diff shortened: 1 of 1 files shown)


dai-shi:
@sukvvon Would you like to give a review?

tests/immer.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+            state.count += 1
+          }),
+      })),
+    )
+
+    useBoundStore.getState().inc()
+    expect(useBoundStore.getState().count).toBe(1)
+  })
+
+  it('should merge state when given a non-function updater', () => {
```
sukvvon:
```suggestion
  it('should not apply produce when updater is not a function', () => {
```

The name "merge" feels slightly misleading here — merging is actually vanilla `setState`'s behavior, while what this test truly covers is the right-hand branch at `immer.ts:79`, where a non-function updater bypasses `produce`. Something like this would make the intent clearer.

KimHyeongRae0: Thanks for the catch — you're right that "merge" conflates this with vanilla `setState` semantics, when the test is really about the non-function branch at `immer.ts:79` bypassing `produce`. Renamed to `should not apply produce when updater is not a function` in 62a421f.

sukvvon (approved):
@dai-shi Checked out and verified locally — `immer.ts` coverage goes 66% → 100% and tests 207 → 212 exactly as described. The `should produce nested state updates via a draft without mutating the previous state` case nicely covers immer's structural sharing. Looks good to merge.

dai-shi (approved):
Thanks @KimHyeongRae0 @sukvvon.

KimHyeongRae0:
Thanks @dai-shi for the quick merge, and @sukvvon for taking the time to check it out locally — really appreciate the thoughtful rename suggestion and the coverage verification.

PR #3472, opened 2026-04-20
fix(shallow): preserve URLSearchParams duplicates

Bojun-Vvibe:
**Repo:** pmndrs/zustand (⭐ 51000)
**Type:** bugfix
**Files changed:** 2
**Lines:** +36/-0

## What
This change fixes `shallow` so `URLSearchParams` values with duplicate keys are compared without collapsing repeated entries. It adds a dedicated `URLSearchParams` comparison path and covers it with regression tests for both reordered equivalent duplicates and mismatched duplicate-key multiplicity.

## Why
The previous implementation treated `URLSearchParams` like a `Map`, which drops repeated keys and could incorrectly report `new URLSearchParams('a=1&a=2')` as shallow-equal to `new URLSearchParams('a=2')`. That is a real false positive for consumers using query params as selector inputs or derived state.

## Testing
Dependencies are not installed in this clone, so the Vitest suite was not run here. I verified the core failure mode and the normalization logic with lightweight `node` checks, and the added regression test documents the expected runtime behavior for the repo test suite.

## Risk
Low / The change is narrowly scoped to `URLSearchParams` handling and leaves existing array, map, set, and object comparison paths unchanged.

### DIFF
--- a/src/vanilla/shallow.ts
+++ b/src/vanilla/shallow.ts
@@ -26,6 +26,30 @@ const compareEntries = (
   return true
 }
 
+const compareURLSearchParams = (
+  valueA: URLSearchParams,
+  valueB: URLSearchParams,
+) => {
+  const sortEntries = ([keyA, valueA]: [string, string], [keyB, valueB]: [
+    string,
+    string,
+  ]) => (keyA === keyB ? valueA.localeCompare(valueB) : keyA.localeCompare(keyB))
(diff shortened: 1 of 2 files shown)


dai-shi:
While the motivation is understandable, I don't think it's worth the complexity. Implementation-wise, it's two-level comparison. So, let me define it as an implementation limitation. If users need such a capability, they should use their own comparison function.

PR #3475, opened 2026-04-22
docs: explain persist + SSR hydration trap in the SSR guide

MukundaKatta:

None — this fills a documentation gap I hit while wiring `persist` in a Next.js app router project.

## Summary

The SSR and Hydration guide lists generic hydration pitfalls (`typeof window`, `matchMedia`, extra whitespace, etc.) but doesn't cover the most common zustand-specific one: `persist` auto-rehydrating from storage during the first client render, so components that read a persisted store disagree with the server's initial-state render and surface as minified React #418/#423/#425.

I added a dedicated section to `docs/learn/guides/ssr-and-hydration.md` covering:

- **Symptom** — what the production error looks like (minified #425 "Text content did not match"), with links to [react.dev/errors/418](https://react.dev/errors/418).
- **Fix** — the documented `skipHydration: true` + `useEffect`-driven `persist.rehydrate()` pattern, with a copy-pasteable `StoreRehydrator` component for app router root layouts.
- **Multi-store example** — rehydrating several persisted stores from the same effect so initial defaults line up in a single pass.
- **When not to use it** — client-only consumers and existing "mounted" guards are called out so readers don't reach for `skipHydration` unnecessarily.

`skipHydration` is already documented in `docs/reference/middlewares/persist.md`, but only as an API flag — this adds the why/when in the learn guide where SSR users actually land via search.

Docs-only change; no runtime or test changes. No overlap with [discussion #2740](https://github.com/pmndrs/zustand/discussions/2740) (which is scoped to per-request stores in the Next.js guide).

## Check List

- [x] `pnpm run test:format` passes (prettier clean on changed file and repo-wide)
- [x] Conventional commit type `docs:`
- [x] PR stays focused in scope (single file, single section added)

### DIFF
--- a/docs/learn/guides/ssr-and-hydration.md
+++ b/docs/learn/guides/ssr-and-hydration.md
@@ -196,3 +196,92 @@ node server.js
 > React recovers from some hydration errors, but you must fix them like other bugs. In the best case, they’ll lead to a slowdown; in the worst case, event handlers can get attached to the wrong elements.
 
 You can read more about the caveats and pitfalls here: [hydrateRoot](https://react.dev/reference/react-dom/client/hydrateRoot)
+
+## Using `persist` middleware under SSR
+
+A common source of hydration errors when using Zustand with an SSR framework
+(Next.js, Remix, etc.) is the [`persist`](../../reference/middlewares/persist.md)
+middleware. By default `persist` rehydrates from storage **during the first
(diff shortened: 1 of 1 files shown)


dbritto-dev:
@MukundaKatta this docs is not related to persist and shouldn't tight to persist if you want to add more info about SSR and persist middleware that should be updated in the persist api docs not here

PR #3477, opened 2026-04-27
docs(persist): fix signature to require persistOptions

pavan-sh:

Fixes #
-
## Summary
The persist docs showed persistOptions as optional, which is misleading. In practice, persist needs a config object (including a required name), so calling it with only the state creator is not valid. 

This PR updates the TypeScript signature and reference heading in the persist docs to clearly show the correct usage: persist(stateCreatorFn, persistOptions).

Ref: 
https://github.com/pmndrs/zustand/blob/main/src/middleware/persist.ts#L377-L385
https://github.com/pmndrs/zustand/blob/main/src/middleware/persist.ts#L59-L73

## Check List
(docs only change)

### DIFF
--- a/docs/reference/middlewares/persist.md
+++ b/docs/reference/middlewares/persist.md
@@ -31,7 +31,7 @@ const nextStateCreatorFn = persist(stateCreatorFn, persistOptions)
 ### Signature
 
 ```ts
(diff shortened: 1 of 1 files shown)


dbritto-dev (approved):

dbritto-dev:
Thanks for your contribution

pavan-sh:
> Thanks for your contribution

Happy to help!

PR #3481, opened 2026-04-29
perf(shallow): avoid Map allocation for plain object comparison

vanyaxk:

Fixes #

## Summary

This PR replaces compareEntries (Object.entries + new Map) with Object.keys +  direct property access for performance and less GC pressure. Since `shallow` is called on every state change per subscriber - for plain objects it was routing through `compareEntries`, which grows the heap

## Check List


dai-shi:
Feel free to define it on your end, or if you want to, publish a package.

PR #3487, opened 2026-05-07
docs(react): fix useStore signature in reference docs

TheSeydiCharyyev:
The signature in `docs/reference/hooks/use-store.md` doesn't match the actual `useStore` from `src/react.ts`. Three issues on one line:

1. `<StoreApi<T>, U = T>` — `StoreApi<T>` is a type instantiation, not a valid generic parameter name.
2. `) => UseBoundStore<...>` — the arrow `=>` should be `:` for a return type annotation.
3. `UseBoundStore<StoreApi<T>>` is what `create()` returns, not `useStore`. The actual return type is `T` (full state) or `U` (selected slice when a selector is passed).

This PR aligns the signature with the same pattern already used in the sister doc [`use-store-with-equality-fn.md`](https://github.com/pmndrs/zustand/blob/main/docs/reference/hooks/use-store-with-equality-fn.md):

```ts
useStoreWithEqualityFn<T, U = T>(store: StoreApi<T>, selectorFn: (state: T) => U, equalityFn?: (a: T, b: T) => boolean): U
```

### Diff

```diff
- useStore<StoreApi<T>, U = T>(store: StoreApi<T>, selectorFn?: (state: T) => U) => UseBoundStore<StoreApi<T>>
+ useStore<T, U = T>(store: StoreApi<T>, selectorFn?: (state: T) => U): U
```

Docs-only change — no changeset needed.

PR #3489, opened 2026-05-08
docs(react): fix equalityFn type in useStoreWithEqualityFn signature

TheSeydiCharyyev:
Following [#3487](https://github.com/pmndrs/zustand/pull/3487) which fixed the `useStore` signature, the sister doc `use-store-with-equality-fn.md` (which I cited as reference for the correct pattern in #3487) has its own related signature bug.

The current signature types `equalityFn` against `T` (the full state), but the actual implementation in `src/react.ts` types it against `U` (the selected slice).

```ts
// src/react.ts
export function useStoreWithEqualityFn<S extends ReadonlyStoreApi<unknown>, U>(
  api: S,
  selector: (state: ExtractState<S>) => U,
  equalityFn?: (a: U, b: U) => boolean,    // compares the selected slice U
): U
```

The equality function compares the result of the selector (the slice), not the full store state. Using `(a: T, b: T)` is misleading for users who pass a selector returning a primitive or a different shape — the equality check operates on that returned value.

### Diff

```diff
- useStoreWithEqualityFn<T, U = T>(store: StoreApi<T>, selectorFn: (state: T) => U, equalityFn?: (a: T, b: T) => boolean): U
+ useStoreWithEqualityFn<T, U = T>(store: StoreApi<T>, selectorFn: (state: T) => U, equalityFn?: (a: U, b: U) => boolean): U
```

Docs-only change — no changeset needed.

### DIFF
--- a/docs/reference/hooks/use-store-with-equality-fn.md
+++ b/docs/reference/hooks/use-store-with-equality-fn.md
@@ -32,7 +32,7 @@ const someState = useStoreWithEqualityFn(store, selectorFn, equalityFn)
 ### Signature
 
 ```ts
-useStoreWithEqualityFn<T, U = T>(store: StoreApi<T>, selectorFn: (state: T) => U, equalityFn?: (a: T, b: T) => boolean): U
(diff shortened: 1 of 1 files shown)


dbritto-dev (approved):

dbritto-dev:
@TheSeydiCharyyev thanks for your contribution

PR #3499, opened 2026-05-19
feat(persist): add equalityFn for redundant write dedupe

Aniketiitk21:

Related to #1135.

## Summary

- add an optional `persist.equalityFn` hook so apps can skip storage writes when the partialized persisted state has not changed
- dedupe in-flight async writes for equal partialized values, while still preserving later changed writes
- document the new option and add sync/async regression coverage for unchanged partialized state and hydration interactions

This keeps the current default behavior unchanged, but gives applications using expensive async storage backends a first-class way to avoid redundant writes.

## Check List

## Validation

- `pnpm run fix`
- `pnpm test:format`
- `pnpm test:types`
- `pnpm test:lint`
- `pnpm build`
- direct runtime verification against the built output for sync dedupe, hydration merge correctness, and async duplicate in-flight writes

## Notes

- I could not run `pnpm run test` / `vitest` successfully in this Windows environment because the repo currently hits a `jsdom` / `html-encoding-sniffer` worker startup `ERR_REQUIRE_ESM` issue before the test files execute.
- Because of that environment-level blocker, I added the direct built-output runtime verification above to make sure the new persist behavior itself was still exercised end to end in-session.

### DIFF
--- a/docs/reference/integrations/persisting-store-data.md
+++ b/docs/reference/integrations/persisting-store-data.md
@@ -146,6 +146,34 @@ export const useBoundStore = create(
 )
 ```
 
+### `equalityFn`
+
+> Type: `(previousState: Object, nextState: Object) => boolean`
+
+Use `equalityFn` to skip redundant storage writes when the partialized state hasn't changed.
+
+This is especially useful when your store mixes fast-changing transient state with a smaller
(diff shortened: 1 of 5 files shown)


dai-shi:
Please use custom `storage` for this kind of customization.

PR #3507, opened 2026-05-23
docs(use-shallow): add troubleshooting entry for "Maximum update depth exceeded"

hyun907:

Related: #3228, #3225, #2979, #2864, #2771, #2748, #2962

## Summary
The `Troubleshooting` section of the `use-shallow` reference page was a `TBD` placeholder. New v5 users frequently encounter `Maximum update depth exceeded` when selecting multiple store values into an inline object, but there is no docs guidance pointing them to `useShallow`.

Added a troubleshooting entry that explains the root cause (inline object selector creates a new reference every render → equality check always fails → infinite loop), shows the broken pattern, shows the `useShallow` fix, shows the alternative (separate selectors), and links to the v5 migration guide.

## Check List


dbritto-dev (approved):
LGTM

PR #3569, opened 2026-08-19
docs(create): document that state cannot be read during initialization

dai-shi:
The store api passed to the state creator function is not usable for reading state while that function runs: get() and store.getState() return undefined, and store.getInitialState() throws "Cannot access 'initialState' before initialization". Per the discussion, this is a designed limitation rather than a bug, so document it in the create and createStore troubleshooting sections.

Ref: https://github.com/pmndrs/zustand/discussions/3529

Fixes #

## Check List

### DIFF
--- a/docs/reference/apis/create-store.md
+++ b/docs/reference/apis/create-store.md
@@ -21,6 +21,7 @@ const someStore = createStore(stateCreatorFn)
   - [Subscribing to state updates](#subscribing-to-state-updates)
 - [Troubleshooting](#troubleshooting)
   - [I’ve updated the state, but the screen doesn’t update](#i’ve-updated-the-state,-but-the-screen-doesn’t-update)
+  - [Reading the state inside the state creator function throws or returns undefined](#reading-the-state-inside-the-state-creator-function-throws-or-returns-undefined)
 
 ## Types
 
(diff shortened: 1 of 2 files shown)


docs/reference/apis/create-store.md
```diff
@@ -515,3 +516,44 @@ render(personStore.getInitialState(), personStore.getInitialState())
 
 personStore.subscribe(render)
 ```
+
+### Reading the state inside the state creator function throws or returns undefined
```
dai-shi: This section feels too long for such a subtle behavior. Maybe just one line warning should be enough.

dai-shi (commented):
Request changes

docs/reference/apis/create-store.md
dai-shi: @irfanfandi ☝️

irfanfandi: Good point, thanks. Dropped the whole troubleshooting section and replaced it with a one-line caveat on the `stateCreatorFn` parameter in both `create.md` and `create-store.md`:

> Note that `set`, `get` and `store` cannot be used while this function is running, only after the store has been created.

dai-shi:
@dbritto-dev we should apply this and some previous improvements in `./docs` to our new docs.

dbritto-dev:
> @dbritto-dev we should apply this and some previous improvements in `./docs` to our new docs.

Sure, I'm syncing the changes

PR #3580, opened 2026-09-11
fix(devtools): preserve shared connections during store cleanup

dai-shi:

https://github.com/pmndrs/zustand/discussions/3579


src/middleware/devtools.ts
```diff
@@ -243,15 +243,22 @@ const devtoolsImpl: DevtoolsImpl =
       )
       return r
     }) as NamedSet<S>
+    let unsubscribeFromStore: (() => void) | undefined
+    let isCleanedUp = false
```
dai-shi, line 247: I don't believe this flag is very necessary.

src/middleware/devtools.ts
```diff
@@ -243,15 +243,22 @@ const devtoolsImpl: DevtoolsImpl =
...
     ;(api as StoreApi<S> & StoreDevtools<S>).devtools = {
       cleanup: () => {
+        if (isCleanedUp) return
```
dai-shi, line 250: We prefer curly block anyway.

src/middleware/devtools.ts
```diff
@@ -243,15 +243,22 @@ const devtoolsImpl: DevtoolsImpl =
...
+        isCleanedUp = true
+        unsubscribeFromStore?.()
+        unsubscribeFromStore = undefined
+        removeStoreFromTrackedConnections(options.name, store)
         if (
+          (store === undefined || !trackedConnections.has(options.name)) &&
```
dai-shi, line 256: When do we need the first condition?

dai-shi (commented):
Request changes.

src/middleware/devtools.ts
chungyo: I kept the guard because an old store's cleanup can run again after a replacement has registered with the same `name` and `store` key. Without the guard, `removeStoreFromTrackedConnections` removes the replacement's registration. A subsequent store then opens another connection instead of sharing the existing one.

I strengthened the replacement-store test in e947e1f to check connection reuse. Removing the guard makes it fail (3 `connect` calls instead of 2); with the guard it passes. Clearing `unsubscribeFromStore` alone does not prevent the registry removal.

src/middleware/devtools.ts
chungyo: Updated to a curly block in e947e1f.

src/middleware/devtools.ts
chungyo: This handles an untracked store and a tracked store with the same `name`. When `store` is omitted, `extractConnectionInformation` creates a separate connection, so it must still be unsubscribed even if that name exists in `trackedConnections`.

Added a regression test in e947e1f: cleaning up the untracked store closes its connection while the tracked store still receives DevTools updates. Removing `store === undefined` makes the test fail because the untracked connection is never unsubscribed. Build and all checks pass locally, including 228 tests.

src/middleware/devtools.ts
dai-shi: can you pass `api` to `removeStoreFromTrackedConnections` and check `connectionInfo?.stores[store] !== api`?

src/middleware/devtools.ts
```diff
@@ -243,15 +243,24 @@ const devtoolsImpl: DevtoolsImpl =
       )
       return r
     }) as NamedSet<S>
+    let unsubscribeFromStore: (() => void) | undefined
+    let isCleanedUp = false
     ;(api as StoreApi<S> & StoreDevtools<S>).devtools = {
       cleanup: () => {
+        if (isCleanedUp) {
+          return
+        }
```
dai-shi: I just noticed there are many one-liner if-return statement in this file... never mind.

src/middleware/devtools.ts
dai-shi: sounds fair.

src/middleware/devtools.ts
chungyo: Applied in 5b3ae41. `removeStoreFromTrackedConnections` now receives `api` and returns when `connectionInfo?.stores[store] !== api`. I removed `isCleanedUp`.

This also handles a case the flag missed: a replacement registers under the same `name` and `store` before the original store's first cleanup. I added a regression test that fails before this change and verifies that the replacement remains subscribed afterward. The existing repeated-cleanup and connection-reuse tests still pass.

Build, type checks, lint, formatting, and all 229 tests pass locally.

PR #3583, opened 2026-09-16
fix(middleware): persist migration supporting promise-like

dai-shi:

Fixes #3577


src/middleware/persist.ts
```diff
@@ -279,8 +282,10 @@ const persistImpl: PersistImpl = (config, baseOptions) => (set, get, api) => {
                 deserializedStorageValue.state,
                 deserializedStorageValue.version,
               )
-              if (migration instanceof Promise) {
-                return migration.then((result) => [true, result] as const)
+              if (isPromiseLike(migration)) {
+                return Promise.resolve(migration).then(
+                  (result) => [true, result] as const,
+                )
```
dai-shi:
```suggestion
                return migration.then((result) => [true, result] as const)
```

We don't need wrapping, right?

Tnalxmsk: I checked this locally. With synchronous storage, toThenable still checks instanceof Promise. Calling migration.then(...) directly returns another Dexie/foreign-realm Promise, which that helper treats as a synchronous value. The next step then fails with migrationResult is not iterable, and hydration does not complete.

Promise.resolve normalizes the result to a native Promise that the existing helper recognizes. I think we need to keep it for this migration-only change; we could revisit it if normalization is moved into toThenable as well.

dai-shi: Ah, I see. I guess it's okay to change toThenable to use `isPromiseLike`. I'm curious why, in your use case, `getItem` doesn't return Dexie Promise. (Ah, because the storage isn't async. But what if the storage is async?)

Tnalxmsk: Right—the reproduction uses synchronous storage to isolate the migration issue, with Dexie used only inside `migrate`.

I’ve now also tested async storage returning Dexie and cross-realm Promises. Updating `toThenable` fixes the direct storage path, but `createJSONStorage` needs the same treatment to avoid parsing an unresolved Promise as JSON. I’ve made both changes locally using `isPromiseLike` and `Promise.resolve`.

With those changes, calling `migration.then(...)` directly works for Dexie and cross-realm Promises. However, a thenable that resolves a value without returning a chained Promise still fails, so I’ve kept the migration wrapper for that case:

```js
const migratedState = { drafts: ['saved'] }
const thenable = {
  then(resolve) {
    resolve(migratedState)
  },
}

thenable.then((state) => [true, state]) // returns undefined

Promise.resolve(thenable).then((state) => [true, state])
// resolves to [true, migratedState]
```

dai-shi: My take is that it's violating PromiseLike contract. It should behave like Promise and chainable. Isn't Dexie Promise chainable?

But, I think the problem is that PromiseLike doesn't have `.catch`, so we need the wrapper.
I'd do something like:
```ts
type Thenable<Value> = {
  then<V>(
    onFulfilled: (value: Value) => V | PromiseLike<V> | Thenable<V>,
  ): Thenable<V>
  catch<V>(
    onRejected: (reason: Error) => V | PromiseLike<V> | Thenable<V>,
  ): Thenable<V>
}

const isPromiseLike = (value: unknown): value is PromiseLike<unknown> =>
  typeof (value as PromiseLike<unknown>)?.then === 'function'

const toThenable =
  <Result, Input>(
    fn: (input: Input) => Result | PromiseLike<Result> | Thenable<Result>,
  ) =>
  (input: Input): Thenable<Result> => {
    try {
      const result = fn(input)
      if (isPromiseLike(result)) {
        return Promise.resolve(result) as Thenable<Result>
      }
```

But then, this makes me feel like we should migrate `toThenable` to `toPromiseLike`. 🤔

dai-shi: > migrate `toThenable` to `toPromiseLike`.

No, that doesn't work for the sync case.

dai-shi:
```suggestion
                return toThenable(migration.then((result) => [true, result] as const))
```
is my second take.

Tnalxmsk: Thanks for clarifying! You’re right—the void-returning example I shared doesn’t satisfy the PromiseLike contract. I’ve replaced it with a chainable PromiseLike whose `then` returns another PromiseLike without `catch`.

I’ve pushed an update in [9c531dc](https://github.com/pmndrs/zustand/pull/3583/commits/9c531dc01d9a422791f242f5dcfe02755ab9176f):
- Updated `toThenable` to accept PromiseLike and normalize it with `Promise.resolve`, while preserving the existing synchronous path.
- Updated `createJSONStorage` to resolve promise-like storage reads before parsing the JSON.
- Removed the extra wrapper around the migration result.

For the migration step, returning `migration.then(...)` directly works with these changes: the enclosing `toThenable` chain already normalizes the returned PromiseLike, or the native Promise chain assimilates it when storage is asynchronous.

I verified this with Dexie, cross-realm Promises, and a chainable PromiseLike without `catch`, across synchronous and asynchronous storage. 

Thanks for helping me work through this!

src/middleware/persist.ts
```diff
@@ -48,8 +48,8 @@ export function createJSONStorage<S, R = unknown>(
         return JSON.parse(str, options?.reviver) as StorageValue<S>
       }
       const str = storage.getItem(name) ?? null
-      if (str instanceof Promise) {
-        return str.then(parse)
+      if (isPromiseLike(str)) {
+        return Promise.resolve(str).then(parse)
```
dai-shi: Is it bad to return PromiseLike instead of Promise?

src/middleware/persist.ts
```diff
@@ -147,22 +147,25 @@ type StorePersist<S, Ps, Pr> = S extends {
...
-    fn: (input: Input) => Result | Promise<Result> | Thenable<Result>,
+    fn: (input: Input) => Result | PromiseLike<Result> | Thenable<Result>,
   ) =>
   (input: Input): Thenable<Result> => {
     try {
       const result = fn(input)
-      if (result instanceof Promise) {
-        return result as Thenable<Result>
+      if (isPromiseLike(result)) {
+        return Promise.resolve(result) as Thenable<Result>
```
dai-shi: I wonder if this `isPromiseLike(result)` is true for `result=Thenable<Result>`, then we handle sync with async branch, which is bad.

src/middleware/persist.ts
Tnalxmsk: I checked `str.then(parse)` locally, and it works through persist because `toThenable` normalizes the returned chain. I verified this with Dexie, cross-realm Promises, and a chainable PromiseLike without `catch`.

The distinction is standalone usage: `createJSONStorage().getItem()` can return a PromiseLike without `catch` if the underlying storage provides one. The public storage types currently declare Promise rather than PromiseLike, and I haven’t changed those types.

src/middleware/persist.ts
Tnalxmsk: You’re right—I reproduced this when a synchronous storage adapter’s `setItem` returns another store’s synchronous `persist.rehydrate()` result:

`setItem: () => otherStore.persist.rehydrate()`

The state is restored immediately, but `Promise.resolve` defers the completion callback and the update to `hasHydrated()`. My earlier tests missed this path.

I’ve added regression tests and compared explicitly marking internal synchronous wrappers with preserving synchronous settlement directly. The marker approach needs a way to recognize existing unmarked wrappers, while the settlement approach changes when external `then` methods are invoked.

The rejection path also needs care: the existing internal rejected wrapper forwards errors through `catch`, but does not invoke the rejection callback passed to `then`. Treating it as a regular PromiseLike can therefore leave the chain pending.

I’d like to preserve the synchronous behavior of the existing internal chain. Do you have a preferred approach for supporting external PromiseLike values while keeping that behavior?

dai-shi: I don't have any specific idea that I feel I'd pick. Making Thenable special is one option (I also notice our Thenable is slightly different from ES Thenable which doesn't have .catch). Wrapping with Promise is another. I wonder if there are any others.

Tnalxmsk: Thanks. One concrete option I tested is to restore `toThenable` and its internal types to the base implementation, and normalize PromiseLike values specifically at the storage-read and migration boundaries.

For the storage read, the existing hydration chain would start from:

```ts
const getItem = storage.getItem.bind(storage)

return toThenable((name: string) => {
  const result = getItem(name)
  return isPromiseLike(result) ? Promise.resolve(result) : result
})(options.name)
```

For migration:

```ts
if (isPromiseLike(migration)) {
  return Promise.resolve(migration).then(
    (result) => [true, result] as const,
  )
}
return [true, migration] as const
```

And `createJSONStorage` could return `str.then(parse)` directly.

I verified this locally with Dexie and cross-realm Promises. It addresses the reported migration issue and asynchronous storage reads without changing the common helper’s handling of synchronous Thenables.

There is a separate, pre-existing write-path limitation: when reading and migration are synchronous, a non-native PromiseLike returned by `setItem` is not awaited. Hydration can report completion before the write finishes, and a write failure may not reach the hydration callback. The two write-completion tests pass with the current PR implementation, but fail with this narrower approach.

What do you think about handling the write path separately and focusing this PR on migration and storage reads?

dai-shi: that's probably the same as my second thought I said earlier and mine uses `toThenable` instead of `Promise.resove()`.

Tnalxmsk: Right, the wrapping idea is the same. I tried it with the original toThenable, but with synchronous storage, the Dexie/cross-realm Promise was still treated as a synchronous value, causing migrationResult is not iterable.

That’s why I kept Promise.resolve at those boundaries—it converts the result into a native Promise that the existing helper recognizes.

dai-shi: Let's look for a solution that fits my preference.

Tnalxmsk: I’ve prepared a version locally that keeps the base toThenable unchanged and uses a separate converter for external PromiseLike values. Would this separation work?

```ts
const toPromise = <Value>(value: PromiseLike<Value>): Promise<Value> => {
  if (value instanceof Promise) {
    return value
  }
  return new Promise<Value>((resolve, reject) => {
    value.then(resolve, reject)
  })
}
```

The PromiseLike branch in migration becomes:
```ts
return toPromise(migration).then((result) => [true, result] as const)
```

Storage reads use the same converter inside the existing toThenable call. This keeps the external PromiseLike conversion separate from the internal synchronous chain.

I’ve pushed this version in [ff04ebe](https://github.com/pmndrs/zustand/pull/3583/commits/ff04ebec48c5cbb24b3dfde9d12fb609e8dcdf9b)

dai-shi: No, I want to slightly change toThenable which is under our control. Avoid both Promise.resolve and new Promise.

dai-shi: Maybe the common signature is `.then(cb, cb)` without .catch.

Tnalxmsk: Thanks, I’ve pushed the revised implementation in [c868217](https://github.com/pmndrs/zustand/pull/3583/commits/c868217ad5a3c1ccd79760682c9b565966250530), updating `toThenable` to use `.then(onFulfilled, onRejected)` for external PromiseLike values. This removes `toPromise` and avoids both `Promise.resolve` and `new Promise`.

The returned adapter keeps `catch` and `finally` for callers, but incoming PromiseLike values only need `then`.

The write path uses `toThenable(..., false)` to preserve the handling of older synchronous Thenables while allowing an existing asynchronous chain to await the write. Primitive results return directly, avoiding an extra wrapper. The existing limitation remains: a non-native PromiseLike returned by `setItem` is still not awaited when both the read and migration are synchronous.

Does this structure fit the approach you had in mind?

dai-shi: I should have said I don't have concrete answers. My idea is still vague. Will check your idea.

Tnalxmsk: Thanks for taking a look!

dai-shi: I haven't taken a look yet.

src/middleware/persist.ts
```diff
@@ -146,41 +146,72 @@ type StorePersist<S, Ps, Pr> = S extends {
...
-    onRejected: (reason: Error) => V | Promise<V> | Thenable<V>,
-  ): Thenable<V>
+  then<Fulfilled = Value, Rejected = never>(
+    onFulfilled?: ((value: Value) => Fulfilled | PromiseLike<Fulfilled>) | null,
+    onRejected?: ((reason: unknown) => Rejected | PromiseLike<Rejected>) | null,
+  ): Thenable<Fulfilled | Rejected>
+  catch<Rejected = never>(
+    onRejected?: ((reason: unknown) => Rejected | PromiseLike<Rejected>) | null,
+  ): Thenable<Value | Rejected>
+  finally(onFinally?: (() => unknown) | null): Thenable<Value>
```
dai-shi, line 156: This is over complicated. We should have control on this. Why not use only `.then`?

src/middleware/persist.ts
```diff
@@ -146,41 +146,72 @@ type StorePersist<S, Ps, Pr> = S extends {
...
 }
 
+const isPromiseLike = (value: unknown): value is PromiseLike<unknown> =>
+  typeof (value as PromiseLike<unknown>)?.then === 'function'
+
 const toThenable =
   <Result, Input>(
-    fn: (input: Input) => Result | Promise<Result> | Thenable<Result>,
+    fn: (input: Input) => Result | PromiseLike<Result>,
+    acceptPromiseLike = true,
```
dai-shi, line 165: Can we avoid this?

dai-shi (commented):
Thanks for working on it. It looks promising. If you give up, I can try.

src/middleware/persist.ts
Tnalxmsk: Sorry for the slow reply, and thanks for the review.

I’ve removed the custom Thenable type locally and changed toThenable to return PromiseLike, so the internal chain only uses .then.

The base implementation already supports .catch() on synchronous rehydrate() results. To preserve that, I’ve kept a small adapter only for the public return value, using Promise.prototype.catch and Promise.prototype.finally.

Synchronous hydration is preserved. The adapter does not convert hydration into a native Promise, although invoking the native finally method can create Promises.

What do you think about keeping this small adapter only for the public rehydrate() result?

src/middleware/persist.ts
Tnalxmsk: I’ve removed the boolean option locally. The synchronous value wrapping needed by the write path is now handled by toSyncThenable.

